PR #133, opened 2026-03-06
feat: custom prompt overrides + multi-arg spec/learn

### LINKED ISSUE #117: feat: Allow `spec` and `learn` commands to accept multiple arguments
## Summary
Currently, the `totem spec <input>` and `totem learn <pr-number>` commands only accept a single argument. When working on a feature that spans multiple related tickets (or duplicate bugs), it would be highly beneficial to generate a single cohesive spec from multiple issues. Similarly, extracting lessons from several recently merged PRs at once would streamline the `learn` workflow.

## Proposal
Update the CLI signatures to accept variadic arguments:
- `totem spec [inputs...]` (e.g., `totem spec 123 124 https://github.com/org/repo/issues/125`)
- `totem learn [pr-numbers...]` (e.g., `totem learn 101 102 103`)

## Implementation Details
1. **Commander Setup:** Update `packages/cli/src/index.ts` to use variadic arguments (e.g., `<inputs...>`).
2. **Fetching:** Update the commands to loop through the array of inputs and fetch the data for each issue/PR.
3. **Context Assembly:** Ensure the assembled prompts clearly delineate between the different issues/PRs (e.g., using `=== ISSUE #123 ===`).
4. **Token Limits:** Consider adding a reasonable hard cap (e.g., max 5 items) or a warning to prevent blowing out the context window if a user passes too many large issues/PRs.

### LINKED ISSUE #120: feat: Support custom system prompt overrides for built-in commands
## Context
Currently, the built-in orchestrator commands (`spec`, `shield`, `triage`, `briefing`, `handoff`, `learn`) have highly opinionated `SYSTEM_PROMPT` strings hardcoded directly into their TypeScript source files. 

For example, `totem shield` forces the LLM to adopt a \"Red Team Reality Checker\" persona, and `totem triage` forces a \"Product Manager\" persona. While these are excellent defaults for the \"Magic Onboarding\" experience, they become restrictive for power users and enterprise teams. 

A team might want `totem shield` to focus exclusively on an internal OWASP security checklist, or they might want `totem spec` to output a specific JSON structure for their Jira integration.

## Proposal: Custom Prompt Overrides
Allow users to override the hardcoded system prompts by placing markdown files in a `.totem/prompts/` directory.

### How it works
1. Before assembling the prompt for a command (e.g., `shield`), the CLI checks if `.totem/prompts/shield.md` exists in the local workspace.
2. If it exists, the CLI uses the contents of that file as the `SYSTEM_PROMPT` instead of the hardcoded default.
3. If it does not exist, the CLI falls back to the hardcoded default to maintain zero-config onboarding.

This approach perfectly aligns with the Totem philosophy of providing configurable primitives rather than rigid SaaS workflows, serving as the natural sister-feature to the custom workflow runner (`totem run <workflow>`).

## Acceptance Criteria
- [ ] Add a utility function to check for and read override files from `.totem/prompts/<command>.md`.
- [ ] Update all orchestrator commands (`spec`, `shield`, `triage`, `briefing`, `handoff`, `learn`) to use the override utility when assembling their prompts.
- [ ] Ensure the override logic gracefully handles missing files (falling back to the hardcoded defaults).
- [ ] Update `README.md` to document the ability to override built-in personas.

satur8d:
## Summary

- **Custom prompt overrides (#120):** Add `getSystemPrompt()` utility that reads `.totem/prompts/<command>.md` if present, with path traversal protection and empty-file fallback. All 6 orchestrator commands (spec, shield, learn, briefing, handoff, triage) now support custom system prompts.
- **Multi-arg spec (#117):** `totem spec 120 117` now accepts multiple issue numbers/URLs/topics with dedup and a MAX_INPUTS=5 cap.
- **Multi-arg learn (#117):** `totem learn 100 101` processes each PR sequentially, accumulating lessons across all PRs before writing.
- **Shield-caught fix:** `--raw` mode in multi-PR learn loop now correctly processes all PRs instead of exiting on the first.
- **Docs:** Includes Gemini's roadmap additions and proactive memory lesson (absorbed from closed PR #132).

## Test plan

- [x] 5 new unit tests for `getSystemPrompt` (override file, empty file, missing file, path traversal, unreadable file)
- [x] All 159 existing tests pass (36 core + 123 CLI)
- [x] Pre-push checks pass (format, lint, test)
- [x] `totem shield` review — all critical issues resolved

Closes #120
Closes #117

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/cli/src/commands/spec.ts
+++ b/packages/cli/src/commands/spec.ts
@@ -118,7 +118,7 @@ function assemblePrompt(
       }
     } else if (freeText) {
       sections.push('\n=== TOPIC ===');
-      sections.push(freeText);
+      sections.push(wrapXml('topic_text', freeText));
     }
   }
 


satur8d:
@gemini-code-assist Re: comments #2 and #3 (graceful degradation in spec/learn loops) — **declined**.

The fail-fast behavior is intentional. A partial context assembly is highly dangerous because the LLM will confidently generate a response based on incomplete information. If `totem spec 101 102 103` silently drops issue 103 due to a fetch failure, the AI assumes it has the full picture and generates a spec that misses a third of the requirements — and the developer trusts it.

It is infinitely better for the CLI to crash loudly than for the AI to hallucinate silently. The developer sees the error, fixes the ID or checks their network, and retries with complete data.

PR #197, opened 2026-03-08
feat: universal AI developer baseline lessons (#128)

satur8d:
## Summary

Solves the cold-start problem: fresh Totem installs now offer a curated set of 10 foundational AI developer lessons during `totem init`, so agents have useful knowledge from Day 1.

Closes #128

## What's New

- **10 Universal Baseline lessons** covering: prompt injection, secret management, hallucination traps, scope creep, dependency verification, test quality, error handling, shell injection, context window management, and idempotency
- **Opt-in prompt** during `totem init`: "Install Universal AI Developer Baseline lessons? (Y/n)"
- **Non-TTY safety**: defaults to installing without prompting in CI/piped environments
- **Idempotent**: uses `<!-- totem:baseline -->` HTML comment marker — re-running `init` won't duplicate lessons
- **Preserves user content**: appends to existing `lessons.md` without overwriting
- **TS constant approach**: lessons stored as a TypeScript string constant (not raw `.md` file) to avoid build-time asset loss

## Files Changed

- `packages/cli/src/assets/universal-lessons.ts` — NEW: baseline lessons content + marker constant
- `packages/cli/src/commands/init.ts` — extracted `installBaselineLessons()` function, wired into init flow
- `packages/cli/src/commands/init.test.ts` — 12 new tests (6 format/constant, 6 functional)
- `README.md`, `docs/architecture.md`, `docs/roadmap.md` — documentation updates

## Tests

- 285 tests passing (80 core + 196 cli + 9 mcp)
- 12 new tests covering: baseline format validation, idempotency, opt-in/opt-out paths, TTY vs non-TTY behavior, user content preservation
- Shield: PASS

PR #199, opened 2026-03-08
feat: reap orphaned temp files on CLI startup (#108)

### LINKED ISSUE #108: UX: Clean up orphaned temporary prompt files
## Context
During an architectural review of the CLI package (\packages/cli/src/utils.ts\), I reviewed the \invokeShellOrchestrator\ mechanism.

Currently, it writes LLM prompts to a temporary file (\.totem/temp/totem-<tag>-<hash>.md\) to bypass shell escaping limits, and deletes it in a \inally\ block after the shell command completes.

## The Trap
If the user forcefully kills the CLI (e.g., hitting \Ctrl+C\ twice, or the process receives a \SIGKILL\ from the OS), the \inally\ block will not execute. Over time, these orphaned Markdown files will accumulate in the \.totem/temp/\ directory, wasting disk space.

## The Fix
Add a lightweight cleanup step to a high-frequency command (like \	otem sync\ or the start of the CLI \index.ts\) that asynchronously scans the \.totem/temp/\ directory and deletes any file older than 24 hours. Alternatively, use a process exit hook (\process.on('SIGINT')\), though a time-based reaper is generally more reliable.

satur8d:
## Summary

- **Automatic cleanup:** Adds `reapOrphanedTempFiles()` — a fire-and-forget async reaper that removes `totem-*.md` files older than 24h from `.totem/temp/` on every CLI invocation
- **Zero user impact:** Runs asynchronously, never blocks CLI startup, silently swallows permission/race errors
- **Chore:** Extracts lessons from PR #197 and refreshes triage roadmap

## Why

The CLI writes large prompts to temp files for the shell orchestrator (avoids Windows "argument list too long" errors). The `finally` block cleans them up, but if the process is killed (`SIGKILL`, double `SIGINT`, crash), files are orphaned. Over time this clutters `.totem/temp/`.

## Implementation

- `reapOrphanedTempFiles(cwd, totemDir, maxAgeMs)` in `utils.ts` — scans directory, filters by `/^totem-.*\.md$/` pattern and `mtime` age, deletes matches
- Wired into `index.ts` as `reapOrphanedTempFiles(...).catch(() => {})` before `program.parse()`
- Catches `ENOENT` (race conditions) and `EACCES`/`EPERM` (permissions) silently

## Test plan

- [x] 6 new tests in `utils.test.ts` (202 CLI tests total, 291 across all packages)
- [x] Removes files older than threshold
- [x] Keeps files younger than threshold
- [x] Ignores non-totem files in the temp directory
- [x] Returns 0 when temp directory doesn't exist
- [x] Handles mixed old/new files correctly
- [x] Swallows errors from already-deleted files (race condition)
- [x] `totem shield` PASS

Closes #108

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -228,7 +228,7 @@ export async function reapOrphanedTempFiles(
   maxAgeMs: number = TEMP_MAX_AGE_MS,
 ): Promise<number> {
   const tempDir = path.join(cwd, totemDir, 'temp');
-  const { readdir, stat, unlink } = await import('node:fs/promises');
+  const { readdir, stat, unlink } = fs.promises;
 
   let entries: string[];
   try {


satur8d:
@gemini-code-assist Review roundup:

1. **`fs.promises` instead of dynamic import** (round 1) — ✅ Accepted, fixed in `de0eeef`.
2. **`Promise.all` parallelization** (round 2) — Declined. This directory will realistically have 0-3 orphaned files. The sequential `for...of` loop is universally readable, and `Promise.all` + `map` + `reduce` triples the code for zero practical gain. Sequential awaits also give cleaner error semantics — each file's error is caught and swallowed in isolation without needing `Promise.allSettled`.

Issue #214, opened 2026-03-08
Feature: CI Drift Gate (Structural Integrity Check)


satur8d:
## Context
We recently shipped the foundation for drift detection in #181 (`totem sync --prune`). 

## Problem
Currently, this is a local, developer-initiated action. To build a "self-cleaning brain", we need to make this a CI-blocking check.

## Proposed Solution
Extend drift detection into a CI gate. If a PR deletes or renames files referenced in `.totem/lessons.md`, the CI pipeline should fail, forcing the developer to update the lesson alongside the code.

**Note:** As discussed in our strategy review, we should start with easy structural checks (file deletions/renames). Semantic drift ("this lesson's advice no longer applies because the API changed") requires LLM judgment, which contradicts our goal of keeping LLMs out of deterministic CI gates.

## Priority
Ranked as **Priority 3**. This is a natural extension of #181 and relatively low effort.

PR #280, opened 2026-03-10
feat: CI drift gate + adversarial evaluation harness

satur8d:
## Summary
- **`totem drift` command** — Scans `lessons.md` for stale file references and exits non-zero on orphans. CI-safe (no interactivity), works at Lite tier (no embeddings/LLM needed). Wired into `shield.yml` as a new CI step alongside the deterministic shield.
- **Adversarial evaluation harness** — 10 planted-bug test scenarios proving compiled rules catch: npm usage, `error` vs `err`, console.log, empty catches, TODO comments, hardcoded secrets, `any` types, multi-violation diffs, and inline suppression. Deterministic (no LLM calls).
- **Stale ref fix** — Removed hypothetical file path from a lesson that was triggering drift detection.

Closes mmnto-ai/totem#214
Partially addresses mmnto-ai/totem-strategy#8 (deterministic subset — full LLM adversarial harness is future work)

## Test plan
- [x] 15 new tests (5 drift gate CLI, 10 adversarial core)
- [x] All 554 tests pass (216 core + 338 CLI)
- [x] `totem shield --deterministic` passes (24 rules)
- [x] `totem drift` passes on Totem's own lessons (150 lessons, 0 stale refs)
- [x] Lint and format clean

### DIFF
--- a/.totem/lessons.md
+++ b/.totem/lessons.md
@@ -387,7 +387,7 @@ Treat environment variables provided by the host AI (such as `$TOOL_INPUT` in Cl
 
 **Tags:** nodejs, configuration, idempotency
 
-When scaffolding configuration files that store settings in JSON arrays (like `.claude/settings.local.json`), implement deep merging to append entries rather than overwriting the entire key. This allows the tool to maintain idempotency while preserving existing user-defined hooks.
+When scaffolding configuration files that store settings in JSON arrays, implement deep merging to append entries rather than overwriting the entire key. This allows the tool to maintain idempotency while preserving existing user-defined hooks.
 
 ## Lesson — 2026-03-06T18:48:00.895Z
 


satur8d:
@gemini-code-assist Thanks for the second pass. Addressing all comments:

**`log.info` → `log.warn` (drift.ts:60)** — Already applied in 429033b. ✅

**RCE via `loadConfig` executing `totem.config.ts` (critical)** — Declined. This is the standard TypeScript config pattern used by Vite, ESLint, Prettier, Tailwind, etc. The workflow runs on `pull_request` (not `pull_request_target`), so the config change is always visible in the PR diff and subject to review. Every tool with a `*.config.ts` has this characteristic — it's not unique to Totem.

**Path traversal in `detectDrift` (security-medium)** — Valid defense-in-depth observation. `lessons.md` is a committed, reviewed file (not untrusted input), so the practical risk is low — an attacker with commit access can already run arbitrary code. That said, adding a path containment check is cheap insurance. Created follow-up issue.

**TODO regex word boundaries** — Declined. These are test fixture rules for the adversarial harness, not production compiled rules. The test diffs use `// TODO:` and `// FIXME:` which won't false-positive on substrings. Production rules are compiled via `totem compile` with LLM-generated patterns.

**Password regex robustness (repeated)** — Declined for the same reason — test fixture, not a production rule. Also note the suggestion has a bug (uses `"[^"]*"` for both alternatives instead of handling single quotes separately).

PR #338, opened 2026-03-11
fix: totem hooks walks up to git root in monorepo sub-packages

### LINKED ISSUE #333: fix: `totem hooks` should walk up to git root in monorepo sub-packages
## Problem

Running `totem hooks` from a sub-package directory in a monorepo (e.g., `packages/cli/`) fails silently because `path.join(cwd, '.git')` doesn't exist — the `.git` directory is at the monorepo root.

The command should walk up the directory tree to find the git root (like `git rev-parse --show-toplevel`) before checking for `.git/hooks/`.

## Current Behavior
- `totem hooks` from sub-package → "Not a git repository" → silent skip
- Developer falsely assumes hooks were installed

## Expected Behavior
- `totem hooks` finds the git root automatically and installs hooks there
- Works identically whether run from root or sub-package

## Context

Identified during shield review of #332. Low priority since most developers run from the monorepo root, but worth fixing for robustness.

satur8d:
## Summary

- Adds `resolveGitRoot(cwd)` to `git.ts` using `git rev-parse --show-toplevel`
- All hook install/check functions now resolve the git root instead of assuming `.git` exists in `cwd`
- Handles worktrees, submodules, and monorepo sub-packages correctly
- 2 new tests verify subdirectory install + check behavior (42 total)

Closes #333

## Test plan

- [x] `totem hooks` from monorepo sub-package installs hooks at git root
- [x] `totem hooks --check` from sub-package finds hooks at git root
- [x] Non-git directories return null gracefully
- [x] Hook manager detection uses git root (finds `.husky` at root)
- [x] All 42 install-hooks tests pass
- [x] Full suite green (454 CLI tests)

### DIFF
--- a/packages/cli/src/commands/install-hooks.ts
+++ b/packages/cli/src/commands/install-hooks.ts
@@ -392,7 +392,6 @@ export function installHooksNonInteractive(cwd: string): HooksCommandResult | nu
 export function checkHooksInstalled(cwd: string): boolean {
   const gitRoot = resolveGitRoot(cwd);
   if (!gitRoot) {
-    console.error('[Totem] Not a git repository.');
     return false;
   }
   const hooksDir = path.join(gitRoot, '.git', 'hooks');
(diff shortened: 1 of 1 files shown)


satur8d:
@gemini-code-assist Good catches on both. Fixed in 3ca0772:

1. **`checkHooksInstalled` UX** — Resolved git root once at the top of `hooksCommand`, before dispatching to either `--check` or install path. No more confusing double message.
2. **Redundant `resolveGitRoot` call** — Eliminated by the same early guard. Didn't go with the status object refactor since the early exit achieves the same clarity without expanding the API surface.

PR #649, opened 2026-03-17
fix: audit compiled rules to reduce false positives

### LINKED ISSUE #105: Resilience: Add exponential backoff to OpenAI embedder for rate limits
## Context
During an architectural review of the Core package (\packages/core/src/embedders/openai-embedder.ts\), it was noted that the OpenAI embedder relies on a single \client.embeddings.create\ call per batch without any retry logic.

## The Trap
If a user runs \	otem sync --full\ on a very large repository, they may hit the OpenAI API rate limit (HTTP 429 Too Many Requests). Currently, this will cause the entire sync process to crash and fail. 

## The Fix
Implement a retry mechanism with exponential backoff around the OpenAI API call in the \embedBatch\ method. The official \openai\ Node SDK might have some built-in retry options we can enable, or we can use a simple \setTimeout\ backoff loop.

### LINKED ISSUE #123: Epic: Federated Memory (Mothership Pattern)
## Context
When a user relies on Totem across multiple local projects, they inevitably develop a "Mothership" repository. In our dogfooding case, the `totem` repo naturally aggregates all the lessons about *how to use AI effectively* (e.g., prompt injection defenses, LLM tool boundaries, orchestrator patterns). 

Currently, when working in a downstream consumer project (like a local game or API), that project's Totem index is entirely isolated. It cannot benefit from the meta-lessons learned in the mothership.

## The Proposal: Federated Memory (Mothership Pattern)
Allow a `totem.config.ts` to declare external/upstream LanceDB indexes. When `totem search` or an MCP tool runs, it should query the local project's database *and* the federated databases, merging the results.

### Example Configuration
```typescript
// in my-app/totem.config.ts
export default {
  // ...
  federatedIndexes: [
    { 
      name: 'totem-mothership', 
      path: '../totem/.lancedb' // Relative or absolute path to the upstream LanceDB
    }
  ]
}
```

## Why this matters
- **Prevents duplicated learning:** You shouldn't have to re-learn "never use npm" or "how to prevent terminal injection" in every single repo you initialize.
- **Enterprise scale:** This is the foundational primitive for Phase 4 (Enterprise). A platform team can host a read-only `.lancedb` on S3 or a network share, and all developer repos can federate it to inherit company-wide security policies.

## Acceptance Criteria
- [ ] Update `totem.config.ts` schema to support an optional `federatedIndexes` array.
- [ ] Update `packages/core/src/store/lance-store.ts` to allow connecting to and querying multiple database paths simultaneously.
- [ ] Update the MCP `search_knowledge` tool to merge results from both the local and federated indexes, annotating which index the result came from.

### LINKED ISSUE #140: feat: Claude Code integration — hook-based automation
## Parent Epic
Part of #86 — Seamless Host Integration

## Dependencies
Depends on #138 (hookInstaller infrastructure)

## Summary
Scaffold Claude Code hooks so that `totem briefing` runs automatically on session start and `totem shield` runs before commits/pushes — without relying solely on prompt engineering in CLAUDE.md.

## Context
Claude Code supports hooks via `.claude/settings.json` (or project-level `.claude/settings.local.json`). These can trigger shell commands on events like `PreToolUse` and session start via `user_prompt_submit`.

### Session Start Hook
- Configure a hook that runs `totem briefing` at the start of a Claude Code session
- Output should be injected into the conversation context

### Pre-Push/Commit Shield Hook
- Configure a `PreToolUse` hook that intercepts Bash tool calls containing `git push` or `git commit`
- Runs `totem shield` and blocks the tool call if violations are found

### Init Integration
- Implement `hookInstaller` for the Claude Code `AiToolInfo` entry
- Scaffold `.claude/settings.local.json` with hook configuration
- Use marker-based idempotency — do not overwrite existing user hooks

## Files to Modify
- `packages/cli/src/commands/init.ts` (Claude hookInstaller)

## Edge Cases
- Existing `.claude/settings.local.json` with user-configured hooks — merge, don't overwrite
- Claude Code not installed — scaffold anyway (ready when available)
- The existing `AI_PROMPT_BLOCK` in CLAUDE.md serves as fallback for environments without hook support

## Acceptance Criteria
- [ ] `totem init` scaffolds Claude Code hooks when `.claude/` or `CLAUDE.md` detected
- [ ] Re-running `totem init` does not duplicate hooks
- [ ] Briefing runs on session start in Claude Code
- [ ] Shield intercepts git push/commit via PreToolUse hook

satur8d:
## Summary

Audits all 148 compiled rules to reduce noise from false positives that were blocking PRs and cluttering code scanning alerts.

### Changes
- **DELETE rule #33**: Broad `\bimport` matcher flagged stdlib imports like `node:child_process` — redundant with rules #105 and #142 which cover the specific cases
- **FIX #105**: `@mmnto/totem` static import check now excludes `import type` (was false-positive on type-only imports)
- **FIX #123**: `throw new Error` check now excludes `TotemError` variants — the clean error pattern was being flagged as the problem it solves
- **FIX #140**: Updated for TotemError hierarchy, set severity to warning
- **Set severity on 9 rules**: All had `undefined` severity, now set to `warning`

### Result
- 148 → 147 rules
- 0 undefined severity
- False positives on TotemError, type imports, and stdlib imports eliminated

## Test plan
- [x] All 999 tests pass
- [x] `totem lint` — 0 errors, 0 false positives
- [x] Build passes

### DIFF
--- a/.totem/compiled-rules.json
+++ b/.totem/compiled-rules.json
@@ -8,12 +8,7 @@
       "message": "Do not use os.tmpdir() for agent-readable files; use workspace-local paths (e.g., '.totem/temp/') to satisfy MCP boundary restrictions.",
       "engine": "regex",
       "compiledAt": "2026-03-09T02:47:42.931Z",
-      "fileGlobs": [
-        "*.ts",
-        "*.js",
-        "!*.test.ts",
-        "!*.spec.ts"
-      ],
+      "fileGlobs": ["*.ts", "*.js", "!*.test.ts", "!*.spec.ts"],
       "category": "architecture",
       "severity": "error"
     },
@@ -24,10 +19,7 @@
       "message": "Use fs.openSync() instead of fs.createWriteStream() for spawn() stdio. WriteStream.fd is null until the 'open' event fires, which will cause spawn() to fail.",
       "engine": "regex",
       "compiledAt": "2026-03-09T02:50:03.081Z",
-      "fileGlobs": [
-        "packages/cli/**/*.ts",
-        "!**/*.test.ts"
-      ],
+      "fileGlobs": ["packages/cli/**/*.ts", "!**/*.test.ts"],
       "category": "style",
       "severity": "warning"
     },
@@ -48,10 +40,7 @@
       "message": "Anchor input regexes with '^' and include 'https?://' protocol to avoid substring matches in CLI input.",
       "engine": "regex",
       "compiledAt": "2026-03-09T02:53:05.481Z",
-      "fileGlobs": [
-        "packages/cli/**/*.ts",
-        "!**/*.test.ts"
-      ],
+      "fileGlobs": ["packages/cli/**/*.ts", "!**/*.test.ts"],
       "category": "architecture",
       "severity": "error"
     },
@@ -62,10 +51,7 @@
       "message": "Use backticks (`) for column identifiers in LanceDB filters; double quotes (\") cause silent failures in DataFusion.",
       "engine": "regex",
       "compiledAt": "2026-03-09T02:57:21.498Z",
-      "fileGlobs": [
-        "packages/core/**/*.ts",
-        "!**/*.test.ts"
-      ],
+      "fileGlobs": ["packages/core/**/*.ts", "!**/*.test.ts"],
       "category": "architecture",
       "severity": "error"
     },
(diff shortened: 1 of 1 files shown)


satur8d:
@gemini-code-assist Good catch on the `import { type X }` inline syntax. In practice our codebase uses `import type { X }` exclusively (enforced by our ESLint config), so this edge case doesn't produce false positives today. The suggested pattern is correct but adds significant regex complexity for a theoretical case — deferring to keep the rule maintainable. Will revisit if we adopt inline `type` imports.

PR #660, opened 2026-03-18
fix: dimension mismatch detection via index-meta.json (#654)

### LINKED ISSUE #654: fix: dimension mismatch detection via index-meta.json
## Description
Hard-fail \	otem sync\ and \	otem shield\ if the current embedder configuration does not match the vector database's actual dimensions.

## The Problem
LanceDB silently accepts mismatched vectors and returns garbage results if a user switches providers (e.g., from OpenAI 1536d to Gemini 768d). The lazy embedder warns on fallback, but we need a hard structural check before querying.

## Scope
- Store embedder provider and dimensions in \.totem/index-meta.json\ during sync.
- On every query/sync, verify the current config matches this metadata.
- If mismatched, refuse the query and output: \[Totem Error] Index built with OpenAI (1536d) but config uses Gemini (768d). Run 'totem sync --full'.\

## Strategic Context
ADR-051. Prevents silent degradation, which is critical for multi-totem environments where users might change providers frequently.

satur8d:
## Summary

Prevents silent garbage results when a user changes their embedding provider without rebuilding the index.

### Changes
- `totem sync` now writes `.totem/cache/index-meta.json` with provider, model, dimensions, and timestamp
- `verifyIndexMeta()` checks current config against stored metadata
- Throws `TotemDatabaseError` with clear recovery hint if mismatched:
  ```
  [Totem Error] Index was built with OpenAI (1536d) but config now uses Gemini (768d).
    Fix: Run 'totem sync --full' to rebuild the index.
  ```
- Exported from `@mmnto/totem` for use by MCP server and CLI

### Why this matters
LanceDB silently accepts vectors of any dimension. If a user switches from OpenAI (1536d) to Gemini (768d), searches return semantically meaningless results with no error. This is especially critical for multi-totem setups where different repos use different providers.

## Test plan
- [x] 1,008 tests pass
- [x] Build passes
- [x] Format clean

Closes #654

### DIFF
--- a/packages/core/src/ingest/pipeline.ts
+++ b/packages/core/src/ingest/pipeline.ts
@@ -27,7 +27,17 @@ interface IndexMeta {
 function readIndexMeta(totemDir: string): IndexMeta | null {
   const metaPath = path.join(totemDir, INDEX_META_FILE);
   try {
-    return JSON.parse(fs.readFileSync(metaPath, 'utf-8')) as IndexMeta;
+    const parsed = JSON.parse(fs.readFileSync(metaPath, 'utf-8')) as IndexMeta;
+    if (
+      parsed &&
+      typeof parsed.provider === 'string' &&
+      typeof parsed.model === 'string' &&
+      typeof parsed.dimensions === 'number' &&
+      typeof parsed.lastSync === 'string'
+    ) {
+      return parsed;
+    }
+    return null;
   } catch {
     return null;
   }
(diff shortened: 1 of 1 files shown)


satur8d:
@gemini-code-assist All 3 findings addressed in b8cc456:

1. **High (dimension fallback flaw)**: Removed `resolveDefaultDimensions` entirely — no more hardcoded defaults. Provider is compared directly; dimensions are only checked when explicitly set in config. The actual dimensions are captured at sync time from `embedder.dimensions` which already resolves defaults correctly via `LazyEmbedder`.

2. **Medium (unsafe type assertion)**: `readIndexMeta` now validates all fields with `typeof` checks before returning.

3. **Medium (missing error code)**: Added `'DATABASE_MISMATCH'` error code to the `TotemDatabaseError` throw.

PR #752, opened 2026-03-20
feat: Pipeline 1 — manual patterns in lessons (#748)

### LINKED ISSUE #748: feat: Pipeline 1 — manual patterns in lessons (zero LLM compilation)
Part of #747 (Multi-Pipeline Confidence Engine, Proposal 186).

### What
Allow lesson authors to provide the pattern directly in the lesson markdown. When `totem compile` encounters a lesson with a `Pattern:` field, it skips the LLM and extracts the rule directly. Zero LLM cost, 100% compilation rate.

### Lesson format
```markdown
## Lesson — Never use process.env directly
**Tags:** config, env
**Pattern:** process\.env\[
**Engine:** regex
**Scope:** src/**/*.ts, !src/**/config.ts
**Severity:** warning

Use a validated config schema instead of direct env var access.
Fix: Read env vars through a Zod-validated config module.
```

### Implementation
1. Update `parseCompilerResponse` or add a new parser that extracts Pattern/Engine/Scope/Severity from lesson body
2. In `compile.ts`, check for these fields before calling the LLM — if present, skip LLM
3. Validate the pattern (regex syntax check, ReDoS check) same as LLM-generated patterns
4. Write the rule to compiled-rules.json with the extracted fields

### What NOT to do
- Don't change the LLM compilation path (Pipeline 2)
- Don't require these fields — they're optional enrichments
- Don't build the router yet — just check for the fields in compile.ts

satur8d:
## Summary

Lessons can now include `Pattern:`, `Engine:`, `Scope:`, and `Severity:` fields directly in the markdown body. When `totem compile` encounters these fields, it skips the LLM and extracts the rule deterministically.

This is Pipeline 1 from Proposal 186 / ADR-058 (Multi-Pipeline Confidence Engine).

### Example lesson
```markdown
## Lesson — Never use process.env directly
**Tags:** config, env
**Pattern:** process\.env\[
**Engine:** regex
**Scope:** src/**/*.ts, !src/**/config.ts
**Severity:** warning

Use a validated config schema instead of direct env var access.
Fix: Read env vars through a Zod-validated config module.
```

### What this enables
- Zero LLM cost for rules where the author knows the pattern
- 100% compilation rate for manual patterns
- Power users can author precise rules without waiting for Flash to hallucinate

### Implementation
- `lesson-pattern.ts` — parser that extracts manual fields from lesson body
- `compile.ts` — checks for manual pattern before calling LLM; if found, skips LLM entirely
- 8 new tests for the parser

Closes #748

## Test plan
- [x] 1,027 tests pass (392 core + 626 cli + 9 mcp)
- [x] `totem lint` — 0 violations
- [x] `totem shield` — reviewed (backtick edge case noted, not blocking)

### DIFF
--- a/packages/core/src/lesson-pattern.ts
+++ b/packages/core/src/lesson-pattern.ts
@@ -18,7 +18,8 @@ export interface ManualPattern {
 
 function extractField(body: string, field: string): string | undefined {
   // Match: **Field:** value, **Field**: value, Field: value
-  const re = new RegExp(`^(?:\\*{2})?${field}:?(?:\\*{2})?:?\\s+(.+)$`, 'im');
+  // Colon is mandatory to avoid matching prose like "Pattern is important..."
+  const re = new RegExp(`^(?:\\*{2})?${field}:(?:\\*{2})?\\s+(.+)$`, 'im');
   const match = body.match(re);
   return match?.[1]?.trim();
 }


satur8d:
@gemini-code-assist The current regex is correct. In markdown, `**Pattern:**` means the colon is inside the bold markers. The regex `${field}:(?:\*{2})?\s+` handles this by matching `Pattern:` then optionally `**` then whitespace.

Your suggestion `${field}(?:\*{2})?:\s+` would match `**Pattern**: value` but NOT `**Pattern:** value` — which is the standard markdown format. All 8 tests pass with the current regex, including both bold formats.

PR #888, opened 2026-03-23
fix: Phase 2 broken functionality — wrap.ts, AST queries, process.exit, suppression metrics

### LINKED ISSUE #876: fix: wrap.ts error matching, AST query result loss, process.exit, suppression metrics
## Phase 2 — Broken Functionality (Proposal 189)

Four items that are demonstrably not working as intended.

### 2A. `wrap.ts` error name matching is broken
**File:** `packages/cli/src/commands/wrap.ts:49,88`
Checks `err.name === 'NoDocsConfiguredError'` but actual errors are `TotemConfigError`/`TotemError`. Catch blocks never match — any error from docs/compile aborts the wrap pipeline even when benign.
**Fix:** Check `err.code` or use `instanceof`.

### 2B. `matchAstQueriesBatch` result loss
**File:** `packages/core/src/ast-query.ts:213-225`
Results stored in `Map<string, AstMatch[]>` keyed by query string. Duplicate queries overwrite — rules sharing the same AST query against different line ranges silently miss violations.
**Fix:** Key by index (array), matching how `matchAstGrepPatternsBatch` already works.

### 2C. `process.exit(1)` in shield, drift, lint-lessons, test-rules
**Files:** `shield.ts:403,419,465,471`, `drift.ts:53`, `lint-lessons.ts:47`, `test-rules.ts:88`
Kills process immediately — skips cleanup, prevents composability, makes testing impossible. `extract.ts` already uses `process.exitCode = 1` correctly.
**Fix:** Throw typed errors, let top-level handler in `index.ts` exit.

### 2D. AST rule suppression metrics never recorded
**File:** `packages/core/src/rule-engine.ts:199-234`
Suppressed lines filtered before AST query, so `onRuleEvent('suppress',...)` never fires for tree-sitter rules.
**Fix:** Remove pre-filter and rely on post-match suppression check, or record metrics in the pre-filter.

## References

- Proposal 189 — Phase 2
- Claude 1.4.0 code review (W9, W10, C4, W18)

satur8d:
## Summary

Addresses all 4 items in Phase 2 of Proposal 189 (broken functionality that is demonstrably not working as intended).

- **2A: wrap.ts error matching** — `err.name === 'NoDocsConfiguredError'` never matched because actual errors are `TotemConfigError`. Fixed to use `instanceof TotemError` + `.code` checks. Wrap pipeline no longer aborts on benign "no docs configured" errors.
- **2B: AST query result loss** — `matchAstQueriesBatch` stored results in a Map keyed by query string. Duplicate queries overwrote each other. Changed to array-by-index, matching `matchAstGrepPatternsBatch`.
- **2C: process.exit(1)** — Replaced in shield, drift, lint-lessons, test-rules with thrown `TotemError`. Functions are now composable and testable. Top-level `handleError` still calls `process.exit(1)`.
- **2D: Suppression metrics** — Removed pre-filter that stripped suppressed lines before AST queries. `onRuleEvent('suppress', ...)` now fires correctly for tree-sitter and ast-grep rules.

Also fixes:
- wrap.test.ts mock for `node:child_process` (pre-existing `execFile` export error)
- New error codes `DRIFT_FAILED` and `TEST_FAILED` in `TotemErrorCode`

Closes #876

## Test plan

- [x] 1,168 tests passing (461 core + 692 cli + 15 mcp)
- [x] 5 new tests (3 ast-query batch, 2 rule-engine suppression)
- [x] `totem shield` run
- [x] Lint + format clean

### DIFF
--- a/packages/cli/src/commands/drift.ts
+++ b/packages/cli/src/commands/drift.ts
@@ -1,7 +1,5 @@
 import * as path from 'node:path';
 
-import { detectDrift, readAllLessons, TotemError } from '@mmnto/totem'; // totem-ignore
-
 import { bold, errorColor, log, success as successColor } from '../ui.js';
 import { loadConfig, resolveConfigPath, sanitize } from '../utils.js';
 
@@ -12,6 +10,7 @@ const TAG = 'Drift';
 // ─── Main command ───────────────────────────────────────
 
 export async function driftCommand(): Promise<void> {
+  const { detectDrift, readAllLessons, TotemError } = await import('@mmnto/totem');
   const cwd = process.cwd();
   const configPath = resolveConfigPath(cwd);
   const config = await loadConfig(configPath);
--- a/packages/cli/src/commands/shield.ts
+++ b/packages/cli/src/commands/shield.ts
@@ -1,7 +1,6 @@
 import * as path from 'node:path';
 
 import type { ContentType, LanceStore, SearchResult } from '@mmnto/totem';
-import { TotemConfigError, TotemError } from '@mmnto/totem';
 
 import {
   extractChangedFiles,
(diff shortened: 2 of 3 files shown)


satur8d:
@coderabbitai @gemini-code-assist — Addressing all findings:

**Static imports (all 3 bots):** Fixed in `6fa2ff4`. Converted `@mmnto/totem` static imports to dynamic `await import()` inside command functions for shield.ts, drift.ts, and wrap.ts.

**Manifest attestation CI failure:** Root cause found and fixed — Prettier was reformatting `compiled-rules.json` after `totem compile`, invalidating the hash. Added both files to `.prettierignore`. Manifest regenerated. This should be the last time we see this issue.

PR #990, opened 2026-03-26
fix: harden phase-gate hooks + graceful AST query degradation

### LINKED ISSUE #986: fix: harden phase-gate hooks — block (not warn) on missing preflight
## Context

PR #982 shipped with 5 spec gaps because the Claude agent skipped `/preflight`. Root cause: `.claude/hooks/pre-push-check.sh` has two loopholes:

1. **`fix/*` branches are exempted** (line 16) — the agent was on `fix/982-spec-gaps` and got a free pass
2. **Gate 1 is a warning, not a block** — even on non-exempt branches, the hook `echo`s a warning and `exit 0`, which agents ignore

### Layer 1: Claude hook (`.claude/hooks/pre-push-check.sh`)
- Remove `fix/*` from exempt branch list (only `hotfix/*` should bypass)
- Change Gate 1 from warning (`echo` + `exit 0`) to hard block (`exit 2`)
- The only legit exemptions: `main`, `master`, `hotfix/*`, `docs/*`, detached HEAD

### Layer 2: Gemini hook (`.gemini/hooks/BeforeTool.js`)
- Add preflight gate check (currently only does shield)
- Fix `--deterministic` flag (deprecated, should use `lint`)

### Layer 3: Session startup (future)
- Claude needs equivalent of Gemini's `SessionStart.js` that runs `totem briefing`
- `totem briefing` output should include current milestone + required agent protocols
- This establishes the "why" on Turn 1 so hard blocks aren't bypassed via `touch .totem/cache/.spec-completed`

## Upstream
- ADR-063 (Context Engineering Principles — Phase-Gate Enforcement)
- ADR-056 (Spec Output with Inline Invariant Injection)
- PR #982 / #985 (where the gaps were discovered)

### LINKED ISSUE #988: fix: AST query crash on .js files — Bad node name 'required_parameter'
## Context

`totem lint` crashes with `AST query failed: Bad node name 'required_parameter'` when processing `.js` files (e.g., `.gemini/hooks/BeforeTool.js`). The error comes from a tree-sitter S-expression query that uses TypeScript-specific node names against a JavaScript file.

The crash prevents lint from completing — it exits with code 1 and no verdict output.

## Reproduction

```bash
# Any diff that includes a .js file
totem lint  # → [Totem Error] AST query failed: Bad node name 'required_parameter'
```

## Fix

Either:
1. Skip AST rules on `.js` files when the query uses TS-specific node names
2. Catch the error per-rule and continue (graceful degradation, per #855)
3. Validate AST queries against the target language before execution

## Files

- `packages/core/src/rule-engine.ts` — `applyAstRulesToAdditions`
- The specific AST rule with `required_parameter` in its S-expression

satur8d:
## Summary
- **Claude hook**: Remove `fix/*` from exempt branches, convert Gate 1 from warning to hard block (`exit 2`). Only `hotfix/*`, `docs/*`, `main`, `master` bypass.
- **Gemini hook**: Add Gate 1 (block commit without spec), fix regex anchoring for branch names, keep `totem shield` for push gate (was `--deterministic`), increase timeout to 120s.
- **AST engine (#988)**: Per-query error handling in `matchAstQueriesBatch` — invalid queries warn via `onWarn` instead of crashing all of lint. Prevents TS-specific node names from blocking `.js` file diffs.
- **Docs**: workflow-automation.md updated to reflect hard block status.

Closes #986, Closes #988

## Test plan
- [x] All tests passing (710 core + 1,123 cli + 38 mcp)
- [x] `totem shield` — PASS (0 findings)
- [x] `totem lint` — PASS (2 warnings, 0 errors — down from 51 before AST fix)
- [x] `pnpm format:check` — clean
- [ ] CI green on all three platforms

### DIFF
--- a/.claude/hooks/pre-push-check.sh
+++ b/.claude/hooks/pre-push-check.sh
@@ -7,7 +7,7 @@ TOOL_INPUT=$(cat)
 COMMAND=$(echo "$TOOL_INPUT" | grep -o '"command":"[^"]*"' | head -1 | sed 's/"command":"//;s/"//')
 
 # ─── Gate 1: Spec before commit (hard block) ──
-if [[ "$COMMAND" == *"commit"* && "$COMMAND" == *"git"* ]]; then
+if [[ "$COMMAND" =~ (^|[[:space:]])git[[:space:]].*commit ]]; then
   SPEC_FLAG=".totem/cache/.spec-completed"
   BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null)
 
@@ -25,7 +25,7 @@ if [[ "$COMMAND" == *"commit"* && "$COMMAND" == *"git"* ]]; then
 fi
 
 # ─── Gate 2: Shield before push (hard block) ──
-if [[ "$COMMAND" == *"push"* && "$COMMAND" == *"git"* ]]; then
+if [[ "$COMMAND" =~ (^|[[:space:]])git[[:space:]].*push ]]; then
   SHIELD_FLAG=".totem/cache/.shield-passed"
 
   if [ ! -f "$SHIELD_FLAG" ]; then


satur8d:
@gemini-code-assist Addressing review findings:

1. **JSON parsing robustness** (high): Deferred to #991. `jq` availability isn't guaranteed across environments — need to evaluate fallback strategy before switching from grep/sed.

Issue #1007, opened 2026-03-26
fix: github-cli-pr.ts missing GH_PROMPT_DISABLED on direct gh invocation


satur8d:
## Context

CR review on PR #1003 flagged that `github-cli-pr.ts:getRepoNwo()` calls `safeExec('gh', ...)` without passing `env: { ...process.env, GH_PROMPT_DISABLED: '1' }`. This means `gh` could prompt for auth interactively, which hangs in CI/agent contexts.

Other gh calls go through `ghExecOptions()` which includes this env var, but this direct call bypasses it.

## Fix
Add `env: { ...process.env, GH_PROMPT_DISABLED: '1' }` to the `safeExec` options in `getRepoNwo()`.

## Files
- `packages/cli/src/adapters/github-cli-pr.ts`

## Upstream
- CR review on PR #1003 [3]

Issue #1005, opened 2026-03-26
fix: gh-utils error unwrapping doesn't match safeExec error chain structure


satur8d:
## Context

CR review on PR #1003 flagged that `gh-utils.ts:handleGhError` unwraps `err.cause` directly, but `safeExec()` puts stderr on the wrapper error message, not on `cause.stderr`. The error chain structure changed when we migrated to `safeExec` — the unwrapping logic needs to match.

## Fix
Review how `safeExec` structures its errors (`Command failed: ...` with stderr appended) and update `handleGhError` to extract the relevant info from the right level of the cause chain.

## Files
- `packages/cli/src/adapters/gh-utils.ts`

## Upstream
- #995 (safeExec migration)
- CR review on PR #1003 [2]

Issue #991, opened 2026-03-26
chore: use jq for JSON parsing in pre-push-check.sh hook


satur8d:
## Context

GCA review on PR #990 flagged that `.claude/hooks/pre-push-check.sh` parses JSON input with `grep`/`sed`, which breaks if the command string contains double quotes (e.g., commit messages with quotes).

## Current
```bash
COMMAND=$(echo "$TOOL_INPUT" | grep -o '"command":"[^"]*"' | head -1 | sed 's/"command":"//;s/"//')
```

## Fix
```bash
COMMAND=$(echo "$TOOL_INPUT" | jq -r '.command // empty')
```

## Consideration
`jq` may not be available in all environments. Need to either:
1. Add a `jq` availability check with fallback to current grep/sed
2. Document `jq` as a requirement for the hook
3. Use a Node.js wrapper instead of bash for more robust parsing

## Upstream
- #986 (phase-gate hardening)
- GCA review on PR #990

Issue #989, opened 2026-03-26
fix: resolve .spec-completed path relative to git root, not cwd


satur8d:
## Context

Shield review on PR #986 flagged an INFO finding: `.gemini/hooks/BeforeTool.js` checks `fs.existsSync('.totem/cache/.spec-completed')` using a relative path. If an agent runs `git commit` from a subdirectory, this would be a false positive (file exists at repo root but not at cwd).

## Fix

Resolve the path relative to the git root:
```javascript
const gitRoot = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
const specFlag = path.join(gitRoot, '.totem/cache/.spec-completed');
```

Same fix should apply to `.claude/hooks/pre-push-check.sh` which uses the same relative path pattern.

## Files
- `.gemini/hooks/BeforeTool.js`
- `.claude/hooks/pre-push-check.sh`

## Upstream
- #986 (phase-gate hardening)
- Shield review finding on PR for #986

PR #1022, opened 2026-03-27
fix: bug bundle — safeExec, gh-utils, GH_PROMPT, git root paths, jq hook

### LINKED ISSUE #1006: fix: safeExec result.trim() crashes if stdio overridden to non-pipe mode
## Context

CR review on PR #1003 flagged that `safeExec` calls `result.trim()` assuming a string return, but `execFileSync` can return `Buffer` or `null` when `stdio` is overridden (e.g., `stdio: 'inherit'`). Currently `safeExec` allows `stdio` as an option, so a caller could trigger this.

## Fix
Either:
1. Guard `result.trim()` — check type before trimming
2. Remove `stdio` from `SafeExecOptions` — force pipe mode (safer, prevents MCP corruption too)
3. Cast/handle the Buffer case explicitly

Option 2 is probably safest — `safeExec` should always capture output.

## Files
- `packages/core/src/sys/exec.ts`

## Upstream
- CR review on PR #1003 [7]

satur8d:
## Summary
Five independent bug fixes bundled into one PR:

- **#1006**: Remove `stdio` from `SafeExecOptions`, hardcode pipe mode, add type guard on `trim()`
- **#1005**: Update `handleGhError` to match safeExec error chain structure (check wrapper + cause)
- **#1007**: Add `GH_PROMPT_DISABLED` env to `getRepoNwo()` direct `gh` call
- **#989**: Resolve `.spec-completed` and `.shield-passed` from git root, not cwd
- **#991**: Use `jq` for JSON parsing in pre-push hook with `grep`/`sed` fallback

Closes #1006, #1005, #1007, #989, #991

## Test plan
- [x] 1,934 tests passing (2 new)
- [x] totem lint: 344 rules, 0 errors
- [x] totem shield: PASS (0 findings)
- [x] format:check passing

### DIFF
--- a/packages/core/src/sys/exec.ts
+++ b/packages/core/src/sys/exec.ts
@@ -46,7 +46,8 @@ export function safeExec(
       input: rest.input,
     });
 
-    return shouldTrim ? (typeof result === 'string' ? result.trim() : '') : (result as string);
+    const output = typeof result === 'string' ? result : '';
+    return shouldTrim ? output.trim() : output;
   } catch (err) {
     const message = err instanceof Error ? err.message : String(err);
     // Extract stderr from child process error if available


satur8d:
@gemini-code-assist

1. **Unsafe `as string` cast** (exec.ts:49) — Fixed. Replaced `(result as string)` with a proper type guard: `typeof result === "string" ? result : ""` applied to both trim and no-trim paths. Pushed in latest commit.

PR #1157, opened 2026-04-02
fix: pre-push format check + manifest rehash after observation capture

### LINKED ISSUE #1155: bug: totem review observation capture silently mutates compiled-rules.json
## Problem

`totem review` (Pipeline 5) captures observation rules and writes them to `compiled-rules.json`. This silently mutates the file after `totem compile` has set the compile manifest hash, causing the Compile Manifest Attestation (CMA) check to fail in CI.

The pre-push hook runs `verify-manifest` against the **working tree**, which matches because the mutation happened locally. But CI checks the **committed** state — if the developer committed the manifest but not the mutated `compiled-rules.json` (or vice versa), CMA fails.

## Reproduction

1. Run `totem compile` → manifest hash set
2. Run `totem review` → Pipeline 5 captures observation rules, reorders JSON properties in `compiled-rules.json`
3. Commit manifest only → push → CMA fails because committed `compiled-rules.json` hash doesn't match manifest

## Proposed fix

Options:
- **A:** `totem review` should not mutate `compiled-rules.json` directly — write observation rules to a separate file (e.g., `.totem/cache/observations.json`)
- **B:** After observation capture, automatically re-hash and update the manifest
- **C:** Make `verify-manifest` check the git index (committed state), not the working tree

Discovered during #1039 / PR #1154.

### LINKED ISSUE #1156: fix: pre-push hook should run format:check before push
## Problem

The pre-push hook runs `totem lint` + `totem verify-manifest` + tests, but does **not** run `pnpm format:check`. CI runs `prettier --check .` and fails when docs or config files are committed unformatted.

This is easy to hit when committing non-code files (docs, JSON) that aren't auto-formatted by the pre-commit hook (which only formats staged `.ts` files).

## Proposed fix

Add `pnpm format:check` to the pre-push hook, or to the `/prepush` skill's checklist. This is a fast check (~2s) and would catch formatting issues before they hit CI.

Discovered during #1039 / PR #1154 — `docs/active_work.md` was committed unformatted and failed CI on all 3 platforms.

satur8d:
## Summary

Closes #1155, closes #1156

Two workflow bugs that caused CI failures on every PR:

**1. Format check in pre-push hook (#1156)**
The pre-commit hook only auto-formats staged `.ts` files, so non-code files (docs, JSON, YAML) could be committed unformatted. CI's `prettier --check` caught them but the developer had no local signal. Now the pre-push hook runs `format:check` — but only if the project defines that script (respects "no workflow opinions" principle).

**2. Manifest rehash after observation capture (#1155)**
`totem review`'s Pipeline 5 captures observation rules and writes them to `compiled-rules.json`, which changes the file hash. The compile manifest's `output_hash` then mismatches, causing CMA to fail in CI. Now the manifest is updated in-place after observation capture.

## Test plan

- [x] New test: `buildPrePushHook includes format:check`
- [x] New test: `captureObservationRules updates compile manifest hash`
- [x] Updated `extractTotemBlock` helper for 3-block hook format
- [x] 1,522 tests pass
- [x] `totem lint` — PASS
- [x] `totem review` — PASS

### DIFF
--- a/packages/cli/src/commands/install-hooks.ts
+++ b/packages/cli/src/commands/install-hooks.ts
@@ -338,12 +338,24 @@ fi
 
 # Format check — catch unformatted files before CI does
 # Only runs if the project defines a format:check script (no workflow opinions)
-if [ -f "package.json" ] && command -v pnpm >/dev/null 2>&1; then
-  if node -e "const p=require('./package.json'); process.exit(p.scripts && p.scripts['format:check'] ? 0 : 1)" 2>/dev/null; then
-    if pnpm run format:check > /dev/null 2>&1; then
+# Detects package manager from lockfile presence
+if [ -f "package.json" ]; then
+  FORMAT_CMD=""
+  if [ -f "pnpm-lock.yaml" ] && command -v pnpm >/dev/null 2>&1; then
+    FORMAT_CMD="pnpm run"
+  elif [ -f "yarn.lock" ] && command -v yarn >/dev/null 2>&1; then
+    FORMAT_CMD="yarn run"
+  elif [ -f "bun.lockb" ] || [ -f "bun.lock" ]; then
+    command -v bun >/dev/null 2>&1 && FORMAT_CMD="bun run"
+  elif command -v npm >/dev/null 2>&1; then
+    FORMAT_CMD="npm run"
+  fi
+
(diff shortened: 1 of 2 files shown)


satur8d:
Addressing CodeRabbit comment:

**install-hooks.ts:350 (pnpm-only format check)** — ✅ Fixed. Now detects package manager from lockfile presence: pnpm-lock.yaml → pnpm, yarn.lock → yarn, bun.lockb/bun.lock → bun, fallback → npm. Error message also reflects the detected manager.

Issue #1131, opened 2026-03-31
feat: rule refinement diagnostic via context telemetry


satur8d:
## Context

Follow-up from #1132 (context telemetry in rule metrics, shipped via #1227). Now that `RuleMetric.contextCounts` tracks code/string/comment/regex match distribution, we can automatically flag regex rules whose matches predominantly land in non-code contexts — a high-signal indicator the pattern is too broad and should be upgraded to `ast-grep`.

**This supersedes the original GHAS-alert-driven proposal.** Internal telemetry is a more immediate and universal signal than external scan dismissals. Scan alert extraction has moved to **Strategy #50**.

### 1. `totem doctor` — upgrade candidate detection

Add `checkUpgradeCandidates()` to `doctor.ts`:
- Load `RuleMetric` entries via `loadRuleMetrics()`
- Filter to rules where `engine === 'regex' || engine === 'ast'` (skip already-structural)
- Skip rules where `contextCounts` is undefined (no telemetry yet)
- Skip rules where `sum(contextCounts) < MIN_CONTEXT_EVENTS` (default 5)
- Compute `nonCodeRatio = (sum - contextCounts.code) / sum`
- Flag rules where `nonCodeRatio > NON_CODE_THRESHOLD` (default 0.2)
- Report as a `DiagnosticResult` with remediation: `totem compile --upgrade <hash>`

### 2. `totem compile --upgrade <hash>`

Add `upgrade?: string` to `CompileOptions`:
- Filter `lessons` to the matching `lessonHash`
- Bypass the `existingByHash` cache check (like `--force`)
- Thread a telemetry-prefixed system prompt addendum into `compileLesson`:
  > "This rule was flagged because N% of its matches occur in strings/comments/regex literals. Please prefer an `ast-grep` structural pattern that only matches code."
- Route through Claude Sonnet (already the default per `totem.config.ts`)

### 3. `--pr` self-healing extension

Follow the existing downgrade phase precedent in `runSelfHealing()` — add an upgrade-recommendation phase gated behind `--pr` that can auto-run `compile --upgrade` on flagged rules and open a PR with the diff. Read-only by default.

## Constraints

- **ADR-027:** Never auto-apply. `doctor` reports read-only; `compile --upgrade` only runs when explicitly invoked; `--pr` requires explicit flag.
- **Semantic trap:** Use `contextCounts.code` for the authoritative violation count, NOT `triggerCount` (which includes all matches including strings/comments).
[...]

PR #1223, opened 2026-04-06
feat: upgrade 8 empty-catch rules to ast-grep + 7 new Sonnet compilations

### LINKED ISSUE #664: feat: AST-based empty catch detection (multi-line support)
## Summary
The regex-based empty catch rule only catches single-line `catch(e) {}`. Multi-line empty catches (`catch {\n}`) are missed because regex matching is line-scoped.

## Context
GCA suggested using the Tier 2 AST engine for this (#663 review). Tree-sitter can detect `statement_block` nodes with zero children, but the S-expression predicate `#eq?` only matches exact text. Need either:
- A custom predicate for `namedChildCount === 0`
- ast-grep integration (supports `has` operator natively)

## Depends on
- ast-grep migration proposal (in progress)

## Acceptance Criteria
- [ ] Catches multi-line empty catch blocks
- [ ] Does not flag catch blocks with comments (intentionally empty)
- [ ] Works across TS/JS/TSX/JSX

satur8d:
[UPDATE-RULES]

## Summary
- Upgrade 7 empty-catch rules from `ast` engine (`#eq?` predicate) to `ast-grep`
- The Tree-sitter `#eq?` only matches literal `{}` — misses multi-line empty catches
- ast-grep pattern `try { $$$BODY } catch ($ERR) { }` handles both structurally
- Restored 1 mismatched rule (491df4448463dd87) that was incorrectly caught in bulk upgrade
- Sonnet recompile picked up 7 additional rules Gemini couldn't compile (438 total)
- Updates `docs/active_work.md` for 1.13.0 milestone

Closes #664

## Test plan
- [x] Wind tunnel: 11 rule tests pass, 61 TODO fixtures correctly skipped
- [x] Full suite: 2571 tests pass
- [x] 438 rules, 0 lint violations
- [x] Manifest verified locally

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.totem/compile-manifest.json
+++ b/.totem/compile-manifest.json
@@ -1,7 +1,7 @@
 {
-  "compiled_at": "2026-04-06T00:25:13.873Z",
+  "compiled_at": "2026-04-06T00:40:46.246Z",
   "model": "gemini-3-flash-preview",
   "input_hash": "da5fb64ebe3fbfbf72e2cc841ed4b5550eab3f39df8107be622dda6982a876cf",
-  "output_hash": "6066814e1ac7419f9133080ed25ac7ce5202c141a7d6eb84e72dbe20ccd6c6db",
+  "output_hash": "ba841f768ea1194b0cf7dc2bf99cd93f0b319d5c63efb631ae49cff1a25d04ea",
   "rule_count": 438
 }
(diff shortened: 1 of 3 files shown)


satur8d:
Addressed in e26665ae:

**@gemini-code-assist**
1. **Mismatched re-throw rule (491df4448463dd87)** — restored to original AST engine. My bulk upgrade filter was too broad (matched all `#eq?` in `catch_clause`, but this one's `#eq?` was on the constructor name, not the block body). Good catch.
2. **Overly broad new compilations** — these are Sonnet's first pass. Refinement is tracked in #1131.

**@coderabbitai**
1. **Parameterless `catch {}`** — valid point. The pattern `catch ($ERR) { }` won't match ES2019 `catch {}`. Our codebase uses TypeScript strict mode with typed bindings, so this is a low-risk gap. Filed as a follow-up for the ast-grep YAML rule format which supports `any` combinators.
2. **Rule count 431 → 438** — fixed in active_work.md.

PR #1234, opened 2026-04-06
feat: rule refinement diagnostic via context telemetry

satur8d:
## Summary

Closes mmnto/totem#1131. Last ticket for **1.13.0 — The Refinement Engine**.

Adds a telemetry-driven rule refinement loop: `totem doctor` now flags regex/ast rules whose matches predominantly land in non-code contexts (strings, comments, regex literals), and a new `totem compile --upgrade <hash>` flow re-compiles a specific rule through Claude Sonnet with a non-code-ratio directive that nudges it toward a structural `ast-grep` pattern.

### Phase 1 — `totem doctor` upgrade candidate detection

- New `findUpgradeCandidates(cwd, totemDir)` pure helper reads `RuleMetric.contextCounts` (not `triggerCount` — see the semantic trap note in `packages/core/src/rule-metrics.ts`) and joins against `compiled-rules.json`
- New `checkUpgradeCandidates()` diagnostic returns `warn` when any rules exceed `NON_CODE_THRESHOLD = 0.2` with at least `MIN_CONTEXT_EVENTS = 5` matches
- Remediation line points users at `totem compile --upgrade <hash>`
- Skips already-structural `ast-grep` rules, skips rules without telemetry, skips low-confidence (<5 events)

### Phase 2 — `totem compile --upgrade <hash>`

- New `upgrade?: string` flag on `CompileOptions`, wired into both `program.command('compile')` and `lessonCmd.command('compile')`
- Supports full hash or short prefix match (ambiguous prefixes throw `UPGRADE_HASH_AMBIGUOUS`)
- **Scoped cache eviction** — deletes the target from `existingByHash` and `nonCompilableSet` so it alone re-compiles; all 389 other rules pass through unchanged
- Also filters the stale target copy out of `newRules` before the compile loop to prevent duplicate-hash entries on save
- New exported `buildTelemetryPrefix(contextCounts)` helper builds a directive string: _\"This rule was flagged because N% of its matches occur in non-code contexts (strings: X, comments: Y, regex literals: Z). Please prefer an ast-grep structural pattern...\"_
- Telemetry prefix is applied **per-lesson** in the batch map, so it only affects the `--upgrade` target — not every lesson being compiled
- Rejects `--upgrade` + `--cloud` with `UPGRADE_CLOUD_UNSUPPORTED` (cloud worker can't thread per-lesson prompts today — refs mmnto/totem#1221)
- Only affects Pipeline 2 (Sonnet). Pipeline 1 (manual `**Pattern:**`) and Pipeline 3 (example-based) are untouched

### Phase 3 — `runSelfHealing` upgrade extension

- New upgrade phase slots after the existing downgrade + GC phases in `totem doctor --pr`
- Calls `compileCommand({ upgrade })` in-process (avoids hardcoding `pnpm exec`, works for any package manager / global install)
- Bundles upgraded rules into the existing branch + commit + PR flow alongside downgrades and archives
- PR title becomes `chore(doctor): upgrade N rule(s) to ast-grep via telemetry diagnostic` when only upgrades are present
- PR body gets a new `### Upgrades` section listing each rule with hash and non-code ratio

## Data-loss bug caught by `totem review`

The first-pass implementation truncated the `lessons` array to 1 and set `options.force = true` globally, which would have caused `compile --upgrade <hash>` to write a `compiled-rules.json` containing only the upgraded rule — **deleting the other 389**. `totem review` caught it. Fixed by:

1. Removing the truncation
2. Removing `options.force = true`
3. Using scoped `existingByHash.delete(upgradeTargetHash)` + `nonCompilableSet.delete(upgradeTargetHash)` instead
4. Filtering the stale target out of `newRules` before compile
5. Moving `telemetryPrefix` out of shared `coreDeps` and into per-lesson deps
6. Adding the `UPGRADE_CLOUD_UNSUPPORTED` error for the one path where per-lesson prompt threading isn't possible yet

## Test plan

- [x] `pnpm run format` — clean
- [x] `pnpm build` — 3 packages successful
- [x] `pnpm test --filter @mmnto/cli --filter @mmnto/totem` — 1581 passing (+1 for cloud rejection test)
- [x] `pnpm exec totem lint` — PASS, 0 errors, 136 warnings (all pre-existing or test-file false positives)
- [x] `pnpm exec totem verify-manifest` — PASS, 389 rules, hashes match
- [x] `pnpm exec totem review --no-auto-capture` — **PASS** (data-loss CRITICAL resolved; one WARN about `cwd` threading tracked as mmnto/totem#1232)

## Follow-ups filed

- mmnto/totem#1232 — Thread explicit `cwd` through `compileCommand` (flagged by review, pre-existing pattern, not a #1131 regression)
- mmnto/totem#1233 — Stray `packages/core/{}` file created during `pnpm build` (bash brace-expansion accident, cosmetic)

## Stats

\`\`\`
 docs/active_work.md                               |  21 +
 packages/cli/src/commands/compile-upgrade.test.ts | 151 +++++ (new)
 packages/cli/src/commands/compile.ts              | 129 +++++
 packages/cli/src/commands/doctor.test.ts          | 310 +++++++-
 packages/cli/src/commands/doctor.ts               | 193 ++++++
 packages/cli/src/index.ts                         |  10 +
 packages/core/src/compile-lesson.test.ts          |  40 +
 packages/core/src/compile-lesson.ts               |  16 +-
 packages/core/src/errors.ts                       |   5 +-
 9 files changed, 855 insertions(+), 20 deletions(-)
\`\`\`

After this lands, **1.13.0 — The Refinement Engine** is feature-complete and ready for the docs/wiki/playground/binary refresh pass before publish.

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -475,18 +475,26 @@ export async function findUpgradeCandidates(
       // Exclude `unknown` from both numerator and denominator — it represents
       // historical / unclassified telemetry (pre-context-aware hits, or seeding
       // via `triggerCount - 1`) and is not evidence of non-code leakage.
+      // The `?? 0` defaults are defensive: the Zod schema at
+      // packages/core/src/rule-metrics.ts declares every contextCounts field
+      // as a non-negative integer, but a hand-edited rule-metrics.json could
+      // bypass that and produce NaN in the arithmetic below.
       const cc = metric.contextCounts;
-      const classifiedTotal = cc.code + cc.string + cc.comment + cc.regex;
+      const code = cc.code ?? 0;
+      const strings = cc.string ?? 0;
+      const comments = cc.comment ?? 0;
+      const regexes = cc.regex ?? 0;
+      const classifiedTotal = code + strings + comments + regexes;
       if (classifiedTotal < MIN_CONTEXT_EVENTS) continue;
 
-      const nonCodeRatio = (cc.string + cc.comment + cc.regex) / classifiedTotal;
+      const nonCodeRatio = (strings + comments + regexes) / classifiedTotal;
       if (nonCodeRatio > NON_CODE_THRESHOLD) {
         candidates.push({
           lessonHash: rule.lessonHash,
           heading: rule.lessonHeading ?? rule.lessonHash,
-          engine: rule.engine,
+          engine: 'regex',
           total: classifiedTotal,
-          codeCount: cc.code,
+          codeCount: code,
           nonCodeRatio,
         });
       }
@@ -609,7 +617,12 @@ export const MIN_CONTEXT_EVENTS = 5;
 export interface UpgradeCandidate {
   lessonHash: string;
   heading: string;
-  engine: 'regex' | 'ast';
+  /**
+   * Always `'regex'` — `findUpgradeCandidates` filters to regex rules only
+   * because only they carry trustworthy non-code telemetry. Narrowed from
+   * the broader engine union so the type matches the implementation.
+   */
+  engine: 'regex';
   total: number;
   codeCount: number;
   nonCodeRatio: number;
@@ -729,7 +742,8 @@ export async function runSelfHealing(cwd: string): Promise<void> {
   if (gcConfig && gcConfig.enabled !== false && !gitDirty) {
     console.error(`\n${pc.cyan('[Auto-Healing]')} Checking for stale rules to archive...`);
 
-    const fs = await import('node:fs');
+    // fs is statically imported at the top of this file (line 2); no need to
+    // re-import dynamically here (mmnto/totem#1234 CR cleanup).
     if (!fs.existsSync(rulesPath)) {
       console.error(pc.dim('  No compiled-rules.json found. Skipping GC.'));
     } else {


satur8d:
@gemini-code-assist thanks for the second-round review — all four findings addressed in b1430181:

**1. `existingByHash.delete(upgradeTargetHash)` loses `createdAt` (compile.ts:362)** — **fixed**. You were right: core's `buildCompiledRule` / `buildManualRule` read the existing entry from that map to preserve metadata. Deleting made upgraded rules look brand-new and broke GC / audit trails. Switched to a loop-level cache bypass: when `hash === upgradeTargetHash`, the `toCompile` construction skips the `existingByHash.has` / `nonCompilableSet.has` checks without mutating the map. The target falls through to re-compile while its old entry stays in `existingByHash` so core can copy `createdAt` onto the new rule.

**2. Stale-copy removal nested inside `case 'compiled':` (compile.ts:628)** — **fixed**. Correct catch — if the LLM decided the upgrade target was non-compilable (`skipped` status), the old rule persisted in `newRules` AND the hash was added to `nonCompilableSet`, creating an inconsistent state. Moved the splice-out logic so it runs for both terminal outcomes (`compiled` and `skipped`) but NOT for `failed` (transient errors) or `noop` — so a flaky network / rate-limit never silently deletes work. `compiled` also clears any stale `nonCompilableSet` entry for the target.

**3. Fatal error check doesn't use log.error pattern (doctor.ts:628)** — **resolved by removal**. I pulled the upfront `gh --version` check entirely. The existing `try/catch` around `gh pr create` at doctor.ts:914-925 already prints a helpful `Failed to create PR: ... Push manually with: git push -u origin ...` message, which is the graceful early-exit behavior you asked for — just at the end of the flow instead of the beginning. Removing the upfront check also fixed the Windows CI regression where the three `runSelfHealing` doctor tests were failing because Windows runners don't have `gh` in PATH. The diagnostic + downgrade + upgrade work is still valuable on a machine without `gh` installed — the user just can't auto-open a PR — so aborting up front was worse UX than letting the work complete and failing gracefully at the PR step.

**4. Loop calls `compileCommand` per candidate (doctor.ts:787)** — **still deferred to mmnto-ai/totem#1235**. Same reasoning as the first-round reply: perf concern bounded at expected N (≤10 candidates), and the batching refactor (array-of-hashes on `CompileOptions`) is a structural change that warrants its own ticket so it can be scoped and reviewed independently of this PR.

Refs mmnto-ai/totem#1131 / mmnto-ai/totem#1234.

Issue #1329, opened 2026-04-11
safeExec leaks shell metacharacters to cmd.exe on Windows


satur8d:
## Root cause (discovered during #1233)

`packages/core/src/sys/exec.ts:40` sets `shell: IS_WIN` so that \`.cmd\`/\`.bat\` shims (most notably \`git.cmd\`) can be resolved without manual extension handling. But \`shell: true\` on Windows means Node passes the constructed command line to \`cmd.exe\` **without escaping arguments for shell metacharacters**.

The caller is expected to pre-escape — but nothing in the signature or JSDoc flags that requirement, and every callsite in the codebase passes raw arguments.

## How #1233 surfaced this

\`packages/core/src/sys/exec.test.ts\` was passing this argument to \`node -e\`:

\`\`\`ts
safeExec('node', ['-e', 'setTimeout(() => {}, 30000)'], { timeout: 100 });
\`\`\`

cmd.exe parsed \`=>\` as \`=\` followed by \`>\` (output redirection), then redirected stdout to a literal file named \`{}\` in the cwd. Result: a 0-byte \`packages/core/{}\` file appeared on every test run.

The test has been rewritten to avoid shell metacharacters, but the underlying correctness problem remains: **\`safeExec\` is only safe when callers happen to pass arguments without shell special characters**.

## Impact

- **Correctness:** any caller passing a string with \`<\`, \`>\`, \`&\`, \`|\`, \`^\`, \`(\`, \`)\`, \`%\`, \`!\`, \`\"\` on Windows will see cmd.exe reinterpret it.
- **Security:** if a caller ever interpolates untrusted input into an arg, it becomes a shell injection vector. There are no known user-controlled callsites today, but the foot-gun is live.
- **Cross-platform divergence:** the same \`safeExec\` call that works on macOS/Linux (no shell) may behave differently on Windows (shell interprets metacharacters).

## Fix options

1. **Stop using \`shell: true\` by default.** Detect \`.cmd\`/\`.bat\` shims explicitly and invoke via \`cmd /c\` with properly quoted args. More code but matches POSIX behavior.
2. **Escape args on the Windows path.** Wrap each arg in double quotes and double any internal quotes, per cmd.exe's argument-parsing rules. Known-brittle — cmd.exe's quoting rules are famously inconsistent.
3. **Switch to a library that already solves this.** \`cross-spawn\` is the standard answer here; it handles Windows shim resolution without \`shell: true\`.

[...]

PR #1330, opened 2026-04-11
fix(test): avoid shell metacharacters in safeExec timeout test (#1233)

### LINKED ISSUE #1233: bug: stray 'packages/core/{}' file created during pnpm build
## Symptom

Running \`pnpm build\` from the monorepo root reproducibly creates a 0-byte file at literal path \`packages/core/{}\` (braces in the filename). The file appears after every build and has to be deleted manually before committing.

Reproduced twice during #1131 work (2026-04-06). First observed by a background agent at ~10:20 local time.

\`\`\`
$ ls -la packages/core/{}
-rw-r--r-- 1 jmatt 197609 0 Apr  6 10:20 packages/core/{}
\`\`\`

## Suspected cause

Almost certainly a bash brace-expansion accident in a script. Most likely culprits:
- \`packages/core/package.json\` — a \`clean\` or \`prebuild\` script with something like \`rm -rf dist/**/{.tsbuildinfo,*.map}\` where the braces aren't quoted, causing the literal \`{}\` to be created on a shell that doesn't expand it
- Windows Git Bash interaction with tsc output
- A pre-commit / pre-build hook

## Impact

- Cosmetic — the file is 0 bytes and does nothing
- Annoying — shows up in \`git status\` as untracked noise before every commit
- Low priority — no functional impact on build output or tests

## Repro

1. \`git clean -fxd packages/core\` (to start from a known state)
2. \`pnpm build\` from repo root
3. \`ls packages/core/\` — observe the literal \`{}\` file

## Proposal

Grep \`packages/core/package.json\` + any turbo / tsc config for \`{\` in a script context; find the unquoted brace expression and fix it.

satur8d:
## Summary

Closes #1233. Refs #1329.

The reproducible 0-byte `packages/core/{}` file that #1233 tracked is caused by `exec.test.ts` passing `setTimeout(() => {}, 30000)` as a `node -e` argument. On Windows, `safeExec` runs with `shell: true` (to resolve `.cmd`/`.bat` shims), so cmd.exe re-parses the command line and splits the `=>` token into `=` + `>` (output redirection) — then redirects stdout to a file literally named `{}` in the cwd.

This replaces the arrow-function body with `setInterval(Object, 30000)`, which is equivalent for the test's purpose (long-running to prove the timeout triggers) but contains no shell metacharacters. An inline comment documents the pitfall so the test isn't "fixed" back into the broken state by a well-meaning future refactor.

## Scope note

This fixes the **symptom** — the stray file stops appearing. The **architectural issue** — `safeExec` leaks unescaped args to cmd.exe whenever `shell: true` is active on Windows, which is a potential shell-injection vector and a cross-platform behavior divergence — is filed separately as #1329. Deliberately kept out of scope here because the `safeExec` fix has wider blast radius and deserves its own PR with proper alternatives discussion (cross-spawn vs. manual escaping vs. explicit `.cmd` detection).

## Verification

Before the fix:
```
$ rm -f packages/core/\{\}
$ pnpm exec vitest run src/sys/exec.test.ts
  ✓ src/sys/exec.test.ts (7 tests) 229ms
$ ls packages/core/\{\}
packages/core/{}   # ← 0-byte file created
```

After the fix:
```
$ rm -f packages/core/\{\}
$ pnpm exec vitest run src/sys/exec.test.ts
  ✓ src/sys/exec.test.ts (7 tests) 229ms
$ ls packages/core/\{\}
ls: cannot access 'packages/core/{}': No such file or directory  # ← gone
```

## Noted lint warning

Pre-push lint flagged one warning on line 53 (catch block fallback comment rule). It's a false positive — the catch block intentionally swallows the timeout error because that *is* what the test is asserting. Leaving as-is; happy to rephrase if reviewers want.

## Test plan

- [x] `exec.test.ts` still passes (7/7)
- [x] No stray `{}` file after test run (verified locally)
- [x] Pre-push hook passed (lint + 2725 tests + compiled rules)
- [ ] CI green on Windows runner (real test for this fix)

### DIFF
--- a/packages/core/src/sys/exec.test.ts
+++ b/packages/core/src/sys/exec.test.ts
@@ -2,6 +2,14 @@ import { describe, expect, it } from 'vitest';
 
 import { safeExec } from './exec.js';
 
+// Long-running interval used to ensure the child process outlives the
+// timeout window. Any value safely larger than TIMEOUT_MS works; 30s is
+// generous enough that a slow CI runner won't race the check.
+const LONG_RUNNING_INTERVAL_MS = 30_000;
+// Timeout for the test that asserts safeExec honors its timeout option.
+// Short enough to keep the suite fast.
+const TIMEOUT_TEST_MS = 100;
+
 describe('safeExec', () => {
   it('executes a command and returns trimmed output', () => {
     // Use single quotes inside the JS expression — cmd.exe strips double quotes
@@ -50,7 +58,9 @@ describe('safeExec', () => {
       // shell: true to resolve .cmd/.bat shims, and cmd.exe would parse
       // the `=>` token in an arrow function as `=` + `>` output redirection,
       // creating a stray file named `{}` in the cwd. See mmnto/totem#1233.
-      safeExec('node', ['-e', 'setInterval(Object, 30000)'], { timeout: 100 });
+      safeExec('node', ['-e', `setInterval(Object, ${LONG_RUNNING_INTERVAL_MS})`], {
+        timeout: TIMEOUT_TEST_MS,
+      });
       expect.unreachable('should have thrown');
     } catch (err) {
       expect(err).toBeInstanceOf(Error);
--- a/packages/core/src/sys/exec.test.ts
+++ b/packages/core/src/sys/exec.test.ts
@@ -45,7 +45,12 @@ describe('safeExec', () => {
 
   it('respects timeout option', () => {
     try {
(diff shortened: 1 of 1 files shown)


satur8d:
@coderabbitai addressed the nit in `60fc5ba5` — extracted `LONG_RUNNING_INTERVAL_MS` and `TIMEOUT_TEST_MS` constants with short docstrings explaining what each value protects against. All 7 tests still pass.

Issue #1286, opened 2026-04-07
Track refinement of 6 over-broad lessons firing on test code (Refinement Engine accountability)


satur8d:
## Summary

Track the natural refinement of 6+ specific over-broad lessons that fire as false positives on test code (test assertions, test fixtures, jest expect calls). The premise of the 1.13.0 Refinement Engine is that `totem doctor --pr` will naturally flag rules with high non-code hit ratios as upgrade candidates. **This ticket exists to hold the system accountable to that loop**: if these lessons don't naturally float to the top of the doctor hit list by the end of 1.14.0, our bypass math is wrong.

## The 6+ noisy lessons

All discovered via PR #1278 and PR #1282 lint output, all firing on test files where they don't apply:

1. **`Avoid using partial matchers like toContain when testing`**
   - Fires on: `expect(result?.message).toContain('Hydration errors...')` in `lesson-pattern.test.ts`
   - Actual scope: secret-masking assertions
   - Current scope: any `toContain` in any test

2. **`Naive substring matching like includes('dir/') can trigger false positives`**
   - Fires on: `expect(warnings.some((w) => w.includes('lesson-empty.md')))` in `lesson-io.test.ts`
   - Actual scope: path-segment matching in production
   - Current scope: any `.includes()` in test assertions

3. **`Check that author field is not null/undefined before calling .includes()`**
   - Fires on: same assertions
   - Actual scope: third-party review metadata
   - Current scope: any `.includes()` regardless of context

4. **`When detecting markers in user-authored files, use line-start regex`**
   - Fires on: same assertions
   - Actual scope: parsing user-authored markdown
   - Current scope: any string-marker check

5. **`When identifying files by naming conventions (e.g., '.test.', '.spec.')`**
   - Fires on: `path.includes('lesson-empty.md')` test assertions
   - Actual scope: production code classifying files by convention
   - Current scope: any path string match

6. **`Using .includes() to identify target files like READMEs`**
   - Fires on: same path-string assertions
   - Actual scope: production code identifying canonical files
   - Current scope: any string compare with a filename

7. **`Detect existing hook managers and provide manual guidance`** — fires on `fs.writeFileSync(...)` in test fixture setup. Wildly off-target (lesson is about not writing to .git/hooks).

## The verification loop

Per the 1.13.0 Refinement Engine design:
1. `RuleMetric.contextCounts` tracks per-context match distribution
2. `totem doctor --pr` flags regex rules where >20% of matches land in non-code contexts
[...]

PR #1331, opened 2026-04-11
fix(cli): prune stale nonCompilable entries on no-op compile runs (#1281)

### LINKED ISSUE #1281: compile.ts: prune nonCompilable on no-op runs too
## Summary

`compile.ts` only drains stale `nonCompilable` entries when `toCompile.length > 0`. If a project's lessons are already fully compiled and there are stale entries in `nonCompilable` (e.g., from before a parser bug fix, or from a removed lesson that previously failed compilation), running `totem lesson compile` reports the stale entries as still present and never drains them.

## Reproduction

1. Pre-seed `.totem/compiled-rules.json` with stale entries:
```json
{
  "version": 1,
  "rules": [...],  // some already-compiled rule
  "nonCompilable": ["deadbeefdeadbeef", "feedfacefeedface"]
}
```
2. Run `totem lesson compile` (no edits to lessons, nothing needs recompiling)
3. **Expected**: stale entries are pruned (they don't match any current lesson hash)
4. **Actual**: 
```
[Compile] Found N lessons
[Compile] All N lesson(s) in scope already processed (N compiled, 2 non-compilable). Use --force to recompile.
```
   The post-run `compiled-rules.json` still contains both stale entries.

## Root cause

The prune logic at `packages/cli/src/commands/compile.ts:729` lives inside the `} else {` branch that executes only when `toCompile.length > 0`:

```ts
if (toCompile.length === 0) {
  log.success(TAG, `All ${lessonsInScope.length} lesson(s) in scope already processed (...).`);
} else {
  // ... compile loop ...
  
  // Prune stale non-compilable hashes (lesson was edited or removed)
  const freshNonCompilable = [...nonCompilableSet].filter((h) => currentHashes.has(h));
  saveCompiledRulesFile(rulesPath, {
    version: 1,
    rules: newRules,
    nonCompilable: freshNonCompilable,
  });
}
```

When everything is already compiled, the prune is skipped entirely. Stale entries persist until some future run actually has work to do.

## Impact

**Low.** The bug self-heals the moment any lesson needs work — edit any lesson, run compile, prune fires, stale entries drained. The stale-noop case is a niche scenario:
- Cleanup PRs that removed a lesson that had previously failed compilation
- Migration windows after a parser fix (like #1278) where old non-compilable entries from pre-fix runs need to age out

But it's a real correctness gap and a one-line fix.

## Suggested fix

Either:
1. **Hoist the prune outside the if/else**: always drain stale entries on every compile run, regardless of whether new work happened. The save would need to be conditional (skip if nothing changed in the rules array AND nothing changed in nonCompilable) to avoid touching the file unnecessarily.
2. **Add a "no-op cleanup" pass**: keep the current structure but add an explicit `if (toCompile.length === 0 && nonCompilableSet.size > 0)` block that runs the prune and saves if any entries got drained.

Option 1 is cleaner. The "save only if changed" guard is easy: compare `freshNonCompilable.length` to `existingFile.nonCompilable?.length ?? 0`, skip save if equal AND `newRules` is identical to `existingRules`.

## Discovered during

E2E reproduction of #1264 phantom bug against current `main` (post-#1278). See [the closing comment on #1264](https://github.com/mmnto-ai/totem/issues/1264) for the full session — I noticed this when seeding `nonCompilable` with two fake stale hashes and observing they survived a no-op compile run.

## Out of scope

- This is NOT the same bug as the now-closed #1264. That ticket described a "stale skip on edited lesson" symptom, which was a phantom of #1263 (the em-dash silent skip). #1278 fixed that. This ticket is a different, smaller correctness gap.
- The schema upgrade tracked at #1280 (`{hash, title}` tuples) is independent of this. Both can ship in either order.

## Acceptance criteria

- [ ] Stale `nonCompilable` entries are drained on every compile run, regardless of `toCompile.length`
- [ ] No-op runs (no lesson edits, no stale entries to drain) do not touch the file unnecessarily
- [ ] Test fixture: pre-seed stale entries, run compile with no lesson changes, verify entries are drained and file is rewritten exactly once

satur8d:
## Summary

Closes #1281.

`totem lesson compile` was only draining stale entries from the `nonCompilable` cache when `toCompile.length > 0`. On a no-op run (all lessons already compiled), stale entries left over from lessons that had been edited or removed in a previous run survived forever until some future run happened to have real work to do.

## Fix

Extract the prune logic into a pure helper `pruneStaleNonCompilable` and call it from both branches:

- **No-op branch** (`toCompile.length === 0`) now prunes and rewrites `compiled-rules.json` *only* if something was actually drained. A genuinely idle run (no lesson changes, no stale entries) still does not touch the file, matching the acceptance criterion in the ticket.
- **Active-compile branch** uses the same helper instead of its previous inline prune, keeping both code paths in lockstep.

The helper is pure and non-mutating — the six unit tests cover empty map, all-current, all-stale, partial drain, legacy title preservation, and explicit non-mutation of the input map.

## Shield finding addressed inline

Shield caught that the success log message was using `nonCompilableMap.size` (the pre-prune size), so it would report the stale count even after draining entries. Fixed by computing `reportedNonCompilable` after the prune and threading it through to the log line.

## Test plan

- [x] New unit tests: 6/6 pass (`compile-prune.test.ts`)
- [x] Full cli test suite: 1627/1627 pass
- [x] Shield PASS (second run after fixing the log-count finding)
- [x] Pre-push hook passed (format, lint, tests, compiled rules)
- [ ] CI green

## Noted lint warning

Pre-push lint flagged one warning: `orchestrator test suites should set an explicit 15s timeout`. This is a false positive — the test suite is for a pure synchronous function with no orchestrator involvement. Leaving as-is.

### DIFF
--- a/.changeset/pr-1281-compile-prune-noop.md
+++ b/.changeset/pr-1281-compile-prune-noop.md
@@ -1,0 +1,11 @@
+---
+'@mmnto/cli': patch
+---
+
+Prune stale `nonCompilable` entries on no-op compile runs (#1281)
+
+`totem lesson compile` was only draining stale entries from the `nonCompilable` cache when there was actual compile work to do. On a no-op run (all lessons already compiled), stale entries — left over from lessons that had been edited or removed in a previous run — survived forever until some future run happened to have real work.
+
+The prune logic is now extracted into a pure helper (`pruneStaleNonCompilable`) and called from both branches. The no-op path only rewrites `compiled-rules.json` when there's actually something to drain, so genuinely idle runs still don't touch the file.
+
+Closes #1281. Discovered during the #1264 E2E reproduction.
--- a/packages/cli/src/commands/compile-prune.test.ts
+++ b/packages/cli/src/commands/compile-prune.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from 'vitest';
+
+import { pruneStaleNonCompilable } from './compile.js';
+
+describe('pruneStaleNonCompilable', () => {
+  it('returns empty result when the map is empty', () => {
+    const result = pruneStaleNonCompilable(new Map(), new Set(['abc', 'def']));
+    expect(result.fresh).toEqual([]);
+    expect(result.drained).toBe(0);
+  });
+
+  it('returns all entries when every hash is still current', () => {
+    const map = new Map<string, string>([
+      ['abc', 'First lesson'],
+      ['def', 'Second lesson'],
+    ]);
+    const current = new Set(['abc', 'def']);
+
+    const result = pruneStaleNonCompilable(map, current);
+
+    expect(result.fresh).toEqual([
+      { hash: 'abc', title: 'First lesson' },
+      { hash: 'def', title: 'Second lesson' },
+    ]);
+    expect(result.drained).toBe(0);
+  });
+
+  it('drops entries whose hashes are no longer present in current lessons', () => {
+    const map = new Map<string, string>([
+      ['abc', 'Kept lesson'],
+      ['stale1', 'Removed lesson A'],
+      ['stale2', 'Removed lesson B'],
+    ]);
+    const current = new Set(['abc']);
+
+    const result = pruneStaleNonCompilable(map, current);
+
+    expect(result.fresh).toEqual([{ hash: 'abc', title: 'Kept lesson' }]);
+    expect(result.drained).toBe(2);
+  });
+
+  it('drains everything when no hashes are current', () => {
+    const map = new Map<string, string>([
+      ['stale1', 'Removed A'],
+      ['stale2', 'Removed B'],
+    ]);
+    const current = new Set<string>();
+
+    const result = pruneStaleNonCompilable(map, current);
+
+    expect(result.fresh).toEqual([]);
+    expect(result.drained).toBe(2);
+  });
+
+  it('preserves tuple shape including titles from legacy-normalized entries', () => {
+    // Legacy string-form entries get normalized to {hash, title: '(legacy entry)'}
+    // by the schema transform. The prune helper must preserve that title verbatim.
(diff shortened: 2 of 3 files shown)


satur8d:
@gemini-code-assist @coderabbitai all three review findings addressed in commit \`e67e7cad\`:

**GCA #1 (high) — prune stale rules too, not just nonCompilable.** Valid — same no-op stall pattern applied to compiled rules when a lesson is removed. Extracted a symmetrical \`pruneStaleRules\` helper with six unit tests covering empty / all-current / partial / all-stale / identity preservation / non-mutation. Wired into the no-op branch alongside \`pruneStaleNonCompilable\`. Both counts flow into the success-log message so reported state matches disk state.

**CR (major) — refresh compile manifest when no-op rewrites rules.** Valid and a real provenance hazard. Added the same manifest-refresh block from the active-compile branch to the no-op path — \`generateInputHash\` → \`generateOutputHash\` → \`writeCompileManifest\` → \`Manifest: xxx→yyy\` trace log. Now both branches leave the manifest consistent with on-disk state.

**GCA #2 (high) — log drained nonCompilable in active branch + \`totem-context:\` vs \`totem-ignore\`.** Valid. The active-compile branch was silently draining while the no-op branch logged — made it symmetric. For the styleguide point, I updated all suppressions I added in this PR to \`// totem-context: <reason>\`. Pre-existing \`// totem-ignore\` comments in the file are left untouched — that's a broader styleguide migration that belongs in its own PR, not this one.

Test counts: 1627 → 1633 (+6 new \`pruneStaleRules\` tests). Full cli test suite passes. Shield re-ran clean.

The two lint warnings on line 525 about \`await import('@mmnto/totem')\` are a known rule collision (#1286 territory) — one rule says "don't dynamic import outside entry points", a sibling rule says "use dynamic imports in command handlers". The code matches the second rule's guidance since it's inside a CLI command handler, and mirrors the pattern the active-compile branch has used for several releases.

PR #1332, opened 2026-04-11
fix(mcp): prevent double-heading when add_lesson body starts with canonical heading (#1284)

### LINKED ISSUE #1284: MCP add_lesson tool double-prepends the ## Lesson — heading
## Summary

The `mcp__totem-dev__add_lesson` MCP tool double-prepends a `## Lesson — ` heading when the input `lesson` text already starts with one. The result is a malformed lesson file with two duplicate `## Lesson —` lines, which the parser correctly identifies as TWO separate lessons in the same file.

## Reproduction (from PR #1282 dogfooding session, 2026-04-07)

Called `mcp__totem-dev__add_lesson` with a `lesson` parameter that started with the canonical totem heading format:

```markdown
## Lesson — Read-path schema changes break write-path invariants

When modifying parsing logic that produces a data structure...
```

Expected: file written with my heading, my body, and tags from `context_tags`.

Actual: file `lesson-400fed87.md` was written with **two** lesson headings:

```markdown
## Lesson — Lesson — Read-path schema changes break write-path  ← AUTO-GENERATED, truncated to 60 chars
**Tags:** architecture, parsing, compiler, schema-evolution, shield
## Lesson — Read-path schema changes break write-path invariants ← MY ORIGINAL HEADING
[my original body]
**Source:** mcp (added at 2026-04-07T20:48:20.005Z)              ← AUTO-GENERATED footer
```

The first heading is auto-generated by the MCP tool, derived from the first words of my body — including the literal `Lesson —` prefix. So instead of generating `## Lesson — Read-path schema changes...`, the auto-generation produced `## Lesson — Lesson — Read-path schema changes...` with the duplicated `Lesson —`.

The next compile run correctly identified TWO lessons in the file:

```
[Compile] Found 2 lessons        ← should have been 1
[Compile] [Read-path schema changes break write-path invariants] Not compilable — skipping
[Compile] [Lesson — Read-path schema changes break write-path] Not compilable — skipping
```

I corrected the file manually using the Write tool with the canonical single-heading format and re-ran compile, which then correctly produced 1 lesson.

## Suggested fix directions

1. **Detect existing heading and skip auto-prepend**: if the body already starts with a `## Lesson — ` line, skip the auto-generation entirely and use the body as-is. Simplest backward-compatible fix.
2. **Strip the leading heading from the body before auto-generating**: if the body starts with `## Lesson — Foo`, strip that line, derive the title from "Foo", then re-prepend with the canonical format.
3. **Document the contract clearly**: the tool's docstring should specify whether the `lesson` parameter expects "body only" or "full lesson with heading".

## Why it matters

This makes the MCP tool actively annoying to use for dogfooding. Every lesson I add via the tool needs manual cleanup. Defeats the point of having a programmatic add_lesson tool, and makes me question whether other MCP tools have similar silent-malformation behaviors.

## Acceptance criteria

- [ ] Calling `add_lesson` with a body that already starts with `## Lesson — Foo` produces a file with exactly one `## Lesson — Foo` heading
- [ ] Calling `add_lesson` with a body that does NOT start with a heading still produces a properly-formatted lesson file (existing behavior preserved)
- [ ] Test fixture: both shapes round-trip through the parser correctly (one lesson per call)

## Discovered during

PR #1282 (Pipeline 1 Message field + nonCompilable observability tuples), commit `918a4166`. See journal entry `2026-04-07-lesson-edit-feedback-loop.md` for the full context.

satur8d:
## Summary

Closes #1284.

When a caller passes a pre-formatted lesson to `add_lesson` whose body already starts with `## Lesson — Foo`, the tool was deriving a title from the first line of the body — which included the literal `Lesson —` prefix — and producing a file with `## Lesson — Lesson — Foo` as the wrapper heading, **with the original `## Lesson — Foo` still intact inside the body**. The parser then correctly counted that as two separate lessons.

## Fix

Detect a pre-existing canonical heading (em-dash, en-dash, or hyphen variants, consistent with the parser's `LESSON_HEADING_RE` in #1278), extract the title, and strip the heading line from the body before the wrap step. Callers who pass plain body text with no leading heading see unchanged behavior.

## Shield edge case caught and fixed inline

Shield caught that the initial regex required `\n+` at the end, which would fall through for a single-line input like `## Lesson — Foo` (no body, no trailing newline) and reproduce the bug. Added a test for that case and widened the terminator to `(?:\s*\n+|\s*$)`.

## `match` → `matchAll` refactor

Project-wide lint rule "use `matchAll()` instead of `match()` for security validations" flagged the initial use of `safeLesson.match(LESSON_HEADING_RE)`. The rule is defending against attackers shadowing malicious payloads with a safe prefix — not actually a hazard here because the regex is `^`-anchored (matches at position 0 only), but the rule is error-level. Refactored to `matchAll(...).next().value` to satisfy the rule without semantic changes. Inline comment documents that the `^` anchor already bounds this to one match.

## Test plan

- [x] New tests: 7/7 pass (`add-lesson.test.ts` — em-dash / en-dash / hyphen / no-heading / body-preservation / tags+provenance / single-line)
- [x] Full MCP test suite: 80/80 pass
- [x] Shield PASS (second run after fixing the single-line edge case)
- [x] Pre-push hook passed (20 unrelated warnings on pre-existing code, 0 errors)
- [ ] CI green

### DIFF
--- a/.changeset/pr-1284-mcp-double-heading.md
+++ b/.changeset/pr-1284-mcp-double-heading.md
@@ -1,0 +1,11 @@
+---
+'@mmnto/mcp': patch
+---
+
+Fix `add_lesson` MCP tool double-prepending `## Lesson — ` heading (#1284)
+
+When a caller passed a pre-formatted lesson to the `add_lesson` MCP tool whose body already started with a canonical `## Lesson — Foo` heading, the tool derived a title from the first line of the body — which included the literal `Lesson —` prefix — and produced a file with `## Lesson — Lesson — Foo` as the wrapper, with the original `## Lesson — Foo` still intact inside the body. The parser correctly read that as two separate lessons.
+
+The tool now detects a pre-existing canonical heading (em-dash, en-dash, or hyphen variants, consistent with the parser fix in #1278), extracts the title, and strips the heading line from the body before wrapping. Callers who pass plain body text with no leading heading see unchanged behavior.
+
+Closes #1284. Discovered during PR #1282 dogfooding.
--- a/.changeset/pr-1284-mcp-double-heading.md
+++ b/.changeset/pr-1284-mcp-double-heading.md
@@ -2,7 +2,7 @@
 '@mmnto/mcp': patch
 ---
 
-Fix `add_lesson` MCP tool double-prepending `## Lesson — ` heading (#1284)
+Fix `add_lesson` MCP tool double-prepending `## Lesson —` heading (#1284)
 
 When a caller passed a pre-formatted lesson to the `add_lesson` MCP tool whose body already started with a canonical `## Lesson — Foo` heading, the tool derived a title from the first line of the body — which included the literal `Lesson —` prefix — and produced a file with `## Lesson — Lesson — Foo` as the wrapper, with the original `## Lesson — Foo` still intact inside the body. The parser correctly read that as two separate lessons.
 
--- a/packages/mcp/src/tools/add-lesson.test.ts
+++ b/packages/mcp/src/tools/add-lesson.test.ts
@@ -196,3 +196,108 @@ describe('add_lesson auth model (#844)', () => {
     expect(typeof opts.shell).toBe('boolean');
   });
 });
+
+// ---------------------------------------------------------------------------
+// Double-heading bug (#1284)
+// ---------------------------------------------------------------------------
+
+describe('add_lesson double-heading guard (#1284)', () => {
+  let handle: (args: Record<string, unknown>) => Promise<unknown>;
+
+  beforeEach(() => {
+    _resetRateLimit();
+    lastWrittenEntry = '';
+    handle = setup();
+  });
+
+  /** Count occurrences of `## Lesson —` (all dash variants) in a string. */
+  function countLessonHeadings(entry: string): number {
+    const matches = entry.match(/^## Lesson[\s\u2014\u2013-]+/gm);
+    return matches ? matches.length : 0;
+  }
+
+  it('does not duplicate heading when body starts with canonical em-dash heading', async () => {
+    await handle({
+      lesson:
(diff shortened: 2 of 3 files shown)


satur8d:
@gemini-code-assist @coderabbitai addressing all three findings in commit \`38f5e05f\`:

**GCA (high) + CR (minor) — add \`^\s*\` for leading whitespace.** Valid and agreed. Added \`\s*\` after the \`^\` anchor so callers (especially LLMs) that emit pre-formatted lessons prefixed by blank lines or whitespace still get detected. New test \`handles pre-formatted lesson with leading blank lines or whitespace\` exercises the path; it failed RED against the old regex (count = 3 \`## Lesson\` lines in the written entry) and passes GREEN now (count = 1).

**GCA suggestion to add the \`/i\` flag — declined with rationale.** The canonical parser regex in \`packages/core/src/drift-detector.ts:84\` is \`^## Lesson [—–-] \` with capital L — case-sensitive. If this detection accepted lowercase \`## lesson\` and stripped it, the parser would still not recognize the resulting content as a heading, and round-trip semantics would break: we'd delete a line the parser treats as body text. Added an explicit test \`does NOT treat lowercase "## lesson" as a canonical heading\` that asserts the lowercase form is preserved verbatim in the body via the fallback path. If the parser's canonical form ever changes to accept lowercase, this detection should follow — but that's a coordinated change across both files, not a unilateral edit here.

**CR (minor) — markdownlint MD038 trailing space in code span.** Valid. Dropped the trailing space inside the \`\`\`## Lesson —\`\`\` span in the changeset.

Test count: 14 → 16 (+2 for leading-whitespace + lowercase-rejection). All 80 MCP tests still pass. Shield re-ran clean.

On the \`matchAll\` + \`g\` flag warning — the \`headingMatch\` const is computed once per handler invocation from a fresh \`LESSON_HEADING_RE\` literal, so there's no shared \`lastIndex\` state to corrupt across calls. No loop involved.

PR #1327, opened 2026-04-11
chore: backfill changesets for 1.14.1

satur8d:
## Summary

Backfills the seven missing changesets for code-bearing PRs merged to main since `@mmnto/*@1.14.0`. Prepares the ground for the 1.14.1 Version Packages PR once the remaining Phase 1 papercuts land.

**Changesets written** (all patch bumps per the incremental-release rule):

| PR | Scope | Title |
|---|---|---|
| #1311 | core + cli | 1.14.1 Hotfix Sweep |
| #1313 | cli | Queue drain: Shield branding + preflight + styleguide |
| #1314 | cli | Resolve non-staged AST paths against repo root |
| #1316 | cli | `totem handoff` deterministic scaffold refactor |
| #1324 | core | Pipeline 5 sanity gate |
| #1325 | cli | `--no-edit` → `--stdout` rename |
| #1326 | core | Tilde-fence support |

Four PRs are docs-only and do not ship to npm, so no changesets needed: #1315 (dropped lesson files), #1320 (README), #1322 (wiki), #1323 (wiki).

## Verification

`changeset status` confirms all three packages will bump to **1.14.1**:

```
info Packages to be bumped at patch:
- @mmnto/cli
- @mmnto/totem
- @mmnto/mcp
```

The `fixed` group in `.changeset/config.json` pulls `@mmnto/mcp` along even though no changeset names it directly.

## Noted lint warning

Pre-push lint flagged one warning on `.changeset/pr-1325-stdout-rename.md:7` for the phrase "never worked" — the rule fires on "never" as an absolute claim, but this is a factual statement about a flag that was broken by Commander.js's `--no-*` negation convention. Leaving as-is since the phrasing is accurate; happy to rephrase if reviewers disagree.

## Test plan

- [x] `changeset status` shows expected bumps
- [x] Pre-push hook passes (lint + tests + compiled rules)
- [ ] CI green
- [ ] Version Packages PR opens automatically and shows the expected 1.14.1 bump once merged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1317, opened 2026-04-10
bug(cli): --no-edit flag on totem handoff is silently ignored (Commander.js --no- prefix)


satur8d:
## Summary

`totem handoff --no-edit` does not suppress the editor — it writes the scaffold to disk AND opens `$EDITOR`, identical to the default (no flags) behavior.

## Root cause

Commander.js treats `--no-<flag>` as a boolean negation prefix. When we register `.option('--no-edit', '...')`, Commander interprets it as negating a `--edit` boolean, setting `options.edit = false` instead of `options.noEdit = true`. The `handoffCommand` checks `options.noEdit` which is always `undefined`.

## Reproduction

```bash
totem handoff --no-edit
# Expected: scaffold printed to stdout
# Actual: scaffold written to .totem/journal/YYYY-MM-DD-session.md, editor opens
```

## Fix options

1. **Rename to `--print` or `--stdout`** — avoids the `--no-` prefix entirely
2. **Register as `--edit` (default true)** and let `--no-edit` be the Commander negation — check `options.edit === false`
3. **Use `.option('-n, --no-edit')`** with short flag and handle Commander's convention

Option 2 is probably cleanest — it uses Commander's convention correctly. The `--lite` alias should map to the same behavior.

## Discovered

First real use of the new scaffold during the 2026-04-10 queue drain session handoff. The feature shipped in PR #1316 (#1310).

## Scope

Small — rename/re-register the flag, update the check in `handoffCommand`, update tests.

Issue #1279, opened 2026-04-07
Pipeline 5 auto-capture produces nonsense regex from shield prose findings


satur8d:
## Summary

Pipeline 5 (the Shield AI observation auto-capture path) compiles shield findings into nonsense regex rules. The compile path takes the prose `message` field from a shield verdict and derives a pattern from it via the standard compile prompt — but the prose doesn't have any code to anchor on, so the model emits patterns that match arbitrary syntactic noise (e.g. `\*/` against `**/*.ts`, which matches the end of every block comment in every TypeScript file).

These rules then go straight into `.totem/compiled-rules.json` without a sanity gate. The next time anyone runs `totem lint`, the trap ledger gets flooded with hits that have nothing to do with the original shield finding.

## Reproduction (just happened on PR #1278)

1. On a branch with a real architectural change (in this case: widening `LESSON_HEADING_RE`), run `pnpm totem review`
2. Shield emits a legitimate WARN finding — for #1278 it was: *"`enforceHeadingLimit` hardcodes the em-dash (`## Lesson — `) and will fail to truncate headings that use a hyphen or en-dash. Consider updating its regex to match the new `LESSON_HEADING_RE`."*
3. Pipeline 5 captures this as an "observation rule" via `claude-sonnet-4-6` in the compiler path
4. Result: a new entry appears in `.totem/compiled-rules.json`:

```json
{
  "lessonHash": "c02cc47ceafb88a6",
  "lessonHeading": "Pipeline 5: observation from shield",
  "pattern": "\\*/",
  "message": "`enforceHeadingLimit` hardcodes the em-dash (`## Lesson — `) and will fail to truncate headings that use a hyphen or en-dash. Consider updating its regex to match the new `LESSON_HEADING_RE`.",
  "engine": "regex",
  "severity": "warning",
  "fileGlobs": [
    "**/*.ts"
  ],
  "compiledAt": "2026-04-07T19:17:20.344Z",
  "createdAt": "2026-04-07T19:17:20.344Z"
}
```

The pattern `\*/` matches the literal characters `*/` in any file under `**/*.ts` — which is **every closing block comment in every TypeScript file in the repo**. The message correctly preserves the shield's text but the pattern has zero semantic relationship to it.

## Why it's wrong

[...]

PR #1334, opened 2026-04-11
chore: postmerge lessons and rules from 1.14.1 cycle

satur8d:
## Summary

Postmerge workflow output from the 1.14.1 release cycle (PRs #1327, #1328, #1330, #1331, #1332). Ships now that 1.14.1 is published to npm — these files are repo-internal dogfood state and do not affect the tarballs.

## What's in the bundle

**16 new lesson files** in `.totem/lessons/` extracted via `totem lesson extract` from the merged PRs.

**6 new compiled rules** (394 → 400, all ast-grep, warning severity):

| Hash | Heading | Source |
|---|---|---|
| `4ac94d6f` | Account for leading whitespace in line-start parsers | #1284 |
| `1abb90c2` | Handle single-line inputs in regex terminators | #1284 |
| `9ea5abc3` | Avoid `--no-` prefix for standalone flags | #1317 |
| `91baf4a3` | Avoid shell metacharacters in exec arguments | #1233 |
| `ab3bc053` | Use shell-safe snippets for timeout tests | #1233 |
| `6f248090` | Use `matchAll` for security-sensitive regex matching | #1332 |

**9 conceptual lessons skipped** as non-compilable (kept as searchable memory in LanceDB, not enforced as tripwires).

**1 parse failure** — `Escape triple backticks in Markdown regex` couldn't be expressed as a regex or ast-grep pattern.

## Curation

Per Gemini's strategic triage verdict on the fresh rule set:

- **`matchAll` rule kept** — compiled as a `warning`, not an `error`, so it won't block pushes. Letting it run for a cycle so `totem doctor` can refine it via real telemetry rather than silencing it preemptively.
- **Duplicate `safeExec` rule deleted** — extract produced two rules for the same underlying issue (`Avoid shell metacharacters in exec arguments` and `Address safeExec shell leakage on Windows`). The narrower `spawn`-only variant (lesson `ac59b517` / hash `63011271`) was removed in favor of the broader `{ shell: true }` selector, which catches `execFileSync`/`execSync`/`exec`/`spawn` equally — matching the actual site of the #1233 bug.
- **9 conceptual lessons retained** — not every lesson compiles to a tripwire, and that's fine. They feed the memory layer, not the enforcement layer.
- **Pipeline 5 output not reverted** — #1279 shipped in 1.14.1 and the fresh rules look clean. If Pipeline 5 generates nonsense again, we reopen #1279 rather than reverting blindly.

## Incident note

First attempt to remove the duplicate rule used a Python script that opened `compiled-rules.json` in text-write mode and hit a `UnicodeEncodeError` mid-dump on a surrogate-escaped emoji in a lesson body. Python had already truncated the file via `open('w')` so the exception left a corrupt half-written JSON at line 6132. Recovery: `git checkout HEAD -- .totem/compiled-rules.json .totem/compile-manifest.json` + re-run compile. Saved as a feedback memory (`feedback_python_json_surrogates.md`) so I default to Node for JSON mutations next time.

## Test plan

- [x] Pre-push hook passed (format + lint + review)
- [x] 400 rules verified (394 baseline + 6 new, no duplicates)
- [x] No npm package changes (this is repo-internal dogfood state)
- [ ] CI green

### DIFF
--- a/.totem/compile-manifest.json
+++ b/.totem/compile-manifest.json
@@ -1,7 +1,7 @@
 {
   "compiled_at": "2026-04-11T05:28:26.417Z",
   "model": "gemini-3-flash-preview",
-  "input_hash": "4ffac7c3592328d135116a288ad8a1ecd28c768a8200003d7f82c8d81de507da",
-  "output_hash": "29e42b5aceb3b8daf5f9e2aacb52b101eb5b077bbde2cc2200c29f210ed64ae9",
-  "rule_count": 400
+  "input_hash": "c5eecb0bbd86210f7173b09dd8697ccdd0eea00bef6f6fd560486e6a39371f12",
+  "output_hash": "b56308fd1f5f79a448b2851eae4f59f0edf2f3fc9decc6403f93268a56987ce9",
+  "rule_count": 397
 }
--- a/.totem/compile-manifest.json
+++ b/.totem/compile-manifest.json
@@ -1,7 +1,7 @@
 {
-  "compiled_at": "2026-04-11T05:06:26.002Z",
+  "compiled_at": "2026-04-11T05:28:26.417Z",
   "model": "gemini-3-flash-preview",
-  "input_hash": "9bec3a9877736135ecaa4fe5131f5d2bc4708bdd20c9dee60595c083b374307c",
-  "output_hash": "a7d4ed13604779e15529dcaf6c792ac9d9a6cc704c681a946499b04918618b9c",
+  "input_hash": "4ffac7c3592328d135116a288ad8a1ecd28c768a8200003d7f82c8d81de507da",
+  "output_hash": "29e42b5aceb3b8daf5f9e2aacb52b101eb5b077bbde2cc2200c29f210ed64ae9",
   "rule_count": 400
 }
--- a/.totem/compile-manifest.json
+++ b/.totem/compile-manifest.json
@@ -1,7 +1,7 @@
 {
-  "compiled_at": "2026-04-09T07:29:13.410Z",
+  "compiled_at": "2026-04-11T05:06:26.002Z",
   "model": "gemini-3-flash-preview",
-  "input_hash": "7fd343e931e04a9d4231031f945d58eb9581c6a7234ab3a732c29df09b7b6ed1",
-  "output_hash": "a8e05eb5dd33abfbfeaf0cfe672535f3ba84e77e61c8740cc2c381ee7e3de46f",
-  "rule_count": 394
+  "input_hash": "9bec3a9877736135ecaa4fe5131f5d2bc4708bdd20c9dee60595c083b374307c",
+  "output_hash": "a7d4ed13604779e15529dcaf6c792ac9d9a6cc704c681a946499b04918618b9c",
+  "rule_count": 400
 }
--- a/.totem/compiled-rules.json
+++ b/.totem/compiled-rules.json
@@ -6344,6 +6344,95 @@
         "packages/mcp/src/smoke-test.ts"
       ],
       "severity": "warning"
+    },
+    {
+      "lessonHash": "4ac94d6f9c387e1e",
+      "lessonHeading": "Account for leading whitespace in line-start parsers",
+      "message": "Start-of-string anchor (^) used without accounting for leading whitespace. Use trimStart() on input or include optional whitespace (^\\s*) to prevent failures from leading blank lines.",
+      "engine": "regex",
+      "severity": "warning",
+      "pattern": "/\\^[^\\s\\\\]",
+      "compiledAt": "2026-04-11T05:06:18.175Z",
+      "createdAt": "2026-04-11T05:06:18.175Z",
+      "fileGlobs": [
+        "packages/mcp/src/tools/**/*.ts"
+      ]
+    },
+    {
+      "lessonHash": "1abb90c2493fa2b5",
+      "lessonHeading": "Handle single-line inputs in regex terminators",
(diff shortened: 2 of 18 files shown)


satur8d:
@gemini-code-assist addressed in commit \`f2904fc6\`, but with a partial push-back on the specific wording.

**The underlying concern is valid** — \`setInterval(Object, ms)\` is unidiomatic and a reader without context could reasonably be confused. Rewrote the lesson body to lead with the principle (avoid \`=>\` in shell-routed \`-e\` scripts on Windows because cmd.exe parses it as output redirection) and named \`function() {}\` as the idiomatic form. Kept \`setInterval(Object, ms)\` as a noted alternative because the test code in \`packages/core/src/sys/exec.test.ts\` still uses it and both forms work.

**Push-back on the specific suggestion:** your proposed text — \"Use \`setInterval(function() { ... }, ms)\" — implies there's intended logic in the callback body (\`{ ... }\` reads as \"some code\"). The original pattern was an empty arrow function \`() => {}\` used as a keep-alive no-op callback. There is no logic to preserve. \`setInterval(Object, ms)\` preserves that exactly — \`Object\` called without \`new\` is a no-op callable that returns an empty object which gets discarded. It's a trick, but it's a semantically-faithful one.

The new body says: \"Use any callable without \`=>\` as the keep-alive callback — a plain \`function() {}\` is the most idiomatic form; \`setInterval(Object, ms)\` also works because \`Object\` is a no-op callable. The principle is the important part: no \`=>\` in shell-routed \`-e\` strings on Windows.\"

Recompile regenerated the rule (\`ab3bc053\` → \`075012b9\`). The ast-grep selector is unchanged; only the violation message was refreshed to match the principled framing.

satur8d:
@gemini-code-assist thanks for the second-pass review. Verdicts + resolution in commit \`afad20bf\`:

## 1. \`lesson-d879ab4d.md\` — \`setInterval(Object, ms)\` is nonsensical

**Deleted the lesson entirely** (and the rule it compiled into). Your first-round review prompted me to rewrite the lesson body in \`f2904fc6\` to lead with \`function() {}\` as the idiomatic form and keep \`setInterval(Object, ms)\` only as a documented alternative. This second-round comment appears to be flagging the same lesson against your preferred single-line wording, but the updated body was more precise than a one-line diff suggestion could capture. Since comment #4 below makes the whole lesson moot (the ast-grep rule can't see inside the string literal that's the actual source of the bug), I'm dropping both the lesson and the rule rather than refining further.

## 2. Rule \`91baf4a30b32ce18\` — pattern too narrow (\`compiled-rules.json:6394\`)

**Agreed. Deleted rule + \`lesson-9a5f6021.md\`.** The compiled pattern is \`spawn($CMD, $$$ARGS, { $$$BEFORE, shell: true, $$$AFTER })\` — spawn-only. The actual #1233 bug was in \`execFileSync\`, which this rule would completely miss. Broadening the pattern to match the \`{ shell: true }\` options object across all exec variants is a different lesson shape that deserves a deliberate, CI-verified capture in a future cycle rather than an iterative LLM refinement.

## 3. Rule \`6f2480907c256629\` — matchAll rule scope too broad (\`compiled-rules.json:6410\`)

**Agreed. Deleted rule + \`lesson-f6867216.md\`.** Earlier today Gemini in chat suggested keeping this rule and letting the self-healing loop refine it via telemetry (\`totem doctor --pr\` was supposed to archive it if noise rate crossed a threshold). But while debugging PR #1338 just now I discovered **#1336** — the "Archive Lie": \`loadCompiledRules\` doesn't honor \`rule.status === 'archived'\`. The self-healing loop is a placebo. Without a working archive path, shipping this rule would be shipping permanent noise. Deletion is the only honest option until #1336 is fixed.

## 4. Rule \`075012b99dbf1849\` — conceptually broken rule (\`compiled-rules.json:6428\`)

**Agreed, and this is the most interesting of the four.** You're exactly right that ast-grep operates on the host code's AST and can't look inside string literal contents. The lesson was about \`=>\` being interpreted as output redirection by \`cmd.exe\` *when parsing a \`-e\` string argument passed to \`node\`*. That's a **string-content concern**, not an AST-node concern. The compiled pattern \`setInterval(() => { $$$BODY }, $MS)\` would fire on real \`setInterval\` arrow-function calls in production code — the exact opposite of what's intended. Classic semantic-level mismatch between the lesson and the engine it was compiled for. Deleted both.

## On "don't manually edit \`compiled-rules.json\`"

Noted and agreed as a principle. In this case the edits were:
1. The initial \`d2623e7c\` commit (6 new rules generated by \`totem lesson compile\` — machine-generated, not hand-edited)
2. This \`afad20bf\` commit (deletion of 3 entries + manifest refresh via a one-off Node script using core's \`generateInputHash\`/\`generateOutputHash\`)

The deletion path was necessary because:
- \`totem doctor --pr\` can't archive (#1336 — archive status is ignored)
- \`totem lesson compile\` on no-op runs doesn't refresh the manifest when the only change is a deleted lesson with no stale rule (#1337 — filed today as a P0)
- The only alternative was \`totem lesson compile --force\` which in this repo is ~19 minutes of non-deterministic LLM calls that dropped 393→386 rules on the one time I tried it

Both root causes are filed as P0 in 1.15.0. Once #1336 and #1337 land, the sequence "rewrite lesson → compile → archive bad rule" will actually work end-to-end and the "don't touch the artifact" rule will have teeth. Until then, surgical deletes are the pragmatic path.

## Net change

- \`compiled-rules.json\`: 400 → 397 (3 rules deleted)
- \`.totem/lessons/\`: 3 lesson files removed
- Commit message \`afad20bf\` has the full rationale for each deletion

PR #1341, opened 2026-04-11
docs(readme): surface the planning layer (totem spec + preflight)

satur8d:
## Summary

Reframes the "What Works and What Doesn't" section from two layers to three, and adds \`totem spec\` to the command table. Motivated by a conversation in chat where I flagged that the previous section undersold the planning layer — \`totem spec\` and the \`/preflight\` workflow — which have been load-bearing for velocity and architectural consistency in practice.

## What changed

**1. Three-layer framing in "What Works and What Doesn't"**

Before: two layers (enforcement + memory).

After: three layers (enforcement + planning + memory). The new middle layer calls out \`totem spec\` and \`/preflight\` as the upstream gate that catches architectural drift *before* code is written, while preserving the "hard tripwire vs soft gate" distinction that the Pragmatic Builder voice demands.

Key honest-assessment line I'm trying to land: **"A meaningful chunk of the velocity and architectural consistency I've been hitting comes from this upstream gate, more than I expected when I first added it."** Grounded in the reality that neither \`totem spec\` nor \`/preflight\` is enforced mechanically — they're just structured prompts — but the structure turns out to be surprisingly load-bearing when the agent is deep in a task and starting to drift.

**2. Added \`totem spec\` to the command table**

Placed next to \`totem review\` since both are LLM-powered and require an API key. Also tightened the \`totem review\` description to say "on an uncommitted diff" so the two rows don't read as overlapping.

## Voice check

Grounded in \`.strategy/voice-tuning-dataset.md\`:

- **Honest Assessment Pattern** (rule 5) — three layers, each categorized as "works", "works too, to my surprise", or "real infrastructure but agent discipline is not deterministic"
- **Harmonize Memory and Enforcement** (rule 6) — the planning layer makes the memory layer + enforcement layer reinforce each other (planning forces the memory query upstream of code; enforcement catches what both upstream layers miss)
- **Humility and Curiosity** (rule 1) — "to my surprise", "more than I expected"
- **No Role-Assignment Metaphors** (rule 7) — no "flight controller" language, just named tools

## Test plan

- [x] Pre-push hook passes (lint, no relevant changes, no test failures)
- [ ] Gemini review on the voice and the three-layer framing
- [ ] Readability check: does the planning-layer paragraph land or does it feel like feature bloat?

## Related

- Prior README voice refresh: #1320
- Voice tuning dataset: \`.strategy/voice-tuning-dataset.md\` (added 2026-04-10)


satur8d:
@gemini-code-assist @coderabbitai both valid — addressed in commit \`13c0edcd\`.

Both of you flagged the \`/preflight\` reference. You're right for different but reinforcing reasons:

- **@gemini-code-assist** — Tenet 13 ("Sensors, Not Actuators") + styleguide rule 71 against referencing unshipped Claude Code hooks. Describing an agent-specific skill as a core product layer contradicts the "platform of primitives" positioning.
- **@coderabbitai** — editorial clarity. Users can't tell if \`/preflight\` is a Totem built-in, an agent-specific skill, or a personal customization. Blending it with \`totem spec\` creates confusion about what the product actually ships.

## What changed

Rewrote the planning-layer paragraph to describe **only** \`totem spec\`, which is the shipped CLI command. Specifically:

**Removed:**
- "\`/preflight\` skill forces a knowledge-base query"
- "For architectural changes it drafts a 1-page design doc and waits for explicit approval"

**Added** (accurate description of what \`totem spec\` actually does, verified against \`packages/cli/src/commands/spec-templates.ts\`):
- "pulls the GitHub issue body"
- "queries the knowledge base for relevant lessons and ADRs"
- "dumps a structured implementation spec to \`.totem/specs/<issue>.md\`"
- "architectural context, files to examine, edge cases the issue description missed, and task-by-task TDD directives with retrieved lessons injected inline as invariants"

The knowledge-base query isn't a separate skill — it's baked into the \`totem spec\` command's prompt assembly step. That was my mistake in the first version; I was conflating the CLI command with the \`/preflight\` skill that orchestrates it. \`totem spec\` does the KB query on its own.

Also extended the closing paragraph — "each catches a different class of failure, at a different stage of the workflow" — to make the three-layer argument more explicit now that only one upstream layer is referenced (the "upstream layers" plural in the original didn't make sense after dropping one of them).

Issue #1352, opened 2026-04-11
archive: 3 over-broad Pipeline 5 split() rules blocking legitimate code


satur8d:
## Summary

Three Pipeline 5 observation rules are firing as errors on legitimate first-line-extraction code in catch blocks, blocking shipping of mmnto/totem#1349. All three are archive candidates for the #1345 lifecycle filter (same pattern as mmnto/totem#1347 did for the diff-header rule).

## The three rules

| Hash | Heading | Engine |
|---|---|---|
| `f1e6b70ba8796049` | Prefer Explicit Metadata Tokens from LLMs Over Heuristics | regex |
| `dc656c9238a0f80e` | Performing content.split('\n') inside a loop over line | ast-grep |
| `b0ad61d6fd0eafe4` | Performing content.split('\n') inside a loop over line | regex |

The last two are **duplicates** of the same lesson — one ast-grep, one regex — a Pipeline 5 double-compilation artifact.

## Why they should be archived

**`f1e6b70ba8796049` (metadata tokens):** The lesson is about parsing LLM output and preferring explicit metadata tokens over heuristic splitting. It has nothing to do with general-purpose string splitting in catch blocks, error handlers, or utility code. The regex pattern `/\.split\(['"/](?:\n|\r\n)['"/]\)\s*\[\s*0\s*\]/` fires on any `.split('\n')[0]` literal regardless of the surrounding context — including inside comments that happen to mention the idiom.

**`dc656c9238a0f80e` and `b0ad61d6fd0eafe4` (loop over line):** The lesson is about using `.split('\n')` as a per-iteration operation inside a loop that iterates lines — a performance anti-pattern for large content. The rule fires on any `.split('\n')[` literal regardless of whether there's a loop context anywhere in the function. In mmnto/totem#1349 it fired on a one-shot catch block that extracts the first line of an error message — zero loop, zero large content, zero performance concern.

## Concrete bite

On mmnto/totem#1349's review cycle, I rewrote an ast-grep parser error handler to use `split('\n')` instead of `split(/[.\n]/)` (legitimate fix addressing a real bug GCA flagged: splitting on `.` truncated the verbatim pattern source from ast-grep error messages, discarding the single most useful debugging signal). Pre-push hook then failed with 2 errors from these rules, THEN failed again with 3 errors after I added a comment explaining the workaround — because the comment itself contains the literal string `raw.split('\n')[` and the rules match against any line containing that regex.

[...]

PR #1362, opened 2026-04-11
fix(cli): retire totem wrap pending fix for destructive docs overwrite of active_work.md

### LINKED ISSUE #1361: fix(cli): retire totem wrap pending fix for destructive docs overwrite of active_work.md
## Problem

`totem wrap` step 4 (`totem docs`) silently overwrites hand-crafted committed documentation via LLM rewrite. Surfaced during the 2026-04-11 PM `/signoff` session when I hand-crafted a 1.14.5 refresh in `docs/active_work.md` and realized running `/postmerge` would have destroyed it.

## The exact failure mode

1. `totem wrap <prs>` runs 6 steps. Step 4 is `totem docs` which iterates every target in `config.docs`.
2. `totem.config.ts:42-50` registers `docs/active_work.md`, `docs/roadmap.md`, and `docs/architecture.md` as doc targets.
3. `packages/cli/src/commands/docs.ts:450-458` has a dirty-file guard:
   ```ts
   const dirtyFiles = targets.filter((d) => isFileDirty(cwd, d.path));
   if (dirtyFiles.length > 0 && !options.dryRun) { throw ... }
   ```
4. **This guard only catches UNCOMMITTED changes.** Recently committed hand-crafted edits are fair game.
5. The LLM prompt at `docs.ts:38-42` explicitly tells the model to "aggressively rewrite forward-looking sections (roadmaps, active work, upcoming features) to match the Ground Truth."
6. Result: any committed human edit to a registered doc target will be silently rewritten the next time `totem wrap` runs. Precedent exists: `d59af2f6 docs: refresh active_work for 1.14.1 hotfix sweep` is a prior manual refresh around wrap.

This is a "Sensors Not Actuators" (Tenet 5) violation. A tool should never destroy human-authored context without an explicit `--force` flag.

## Decision: retire via hard error

Per the 2026-04-11 `/signoff` conversation: Option C (hard error, not deprecation warning). A command that actively corrupts committed work deserves an honest "this is turned off" signal, not a soft warning. The code stays in place to preserve institutional memory; the command surface is blocked.

Paired with this retirement:
- Remove `totem wrap` from `--help` output
- Rewrite `.claude/skills/postmerge/SKILL.md` to run the manual sequence directly
- Update README / wiki / CLAUDE.md references
- Leave `wrap.ts` + `wrap.test.ts` source intact, bodies replaced with the retirement error

## Return conditions

Before `totem wrap` comes back, all three must be true:

1. **`--skip-docs` flag exists on wrap**, so post-release refresh workflows can opt out of Step 4 entirely.
2. **Freshness guard on `totem docs`** — if the target's last git author date is within the last N hours (default 24), refuse to rewrite and require `--force-regen`. This protects hand-crafted committed edits in the same window they were made.
3. **End-to-end regression test for wrap** that seeds a hand-crafted `active_work.md`, runs wrap, and asserts the file was not rewritten. Locks the invariant mechanically so future refactors can't re-introduce the gap.

## Manual post-merge workflow (until wrap returns)

```bash
# 1. extract lessons from merged PRs
pnpm exec totem extract <pr-numbers> --yes

# 2. sync index (often already handled by post-merge git hook)
pnpm exec totem sync

# 3. compile new rules locally (NOT --cloud per policy)
pnpm exec totem compile --export

# 4. revert compiled-rules.json to curated set (LLM compile still produces noise)
git checkout HEAD -- .totem/compiled-rules.json

# 5. format + stage lessons/exports + commit
pnpm run format
git add .totem/lessons/ .github/copilot-instructions.md .junie/skills/totem-rules/rules.md
git commit -m "chore: totem postmerge lessons for <PRs>"
```

This manual sequence will also become the new `.claude/skills/postmerge/SKILL.md` body as part of this retirement PR.

## Related

- Surfaced by the `/signoff` session on 2026-04-11 (journal: `.strategy/.journal/2026-04-11-four-p0-sweep-and-1.14.5-shipped.md`)
- Tenet 5 (Sensors Not Actuators) from the strategy repo
- Historical: #786 (prevent doc regen from persisting stale issue references) — related class of bug, closed

satur8d:
Closes #1361

## Summary

- Retire `totem wrap` via a hard `TotemError` throw. Surfaced during the 2026-04-11 PM `/signoff` session: step 4 (`totem docs`) silently overwrites hand-crafted committed documentation because the dirty-file guard at `docs.ts:450` only catches uncommitted changes, not recent committed edits. This is a Tenet 5 ("Sensors Not Actuators") violation.
- Keep the command wired behind `{ hidden: true }` so invocation fires the error instead of commander's "unknown command" fallback. The error's `recoveryHint` lists the full manual workaround sequence so users are not blocked.
- Rewrite `.claude/skills/postmerge/SKILL.md` to run the individual steps directly (`totem extract` → `totem sync` → `totem compile --export` → revert compiled-rules.json → format → stage → commit).
- Update all user-facing doc references: `docs/wiki/cli-reference.md`, `docs/wiki/agent-gemini-code-assist.md`, `docs/architecture.md`, `docs/reference/workflow-automation.md`, `docs/reference/supported-models.md`.

## Context

The signoff session on 2026-04-11 hand-crafted a 1.14.5 refresh in `docs/active_work.md`. Before running `/postmerge` for the six marathon PRs, I checked what `totem wrap` would actually do and found:

1. `totem wrap <prs>` runs 6 steps. Step 4 is `totem docs` iterating every target in `config.docs`.
2. `totem.config.ts:42-50` registers `docs/active_work.md`, `docs/roadmap.md`, and `docs/architecture.md` as doc targets.
3. `packages/cli/src/commands/docs.ts:450-458` has a dirty-file guard, but it only catches uncommitted changes.
4. The LLM prompt at `docs.ts:38-42` explicitly tells the model to "aggressively rewrite forward-looking sections (roadmaps, active work, upcoming features) to match the Ground Truth."
5. Result: the signoff refresh would have been silently destroyed on the first postmerge wrap run.

Per the `/signoff` conversation (Claude + Gemini agreement), Option C was chosen: hard error, not deprecation warning. A command that actively corrupts committed work deserves an honest "this is turned off" signal.

## Return conditions (tracked in #1361)

Before `totem wrap` comes back, all three must ship:

1. `--skip-docs` flag on wrap so post-release refresh workflows can opt out of Step 4 entirely.
2. Freshness guard on `totem docs`: skip targets whose git author date is within the last 24 hours without `--force-regen`.
3. End-to-end regression test for wrap that seeds a hand-crafted `active_work.md`, runs wrap, and asserts the file survives the pipeline unmodified.

## What is NOT changed

- `wrap.ts` source is preserved (only the body of `wrapCommand` is replaced with the throw). Function signature, `WrapOptions` interface, and `TAG` constant all stay as named exports so the un-retirement PR can restore the original 6-step implementation without re-deriving the shape.
- The original orchestration tests in `wrap.test.ts` are not preserved inline but they are in git history (look at the parent of this commit). The un-retirement PR should restore them alongside the new regression test for the freshness guard.
- `packages/cli/src/commands/docs.ts` is untouched. The underlying `totem docs` drift is a separate fix that lands as part of return condition #2.

## Test plan

- [ ] CI green on the branch
- [ ] `totem wrap 1234` from the command line prints the retirement error and the manual workaround (verify locally)
- [ ] `totem --help` does not show `wrap` in the command list
- [ ] `pnpm exec vitest run packages/cli/src/commands/wrap.test.ts` passes the four new retirement-contract assertions
- [ ] `docs/wiki/cli-reference.md` rendered preview shows the "RETIRED" marker and the workaround sequence
- [ ] `.claude/skills/postmerge/SKILL.md` no longer references `totem wrap`

### DIFF
--- a/.claude/skills/postmerge/SKILL.md
+++ b/.claude/skills/postmerge/SKILL.md
@@ -1,13 +1,51 @@
 ---
 name: postmerge
-description: Post-merge workflow — extract lessons and wrap
+description: Post-merge workflow — extract lessons and compile rules (manual sequence, wrap is retired)
 ---
 
-After merging PRs, run the post-merge workflow:
+`totem wrap` is retired pending mmnto-ai/totem#1361 (it silently
+overwrites hand-crafted committed docs via the `totem docs` step).
+Run the post-merge steps directly instead.
 
-1. Run `pnpm exec totem wrap $ARGUMENTS --yes` (pass PR numbers as arguments)
-2. Revert compiled rules to curated set: `git checkout HEAD -- .totem/compiled-rules.json`
-3. Format wrap output: `pnpm run format`
-4. Stage lessons, docs, and exports: `git add .totem/lessons/ README.md docs/ .github/copilot-instructions.md .junie/skills/totem-rules/rules.md`
-5. Commit: `git commit -m "chore: totem wrap for PRs $ARGUMENTS"`
-6. Report: how many lessons extracted, any doc failures, rule compilation stats (note: new compiled rules were reverted)
+After merging PRs, run the following sequence. Replace `$ARGUMENTS`
+with the merged PR numbers (space-separated, e.g. `1345 1347 1348`).
+
+1. Extract lessons from the merged PR(s):
+   `pnpm exec totem extract $ARGUMENTS --yes`
+
+2. Sync the semantic index (usually already handled by the post-merge
+   git hook, but running it explicitly is cheap and safe):
+   `pnpm exec totem sync`
+
+3. Compile new rules locally and export to AI tool configs. Do NOT
+   pass `--cloud`; the cloud worker is still Gemini-only per
+   mmnto-ai/totem#1221. Local compile routes to Sonnet 4.6:
+   `pnpm exec totem compile --export`
+
+4. Revert `compiled-rules.json` to the curated set. Compile will
+   produce new rules from the extracted lessons, but the curated set
+   in main is the source of truth. The new rules should be reviewed
+   and cherry-picked by hand rather than auto-merged (empirically
+   4/6 auto-compiled rules from the 1.14.1 postmerge were bad, and
+   mmnto-ai/totem#1349 now catches the syntactic-invalid cases but
+   not the over-broad ones):
+   `git checkout HEAD -- .totem/compiled-rules.json`
+
+5. Format everything wrap might have touched:
+   `pnpm run format`
+
+6. Stage only the artifacts we keep (lessons, exports, docs if they
+   were intentionally updated — NOT `docs/active_work.md`,
+   `docs/roadmap.md`, or `docs/architecture.md` unless you hand-edited
+   them deliberately):
+   `git add .totem/lessons/ .github/copilot-instructions.md .junie/skills/totem-rules/rules.md`
+
+7. Commit:
+   `git commit -m "chore: totem postmerge lessons for $ARGUMENTS"`
+
+8. Report: how many lessons extracted, which rules compiled and were
+   reverted, whether any over-broad rules warrant follow-up tickets.
+
+The retirement error from `totem wrap` produces this same workaround
+text at runtime, so if you forget the sequence, just run
+`pnpm exec totem wrap <prs>` and copy the hint.
--- a/.claude/skills/postmerge/SKILL.md
+++ b/.claude/skills/postmerge/SKILL.md
@@ -11,7 +11,7 @@ After merging PRs, run the following sequence. Replace `$ARGUMENTS`
 with the merged PR numbers (space-separated, e.g. `1345 1347 1348`).
 
 1. Extract lessons from the merged PR(s):
-   `pnpm exec totem extract $ARGUMENTS --yes`
+   `pnpm exec totem lesson extract $ARGUMENTS --yes`
 
 2. Sync the semantic index (usually already handled by the post-merge
    git hook, but running it explicitly is cheap and safe):
@@ -20,7 +20,7 @@ with the merged PR numbers (space-separated, e.g. `1345 1347 1348`).
 3. Compile new rules locally and export to AI tool configs. Do NOT
    pass `--cloud`; the cloud worker is still Gemini-only per
    mmnto-ai/totem#1221. Local compile routes to Sonnet 4.6:
-   `pnpm exec totem compile --export`
+   `pnpm exec totem lesson compile --export`
 
 4. Revert `compiled-rules.json` to the curated set. Compile will
    produce new rules from the extracted lessons, but the curated set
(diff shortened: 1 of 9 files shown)


satur8d:
@gemini-code-assist @coderabbitai — both findings valid, both fixed in `1ca06d8c`.

## Dynamic import of `TotemError`

Both of you caught that my top-level `import { TotemError } from '@mmnto/totem';` violated the CLI lazy-load convention that every other command in `packages/cli/src/commands/` follows. You're right — and the original pre-retirement wrap.ts also did the dynamic import, so my simplification actually broke precedent.

Fix: moved `TotemError` to a dynamic import inside `wrapCommand`:

```ts
export async function wrapCommand(_prNumbers: string[], _options: WrapOptions): Promise<void> {
  const { TotemError } = await import('@mmnto/totem');
  throw new TotemError(...);
}
```

This also gives the async function a legitimate `await`, which let me drop the `eslint-disable-next-line @typescript-eslint/require-await` suppression I had to add. Win all around.

Side benefit: local lint warning count dropped from 34 to 4 after this change. The 4 remaining warnings are all false positives on the async-test-callback rule firing on tests that legitimately `await wrapCommand(...)` — same noise class tracked in #1352.

On the `[Totem Error]` auto-prefix note @gemini-code-assist raised: confirmed I'm NOT manually prefixing. The message string passed to the `TotemError` constructor is plain, and the `[Totem Error]` prefix is added by the constructor itself (`errors.ts:39`). No action needed on that point, but appreciate the belt-and-suspenders reference to the convention.

## Canonical noun-verb commands (CR nit)

@coderabbitai absolutely right that `totem extract` and `totem compile` are the deprecated hidden aliases, not the canonical forms. Verified against `packages/cli/src/index.ts`:

- Line 410: `.command('compile', { hidden: true })` ← hidden alias
- Line 495: `.command('extract [pr-numbers...]', { hidden: true })` ← hidden alias
- Line 747: `program.command('lesson')` ← canonical entity command
- Line 774: `.command('compile')` under lesson ← canonical noun-verb
- Line 817: `.command('extract [pr-numbers...]')` under lesson ← canonical noun-verb

Fixed across every file touched by this PR:

- `wrap.ts` — the retirement error's `recoveryHint` now prints `totem lesson extract` and `totem lesson compile --export`, so users who trip the error at runtime get the up-to-date workaround
- `wrap.test.ts` — assertions updated with a comment citing the `index.ts` line numbers as the source of truth
- `.claude/skills/postmerge/SKILL.md` — both steps use canonical forms
- `docs/wiki/cli-reference.md` — workaround code block uses canonical forms
- `docs/reference/workflow-automation.md` — Phase 6 command list uses canonical forms

## Out of scope (deferred, not ignored)

The wider codebase still uses the bare `totem extract` / `totem compile` forms in several places I am NOT touching on this PR:

- `README.md` (lines 62, 66, 75, 94, 95)
- `COVENANT.md`
- `GEMINI.md`
- `.junie/guidelines.md`
- `.claude/docs/contributing.md`
- `.claude/docs/architecture.md`
- `.gemini/styleguide.md`

A tactical follow-up should migrate the wider doc surface to `totem lesson extract` / `totem lesson compile` for consistency. I'll file a ticket if one doesn't already exist, but it's deliberately out of scope for the retirement PR to keep the diff focused on the one mechanical change (retire `totem wrap`).

Issue #62, opened 2026-03-03
Refactor: Unify GitHub data fetching and JSON validation


satur8d:
## Background
Our current GitHub CLI (gh) helper functions (in spec.ts, 	riage.ts, riefing.ts, and the upcoming learn.ts) use a mixed pattern where JSON.parse and Zod validation are often tightly coupled. While functional, this makes error handling for malformed JSON vs. schema mismatches difficult to distinguish and leads to duplicated boilerplate across command files.

## Goal
Extract a unified, robust GitHub fetching utility that handles the execFileSync execution, JSON parsing, and schema validation as distinct, observable steps.

## Tasks
- [ ] Create a shared GitHub fetch utility (possibly in packages/cli/src/git.ts or a new gh.ts) that isolates JSON.parse from Zod validation.
- [ ] Implement standardized error reporting for:
    - Missing gh CLI.
    - GitHub API timeouts/failures.
    - Malformed JSON responses.
    - Schema validation (Zod) errors.
- [ ] Refactor spec.ts, 	riage.ts, riefing.ts, and learn.ts to use this unified utility.
- [ ] Ensure consistent use of GH_TIMEOUT_MS and IS_WIN across all fetchers.

## Relationship
This addresses technical debt identified during the review of **Issue #18** (Learning Loop) to ensure architectural consistency across the CLI's data layer.

PR #1366, opened 2026-04-11
chore: postmerge lessons and rules from 1.14.3-1.14.5 cycle

satur8d:
Postmerge workflow for the 2026-04-11 PM marathon PRs: #1345, #1348, #1349, #1353, #1356, #1362. First production run of the new manual postmerge sequence after `totem wrap` retirement (#1362), and **Day Zero validation of the archive filter from #1345**.

## Summary

- **25 new lesson files** extracted via `totem lesson extract` (2 batches of 5 + 1, hitting the 5-PR cap in one session)
- **5 new rules compiled** via `totem lesson compile --export` (local, routed to Sonnet 4.6 via `anthropic:claude-sonnet-4-6`)
- **20 conceptual/architectural lessons** kept as memory-only context (not compilable into tripwire rules)
- **1 lesson failed to parse** — "Escape triple backticks in Markdown regex", a known compile-worker edge case already seen on the 1.14.1 postmerge
- **Rule count: 394 → 399 total** (1 → 5 archived, 393 → 394 active)

## Curation via the #1345 archive filter

Per the Option B strategy (archive, not delete) decided with Gemini during the postmerge session: of the 5 new rules, 4 were flagged as over-broad or hallucinated during post-compile human review and archived via `status: 'archived'` with detailed `archivedReason` fields. The 5th was kept active.

**This is the first production use of the archive mechanism on freshly compiled rules.** The `#1345` filter that shipped this morning is doing exactly what we built it for: letting us silence bad auto-compile output without destroying the provenance chain. The archived rules stay in `compiled-rules.json` with telemetry intact for future compile-worker prompt regression analysis.

### Archived rules (preserved for telemetry, not deleted)

| Hash | Lesson Heading | Reason |
|---|---|---|
| `48f755d9a388e87f` | Prefer regex over split for line extraction | Over-broad. `$STR.split('\n')[0]` fires on legitimate first-line extraction. Same class as #1352. |
| `0d0bbae4392c0255` | Lazy-load core libraries in CLI commands | **Hallucinated package name.** References `@mcp-b/core`, which does not exist in this repo. Canonical is `@mmnto/totem`. |
| `eb59a093e36afaf2` | Consolidate dynamic imports in CLI handlers | Matches every dynamic import in the CLI. Would produce thousands of false positives. |
| `3c4931d9071af448` | Never split error messages on periods | `$STR.split('.')` fires on filename extensions, version strings, and all dot-notation. |

### Active new rule (1 of 5)

| Pattern | Purpose |
|---|---|
| `spawn($CMD, $ARGS, { $$$BEFORE, shell: true, $$$AFTER })` | Encodes the #1329 `safeExec` security concern as a durable tripwire. The only direct path to prevent a future reintroduction of `shell: true` in spawn calls. |

## Empirical quality ratio

**1 usable rule out of 5 compiled = 20% pass rate.** Worse than the 1.14.1 cycle's 2/6 = 33%, and both are far below the aspirational "compile gate as quality filter" target. This is a real and compounding problem with the compile worker's prompt. Each archived rule's `archivedReason` is deliberately written to feed the prompt rewrite conversation tracked under Strategy #73 (compilation quality benchmark) and Strategy #62 (model-specific prompt adapters).

## Mechanical verification

- `totem verify-manifest` → **PASS** (399 rules, hashes match)
- `totem lint` → **PASS** (394 active rules, 0 violations)
- `totem review` → deterministic fast-path (non-code changes only)
- `prettier --check` → PASS (after explicit `--write` pass on the export files, which are auto-generated and don't always match prettier's style on first write)

## Workflow note (deferred to #1366)

This PR is the first run of the new manual postmerge sequence after #1362 retired `totem wrap`. During execution I discovered a gap in the `.claude/skills/postmerge/SKILL.md` step 4 ("Revert compiled-rules.json to the curated set"): reverting the rules file while keeping the new lessons creates a manifest inconsistency (the manifest's `input_hash` still reflects the new lessons but the rules file reverts to the old set). The Option B archive approach taken here sidesteps the bug entirely by mutating rather than reverting, keeping the manifest mathematically consistent throughout.

Filing #1367 to update SKILL.md step 4 from "revert" to "review and mark over-broad rules as archived inline", matching the precedent established by #1334 and validated by this PR.

## Files changed

- 25 `.totem/lessons/lesson-*.md` — new lessons
- `.totem/compiled-rules.json` — 5 rules added, 4 archived with detailed reasons
- `.totem/compile-manifest.json` — refreshed hash `afbfbe67…→b31fa321…`
- `.github/copilot-instructions.md` — auto-regenerated export (archived rules filtered out)
- `.junie/skills/totem-rules/rules.md` — auto-regenerated export
- `scripts/archive-bad-postmerge-rules.cjs` — new one-off mutation script preserved per the `scripts/` precedent (`clean-rules.js`, `kill-broad-astgrep.js`, `scope-rules*.js` are all prior one-off rule-mutation scripts in the same directory). Documents the exact archive operation for institutional memory.

## Test plan

- [ ] CI green on the branch
- [ ] `totem lint` on the branch shows 394 active rules (not 399)
- [ ] `git log --stat` shows 25 new lesson files + the export refreshes + the mutation script
- [ ] Archived rules' `archivedReason` fields render cleanly in `compiled-rules.json`


satur8d:
@coderabbitai @gemini-code-assist — both findings processed. One fixed inline (`bbac78aa`), one deferred with a scoped tracking ticket. Details below.

## @coderabbitai — use `lessonHash` for target matching

**Valid, fixed in `bbac78aa`.** You caught a real preventative hardening gap: I had the hashes right in front of me (my own script printed them during the first run) and still matched by `lessonHeading`, which is explicitly documented as auto-truncated and collision-prone in the extract pipeline. The script happened to work this cycle because my 4 targets had unique headings, but it was a latent correctness trap for any future postmerge run.

The rewrite:

- Matches each target by `lessonHash` (stable unique key)
- Builds a hash-keyed lookup map up front with explicit collision detection
- Collects all missing targets before aborting, so a partial compile cycle reports every missing hash at once instead of one per run
- Warns (not errors) if a target's `lessonHeading` no longer matches the heading currently in `compiled-rules.json` (hash stable but heading drifted)
- Verified idempotent against the already-archived state: running the hardened script against main now prints 4 "SKIP (already archived)" lines and exits cleanly with zero mutations

## @gemini-code-assist — `spawn` rule pattern is under-inclusive

**Technically correct, deferred via mmnto-ai/totem#1368.** You're right that the auto-compiled pattern misses:

- 2-argument form: `spawn(cmd, { shell: true })`
- Sync variants: `spawnSync`, `execSync`, `execFileSync`, `exec`
- Property-access forms: `cp.spawn`, `childProcess.spawn`

The source lesson (`.totem/lessons/lesson-7820f501.md`) is pure prose with no explicit `**Pattern:**` block, so it went through Pipeline 2 (LLM compiles natural language to ast-grep) and the compile worker produced the narrow pattern. Refining the pattern means either:

1. Adding an explicit manual `**Pattern:**` block to the source lesson (Pipeline 1)
2. Rewriting the prose to enumerate the idioms and re-running compile (Pipeline 2)

Both are in-scope for mmnto-ai/totem#1368. The ticket recommends **Option 1 (manual pattern)** because this is a security-critical tripwire and determinism matters more than prose elegance. Target: a 1.15.x tactical cycle.

**On the meta-point about editing `compiled-rules.json`:** the rule you cite is about content edits (pattern corrections, message tweaks) which should flow through the source lesson. The `status: 'archived'` mutations in this PR are **lifecycle state**, not content. The archive workflow is specifically what mmnto-ai/totem#1345 (landed this morning) was built to support — it lets us silence bad auto-compile output via lifecycle state changes without destroying the provenance chain. The `archivedReason` field carries the curation rationale in the ledger for future compile-worker prompt analysis. This distinction (lifecycle vs content) is worth codifying into the CLAUDE.md conventions; I'll surface it as a follow-up if the distinction turns out to confuse future bot reviews.

## Why the `spawn` rule stays active despite being narrow

Per the Option B philosophy validated on this PR: the narrow pattern still catches something (literal 3-arg `spawn(cmd, args, { shell: true })` calls, which was the actual bug site from mmnto-ai/totem#1329 / safeExec pre-fix). **Zero coverage is worse than narrow coverage.** The rule stays active while mmnto-ai/totem#1368 tracks the broadening work. When the refined pattern ships, the old narrow rule gets replaced in-place by the un-retirement compile, not archived.

## Follow-ups filed from this PR's bot round

- **mmnto-ai/totem#1368** — broaden the `spawn` rule pattern (GCA finding, deferred)

The CR finding is fixed inline, no ticket needed.

Issue #1337, opened 2026-04-11
No-op compile doesn't refresh manifest when a lesson is deleted with no stale prune trigger


satur8d:
## Summary

When a user deletes a lesson file that was already absent from the compile cache (or was manually removed from `compiled-rules.json`), `totem lesson compile` takes the no-op path, does not rewrite `compiled-rules.json`, and therefore does not refresh `compile-manifest.json`. The manifest's `input_hash` still reflects the pre-deletion lessons dir, so `verify-manifest` fails on the next `git push` with:

```
[totem] Push blocked: compile manifest is stale. Run 'totem lesson compile'.
```

`totem lesson compile` then runs again, hits the same no-op path, and does nothing. The only recovery today is `totem lesson compile --force`, which re-invokes the LLM for every lesson in the project (1,200+ lessons in our repo, ~19 minutes of wall clock) just to refresh a single SHA hash.

## Reproduction

1. Start from a clean working tree with a matching compile manifest.
2. Delete any `.totem/lessons/lesson-*.md` that is already absent from `.totem/compiled-rules.json` (either `nonCompilable` or the active rules).
3. Run `pnpm exec totem lesson compile`.
4. **Expected:** manifest's `input_hash` updated to reflect the new lessons dir.
5. **Actual:** compile reports `All N lesson(s) in scope already processed` and exits. `verify-manifest` fails.
6. Workaround: `totem lesson compile --force` — takes ~19 minutes on this repo and is non-deterministic (in our case, re-running compile dropped 393 → 386 rules because some lessons flipped from "compiled" to "failed" on the new LLM pass).

## Root cause

The #1281 fix that I shipped earlier in this release cycle added a no-op prune path to `packages/cli/src/commands/compile.ts` that looks like this:

```ts
if (toCompile.length === 0) {
  let reportedNonCompilable = nonCompilableMap.size;
  let reportedCompiled = existingRules.length;
  if (!options.raw) {
    const currentHashes = new Set(lessons.map((l) => hashLesson(l.heading, l.body)));
    const { fresh: freshRules, pruned: rulesPruned } = pruneStaleRules(existingRules, currentHashes);
    const { fresh: freshNonCompilable, drained } = pruneStaleNonCompilable(nonCompilableMap, currentHashes);
    if (rulesPruned > 0 || drained > 0) {
      saveCompiledRulesFile(rulesPath, { ... });
      // refresh manifest
      writeCompileManifest(manifestPath, { ... });
    }
  }
[...]

Issue #1221, opened 2026-04-05
feat: update cloud compile worker to route through Claude Sonnet


satur8d:
## Context

Strategy #73 benchmark showed Claude Sonnet 4.6 at 90% correctness / 2.4s avg vs Gemini Pro at 57-73%. PR #1220 routed local `totem compile` to `anthropic:claude-sonnet-4-6`.

The Cloud Run compile worker (`totem-compile-worker` in project `totem-488722`) still uses Gemini. Cloud compile is now less accurate than local compile.

## Proposed fix

Update the Cloud Run worker to:
1. Accept `model` field in the compile request payload (the CLI already sends it — line 368 of compile.ts)
2. Add `@anthropic-ai/sdk` dependency to the worker
3. Route to Claude Sonnet when the request specifies `anthropic:claude-sonnet-4-6`
4. Set `ANTHROPIC_API_KEY` as a Cloud Run secret

## Workaround

Until this ships, compile locally — Sonnet at 2.4s/lesson is fast enough for most runs. Cloud is only needed for bulk (100+ lessons).

## Relates to
- #1220 (routing change)
- Strategy #73 (benchmark evidence)
- #1212 (closed — local gemma4 not viable, but cloud+Sonnet is the real fix)

satur8d:
Closing as **orphaned by ADR-103** (Class 1 architecture pivot — Convergent Spine: DSL → Formal IR → SMT → WASM/NPM).

**Disposition rationale:** Cloud compile worker to Sonnet; the cloud compile worker is deprecated. There is no LLM-backed compile path under V1.

The LLM-compiler this issue depended on is being replaced by formally-verified deterministic enforcement. Per the Tenet 15 corollary ("LLMs may draft; constraints must verify"), the failure mode is structurally eliminated by the new architecture rather than patched in place.

**Rule intent preserved for Baseline 5 curation:** mmnto-ai/totem-strategy#288 (rule corpus migration triage) consults this ticket as source corpus when Baseline 5 selects the top ~30 invariants for hand-authored DSL.

**Refs:**
- mmnto-ai/totem-strategy/adr/adr-103-deterministic-rule-compilation.md
- mmnto-ai/totem-strategy/proposals/accepted/269-deterministic-rule-compilation.md
- mmnto-ai/totem-strategy/audits/internal/2026-05-11-post-adr-103-roadmap-disposition.md
- mmnto-ai/totem-strategy#288 — Baseline 5 rule corpus migration tracking

[...]

PR #1369, opened 2026-04-11
docs(skill): replace postmerge SKILL.md step 4 revert with inline archive workflow

### LINKED ISSUE #1367: docs(skill): postmerge SKILL.md step 4 'revert compiled-rules.json' creates manifest inconsistency — replace with inline archive workflow
## Problem

The `.claude/skills/postmerge/SKILL.md` step 4 currently says:

> 4. Revert `compiled-rules.json` to the curated set. Compile will produce new rules from the extracted lessons, but the curated set in main is the source of truth. The new rules should be reviewed and cherry-picked by hand rather than auto-merged (empirically 4/6 auto-compiled rules from the 1.14.1 postmerge were bad, and mmnto-ai/totem#1349 now catches the syntactic-invalid cases but not the over-broad ones):
>    `git checkout HEAD -- .totem/compiled-rules.json`

**This creates a manifest inconsistency.** Reverting `compiled-rules.json` while keeping the new lessons on disk leaves the `compile-manifest.json` file in a broken state: its `input_hash` reflects the new lessons but its `output_hash` reflects the reverted rules. Running `totem verify-manifest` fails on push.

Discovered during mmnto-ai/totem#1366 (2026-04-11 PM postmerge for PRs #1345, #1348, #1349, #1353, #1356, #1362), the first production run of the manual postmerge sequence after `totem wrap` retirement (#1362).

## The gap is symmetric to #1337

`#1337` was the "lessons deleted but not in cache, manifest stays stale" bug. The SKILL.md revert step creates the opposite case: "rules reverted but lessons still present, manifest stays stale." Both have the same remediation at the manifest layer (refresh the manifest to match current state), but the SKILL.md instruction leads the user into a state that cannot be committed cleanly without surgical intervention.

## Precedent

The 1.14.1 postmerge (`0c1a0f75 chore: postmerge lessons and rules from 1.14.1 cycle (#1334)`) **did not follow the revert instruction**. Matt's commit message on that one explicitly described the curation as an inline process: reviewing rules by hand, deleting one, keeping another, leaving the Pipeline 5 output intact. The #1334 outcome validated "keep and curate" over "revert."

The 2026-04-11 PM postmerge (mmnto-ai/totem#1366) pushed this pattern further by using the `#1345` archive filter (which only shipped that morning) to silence bad rules via `status: 'archived'` rather than deleting them. This is the Option B approach from the session conversation. It preserves provenance, keeps telemetry intact for the compile-worker prompt rewrite, and sidesteps the manifest consistency bug entirely.

## Proposed fix

Replace SKILL.md step 4 with an inline-archive instruction:

> 4. **Review the newly compiled rules.** Run `node -e 'console.log(JSON.stringify(require("./.totem/compiled-rules.json").rules.slice(-N), null, 2))'` (where N is the number of new rules reported by step 3) to see exactly what compiled. For each rule, verify:
>
>    - The `astGrepPattern` is not over-broad (does not fire on legitimate patterns the rule is not trying to flag)
>    - The `astGrepPattern` does not reference hallucinated package names, type names, or file paths
>    - The `lessonHeading` accurately describes the rule's behavior
>
>    For any rule that fails these checks, mutate `compiled-rules.json` to add `status: 'archived'` and a detailed `archivedReason` explaining the failure mode. The mmnto-ai/totem#1345 archive filter in `loadCompiledRules` will silence the rule at lint time while preserving it in the ledger for future compile-worker prompt regression analysis. See `scripts/archive-bad-postmerge-rules.cjs` on #1366 for the canonical script shape.
>
> 5. **Refresh the manifest.** After any mutation to `compiled-rules.json`, re-run `totem lesson compile --export` (which will detect the rules file drift and refresh `compile-manifest.json` automatically, using the #1348 no-op refresh path) AND regenerate the copilot/junie exports so they filter the newly archived rules.

This replaces the broken revert path with a workflow that is:

- **Mathematically consistent** — manifest stays in sync throughout
- **Telemetry-preserving** — bad rules stay in the ledger with `archivedReason` fields that feed the prompt rewrite conversation
- **Single-PR** — no "land and then clean up" two-PR dance
- **Precedent-matching** — validates the #1345 archive mechanism as the intended curation surface

## Return conditions

None. This is a pure documentation fix. The code path (archive filter, manifest refresh, exports regeneration) already exists and works correctly.

## Related

- mmnto-ai/totem#1345 — The archive filter that made this workflow possible
- mmnto-ai/totem#1348 — The no-op manifest refresh that makes the re-compile step work
- mmnto-ai/totem#1362 — The `totem wrap` retirement that required rewriting postmerge/SKILL.md in the first place
- mmnto-ai/totem#1366 — The postmerge PR where this gap was discovered and the correct workflow was validated
- Precedent: `0c1a0f75 chore: postmerge lessons and rules from 1.14.1 cycle (#1334)` — the earlier postmerge that already deviated from the revert instruction in practice

satur8d:
Closes #1367

## Summary

- Replace `.claude/skills/postmerge/SKILL.md` step 4 (revert `compiled-rules.json`) with an inline-archive workflow that uses the `#1345` archive filter as the intended curation surface.
- Add a new step 5 that re-runs `totem lesson compile --export` after the archive mutation to refresh the manifest via `#1348`'s no-op refresh path and regenerate the copilot/junie exports.
- Renumber steps 5-8 to 6-9 and expand the staging command in step 7 to include `compiled-rules.json` + `compile-manifest.json`.
- Add an explicit "Do NOT revert" callout naming the manifest inconsistency failure mode and the symmetric relationship to `#1337`.

## Why

Surfaced empirically while running the 2026-04-11 PM postmerge on `#1366` (the first production run of the manual sequence after `totem wrap` retirement on `#1362`). The old step 4 said to revert `compiled-rules.json` via `git checkout HEAD -- .totem/compiled-rules.json`, but reverting the rules file while keeping the new lessons on disk creates a manifest inconsistency:

- `manifest.input_hash` reflects the new lesson set
- `output_hash` reflects the reverted rules
- `verify-manifest` fails on push

This is the symmetric counterpart of the `#1337` bug we fixed this morning (`#1337` was "lessons deleted but manifest stays stale"; the revert path creates "rules reverted but lessons still present"). The old skill instruction led users directly into a state that cannot be committed cleanly without surgical intervention.

The `#1334` precedent already deviated from the revert instruction in practice, and `#1366` extended the curation approach by using the `#1345` archive filter for Day Zero silencing of bad auto-compile output. This PR codifies that precedent into the skill so the next person running `/postmerge` does not re-discover the bug the hard way.

## The three validation checks in the new step 4

The new step 4 tells the user to inspect each newly compiled rule against three concrete checks:

1. The `astGrepPattern` or `pattern` is not over-broad (does it fire on legitimate code the rule is not trying to flag?)
2. The pattern does not reference hallucinated package names, type names, or file paths that do not exist in the repo
3. The `lessonHeading` accurately describes the rule's behavior

All three checks were used on `#1366` to identify the 4 rules that got archived. The failure modes they caught map directly to the archivedReason fields we wrote: over-broad split rule (check 1), hallucinated `@mcp-b/core` package (check 2), over-broad dynamic import pattern (check 1), over-broad dot-split (check 1).

## The archive mutation pattern

Step 4 cites `scripts/archive-bad-postmerge-rules.cjs` (committed on `#1366`) as the canonical mutation script shape. That script was hardened during the `#1366` bot review round to match target rules by `lessonHash` instead of `lessonHeading` (per CodeRabbit's finding that headings are auto-truncated and collision-prone). The SKILL.md incorporates that learning inline so future scripts do not repeat the mistake.

## The new step 5

After any mutation to `compiled-rules.json`, re-run `totem lesson compile --export`. This triggers the `#1348` no-op refresh path to:

- Update `compile-manifest.json` hashes to match the mutated rules file
- Regenerate `.github/copilot-instructions.md` and `.junie/skills/totem-rules/rules.md` so they filter the newly archived rules out of the exported markdown

Previously this refresh was implicit in step 3 of the 1.14.1-era skill, but when the step pattern changed to "compile once, then curate, then refresh," the refresh became a distinct operation that deserved its own step.

## What the skill still does

- Still uses `totem lesson extract` and `totem lesson compile --export` (canonical noun-verb forms from `#1362` / `#1366` bot rounds)
- Still says to not pass `--cloud` per the `#1221` cloud compile restriction
- Still explicitly names the three doc targets that should NOT be staged unless hand-edited (`docs/active_work.md`, `docs/roadmap.md`, `docs/architecture.md`) because those are `totem docs` rewrite targets
- Still cites the `totem wrap` retirement (`#1361`) so future users understand why the manual sequence exists

## Test plan

- [ ] CI green on the branch (docs-only, should be fast)
- [ ] `totem review` fast-paths (no code changes)
- [ ] `totem lint` reports no changes detected (`.claude/` is in ignorePatterns)
- [ ] The next `/postmerge` invocation produces a cleaner run than #1366 did, because the workflow bug has been codified out

### DIFF
--- a/.claude/skills/postmerge/SKILL.md
+++ b/.claude/skills/postmerge/SKILL.md
@@ -70,13 +70,18 @@ with the merged PR numbers (space-separated, e.g. `1345 1347 1348`).
    `pnpm run format`
 
 7. Stage only the artifacts we keep: new lessons, the mutated rules
-   file, the refreshed manifest, the regenerated exports, and any
-   curation scripts preserved for institutional memory. Do NOT stage
-   `docs/active_work.md`, `docs/roadmap.md`, or
+   file, the refreshed manifest, and the regenerated exports. Do NOT
+   stage `docs/active_work.md`, `docs/roadmap.md`, or
    `docs/architecture.md` unless you hand-edited them deliberately
    (those are `totem docs` targets and a postmerge run should not
    rewrite them):
    `git add .totem/lessons/ .totem/compiled-rules.json .totem/compile-manifest.json .github/copilot-instructions.md .junie/skills/totem-rules/rules.md`
+
+   If you wrote a one-off curation script during step 4 (as was done
+   on mmnto-ai/totem#1366 with `scripts/archive-bad-postmerge-rules.cjs`),
+   stage it explicitly so the institutional memory lands with the
+   commit:
+   `git add scripts/archive-bad-postmerge-rules.cjs  # or whatever you named yours`
 
 8. Commit:
    `git commit -m "chore: totem postmerge lessons for $ARGUMENTS"`
--- a/.claude/skills/postmerge/SKILL.md
+++ b/.claude/skills/postmerge/SKILL.md
@@ -22,29 +22,69 @@ with the merged PR numbers (space-separated, e.g. `1345 1347 1348`).
    mmnto-ai/totem#1221. Local compile routes to Sonnet 4.6:
    `pnpm exec totem lesson compile --export`
 
-4. Revert `compiled-rules.json` to the curated set. Compile will
-   produce new rules from the extracted lessons, but the curated set
-   in main is the source of truth. The new rules should be reviewed
-   and cherry-picked by hand rather than auto-merged (empirically
-   4/6 auto-compiled rules from the 1.14.1 postmerge were bad, and
-   mmnto-ai/totem#1349 now catches the syntactic-invalid cases but
-   not the over-broad ones):
-   `git checkout HEAD -- .totem/compiled-rules.json`
+4. Review the newly compiled rules. Step 3's output prints a count
+   like `N/M (100%) ... X compiled, Y skipped, Z failed`. For each
+   of the X newly compiled rules, inspect them in
+   `.totem/compiled-rules.json` and verify:
+   - The `astGrepPattern` or `pattern` is not over-broad (does it
+     fire on legitimate code the rule is not trying to flag?)
+   - The pattern does not reference hallucinated package names, type
+     names, or file paths that do not exist in the repo
+   - The `lessonHeading` accurately describes the rule's behavior
 
-5. Format everything wrap might have touched:
+   For any rule that fails these checks, mutate
+   `.totem/compiled-rules.json` to add `status: 'archived'` and a
+   detailed `archivedReason` explaining the specific failure mode.
+   The mmnto-ai/totem#1345 archive filter in `loadCompiledRules`
+   silences archived rules at lint time while preserving them in the
+   ledger for future compile-worker prompt regression analysis. See
+   `scripts/archive-bad-postmerge-rules.cjs` (committed on
+   mmnto-ai/totem#1366) for the canonical mutation script shape.
+   Match target rules by `lessonHash`, not `lessonHeading` (headings
+   are auto-truncated to 60 chars and are not guaranteed unique).
+
+   **Do NOT** use `git checkout HEAD -- .totem/compiled-rules.json`
+   to revert the entire rules file. Reverting rules while keeping
+   the new lessons on disk creates a manifest inconsistency
+   (manifest.input_hash reflects the new lessons, output_hash
+   reflects the reverted rules, verify-manifest fails on push). This
+   is the symmetric counterpart of the mmnto-ai/totem#1337 bug.
+   Archive-in-place is the intended curation surface; reverting is
+   not.
+
+   Empirical baseline: approximately 2 of every 6 auto-compiled
+   rules are bad (1.14.1 postmerge), and the 2026-04-11 PM postmerge
+   hit 4 of 5. The compile-worker prompt rewrite conversation lives
+   under Strategy #73 and Strategy #62. Every archivedReason is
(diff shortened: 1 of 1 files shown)


satur8d:
@coderabbitai — valid nit, fixed in \`b1fb23fd\`.

You caught a real prose/command mismatch. Step 7 mentioned staging "any curation scripts preserved for institutional memory" but the concrete \`git add\` command did not include any script path, so the text implied a staging action the one-liner did not actually perform.

Fixed by splitting step 7 into two staging commands:

1. **Always runs** — stages the canonical postmerge artifact set (lessons + mutated rules + refreshed manifest + regenerated exports)
2. **Runs only if applicable** — explicit optional command for curation scripts with a concrete example pointing at \`scripts/archive-bad-postmerge-rules.cjs\` (the mmnto-ai/totem#1366 precedent)

The split makes both the prose and the commands mechanically accurate: the main command always runs, the script command runs only when a mutation script was written during step 4. No more ambiguity about whether "stage any curation scripts" is a do-it or a describe-it.

Thanks for the catch — this kind of prose-vs-command drift is exactly what I wanted to avoid in a skill file that's meant to be followed literally on the next postmerge cycle.

Issue #1386, opened 2026-04-13
Export pipeline corrupts glob wildcards and emphasis markers in copilot/junie markdown


satur8d:
## Problem

`totem lesson compile --export` renders lesson globs and tags incorrectly in the copilot-instructions.md and junie rules.md exports:

1. **Glob wildcards mangled**: `!**/*.test.*` becomes `!**/_.test._` -- the `*` characters are being consumed by markdown formatting instead of escaped
2. **Trailing emphasis broken**: Tag lists like `_(linting, documentation)_` render as `_(linting, documentation)\*` with a stray escaped asterisk, triggering markdownlint MD037 (spaces inside emphasis markers)

Both issues are in the export renderer, not in the source lessons or compiled rules. The source data is correct; only the markdown serialization is wrong.

## Reproduction

Run `totem lesson compile --export` on any lesson with glob patterns containing `*` in its scope (e.g., `docs/wiki/*.md` or `!**/*.test.*`). Inspect the generated `.github/copilot-instructions.md`.

## Impact

- CR flags these on every postmerge PR (#1384, #1385) as formatting issues
- The corrupted globs mean AI agents reading the exported instructions get wrong scope information
- The broken emphasis causes markdownlint warnings

## Related

- #1384, #1385 (bot comments flagging these as findings on postmerge PRs)
- The export renderer lives in `packages/cli/src/commands/compile.ts` (the `--export` code path)

PR #1403, opened 2026-04-14
chore: 1.14.8 release - Perf Follow-up

satur8d:
## Mechanical Root Cause

Closes the 1.14.x cycle with the final patch. Two compile-path perf/correctness fixes from #1401 and the PR template from #1402 are ready to ship.

## Fix Applied

- Release changeset for 1.14.8 (patch bump, all three packages in the fixed group)
- 7 lessons extracted from #1401 and #1402
- 1 rule compiled (over-broad `process.cwd()` pattern) and archived with `upgradeTarget: compound` per Proposal 226
- `docs/active_work.md` updated with 1.14.8 summary
- Export files regenerated

## Out of Scope

- Starting 1.15.0 Pack Distribution work. That waits for the Pre-1.15 deep review gate.

## Tests Added/Updated

- [x] N/A - this is a release prep PR, no code changes. 2785 tests pass from the commits being released.

## Related Tickets

No tracking issue. Release flow for the merged #1401 and #1402 PRs.


satur8d:
Valid nit. You correctly identified that this must be fixed at the exporter layer, not in the generated file. Tracked as #1404. Builds on #1386/#1388 which fixed glob wildcard escaping; #1404 covers the remaining underscore path where `pull_request_template.md` in the Scope field creates MD037 artifacts. Not blocking this release since the generated file is regenerated on every `compile --export` and the fix will flow through automatically.

Issue #1441, opened 2026-04-15
core: shared mutable global state in rule-engine.ts breaks concurrent safety


satur8d:
## Mechanical Root Cause

`packages/core/src/rule-engine.ts` declares `shieldContextDeprecationWarned` and `coreLogger` as top-level mutable `let` variables (module-level state). When `applyAstRulesToAdditions` or rule executions run concurrently (batch PR lint across workers, async ticks during a federated search), the shared state leads to interleaved logger configurations and race conditions on deprecation warnings.

## Why This Matters

Breaks functional isolation of the AST linting phase. Rule execution is supposed to be pure per-rule — a rule evaluates additions and returns violations. Shared mutable state between rule evaluations means one rule's side effect (warn once, swap logger) can bleed into another rule's behavior.

Pack Distribution concern: when consumers federate across packs, their rules run through this same engine. Shared state between packs is a trust violation — pack A can influence pack B's logging behavior.

## Fix

1. Move `coreLogger` into a per-call parameter or a module-scoped `Context` passed through every public entry point.
2. `shieldContextDeprecationWarned` should be per-call state, not module-level. A warning cache tied to the call's lifetime (or just emit every time — deprecation warnings are cheap).

## Related

Candidate for Gemini's proposed compound ast-grep rule "top-level mutable state ban" — a rule flagging `let` at module scope would catch this class.

## Source

Pre-1.15.0 deep review (#1421), Gemini audit (Finding #3). `.strategy/audits/internal/2026-04-14-repomix-architectural-teardown.md`.

Issue #1421, opened 2026-04-15
chore: pre-1.15.0 deep code review (4-surface independent pass before Pack Distribution starts)


satur8d:
## Context

Before the first 1.15.0 Pack Distribution ticket touches code, run a four-surface independent deep review on `main` to catch accumulated drift from the 1.14.x cycle. The 1.14.x arc moved fast: 11 PRs in the quality sweep, plus the #1406 → #1407 → #1408 → #1409 Precision Engine arc with its compound schema + smoke gate + prompt rewrite triangle. Per-PR review catches local issues; fresh eyes catch compound drift that per-PR cycles miss.

Pack Distribution is the first time rules leave our repo. Every pack reflects the current codebase state as a foundation. Foundation bugs get distributed to every downstream consumer. Catching at the foundation layer is orders of magnitude cheaper than retro-fixing every pack consumer.

This ticket tracks the review gate and its findings.

## Trigger

Blocked by 1.14.9 release cutting. Unblocks the moment 1.14.9 ships and BEFORE any 1.15.0 implementation ticket is picked up.

Current 1.15.0 backlog (do NOT start these until this ticket is closed or explicitly parked):

- #1059 — Pack Distribution Gap: Playground lessons should be installable as `@totem/pack-nextjs-security`
- #1221 — Route cloud compile worker through Claude Sonnet
- #1243 — Totem Pack package schema (ADR-085 headline)
- #1244 — `extends` array in `totem.config.ts` to consume packs (ADR-085 headline)
- #1245 — Merge extended pack rules with local rules at lint time (ADR-085 headline)
- #1246 — Initial pack trust mechanism — explicit-trust list (ADR-085 headline)
- #766 — `totem install pack/*` registry command
- #767 — Rule pack format spec
- #913 — Expand baseline packs with community-sourced lessons
- Strategy #35 — Distributing Totem rules as Agent Skills

## Scope

Four independent passes on current `main`, none informed by the others:

1. **Claude review** — full architectural pass. Scope: compile pipeline, engine runtime, smoke gate, mesh, MCP surface, test coverage.
2. **Gemini review** — full architectural pass. Same scope.
3. **CodeRabbit review** — triggered via a trivial wrapper PR that forces CR across the whole diff surface. Can be a doc-only PR touching `docs/architecture.md` or `README.md`.
4. **GCA review** — same wrapper PR.

Each reviewer produces a findings list. Findings triage happens in this ticket's comment thread:
[...]

Issue #1407, opened 2026-04-14
feat(core): extend CompiledRule schema with astGrepYamlRule field


satur8d:
## Context

Proposal 226 requires `CompiledRule` to hold compound ast-grep rules alongside the existing flat `astGrepPattern` string. This ticket lands the schema change and extends compile-time validation.

## Dependencies

Blocked by the `@ast-grep/napi` YAML validation spike (first ticket in this milestone). We need the spike's findings to confirm the exact shape of YAML rule objects before locking the schema.

## Scope

1. Add `astGrepYamlRule?: AstGrepYamlRule` to `CompiledRule` (where `AstGrepYamlRule` is a Zod schema covering the YAML rule object shape validated by the spike).
2. Preserve `astGrepPattern` as-is. The two are mutually exclusive per rule; validator rejects rules that set both or neither for ast-grep-mode rules.
3. Extend `validateAstGrepPattern` (introduced in #1349) to accept and validate YAML rule objects via `@ast-grep/napi`.
4. Ensure compile manifest hash input includes `astGrepYamlRule` content so manifest drift detection (#1337 / #1348) still works for YAML rules.

## Out of Scope

- Engine runtime execution of compound rules (separate ticket).
- Compiler prompt tuning (separate ticket).
- Bulk recompile of archived rules tagged `upgradeTarget: compound` (phase 4 on the epic).

## Acceptance Criteria

- [ ] `CompiledRule` schema compiles with strict TypeScript; new field documented in schema comments.
- [ ] Validator rejects rules with both `astGrepPattern` and `astGrepYamlRule` set with a fail-loud `TotemError`.
- [ ] Validator rejects syntactically invalid YAML rules via the spike-proven path.
- [ ] Compile manifest hash input includes `astGrepYamlRule` when present.
- [ ] Unit tests cover both the dual-field rejection and the YAML validation path.

## References

- Epic: mmnto-ai/totem-strategy#81
- Proposal: `.strategy/proposals/active/226-compound-ast-grep-rules.md`
- Prior validation work: #1349

satur8d:
## Design notes from the #1406 spike (pre-preflight)

Two constraints for the Phase 3 design doc when `/preflight 1407` runs:

### 1. Compiler output schema now requires `badExample` (locked by Gemini)

[...]

PR #1454, opened 2026-04-15
feat(core): Pipeline 1 compound rules + fail-loud audit fixes (closes #1440, #1442)

### LINKED ISSUE #1440: core: fail-open empty catches in sys/git.ts swallow git binary failures (Tenet 4)
## Mechanical Root Cause

`packages/core/src/sys/git.ts` — methods `isFileDirty`, `resolveGitRoot`, `getGitBranch`, `getGitStatus` swallow `safeExec` throwables inside empty `catch {}` blocks. If the `git` binary fails (OOM, missing permissions, corrupted `.git/index`), `isFileDirty` silently returns `false` (treated as clean).

## Why This Matters

Tenet 4 (Fail Loud, Never Drift) violation at foundation utilities. Totem's deterministic gate depends on accurate git state. Swallowing errors and returning a "safe" state fails open — a corrupted `.git/index` becomes indistinguishable from a clean tree. Any caller that branches on `isFileDirty` returning `false` can now take the clean-tree path against corrupted state.

Compound concern with Pack Distribution: if a pack install writes to the working tree and something corrupts the index, downstream checks that rely on `isFileDirty` silently succeed while the repo is actually broken.

## Fix

All four methods should preserve cause chains and throw `TotemGitError` with an appropriate code. Callers that currently treat "clean" as the default on failure should explicitly handle the error rather than hiding it in the utility.

Candidate for Gemini's proposed compound ast-grep rule "fail-open catch ban" — a rule that flags `catch_clause` without a `throw_statement` inside it would catch this class at lint time.

## Source

Pre-1.15.0 deep review (#1421), Gemini audit (Finding #2). `.strategy/audits/internal/2026-04-14-repomix-architectural-teardown.md`.

### LINKED ISSUE #1442: core: regex compilation failure silently continues in applyRulesToAdditions (Tenet 4)
## Mechanical Root Cause

`packages/core/src/rule-engine.ts` — inside `applyRulesToAdditions`, a `new RegExp(pattern)` that throws because the pattern is malformed gets caught in a `catch` block that simply `continue`s to the next rule.

A corrupt compiled manifest entry (hand-edited, mid-migration, or bypassing the validation gate) means the execution engine silently IGNORES the rule instead of crashing. The lint verdict reports "PASS — 0 violations" when the reality is "PASS — 1 rule couldn't evaluate."

## Why This Matters

Another Tenet 4 fail-open. The whole premise of the deterministic lint engine is "if a rule is installed, it evaluates." Silently skipping a rule because its pattern is corrupt means the gate is lying about coverage.

Pack Distribution concern: if a pack ships a rule that happens to be corrupt under a specific Node version, downstream consumers get silent degradation with no signal.

## Fix

Throw loud with a `TotemCompileError` carrying the rule's `lessonHash` and the underlying RegExp error message. Upstream callers (lint CLI, CI workflow) should surface this as a failure: the governance gate cannot verify against a corrupt rule.

Alternative: report it as a first-class diagnostic in the lint output (parallel to errors and warnings — a new "skipped-due-to-error" category) so the user at least sees "rule X couldn't run."

## Source

Pre-1.15.0 deep review (#1421), Gemini audit (Finding #4). `.strategy/audits/internal/2026-04-14-repomix-architectural-teardown.md`.

satur8d:
## Summary

- Extend Pipeline 1 to parse yaml-tagged fenced code blocks after `**Pattern:**` into `astGrepYamlRule` objects. Zero-LLM compound rule authoring for pack authors.
- Ship the first three production compound ast-grep rules (fail-open catch ban, spawn-with-shell-true ban, export-let ban) via the new path.
- Fix the two pre-1.15.0 deep review audit findings the new rules were motivated by: #1440 (`isFileDirty` / `resolveGitRoot` silent fallbacks) and #1442 (regex-engine rule swallow).

## Background

First production use of 1.14.9's compound ast-grep rule support (`astGrepYamlRule` on `CompiledRule`). Until this PR, `compiled-rules.json` had zero compound rules and Pipeline 1 (manual, zero-LLM) had no path to author them — only Pipeline 2 (LLM-driven) could emit `astGrepYamlRule`. That made compound rule precision dependent on Sonnet's YAML output on the first try, which is a shaky footing for the rules we want to be the ones enforcing architectural invariants.

Gemini's pre-1.15.0 deep review (#1421) surfaced the RCE we shipped as 1.14.10 (PR #1429), plus four critical architectural findings. Three of those findings are mechanically fixable with compound ast-grep rules — this PR ships those rules and, while the compiler is hot, closes the specific #1440 / #1442 audit findings the rules are the general form of.

### Pipeline 1 compound extension

- New `extractYamlRuleAfterField(body, field)` export on `@mmnto/totem`. Scans for ` ```yaml ` / `~~~yaml` fences below a bold field marker, stops at next field marker or EOF, rejects bare untagged fences so prose code blocks pass through.
- `ManualPattern.astGrepYamlRule?: Record<string, unknown>` field. Mutually exclusive with `pattern`. Only valid when `engine: ast-grep`.
- `buildManualRule` routes the parsed object through `validateAstGrepPattern` (which already accepts object shapes from the #1407 schema work) and then `engineFields` to produce a `CompiledRule` with `astGrepYamlRule` set — mirroring the Pipeline 2 LLM path.

### Three inaugural compound rules

All authored in `.totem/lessons/*.md` via the new yaml-fence path. All manual Pipeline 1. All compile cleanly into `compiled-rules.json` (now 414 rules, +3 compound).

- **Ban fail-open catch blocks** — `kind: catch_clause` + `not.has.throw_statement (stopBy: end)`. Scope: `packages/**/*.ts`, test files excluded. Escape hatch: `// totem-ignore-next-line` above the intentional cleanup catch.
- **Ban spawn / spawnSync with `shell: true`** — `any:` disjunction over the two call names + `has:` descendant check on a `shell: true` pair. Scope excludes `packages/cli/src/orchestrators/shell-orchestrator.ts` (the one legit site, secured by `MODEL_SAFE_RE` / `quoteShellArg` in 1.14.10).
- **Ban `export let` at module level** — `kind: export_statement` with a `lexical_declaration` child matching `let`. Zero current violations; forward protection ahead of 1.15.0 Pack Distribution.

### Audit fixes (closes #1440, #1442)

- `isFileDirty(cwd, filePath)` — previously `return false` on git failure. Now throws `TotemGitError` with cause chain. `docs` data-loss-protection filter can no longer mistake a broken git for a clean working tree.
- `resolveGitRoot(cwd)` — previously `return null` on any git failure, conflating "not in a repo" with "git broke." Now walks the error cause chain via `containsNotAGitRepo` (bounded depth 8, coerces non-Error causes to string) and returns `null` only for the documented "not a git repository" case. All other failures throw.
- `applyRulesToAdditions` regex branch — previously `continue` on invalid pattern. Now throws `TotemError('CHECK_FAILED')` naming the lesson hash and pattern. A corrupted-manifest entry can no longer silently mute a load-bearing rule.

Cosmetic git helpers (`getGitBranch`, `getGitStatus`, `getGitDiffStat`, `getTagDate`, `getLatestTag`, `getGitLogSince`, `getDefaultBranch` probe loop) keep their silent-fallback behavior but now carry `// totem-ignore` comments documenting why the silence is intentional for each site.

### Tests

- +11 new tests in `lesson-pattern.test.ts` for the yaml-fence extractor and the compound manual path (fence-style variants, malformed YAML rejected, next-field terminator, CRLF handling, Engine mismatch rejection).
- +3 new `resolveGitRoot` tests covering success / narrow-null / throw-on-other-failure.
- Two existing tests (isFileDirty error path, applyRules invalid regex) updated to assert the new throw behavior with issue references in the test names.

Total: 2894 tests passing across core (1134 → 1149), cli (1662), mcp (83). Net +15 from the 1.14.9 baseline.

## Test plan

- [x] `pnpm -r test` — all 2894 tests pass
- [x] `pnpm exec totem lint` — passes diff-only (0 errors, 9-28 warnings all pre-existing regex-related noise on the new regex code)
- [x] `pnpm exec totem review` — Shield standard review PASS, no issues
- [x] `pnpm exec totem lesson compile` — three compound rules produced via Pipeline 1, verified in `compiled-rules.json` with correct `astGrepYamlRule` shapes
- [x] Manually verified scope exclusion on Rule 2 works (shell-orchestrator.ts unaffected)
- [x] Manually verified Rule 1 fires on a real new fail-open catch in this very diff (YAML.parse catch in `lesson-pattern.ts`) — fixed with `totem-ignore-next-line` per Gemini's Option C guidance

## Out of scope

- The remaining ~194 legacy fail-open catches across the codebase. Per Totem's "deterministic constraint at the point of change" tenet, `totem lint` is diff-only: legacy violations only fire on PRs that touch those files, and get fixed gradually. Full sweep is not blocking.
- The rule-engine.ts shared mutable state concern from Gemini's audit finding #3 is tracked separately in #1441 — it requires a runtime refactor (class extraction / param-passing), not a static AST rule.

Closes #1440
Closes #1442


satur8d:
@gemini-code-assist @coderabbitai — rebased on main (post-#1429) and addressed every finding from this round in `ea452bb4`. Consolidated reply:

### CI failures (now fixed)

- **Totem Lint drift** — dropped `.strategy/audits/internal/...` path references from lesson bodies. The audit file exists only in a local strategy working copy and was never committed upstream, so CI's drift detector flagged both lessons. Kept the `pre-1.15.0 deep review (#1421)` attribution; the path reference was cosmetic.
- **Wind Tunnel SHA** — regenerated `.wind-tunnel-sha` after fixture renames + real hit/miss content. New hash: `d43c5d1a`.

### GCA findings

- **`git.ts:213, :247` + `rule-engine.ts:226`** — dropped `: ${msg}` concatenation from error messages. Cause chain carries the original; no more stack-trace obscuring.
- **`rule-engine.ts:226`** — switched `TotemError('CHECK_FAILED')` → `TotemParseError`. Invalid regex in a compiled rule is a parse/compilation failure on the rule source, not a generic check failure. `TotemParseError` names it correctly.
- **`lesson-pattern.ts:245`** — switched `// totem-ignore-next-line` → `// totem-context:` per the repo convention (ticket-free when the rationale is in-band).

### CR findings

- **Lesson heading length** — both lessons truncated to ≤60 chars. `export let` → `Ban \ module-level mutable state` (48 chars); fail-open → `Ban fail-open catch blocks that skip re-throwing` (48 chars).
- **YAML scan at markdown headings** — `extractYamlRuleAfterField` now terminates on `#{2,6}\s` too. Two new unit tests pin the behavior. Without this, a lesson that omitted `**Message:**` and jumped to `### Bad Example` with a yaml fence would have the illustration block parsed as the rule body.
- **Test fixtures** — replaced all three TODO placeholders with real hit/miss cases (export-let, fail-open-catch, spawn-shell-true). Lesson-heading shortening changed the lesson hashes, so the old auto-scaffolded fixtures were removed and the new ones filled in with real content.
- **`git.ts` `totem-ignore` comments** — seven cosmetic-fallback sites migrated to `// totem-context:` with inline `mmnto/totem#1440` references. ADR-071 context directives provide visibility without requiring an external ticket gate, matching the `lesson-pattern.ts` update for consistency.

### Side-cleanup

One Pipeline 5 observation auto-captured during my review cycle was archived in-place (same failure mode as #1429 — root cause tracked in mmnto/totem#1279). Manifest reconciles at rule_count 414 / rules.length 414.

Consolidated reply per GCA quota convention.

satur8d:
@gemini-code-assist @coderabbitai — round 3 findings all addressed in `a64e1995` + `1d2dd5ee`. One consolidated reply:

### CR findings

- **`lesson-fail-open-catch-ban.md:22` (major)** — rule message was pointing callers at unticketed `// totem-ignore: intentional cleanup`, which would push devs into the same policy violation the rule is trying to ban. Migrated to `// totem-context: intentional cleanup` to match the ADR-071 migration in `sys/git.ts`.
- **`lesson-pattern.ts:301` (major)** — `extractManualPattern` now throws `TotemParseError` when a yaml-tagged `**Pattern:**` fence is present but `**Engine:**` is not `ast-grep`. A lesson that wrote yaml clearly intended Pipeline 1 compound compilation; silently falling through to Pipeline 2/3 masked the authoring typo. Test flipped from `toBeNull()` to `toThrow()`.
- **`sys/git.ts:200` (minor)** — `isFileDirty` docstring updated to reference `// totem-context:` for consistency with the rest of the file's migrated annotations.
- **`sys/git.ts:213` (minor — not taken)** — `[Totem Error]` prefix is auto-prepended by the `TotemError` base-class constructor (`errors.ts:40`: `super(\`[Totem Error] ${message}\`, { cause })`). Throwing `new TotemGitError('Failed to check dirty status...')` already produces `[Totem Error] Failed to check dirty status...` as the rendered message. Leaving the message strings without the manual prefix so we don't double-prefix.

### GCA findings

- **`sys/git.ts:216` (high)** — real regression. `isFileDirty` was throwing on `not a git repository`, which is a legitimate state for non-git working directories. Applied the narrow-false pattern from `resolveGitRoot`: use `containsNotAGitRepo(err)` to return `false` only in that specific case; throw for all other git failures. New unit test pins the behavior.
- **`lesson-pattern.ts:195` (high)** — switched the `safeField` regex-escape in `extractYamlRuleAfterField` to a replacer function (matches the `shell-orchestrator.ts` convention). Note: the `'\\$&'` form was actually intentional (`$&` -> matched-char IS the idiom for "prefix each regex metachar with a backslash"), but the function form is the safer default in substitution-sensitive contexts, so taking the stylistic fix.

### Side-cleanup

The round-3 lesson-body edit changed one lesson's content hash (fail-open-catch), so its test fixture rescaffolded under a new hash. `1d2dd5ee` renames the existing real fixture content to the new hash and refreshes `.wind-tunnel-sha` to `095e72cb`.

`totem test` reports 15 passing rule fixtures; `totem review` PASS no issues; full suite (2885+) green. Consolidated reply per GCA quota convention.

satur8d:
@coderabbitai — round 4 review: one dupe + one nit. Taken the nit, declining the dupe with receipts. Pushed as `6bf753e0`.

### Nit (taken) — `git.ts:258-265` cause-chain walk depth

Hoisted the `8` into `ERROR_CAUSE_WALK_MAX_DEPTH` in the constants block at the top of `sys/git.ts`, next to `GIT_COMMAND_TIMEOUT_MS` and friends. Documented rationale inline. No behavior change.

### Dupe (declined) — `[Totem Error]` prefix on `TotemGitError` throws

The `TotemError` base-class constructor auto-prepends the prefix:

```ts
// packages/core/src/errors.ts:40
constructor(code, message, recoveryHint, cause?) {
  super(`[Totem Error] ${message}`, { cause });
  ...
}
```

Verified empirically:

```
$ node -e "const { TotemGitError } = require('./dist/errors.js');
           const e = new TotemGitError('Failed to check dirty status for README.md.', 'hint');
           console.log(JSON.stringify(e.message));"
"[Totem Error] Failed to check dirty status for README.md."
```

Codebase audit (excluding tests and `/dist/`):

```
$ grep -rn "new TotemGitError\|new TotemConfigError\|new TotemError\|new TotemParseError" packages/ --include="*.ts" | wc -l
146
$ ... same | grep -c "\\[Totem Error\\]"
1
```

145 of 146 existing `new Totem*Error(...)` call sites let the base class handle the prefix. Following CR's suggestion here would render `[Totem Error] [Totem Error] Failed to check dirty status...` — a double-prefix bug. My two throws match the 145-site convention.

If the intended rule is "check the rendered `.message`, not the source-code literal," both throw sites already satisfy it at runtime. Happy to revisit if the coding guideline wants the manual prefix specifically at the source level — that would require flipping 145 other sites, which is a sweep ticket, not an inline fix.

Consolidated reply per quota convention.

PR #1455, opened 2026-04-15
Version Packages

satur8d:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   Updated dependencies [84bba42]
-   Updated dependencies [6776b11]
    -   @mmnto/totem@1.14.10

### Patch Changes

-   84bba42: Pipeline 1 compound rule authoring + first three production compound rules + fail-loud git.ts / rule-engine.ts

    **Pipeline 1 compound extension (new capability).** `extractManualPattern` now accepts a yaml-tagged fenced code block following the `**Pattern:**` field marker in a lesson body. The parsed NapiConfig object routes through `buildManualRule` into the `astGrepYamlRule` field on the compiled rule (same path the LLM-driven Pipeline 2 uses). Pack authors no longer need Sonnet in the loop to hand-author compound rules with `inside` / `has` / `not` combinators.

    New export from `@mmnto/totem`: `extractYamlRuleAfterField(body, field)`. Accepts ` `yaml \`\``or`\~\~~yaml`fences. Scans stop at the next bold-field marker or EOF so subsequent`**Message:**`/`### Bad Example\` / narrative sections live freely. Bare untagged fences are ignored so prose code blocks below the pattern pass through.

    **Three inaugural compound rules shipped** (all manual Pipeline 1, first production compound rules in the manifest):

    -   `Ban fail-open catch blocks that swallow errors without re-throwing` — matches `catch_clause` where the body subtree has no `throw_statement`. Compound form: `rule.kind: catch_clause, not.has.throw_statement (stopBy: end)`. The escape hatch is `// totem-ignore-next-line` above the offending catch for genuine best-effort cleanup.
    -   `` Ban `spawn()` / `spawnSync()` with `shell: true` `` — `any:` disjunction over the two call names plus `has:` descendant check on a `shell: true` pair. Scope-excludes `packages/cli/src/orchestrators/shell-orchestrator.ts` (the single legitimate `shell: true` site, secured by `MODEL_SAFE_RE` and `quoteShellArg` in 1.14.10).
    -   ``Ban `export let` declarations (exported module-level mutable state)`` — targets `export_statement` with a `lexical_declaration` child matching `let`. Zero current violations; forward protection ahead of 1.15.0 Pack Distribution.

    **Fail-loud audit fixes (closes mmnto/totem#1440 and mmnto/totem#1442).**

    -   `isFileDirty(cwd, filePath)` — previously returned `false` on any git failure. Now throws `TotemGitError` on non-ENOENT git failures so callers like `docs` (data-loss-protection filter) cannot mistake "git broke" for "file is clean." Pure ENOENT "git missing" still routes through the existing `throwIfGitMissing` helper.
    -   `resolveGitRoot(cwd)` — previously returned `null` on any git failure, conflating "not in a git repo" (legitimate) with "git broken" (bug). Now returns `null` only when the error message matches `/not a git repository/i`; other failures throw. The `string | null` contract is preserved for the one documented case.
    -   `applyRulesToAdditions` (regex engine) — previously swallowed invalid-regex errors with `continue`, so a corrupted manifest entry would mark the diff "compliant" with a load-bearing rule mute. Now throws `TotemError` (code `CHECK_FAILED`) naming the lesson hash and pattern, so the operator sees exactly which rule to fix or archive.

    Cosmetic git helpers that legitimately fall back to display strings (`getGitBranch`, `getGitStatus`, `getGitDiffStat`, `getTagDate`, `getLatestTag`, `getGitLogSince`) keep their silent-fallback behavior but now carry `// totem-ignore` comments documenting the exception. The intermediate `getDefaultBranch` probe-loop catch is also marked with `// totem-ignore` (the outer function throws if every candidate fails — intentional control flow).

    **Tests:** +11 in `lesson-pattern.test.ts` for the yaml-fence extractor, +0 net in `git.test.ts` and `compiler.test.ts` (two existing tests were updated to assert the new throw behavior, with issue references in the test names).

    Thanks to Gemini for the pre-1.15.0 deep review audit (mmnto/totem#1421) that surfaced both #1440 and #1442 and drafted the three compound rule YAML shapes.

-   6776b11: Security: Reject shell-injecting `{model}` tokens in the shell orchestrator

    The shell orchestrator used to interpolate `orchestrator.defaultModel` (and per-command overrides) directly into a command string that was then executed with `shell: true`. A poisoned config value like `"gemini-1.5; rm -rf /"` would run as shell. The `{file}` token was already shell-quoted; the `{model}` token was not.

    **Fix is defense in depth:**

    1.  **Shared allow-list at the boundary.** The shell orchestrator now imports `MODEL_NAME_RE` from `orchestrator.ts` and applies the same leading-dash + allow-list check that `resolveOrchestrator` has always used. Single source of truth for model-name validation, symmetric across all orchestrators, no chance of one gate rejecting what another accepts. Validation errors throw `TotemConfigError` matching the sibling validator.
    2.  **Shell-quoting at interpolation.** Even after the allow-list passes, the model token is wrapped in shell-safe quotes (single-quote on Unix, MSVCRT-style `\"` escape on Windows) — same treatment the `{file}` token has always had. A future regression that drops the allow-list cannot re-open the hole alone.

    **Regression tests added:** 12 exploit cases (semicolon, backtick, dollar-subshell, pipe, redirect, newline, ampersand, space, quote, dquote, paren, leading-dash) all rejected before `spawn()` is called. 8 benign model shapes accepted including underscore-containing ollama quantized tags. 1 defense-in-depth assertion that the model is quoted on the spawn command.

    **Exposure assessment:**

    Exploitable only when a user runs any `totem` command that reaches the shell orchestrator AND a poisoned `totem.config.ts` is present. Realistic vector: cloning a malicious repo and running `totem review` / `totem compile` / `totem spec` against it. Trusted-config users (single-developer, audited-config repos) were never at risk.

    **Minor backward-compat note:**

    Users whose `orchestrator.command` string manually wrapped `{model}` in quotes (e.g., `"--model=\"{model}\""`) will see the token double-quoted after this patch. The shell strips the outer quotes, so the final argv is identical, but if anyone has a custom command that depends on the raw unquoted substitution, they should drop the manual quotes from their template. Most users followed the existing `{file}` pattern (no manual quotes — Totem quotes for you), so this affects a narrow slice.

    Thanks to Gemini for catching this in the pre-1.15.0 deep review.

### Patch Changes

-   Updated dependencies [84bba42]
-   Updated dependencies [6776b11]
    -   @mmnto/totem@1.14.10


satur8d:
@coderabbitai — all three nits on the 1.14.10 CHANGELOG fixed in `47fbc730`. One commit, `packages/core/CHANGELOG.md` only (the cli/mcp CHANGELOGs don't carry the verbose text):

- **Line 11** — rewrote the mangled fence-description sentence. Original had broken inline-code nesting that rendered as literal `` `yaml ``` ``` or `~~~yaml` fences. New text: "Accepts both `yaml`-tagged triple-backtick and tilde-fence styles."
- **Line 15** — `MODEL_SAFE_RE` → `MODEL_NAME_RE`. Drift from the initial changeset text; the shell orchestrator security fix unified on `MODEL_NAME_RE` during its own review cycle.
- **Line 21** — `TotemError` (code `CHECK_FAILED`) → `TotemParseError`. Same drift — round-3 review on #1454 switched the class since an uncompilable compiled rule is a parse failure, not a generic check failure. Also clarified that the lesson hash is in the message and the pattern is in the recovery hint (both true in the shipped code).

Text-only edits. No code change, no version bump. Consolidated reply per quota convention.

Issue #1485, opened 2026-04-16
core: `immutable` flag on CompiledRule + pack-merge enforcement


satur8d:
## Context

Implements the ADR-085 Resolved Decision 3 carve-out for ADR-089. The default pack-merge semantic is "Local Supreme Authority" (local `.totem/compiled-rules.json` overrides pack rules). ADR-089 requires an exception: security rules in `@totem/pack-agent-security` must not be locally downgradable.

## Acceptance Criteria

- [ ] `CompiledRule` schema gains optional `immutable: boolean` (default `false`).
- [ ] Pack-merge logic (the `extends` pipeline from ADR-085) respects the flag:
  - If pack rule has `immutable: true` and local `compiled-rules.json` declares the same `lessonHash`, the local entry can only widen (downgrade is blocked).
  - If the local entry tries to change severity from `error` to `warning` on an immutable rule, merge rejects loudly with a clear error.
  - If the local entry sets `status: archived` on an immutable rule, merge rejects with a clear error. Immutable rules cannot be locally archived.
- [ ] When an immutable rule is bypassed via any means (suppress directive, `.totemignore`, etc.), the bypass appends an entry to the Trap Ledger (`.totem/ledger/events.ndjson`) with actor, timestamp, rule hash, and bypass reason.
- [ ] Unit tests: pack rule with `immutable: true` + local override (each of severity-downgrade, archive, suppression-in-code), verify block + ledger entry.
- [ ] Integration test: full pack install → local override attempt → confirm block.

## Out of Scope

- Actual security rules shipping with `immutable: true`. Tickets H-K set this flag during rule authoring.
- `totem doctor` check for existing immutable-rule overrides in legacy compiled-rules.json files. Optional follow-up.

## References

- ADR-085 Resolved Decisions 3 (Local Supreme Authority with immutable carve-out)
- ADR-089 Decision 2 (Enforcement posture)
- Trap Ledger: `.totem/ledger/events.ndjson` (schema in packages/core; may want to cross-reference)

satur8d:
Closing per post-merge audit on 2026-04-18. Shipped across PR #1510 (AC 1: `CompiledRule.immutable` schema) and PR #1515 (AC 2, 3, 4: pack-merge primitive + Trap Ledger audit + 12+1 unit tests). Residual scope: AC 5 full pack install → local override → confirm block e2e integration test is tracked in #1514, blocked on #1491 `totem install` landing.

Issue #1479, opened 2026-04-16
compile: Layer 3 verify-retry loop on LLM-generated patterns


satur8d:
## Context

Implements ADR-088 Phase 1 Layer 3 hardening on `mmnto-ai/totem-strategy`. Adds a verify-retry loop to the existing free-form LLM compilation path (Pipeline 2/3). Today a zero-match pattern ships silently if it parses; ADR-088 specifies that zero-match patterns should be caught and retried up to 3 times before falling through to Layer 4.

## Acceptance Criteria

- [ ] `totem compile` verifies every LLM-generated pattern against the lesson's Example Hit block. Zero-match triggers a retry with pattern + example fed back into the prompt.
- [ ] Retry cap: 3 attempts. On exhaustion, lesson falls through to Layer 4 with `reasonCode: verify-retry-exhausted`.
- [ ] Security rules (pack: `@totem/pack-agent-security` and any rule where `immutable: true` or `severity: error` in a security-scoped pack) are rejected outright on verify failure. No retry fallback. Zero tolerance per ADR-088 Decision 3.
- [ ] Rules without Example Hit/Miss are NOT subject to this gate (that is ticket B). This ticket is scoped to lessons that DO have an example.
- [ ] Integration test: playground stress-test corpus produces zero silent failures. Compare to the pre-change 0/6 failure mode cited in ADR-088 Context.
- [ ] Unit tests on the retry loop: pattern zero-matches on attempt 1, succeeds on attempt 2; all three attempts fail; security rule fails once.

## Out of Scope

- Layer 1 (programmatic AST extraction) and Layer 2 (template classification). Phase 2 (1.16.0 or later).
- The `unverified` flag for rules lacking examples. Ticket B.
- Machine-readable failure reason codes. Ticket C.

## References

- ADR-088 Decision layer 3, Phase 1 section
- Upstream of #1415 (compile-smoke-gate), #1349 (ast-grep parser validation)

satur8d:
Closing per post-merge audit on 2026-04-18. Shipped via PR #1513 (ADR-088 Phase 1 Layer 3 verify-retry on Pipeline 2 compile). Residual scope tracked in dedicated tickets:

- AC 3 (security-pack caller flipping `deps.securityContext: true`) → #1512.
- AC 5 (playground stress-test corpus integration) → #1511.

[...]

PR #1510, opened 2026-04-17
feat(core): add immutable flag to CompiledRule schema for #1485

satur8d:
This PR adds the \immutable: z.boolean().optional()\ field to \CompiledRuleSchema\ to unblock #1479 and the rest of #1485. 

It is kept strictly optional to prevent breaking inference for the 30+ existing tests and adapter fixtures that mock legacy rules. Pack-merge logic will handle the default \alse\ behavior at runtime.

### DIFF
--- a/packages/core/src/compiler-schema.ts
+++ b/packages/core/src/compiler-schema.ts
@@ -98,6 +98,13 @@ const CompiledRuleBaseSchema = z.object({
    * compiled-rules.json files; the legacy heuristic stays as a fallback.
    */
   manual: z.boolean().optional(),
+  /**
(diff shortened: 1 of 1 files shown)


satur8d:
@coderabbitai This is an intentional schema-only unblock PR. The enforcement logic is split across two upcoming PRs: #1485 (pack-merge path refuses downgrade) and #1479 (Layer 3 verify-retry branch rejects outright). 

I have updated the docstring to explicitly state that this is a schema marker for upcoming ADR-089 enforcement paths so the forward-looking intent is clearer. The inert state here is intentional and transient to unblock parallel work.

Issue #1661, opened 2026-04-24
/preflight: add upstream-API verification step to spec phase (methodology refinement)


satur8d:
## Signal

Two distinct preflight cycles in a single day (2026-04-24) shipped design docs whose load-bearing premise was empirically false against the installed package surface, caught only at TDD setup:

1. **#1653 (α Rust-support engine dispatch).** Items 014 + 017 framed the fix as "extend the dispatch map to include `Lang.Rust`." Empirically, `@ast-grep/napi@0.42.0`'s built-in `Lang` enum has only 5 members (Html, JavaScript, Tsx, Css, TypeScript). Multi-language support requires separate language packs (`@ast-grep/lang-rust@0.0.7`, ~12 MB native binary) and `registerDynamicLanguage()` activation. **Ticket parked pending strategy decision** — see [mmnto-ai/totem-strategy#128](https://github.com/mmnto-ai/towel-strategy/pull/128).
2. **#1657 (ReDoS-safe module-path-tolerant regex docs).** Item 016 + the gh issue body name `\b(?:[A-Za-z_]\w*::)*RunState\b` as the canonical safe form to document. Empirically, that pattern fails `safe-regex2@5.0.0` because of nested `*` quantifiers (star height 2). The actual safe forms are different shapes (suffix-anchor `(?:::|\b)Target\b` and bounded-wrapper `\bWrapper\s*<[^<>]{0,256}\bTarget\s*>`). **Ticket scope amended during preflight** — see PR for #1657 (forthcoming).

Both items came from liquid-city session-17 upstream-feedback. Same root cause: the strategy author proposed fixes against the documentation surface or the abstract API shape, not the empirically-installed package.

## Proposal

Extend `.claude/skills/preflight/SKILL.md` Phase 1 (mechanical context gathering) with an explicit **upstream-API verification step**. When the issue or its source upstream-feedback item names a specific external API surface (an enum, a function, a regex-validator, etc.), the preflight skill instructs:

> Before triaging Phase 2, verify the named API surface empirically against the installed package. Run a one-shot `node -e "require('<package>')..."` (or equivalent) that produces the actual surface (enum members, return type, function behavior). If the verification contradicts the issue's premise, surface the finding before triage and re-scope the design before drafting.

Concretely, the skill could add a numbered Phase 1 step:

[...]

Issue #1656, opened 2026-04-24
core: compile LLM emits 'severity: warning' even when lesson prose declares 'Severity: error'


satur8d:
Filed from strategy-side upstream-feedback item at `.strategy/upstream-feedback/015-compile-default-severity-ignores-lesson-prose.md` (full write-up lives there; this issue is the thin tracking mirror).

## Signal

Lessons with explicit `Severity: error` prose compile out to rules with `"severity": "warning"` in `compiled-rules.json`. The LLM ingests the prose but does not honor it.

## Evidence

From `mmnto-ai/liquid-city#77` session-17, compile log stamps:

```
[ADR-008 Rule 2: RunManager must not mutate TacticalState] Compiled (regex, warning): /\b(?:ResMut\s*<\s*TacticalState\s*>|&\s*mut\s*TacticalState\b)/
[ADR-008 Rule 3a: run_manager.rs must not call TacticalDirector-owned functions] Compiled (regex, warning): /\b(tactical_director_tick|update_heat|apply_host_authoritative_heat|advance_sim_tick_count)\s*\(/
```

The `(regex, warning)` stamp is the LLM's emitted severity, not a totem default applied post-hoc. Five rules in this session, all with the same mismatch against their source lessons.

Sample lesson prose the LLM ingested (from `.totem/lessons/lesson-ef0ba0d7.md`):

> **Severity: error.** Violations indicate a direct breach of the publisher/consumer contract the ADR-008 decomposition is built on.

## Downstream cost

5 manual `"severity"` edits per compile round + 1 `totem lesson compile --refresh-manifest` invocation. On `mmnto-ai/liquid-city#77` this happened twice (R2 and R3), burning ~10 edits total that served no authoring purpose. The contract between lesson prose and compiled rule is broken; every compile cycle re-applies the mechanical fix.

Further context: per a CR review learning cited in the upstream-feedback item, `mmnto-ai/totem` collects only `severity: "error"` as CI-blocking — `warning` is advisory. So a warning-severity rule silently degrades to advisory-only governance even when the author declared it an architectural-invariant error.

## Proposed approach

[...]

Issue #1626, opened 2026-04-23
core: compile-worker scope derivation mis-scopes test-contract lessons away from tests


satur8d:
## Signal

Surfaced by bot review of PR #1625 postmerge artifacts (CR/major + CR/minor):

- `.junie/skills/totem-rules/rules.md:2044` — "Normalize temp paths for cross-platform equality" (test-contract lesson) exported with scope `packages/**/*.ts, !**/*.test.*, !**/*.spec.*`. The rule describes assertion behavior in tests; scope excludes tests. Won't fire where intended.
- `.github/copilot-instructions.md:2082` — "Spy on logger contracts in tests" same class. Title says "in tests"; scope excludes tests.

Both are auto-generated exports of lessons that went through the compile pipeline. Direct-edit is wrong per CR ("this file is auto-generated; correct the source lesson/compiler export path instead of editing generated content directly").

## Class of problem

Default-scope heuristic assumes `packages/**/*.ts` minus `**/*.test.*` / `**/*.spec.*`. For test-contract lessons (behavior that executes inside test files), that's the inverse of the correct scope. Compile worker needs a signal to flip the scope to `**/*.test.ts, **/*.spec.ts, ...`.

## Candidate signals

- Lesson heading contains "test", "spec", "assertion", "contract"
- Lesson tags include `testing`
- `badExample` / `goodExample` reference `describe(`, `it(`, `test(`, `expect(`, vitest/jest patterns
- Lesson body references `**/*.test.*` explicitly

## Class interaction

Sibling to #1598 (context-sensitive lesson extraction gap). Both are compile-worker pattern-derivation failure modes. #1598 covers context guards in the lesson prose; this covers scope inversion.

## Severity

Tier-2. Silent misapplication — affected rules shipped in every downstream export since they were compiled, but won't fire on the code they're meant to govern.

## Upstream

- PR #1625 CR review (2026-04-22), findings on `.junie/skills/totem-rules/rules.md:2044` and `.github/copilot-instructions.md:2082`.
- Sibling: #1598 (context-sensitive lesson extraction gap), same compile-worker surface.

## Acceptance

- [ ] Compile worker detects test-contract lessons via heuristic (heading / tags / example signals) and flips scope to test/spec globs
- [ ] Regression test: "Normalize temp paths for cross-platform equality" and "Spy on logger contracts in tests" both compile with `**/*.test.*`-inclusive scope
[...]

PR #1662, opened 2026-04-24
feat(cli+docs): document ReDoS-safe module-path-tolerant regex patterns (#1657)

### LINKED ISSUE #1657: cli: document ReDoS-safe module-path-tolerant regex idiom in compile prompt and authoring docs
Filed from strategy-side upstream-feedback item at `.strategy/upstream-feedback/016-redos-gate-blocks-module-path-patterns.md` (full write-up lives there; this issue is the thin tracking mirror).

## Signal

Totem's compile-time ReDoS gate correctly rejects the canonical regex idiom for module-path-tolerant matching (`(?:[A-Za-z_][A-Za-z0-9_:]*::)?Target`) because the greedy class that embeds `:` combined with a literal `::` terminator is quadratic under adversarial input. But there is no ReDoS-safe idiom documented in the totem rule surface to express this common hardening intent. Contributors forced to either:

1. Drop the hardening and accept bypass via module-qualified paths (`crate::state::RunState`, `super::RunState`).
2. Hand-author a non-ReDoS-vulnerable equivalent, which requires regex-engine internals knowledge (possessive quantifiers, atomic groups) unavailable in Node's built-in regex engine.

Neither option preserves the hardening at the compile-gate-approved cost the contributor was trying to pay.

## Evidence

`mmnto-ai/liquid-city#77` R2 rejection log:

```
[ADR-008 Rule 2: RunManager must not mutate TacticalState] Rejected regex: ReDoS vulnerability detected — skipping
[ADR-008 Rule 1: TacticalDirector must not mutate RunState] Rejected regex: ReDoS vulnerability detected — skipping
[ADR-008 Rule 4: director files allow only the four state-container ResMut types] Rejected regex: ReDoS vulnerability detected — skipping
```

The author wrote (per GCA suggestion):

```
\bResMut\s*<\s*(?:[A-Za-z_][A-Za-z0-9_:]*::)?RunState\s*>
```

Intent (from `lesson-ef0ba0d7.md`): catch module-qualified paths (`crate::state::RunState`, `super::RunState`) so the invariant holds under refactors that change import scoping.

Result: three rules shipped without module-path-tolerance hardening. A real hazard class (bypass via `use crate::director::state::RunState;` → inline qualification) remains weakly enforced.

## Proposed approach

**Option A (recommended for 1.16.0).** Document a ReDoS-safe module-path idiom in the compile-prompt system message and in the authoring docs. Safe form:

```
\b(?:[A-Za-z_]\w*::)*RunState\b
```

- `[A-Za-z_]\w*` is a standard identifier, no `:` embedded.
- `::` is a literal separator outside the class, with unambiguous boundary.
- `(?:...)*` repeats the identifier-then-`::` unit linearly; each unit has its own boundary so there is no backtracking tail.

Matches `crate::state::RunState`, `super::RunState`, and bare `RunState`. Passes the ReDoS gate in principle.

**Option B (1.17.0).** Make the ReDoS gate rejection message actionable. Today: `Rejected regex: ReDoS vulnerability detected — skipping`. Extend to include the detected backtracking class + a link to the safe-idiom doc from Option A.

**Option C (1.18.0+).** Compile-prompt autopromotion: teach the compile LLM to recognize "module-path tolerance" intent in lesson prose and emit the safe form automatically. Requires prompt engineering + regression guards.

## Class interaction

- Composes with #1653 (α engine dispatch), #1654 (β compile-pipeline grammar), #1649 (corpus-count heuristic) in the compile-pipeline quality stack. All are "compile pipeline rejects or skips a legitimate hardening intent due to an internal constraint the contributor does not see ahead of time." Orthogonal mechanisms, shared UX pattern.
- Adjacent to #1648 (regex-safety telemetry field rename). Both touch the `regex-safety/` module surface.

## Severity

Tier-2. N=1 exhibit today, 3 rules with dropped hardening on `mmnto-ai/liquid-city#77`. Low-frequency per-author, but every hardening decision that hits this class forces a decline or hand-authored workaround. Documentation-first fix unblocks the next contributor before they hit the gate.

## Acceptance

- [ ] Compile-prompt system message adds a "Module-path-tolerant regex" section with safe + unsafe examples.
- [ ] Authoring docs (e.g., `docs/wiki/pipeline-engine.md` or similar) carry the same safe-idiom snippet with rationale.
- [ ] Regression test: a rule using `\b(?:[A-Za-z_]\w*::)*RunState\b` passes the ReDoS gate.
- [ ] Regression test: the unsafe form `(?:[A-Za-z_][A-Za-z0-9_:]*::)?RunState` continues to fail the gate (no regression).
- [ ] ReDoS gate rejection message improvements are tracked as the Option B follow-up.

Upstream-feedback write-up: `.strategy/upstream-feedback/016-redos-gate-blocks-module-path-patterns.md`

satur8d:
## Summary

The compile-time ReDoS gate (safe-regex2's star-height heuristic) correctly rejects module-path-tolerant regex shapes contributors naturally write when hardening against `crate::state::Target`-style bypass classes. Until now no safe alternative was documented. This PR ships two empirically-verified safe forms in both compile prompts and a dedicated wiki page.

- **Form 1 (suffix-anchor)**: `(?:::|\b)Target\b` — identifier-anywhere intent
- **Form 2 (bounded wrapper)**: `\bWrapper\s*<[^<>]{0,256}\bTarget\s*>` — typed-container scoped

Both have star height 1 (no nested quantifiers), pass `safe-regex2`, and cover the `mmnto-ai/liquid-city#77` `ResMut<RunState>` hardening intent including bare, `::`-prefixed (`crate::state::RunState`), and `super::`-prefixed identifier forms.

## Empirical scope correction

Item 016 + this issue's stated acceptance both named `\b(?:[A-Za-z_]\w*::)*Target\b` as the canonical safe form. Empirical verification during /preflight against `safe-regex2@5.0.0` showed that pattern itself is rejected (nested `*` quantifiers — same star-height heuristic the original unsafe form failed on):

```text
node> safeRegex(String.raw`\b(?:[A-Za-z_]\w*::)*RunState\b`)
false
```

Strategy-side correction merged via [`mmnto-ai/totem-strategy#129`](https://github.com/mmnto-ai/totem-strategy/pull/129) ahead of this PR; the wiki page and prompt sections cite the corrected forms. Methodology refinement to add an upstream-API verification step to /preflight Phase 1 filed as [#1661](https://github.com/mmnto-ai/totem/issues/1661).

## Changes

- `packages/cli/src/commands/compile-templates.ts`: new "Module-Path-Tolerant Regex Patterns (#1657)" section in both `COMPILER_SYSTEM_PROMPT` and `PIPELINE3_COMPILER_PROMPT`, parity per the established convention from #1626 + #1656.
- `packages/cli/src/commands/compile-templates.test.ts`: 8 new tests pinning section presence + verbatim Form 1 + Form 2 + trap-shape strings in both prompts (so a future docs PR cannot silently re-document the impossible shape on one prompt while fixing the other).
- `packages/core/src/compiler.test.ts`: 9 new tests pinning Form 1 accept, Form 2 accept, item 016 proposed form reject, original `liquid-city#77` unsafe form reject, plus behavioral matchers (`Form 1` vs `MyRunState` collision, `Form 2` scoping inside `ResMut` wrapper).
- `docs/wiki/regex-safety.md`: new focused page for regex-safety authoring guidance. Documents both safe forms with rationale, the gate's star-height heuristic, an anti-pattern note pointing to ast-grep for unbounded-depth path traversal, and a local verification snippet so contributors can check candidate patterns before submission.
- `docs/wiki/cli-reference.md`: cross-link from the `totem lint` section's existing ReDoS prose to the new page.
- `.strategy` submodule bump to `68740a6` (picks up the item 016 correction + the post-1.15.4 strategy commits).

## Out of scope

Per the design doc:

- ReDoS gate rejection-message improvements (item 016 Option B, deferred to a follow-up ticket).
- LLM autopromotion (item 016 Option C, deferred indefinitely).
- ast-grep / tree-sitter language-pack work for non-TS rules (parked at [`totem-strategy#128`](https://github.com/mmnto-ai/totem-strategy/pull/128) pending the four-option Proposal decision).

## Test plan

- [x] Core test suite: 1348 pass (+9 new in `compiler.test.ts`).
- [x] CLI test suite: 1827 pass (+8 new in `compile-templates.test.ts`).
- [x] `pnpm exec totem lint`: PASS, 82 warnings, 0 errors. All warnings are pre-existing over-broad rules firing on this PR's regex test fixtures (e.g., `re.test('&mut RunState')` triggering "use anchored regex for shell command matching" because the rule pattern is literal `//`); none are introduced by this change.
- [x] `pnpm exec totem review`: PASS, no findings.
- [x] Empirical verification of all four documented patterns against installed `safe-regex2@5.0.0` performed during preflight; the verification snippet from `docs/wiki/regex-safety.md` reproduces the result.

Closes #1657.

### DIFF
--- a/.totem/specs/1657.md
+++ b/.totem/specs/1657.md
@@ -20,24 +20,29 @@ This makes the gh issue's stated acceptance — "rule using `\b(?:[A-Za-z_]\w*::
 Both pass `safe-regex2@5.0.0` AND cover the module-path-tolerance intent (`crate::state::RunState`, `super::RunState`, bare `RunState`):
 
 **Form 1 — Suffix-anchor (generic identifier):**
+
 ```regex
 (?:::|\b)RunState\b
 ```
+
 - Star height: 1 (no nested quantifiers)
 - Matches: bare `RunState` (via `\b`), `crate::state::RunState` (via `::` at preceding `state::` boundary), `super::RunState`, `&mut RunState`, `&mut super::RunState`
 - Doesn't match: `MyRunState`, `RunStateExtra`
 - Use when the rule fires on the identifier regardless of its surrounding container.
 
 **Form 2 — Bounded wrapper (typed-container scoped):**
+
 ```regex
 \bResMut\s*<[^<>]{0,256}\bRunState\s*>
 ```
+
 - Star height: 1 (`{0,256}` is bounded; `[^<>]` has no nested quantifier)
 - Matches only when `RunState` appears inside `ResMut<...>` brackets, with bounded path content between
 - Tighter scope: doesn't fire on `RunState` references outside `ResMut<>`
 - Use when the rule fires specifically on a typed-container hardening intent (ADR-008 Rule 1's actual shape on liquid-city#77)
 
 **Generic recipes for the docs:**
+
 - Identifier match: `(?:::|\b)<IDENT>\b`
 - Wrapper match: `\b<WRAPPER>\s*<[^<>]{0,256}\b<IDENT>\s*>`
 
@@ -72,12 +77,12 @@ None. No new state.
 
 ### Failure modes
 
-| Failure | Category | Agent-facing surface | Recovery |
-|---|---|---|---|
-| Wiki page link broken (typo in cross-ref) | docs | renders broken in GitHub wiki | manual fix in follow-up |
-| Safe-regex2 behavior changes in a future bump | runtime | Form 1/2 tests would fail loudly in CI on dependency upgrade | tests pin behavior; intentional regression signal |
-| Compile prompt token-limit overflow | runtime | Sonnet refuses prompt or truncates | new section is ~12 lines / ~100 tokens, well within budget |
-| LLM ignores new section and emits unsafe form anyway | runtime | smoke gate rejects the rule (existing behavior) | rule rejected at compile, Sonnet retries; no new degradation |
+| Failure                                              | Category | Agent-facing surface                                         | Recovery                                                     |
+| ---------------------------------------------------- | -------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| Wiki page link broken (typo in cross-ref)            | docs     | renders broken in GitHub wiki                                | manual fix in follow-up                                      |
+| Safe-regex2 behavior changes in a future bump        | runtime  | Form 1/2 tests would fail loudly in CI on dependency upgrade | tests pin behavior; intentional regression signal            |
+| Compile prompt token-limit overflow                  | runtime  | Sonnet refuses prompt or truncates                           | new section is ~12 lines / ~100 tokens, well within budget   |
+| LLM ignores new section and emits unsafe form anyway | runtime  | smoke gate rejects the rule (existing behavior)              | rule rejected at compile, Sonnet retries; no new degradation |
 
 No new silent-degradation paths. The new wiki page documents existing behavior; if anyone follows the docs, they get gate-passing patterns. If they don't, the existing gate rejection still fires.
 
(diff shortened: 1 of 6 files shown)


satur8d:
@coderabbitai re: the nitpick on `.totem/specs/1657.md:68` (repeated "Will NOT" sentence openers in the Scope section's negative-scope block):

Acknowledged — by design. The parallel construction is intentional. Enumerating scope exclusions in matching form ("Will NOT X. Will NOT Y. Will NOT Z.") signals "these are the explicit non-goals" with rhetorical weight that mixed openers would dilute. Spec files under `.totem/specs/` are voice-exempt per `MEMORY.md` (the `docs/wiki/**` voice constraints don't reach internal docs), so the Rule-of-Three readability concern doesn't apply here either.

Re: the MD040 finding on the fenced block at `.totem/specs/1657.md:9-12` — already fixed in `d78a3a0e`, pushed before this review round landed. CR's re-scan should auto-resolve it.

Issue #1666, opened 2026-04-25
Change `totem triage-pr` dedup identity from body-hash to deterministic comment_id


satur8d:
## Context

Strategy upstream-feedback item 024 (merged via [mmnto-ai/totem-strategy#133](https://github.com/mmnto-ai/totem-strategy/pull/133), strategy main `f1cae9a` 2026-04-25): `npx totem triage-pr` dedup pass collapses distinct GCA findings that share a `(path, line, author)` tuple, hiding 4 of 6 GCA-high findings on liquid-city PR #80 R3 from the headline summary.

Asymmetric across bots: CR's per-finding line-anchoring survives the dedup; GCA's shared-header-anchor doesn't. Recovery currently requires raw GraphQL fallback.

Source artifact: `.strategy/upstream-feedback/024-triage-pr-collapses-same-line-findings.md`

## Scope

1. **Locate** the dedup pass in `packages/cli/src/commands/triage-pr.ts` (or `parsers/`).
2. **Switch dedup identity** from the current `(path, line, author)` tuple to **deterministic `comment_id`** (`thread_id` for inline review-thread findings, top-level `id` for issue-comment findings). Per item 024 R3 refinement: GitHub's REST + GraphQL pulls APIs both expose unique comment identifiers, so the deterministic primitive is available without API gymnastics.
3. **Body-hash fallback** for the rare case where the API doesn't surface an ID (e.g., synthesized `(review body)` findings per the existing `reviewBodyFindings` pattern). Body-hash retains a fallback role only — NOT primary.
4. **Test fixtures** covering: (a) two GCA findings on the same line that should NOT collapse, (b) the same finding posted both inline and in review-body that SHOULD be deduped.

## Note

Body-hash dedup was correctly rejected by GCA in R2 review of #133 because identical-body findings (same generic lint repeated on different lines) would silently collapse — that's data loss disguised as cleanup.

## Tier

**Tier 2.** UX correctness — `triage-pr` is the canonical bot-review triage surface, so headline-summary accuracy is load-bearing for the workflow but not a data-corruption class.

Issue #1665, opened 2026-04-25
Compile pipeline silently drops test/spec exclusion globs from lesson Scope (inverse of #1626)


satur8d:
## Context

Strategy upstream-feedback item 023 (merged via [mmnto-ai/totem-strategy#133](https://github.com/mmnto-ai/totem-strategy/pull/133), strategy main `f1cae9a` 2026-04-25): totem 1.15.4 compile pipeline silently DROPS test/spec exclusion globs that the lesson source declares in its `Scope:` field. Inverse of #1626 (which was about auto-ADDing test inclusions for test-contract lessons).

Confirmed twice on liquid-city PR #80 (rules `5bcc8aad9096c817` and `6c457c82d3945d15`). CR's `[Learnings used]` block on PR #80 explicitly noted a dedicated upstream-feedback item was planned post-merge; that filing is item 023.

Source artifact: `.strategy/upstream-feedback/023-scope-derivation-silently-drops-exclusions.md`

## Scope

1. **Diagnose** where in the compile pipeline `Scope:` exclusion globs (`!**/*.test.*`, `!**/*.spec.*`, etc.) are being silently dropped from the resulting `fileGlobs`.
2. **Fix** so the compiled rule's `fileGlobs` carries the exclusions the lesson source declared.
3. **Smoke-gate check.** Per item 023 R3 refinement: use **set-of-strings equality** between source-declared `Scope:` glob list and compiled `fileGlobs`, NOT a superset check. Equality is unambiguous in both directions (this auto-DROP bug AND the inverse auto-ADD #1626 case). The proposal includes a worked two-direction example.
4. **Test fixtures** covering both directions: (a) lesson with explicit test exclusions → rule preserves them, (b) test-contract lesson per #1626 → rule auto-includes test patterns.

## Tier

**Tier 1.** Correctness bug — same severity class as #1626 (which was tier-1 and shipped in 1.15.4 as strategy item 016). Inverse direction of the same problem.

Issue #1663, opened 2026-04-25
Implement `applies-to:` lesson frontmatter field per strategy item 020


satur8d:
## Context

Strategy upstream-feedback item 020 (merged via [mmnto-ai/totem-strategy#133](https://github.com/mmnto-ai/totem-strategy/pull/133), strategy main `f1cae9a` 2026-04-25): lesson schema needs an `applies-to:` frontmatter field carrying role-of-code taxonomy so bot reviewers can filter lessons by role match instead of grep-by-topic heuristics.

Source artifact: `.strategy/upstream-feedback/020-lesson-scope-tag-for-bot-citation-accuracy.md`

## Scope

1. **Schema extension.** Add `applies-to: [role, ...]` field to lesson frontmatter parser. Role taxonomy: `mutator`, `boundary`, `aggregator`, `hot-path`, `boundary-test`, `infrastructure`, `presentation`, `any`.
2. **Value-shape coercion.** Canonical internal representation is a list. Scalar shorthand `applies-to: any` parses as `[any]`. Both YAML-frontmatter and prose-frontmatter wire formats supported.
3. **Default-to-any.** Lessons without the field default to `[any]` for backwards compatibility. No migration needed for existing 1,159 lessons.
4. **Bot-prompt construction filter.** When constructing prompts for bot reviewers, filter lessons by role-match against the function under review. Lessons declared `[any]` always apply.

## Convention link

`applies-to:` is the **seed schema key for Proposal 248** (Bot Operations Packs). Naming was deliberate per item 020's review — `scope-tag` was rejected to avoid ADR-070 namespace collision with `scope:` (file-globs).

## Out of scope

- Backfilling role declarations on existing lessons (separate sweep, deferred).
- Bot-side enforcement (bots interpret the field; no totem-side gate).
- Heuristic role detection from lesson content (manual declaration only in V1).

## Tier

**Tier 1.** Proposal 248 prereq. Item 020 was the linchpin item in strategy #133.

Issue #1177, opened 2026-04-04
bug: totem-ignore detection rule is self-suppressing


satur8d:
## Context

GCA flagged in PR #1174 that the compiled rule for lesson-2c048a3f ("Use totem-context for inline justifications") can never fire.

## Problem

The rule compiles to pattern `/\btotem-ignore\b/` (warning severity). However, `rule-engine.ts` (line 159) suppresses any line containing the `totem-ignore` string *before* running regex rules. So the rule's target is consumed by the suppression mechanism before the rule can detect it.

Result: the rule always produces a `suppress` event, never a `trigger` (violation). It's effectively dead.

## Options

1. **Engine-level fix:** Check for `totem-ignore` rules specifically and exempt them from pre-suppression
2. **Compiler-level fix:** Detect self-suppressing patterns during compilation and warn
3. **Accept as documentation:** Keep the lesson for human reference even though the compiled rule can't enforce it
4. **Retire the rule:** Remove the lesson since it can't be mechanically enforced

## Source

GCA review on #1174

PR #1688, opened 2026-04-26
fix(core): emit self-suppressing-pattern reasonCode (#1664)

### LINKED ISSUE #1664: Add `reasonCode: self-suppressing-pattern` to compile-worker nonCompilable ledger
## Context

Strategy upstream-feedback item 021 (merged via [mmnto-ai/totem-strategy#133](https://github.com/mmnto-ai/totem-strategy/pull/133), strategy main `f1cae9a` 2026-04-25): totem 1.15.4 deliberately skips lessons whose compiled regex matches `totem-ignore` / `totem-context` (self-suppressing) but does not emit a `nonCompilable` entry, leaving zero trace in `compiled-rules.json`.

Surfaced during liquid-city PR #80 R3 — bot reviewer (GCA-high) inferred the gap from headcount (7 lessons added, 6 in manifest) and synthesized a "missing from manifest" finding because of the audit-trail asymmetry.

Source artifact: `.strategy/upstream-feedback/021-self-suppressing-skips-not-tracked.md`

## Scope

1. Add `'self-suppressing-pattern'` to the `reasonCode` enum on `CompilerOutputBaseSchema` (sibling to existing `'context-required'`, `'semantic-analysis-required'`, etc.).
2. Compile worker emits this `reasonCode` when skipping a lesson because its compiled pattern matches `totem-ignore` / `totem-context` and would self-suppress at runtime.
3. `nonCompilable` ledger entry written with the new code (consistent with the established `{hash, title}` tuple shape).
4. Decide ledger write-policy: should this code be in `LEDGER_RETRY_PENDING_CODES` (transient, retry on next compile) or terminal (mark unfit permanently)? Lean toward terminal — self-suppression is structural, not transient.

## Reproducer

`lesson-a00d6b65` in liquid-city PR #80 R3 was the first confirmed instance; reproduce locally by compiling that lesson against `totem 1.15.4`.

## Tier

**Tier 2.** Audit-trail completeness improvement; not blocking core functionality but eliminates a recurring bot-review false-positive class.

satur8d:
## Summary

Strategy upstream-feedback item 021 substrate. Pre-fix, the compile worker silently dropped lessons whose compiled pattern would match `totem-ignore` / `totem-context` tokens (self-suppression guard, #1177). The rejection mapped to `'pattern-syntax-invalid'` — a retry-pending code — so no `nonCompilable` ledger entry was written and the lesson left zero trace in `compiled-rules.json`. Bot reviewers consequently synthesized "missing from manifest" findings on PRs that included such lessons (`mmnto-ai/liquid-city#80` R3 was the reproducer).

## Changes

- **New enum member** `'self-suppressing-pattern'` on `NonCompilableReasonCodeSchema` in `packages/core/src/compiler-schema.ts`. Sibling to `'context-required'` (#1639) and `'semantic-analysis-required'` (#1640) — both are terminal classifier codes for structural incapacity.
- **Terminal write-policy.** NOT in `LEDGER_RETRY_PENDING_CODES`, so `shouldWriteToLedger('self-suppressing-pattern')` returns `true`. Self-suppression is structural — the same lesson body produces the same self-suppressing pattern on every retry; retry-pending would loop indefinitely without progress.
- **Classification mapping update** in `classifyBuildRejectReason` (compile-lesson.ts): rejection messages containing `'suppression directive'` now map to `'self-suppressing-pattern'` (was: `'pattern-syntax-invalid'`). Other rejection paths (`'Rejected regex'`, `'Invalid ast-grep pattern'`, etc.) keep their existing mappings.
- **Bot-review citation surface.** Reviewers can now cite the explicit `reasonCode: 'self-suppressing-pattern'` entry in `nonCompilable` instead of inferring "missing from manifest" from headcount mismatches.
- **Submodule bump** `.strategy` `b819561` → `8f79b82` rides this PR per the standing pattern (4 strategy commits since last bump: bot-interaction-nuance file landed via `mmnto-ai/totem-strategy#137`, Proposal 248 promoted Draft → Accepted via `#140`, upstream-feedback items 025+026 from LC session 20 via `#141`, ADR-091 Stage 4 decomposition + bot-nuance patterns via `#143`).

## Why now

#1664 is the last tier-2 item from the upstream-feedback batch (`mmnto-ai/totem-strategy#133`) — pairs with today's tier-1 ships (#1663 `applies-to:` + #1665 source-Scope override). #1666 (item 024 `triage-pr` dedup) is the final remaining item from that batch.

## Test plan

- [x] `pnpm --filter @mmnto/totem test` — 1,404 / 1,404 pass (+6 new tests covering enum acceptance, terminal-policy verification against `LEDGER_RETRY_PENDING_CODES`, `shouldWriteToLedger` behavior, Pipeline 3 end-to-end classification of self-suppressing patterns).
- [x] `pnpm exec totem lint` — PASS (2 warnings, no errors).
- [x] `pnpm exec totem review` — PASS, no findings.
- [x] `pnpm exec totem verify-manifest` — PASS.
- [x] Pipeline 3 end-to-end: lesson with pattern `\btotem-ignore\b` produces `status: 'skipped'` with `reasonCode: 'self-suppressing-pattern'` (was `'pattern-syntax-invalid'` pre-fix).
- [x] Existing self-suppression guard tests (#1177) still pass — they assert the rejection itself, which is unchanged; only the downstream classification shifted.

Closes #1664

### DIFF
--- a/.changeset/1664-self-suppressing-reason-code.md
+++ b/.changeset/1664-self-suppressing-reason-code.md
@@ -1,0 +1,14 @@
+---
+'@mmnto/totem': patch
+'@mmnto/cli': patch
+'@mmnto/mcp': patch
+---
+
+fix(core): emit `self-suppressing-pattern` reasonCode for self-suppressing skips (#1664)
+
+Strategy upstream-feedback item 021 substrate. Pre-fix, the compile worker silently dropped lessons whose compiled pattern would match `totem-ignore` / `totem-context` (and self-suppress at runtime) — the rejection mapped to `pattern-syntax-invalid` (a retry-pending code), so the lesson never landed in `nonCompilable`. Bot reviewers reading `compiled-rules.json` would synthesize "missing from manifest" findings because the audit trail was empty.
+
+- New `'self-suppressing-pattern'` member on `NonCompilableReasonCodeSchema`. Sibling to `'context-required'` (#1639) and `'semantic-analysis-required'` (#1640) — both are terminal classifier codes for structural incapacity.
+- Terminal write-policy: NOT in `LEDGER_RETRY_PENDING_CODES`, so `shouldWriteToLedger('self-suppressing-pattern')` returns true. Self-suppression is structural — the same lesson body would produce the same self-suppressing pattern on every retry, so retry-pending would loop forever.
+- `classifyBuildRejectReason` updated: rejection messages containing `'suppression directive'` now map to `'self-suppressing-pattern'` (was: `'pattern-syntax-invalid'`). Other rejection paths (`'Rejected regex'`, `'Invalid ast-grep pattern'`) keep their existing mappings.
+- Bot reviewers can now cite the explicit `reasonCode: 'self-suppressing-pattern'` entry in `nonCompilable` instead of inferring "this lesson is missing" from headcount mismatches.
--- a/.totem/specs/1664.md
+++ b/.totem/specs/1664.md
@@ -1,0 +1,79 @@
+### Problem Statement
+
+The compile pipeline currently skips lessons silently if their compiled regex matches `totem-ignore` or `totem-context` (self-suppressing), leaving no audit trail in the `nonCompilable` ledger. We need to introduce a new `reasonCode` (`self-suppressing-pattern`), intercept these self-suppressing patterns during compilation to emit the new code, and treat it as a terminal (non-retryable) failure so it properly writes to the ledger.
+
+### Architectural Context
+
+None found in provided context.
+
+### Files to Examine
+
+1. `packages/core/src/schemas.ts` (or `packages/core/src/schemas/compiler.ts`) — To locate `CompilerOutputBaseSchema` and the `reasonCode` enum.
+2. `packages/core/src/constants.ts` (or equivalent where `LEDGER_RETRY_PENDING_CODES` is defined) — To verify ledger write-policy lists.
+3. `packages/core/src/compile-lesson.ts` — To locate the existing silent skip logic and update `compileLesson` to intercept self-suppressing patterns.
+4. `packages/cli/src/commands/compile.ts` — To understand how the orchestrator processes the `CompileLessonResult` and writes to the ledger.
+
+### Technical Approach & Contracts
+
+1. **Schema Extension**: Update `CompilerOutputBaseSchema` (a Zod schema) to include `'self-suppressing-pattern'` in its `reasonCode` enum.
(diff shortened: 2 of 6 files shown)


satur8d:
@gemini-code-assist Declined — the finding cites the auto-generated spec, but the spec's "interception in compileLesson + extend `CompilerOutputBaseSchema.reasonCode`" approach was synthesized without reading the architectural prior art. The spec itself notes *"Architectural Context: None found in provided context"* — its recommendation does not reflect how `'context-required'` (#1639) and `'semantic-analysis-required'` (#1640) actually shipped.

**1. The LLM never emits `reasonCode: 'self-suppressing-pattern'`.** Self-suppression is a deterministic engine-side detection: the `isSelfSuppressing()` guard at `compile-lesson.ts:524` (ast-grep path) and `:582` (regex path) walks the compiled pattern string and rejects on `totem-ignore` / `totem-context` markers. The `compile-templates.ts` system prompt does not teach the LLM about these markers — and shouldn't, because they're a runtime detail of the lint engine, not a semantic property of the rule the LLM is producing.

By contrast, `'context-required'` and `'semantic-analysis-required'` ARE in `CompilerOutputBaseSchema.reasonCode` because the LLM analyzes lesson semantics ("does this hazard require enclosing-function scope I can't capture?", "does this need multi-file analysis?") and emits those classifications directly. Self-suppression is structurally different: it's a string-match against system tokens, post-emission.

**2. The cited fallback at `compile-lesson.ts:1159` cannot fire.** The Zod-validation fallback path the finding describes ("any LLM response... emits this reasonCode will fail Zod validation, causing the compile worker to fall back to `'pattern-syntax-invalid'`") only triggers when the LLM emits `reasonCode: 'self-suppressing-pattern'` in its `compilable: false` response. Since the LLM is never instructed to emit this code, the fallback never executes. The granular classification this PR adds happens in `classifyBuildRejectReason` after the engine guard rejects — a different code path.

**3. The spec's interception approach would create a DRY violation.** Adding the override-to-`compilable: false` in `compileLesson` BEFORE `buildCompiledRule` runs would require duplicating the `isSelfSuppressing` check at a second site that must stay in sync with the existing guards. This is exactly the cross-helper-consistency-cascade trap documented in `lesson-400fed87` (the schema-evolution pattern lesson). My approach keeps the single source of truth — one detection site, one ledger-classification translation.

**The PR's outcome matches the spec's intent** (terminal classification, audit-trail entry, no retry-pending loop). Only the implementation path differs, and the chosen path mirrors the established #1639 / #1640 pattern: terminal reasonCode goes in `NonCompilableReasonCodeSchema` (the ledger enum), LLM-emittable reasonCodes go in `CompilerOutputBaseSchema.reasonCode` (a strict subset).

Tests verify the end-to-end behavior: Pipeline 3 lesson with `\btotem-ignore\b` pattern → `status: 'skipped'` with `reasonCode: 'self-suppressing-pattern'` → `shouldWriteToLedger` returns true → ledger entry written.

Issue #1811, opened 2026-05-03
cli/core: totem sync couples pack-resolution to vector-store embedding — CI without API keys cannot reach pack-resolution alone after cohort bump


satur8d:
## Summary

`totem sync` couples two architecturally distinct concerns at the command level — re-resolving packs into `installed-packs.json` AND re-indexing project files into the vector store — and there is no command for "just do the pack-resolution leg." After the 1.26.0 cohort bump (ADR-097 § Q6 engines-field migration), consumers MUST run `pnpm exec totem sync` to re-resolve the manifest against the new `engines` field, otherwise `totem lint` hard-fails with the **same `TotemParseError: no Tree-sitter language is registered for that extension`** symptom that originally drove `#1803`.

The CI escalation makes this an operational-impossibility class: **`totem sync` requires an embedding API key. CI runs without API keys cannot complete the bundled command, so the manifest never gets written, so lint hard-fails — even though pack-resolution itself needs no LLM access.** Workarounds today (inline manifest generation in bash, granting unrelated API access to CI, skipping rust-rule registration entirely) all distribute substrate-resolver logic into downstream scripts (Tenet 15 violation) or weaken the validation signal exactly where *Publishing Without Wiring Is Incomplete Ship* says it matters most.

This is the third documented N-of-3 instance of the *Publishing Without Wiring Is Incomplete Ship* pattern (after `#1794` `loadInstalledPacks()` not invoked at lint time, and bot packs published-but-not-wired). With N=3 the rule becomes canonical operational pattern.

Source: [`mmnto-ai/totem-strategy/upstream-feedback/053-totem-sync-required-after-cohort-bump.md`](https://github.com/mmnto-ai/totem-strategy/blob/main/upstream-feedback/053-totem-sync-required-after-cohort-bump.md).

### Trace 1 — lc-Claude, liquid-city `feat/heatmap-toggle-and-iter`, 2026-05-03T04:51Z

1. Edited `package.json` — bumped `@mmnto/totem`, `@mmnto/cli`, `@mmnto/mcp`, `@mmnto/pack-rust-architecture` `^1.25.0` → `^1.26.0`. `pnpm install` clean.
[...]

PR #1828, opened 2026-05-05
feat(sync): Phase A / Phase B architectural separation (ADR-101)

satur8d:
## Summary

Closes `mmnto-ai/totem#1811`. Decouples deterministic pack-resolution from LLM-dependent vector-store embedding via mutually-exclusive `--packs-only` / `--index-only` flags on `totem sync`.

CI environments without API keys can now refresh `.totem/installed-packs.json` after a `@mmnto/totem` cohort bump without tripping `requireEmbedding`. Reproduced N=2 in `mmnto-ai/totem-strategy:upstream-feedback/053` (Liquid City + totem-strategy itself); first half of the operational impossibility loop closed by `@mmnto/cli` going to 1.28.0.

## Phase A / Phase B

| Phase | Scope | Tier | Flag |
|---|---|---|---|
| **A** | Pack-resolution + `installed-packs.json` write (deterministic) | **Lite** | `sync --packs-only` |
| **B** | `runSync` + embedding sync + side outputs | **Standard** | `sync --index-only` |
| **A + B** | Default behavior (observably equivalent to prior releases) | Standard | `sync` |

`--packs-only` is mutex-exclusive with `--index-only`, `--full`, and `--prune`. Hard `TotemError('FLAG_CONFLICT', ...)` fires BEFORE `loadConfig` and `requireEmbedding` so the diagnostic stays clean of cascading API-key errors.

## Changes per task

| Task | Disposition |
|---|---|
| **T1** Schema + writer + `resolveEngineVersion` export | `InstalledPacksManifestSchema` gains optional `cohort: string` (semver). `writeInstalledPacksManifest` auto-stamps when caller omits. Pre-1.27.0 manifests parse cleanly. |
| **T2** CLI orchestrator decoupling | OQ1 — split lives at the CLI layer, not inside `runSync`. OQ3 — manifest write moves BEFORE `runSync` so `--packs-only` short-circuits cleanly. |
| **T3** Flag wiring + `assertNoPhaseBFlags` helper | Four mutex cases. OQ4 (`--full`) + OQ5 (`--prune`) dispositioned as hard error symmetrically with `--index-only`. |
| **T4** STALE_MANIFEST UX nudge | New `detectStaleManifest()` + `staleManifestError()`. Wired into `rule-engine.ts` parser-error path. Semver-minor cohort match (OQ2 — patch tolerated). Cohort-match falls through to original install-hint `TotemParseError`. |
| **T5** Tier-table doc | OQ6 — `sync --packs-only` in Lite, `sync` / `sync --index-only` in Standard. |

## Invariants locked via tests

- `--packs-only` constructs no embedder and never invokes `runSync`. Mutex blocks fire before `loadConfig`/`requireEmbedding`.
- Default flag-less `totem sync` is observably equivalent to prior releases (existing tests stay green; new tests assert Phase A AND Phase B both fire).
- UX nudge fires on missing manifest, pre-1.27.0 manifest (no cohort), malformed cohort, OR semver-minor mismatch. Patch drift passes. Cohort-match falls through to the original `TotemParseError` install-hint path.
- Pre-1.27.0 manifests round-trip cleanly through the new schema.
- Cohort stamping is automatic on every `writeInstalledPacksManifest` call; tests pass an explicit value to override.

Total test deltas: `+5` writer (cohort stamping + forward-compat), `+12` stale-manifest helper, `+3` rule-engine integration (3 STALE_MANIFEST fan-out cases), `+10` syncCommand (Phase A/B split + 4 mutex assertions + invariant about gate ordering). 1814 core tests + 1982 CLI tests pass.

## ADR-101 cross-reference

ADR-101 (`mmnto-ai/totem-strategy:adr/adr-101-totem-sync-architectural-separation.md`) drafted in parallel by strategy-Claude per OQ7 disposition; their session owns the durable record.

## Test plan

- [ ] CI green
- [ ] `pnpm exec totem lint` clean (pre-push verified)
- [ ] `pnpm exec totem review` clean (pre-push verified)
- [ ] CI smoke fixture without API keys: `bump → install → sync --packs-only → lint` exits 0 (manual verification deferred to follow-up — substrate is in place, harness is not)

### DIFF
--- a/.changeset/sync-architectural-separation-1811.md
+++ b/.changeset/sync-architectural-separation-1811.md
@@ -1,0 +1,34 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+'@mmnto/mcp': minor
+'@mmnto/pack-agent-security': minor
+'@mmnto/pack-rust-architecture': minor
+---
+
+`totem sync` Phase A / Phase B architectural separation (mmnto-ai/totem#1811, ADR-101).
+
+`totem sync` decomposes into two independently-runnable phases:
+
+- **Phase A** — deterministic pack-resolution + `installed-packs.json` write (no API key required, runs in CI).
+- **Phase B** — vector-store embedding sync (still requires the embedding key; unchanged).
+
+New mutually-exclusive flags on `totem sync`:
+
+- `--packs-only` (Lite tier): write the pack manifest only; skip embedding sync, prune, the global registry update, and the `review-extensions.txt` write. Designed for CI environments without API keys after a `@mmnto/totem` cohort bump where pack-resolution alone needs to run before `totem lint` recognizes newly registered Tree-sitter languages.
+- `--index-only` (Standard tier): run only the embedding sync; skip pack-resolution. Use when `installed-packs.json` is already current and only the vector store needs to re-embed.
+
+Both flags hard-error when combined with each other or with `--full` / `--prune`.
+
+The CLI orchestrator now writes `installed-packs.json` BEFORE invoking `runSync` so `--packs-only` can short-circuit cleanly. The default flag-less behavior is observably equivalent to prior releases.
+
+UX nudge for stale manifests: when a rule expects a Tree-sitter language that isn't registered, the rule-engine now consults `installed-packs.json`'s cohort field and surfaces a structured `STALE_MANIFEST` `TotemError` pointing at `totem sync --packs-only` whenever the manifest is missing, pre-1.27.0, or written by an engine whose `major.minor` differs from the running version. Patch-level cohort drift passes (caret-range pack semver tolerance). Cohort-match falls through to the original "install the pack" `TotemParseError`.
+
+Schema: `InstalledPacksManifestSchema` gains an optional `cohort: string` field (semver). Pre-1.27.0 manifests without the field continue to parse cleanly. Stamped at write time by `writeInstalledPacksManifest()` from `resolveEngineVersion()`; tests can pre-populate the field to override the stamp.
+
+New public surfaces (additive):
+
+- `resolveEngineVersion(): string`
+- `detectStaleManifest(opts): StaleManifestDetection | null`
+- `staleManifestError(detection, context): TotemError`
+- `TotemErrorCode` adds `'STALE_MANIFEST'` and `'FLAG_CONFLICT'`.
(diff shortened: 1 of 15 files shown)


satur8d:
@coderabbitai @gemini-code-assist re-review please. `f2824b1a` addresses all three findings:

- **GCA HIGH** (sync.ts:142) + **CR MAJOR** (sync.ts:151) — converged on the same defect. `--packs-only` now throws `TotemError('SYNC_FAILED', ...)` on Phase A failure (matching the in-tree prune-failure pattern at sync.ts:222-227). Default mode keeps best-effort warn-and-continue (Phase A is a side-output; Phase B is the headline). Tightened the non-Error fallback to `String(err)` per GCA's secondary point.
- **CR MINOR** (changeset:21) — relaxed the wording. Only `--packs-only` hard-errors with the others; `--index-only` composes with `--full`/`--prune` (all Phase B modifiers).

Two new sync tests guard the carve-out: `--packs-only` + Phase A failure throws `SYNC_FAILED`; default mode keeps warn-and-continue (regression guard).

satur8d:
@coderabbitai @gemini-code-assist re-review please. `8a1d8b0b` addresses all four R2 findings:

- **GCA HIGH** (sync.ts:82 + 117) — consolidated dynamic imports; the inner `await import('@mmnto/totem')` is gone, `resolveInstalledPacks` + `writeInstalledPacksManifest` now ride alongside `runSync`/`TotemError`/`updateRegistryEntry` in the single top-of-handler destructure.
- **GCA HIGH** (sync.ts:147) — refactored the `SYNC_FAILED` throw to pass `err` as the 4th `TotemError` arg (`cause`) instead of concatenating `err.message` into the user message. Stack trace preserved; styleguide §6 honored.
- **CR MAJOR** (sync.ts:130) — sanitized `warning.name` once into a local `packName`, used in both `reasonText` template and `log.warn` format string. Control-char injection surface closed.
- **CR INFO** (sync.test.ts:300-309) — added `expect(mockLoadConfig).not.toHaveBeenCalled()` to lock the mutex-before-config invariant the comment claimed.

Test delta: SYNC_FAILED regex tightened (message no longer carries err detail — it's in `.cause`); added a second-pass capture asserting `err.cause === original`. Mock `TotemError` extended to mirror the real `(code, message, hint, cause)` signature.

satur8d:
@gemini-code-assist re-review please. `40362d39` addresses the stale-manifest finding with a partial accept + scope split:

**Accepted concern (engine-version asymmetry is real).** `manifestCohort` is validated at L95; `engineVersion` was not. A malformed `@mmnto/totem` package.json would crash `semver.major(engineVersion)`.

**Declined the verbatim throw, applied a graceful-fallback guard instead.** `detectStaleManifest` has an explicit graceful-fallback discipline documented at `stale-manifest.ts:52-71` — every failure mode returns either `null` or a structured detection; throwing here breaks the rule-engine caller's contract (`detection !== null → throw staleManifestError`; otherwise fall through to the original `TotemParseError` unmasked). I added `if (!semver.valid(engineVersion)) return null;` right after `resolve()`. New test asserts the branch.

**Ticket-tracked the source-level fix.** Your architectural-layer argument ("core library functions should throw explicit errors on failure") applies more naturally to `resolveEngineVersion()` itself — that's where the value is sourced and where the implicit-semver assumption is made. Tracked as `mmnto-ai/totem#1829` (Tier-3) for a follow-up PR. Once that lands, the consumer-side guard becomes redundant defense-in-depth.

@coderabbitai also re-review.

Issue #1793, opened 2026-05-02
tier-3: lesson-extractor produces structurally-leaky patterns; auto-archive rate signals quality


satur8d:
**Surfaced by:** PR #1790 postmerge (2/2 rules auto-archived in Stage 4) and #1791 close-investigation.
**Tier:** 3 (substrate quality, no acute incident — but corpus growth is bottlenecked by it).

## Symptom

Newly-compiled rules from the lesson extractor are increasingly being auto-archived by Stage 4 with `reasonCode: stage4-out-of-scope-match` because their patterns fire on files outside the rule's intended `fileGlobs`. Concrete examples from PR #1790:

- `JSON.stringify($OBJ)` (intended: catch un-canonical JSON in a specific file) → matches every `JSON.stringify` call in the repo → archived.
- Pattern intended for non-test code in `cli/src` → matches similar idioms in test files within scope → archived.

Per ADR-091's design (see closed ticket #1791), the auto-archive is **correct** — the patterns are genuinely over-broad. But that means the extractor is producing patterns that can't survive Stage 4 without being structurally discriminating enough.

## What the auto-archive rate is actually telling us

Stage 4's auto-archive rate is **telemetry on extractor-pattern-quality**. Not noise. Healthy corpus = low auto-archive rate (extractor produces discriminating patterns). Stressed corpus = high auto-archive rate (extractor producing structurally-leaky patterns).

The 2/2 = 100% auto-archive rate on #1790 isn't an isolated event — it's the empirical signal that the extractor's pattern-generation prompt (or input context) doesn't push hard enough toward structural specificity. The 19 conceptual-skip lessons in the same compile pass are the OTHER side of this coin: lessons that the compiler couldn't reduce to a deterministic pattern at all, surfacing the same gap from the opposite end.

## The architectural rule the extractor needs to internalize

**Pattern-tightening over path-widening.** A pattern that only works because of path filtering is fragile (misnamed files, idiom collision, drift as the repo grows). The structural cure (tighten the pattern) is more durable than the path cure (wider `fileGlobs` negation). The current extractor prompt produces patterns like `JSON.stringify($OBJ)` and relies on `fileGlobs` to constrain them — that's path-widening masquerading as a rule. The fix is:

1. The extractor needs structurally-discriminating shapes: `JSON.stringify` calls *without preceding `canonicalizeKeys`* / *without an explicit replacer arg* / etc. — patterns that capture the actual antipattern, not the syntactic form.
[...]

PR #1832, opened 2026-05-05
fix(cli): graceful skip when compiled-rules.json is empty (#1831)

### LINKED ISSUE #1831: [Bug] 1.28.0 Regression: totem lint throws when compiled rules are empty (strict-on-empty)
### Description
In `@mmnto/totem@1.28.0` and `@mmnto/cli@1.28.0`, there is a regression where `totem lint` strictly throws an error if the compiled rules array is empty:

```
[Totem Error] No compiled rules found at .totem/compiled-rules.json.
  Fix: Run 'totem compile' to generate rules.
```

If there are no lessons (e.g. `.totem/lessons/*.md`), `totem compile` legitimately outputs an empty rules array. In prior versions (like `1.27.0`), `totem lint` would gracefully skip execution (the "no-rules gate" behavior). The 1.28.0 release notes did not document this breaking change.

### Impact
For repositories like `totem-status` that have a valid lint configuration but are pre-lessons (aspirational corpus setup), their CI and local lint scripts now fail completely instead of skipping gracefully.

### Requested Resolution
Either:
1. Restore the graceful skip-when-empty behavior.
2. Add a configuration flag like `lint.allowEmpty: true` to explicitly opt-in to this behavior if strict-mode is the new intentional default.

satur8d:
## Summary

`totem lint` no longer hard-throws `NO_RULES` when `.totem/compiled-rules.json` is empty. The runner now logs the empty-corpus state and returns an empty result so CI exits cleanly. Empty corpus is a legitimate state for early-adoption / aspirational repos like `totem-status` that have a valid lint configuration but have not yet compiled any lessons.

Closes mmnto-ai/totem#1831.

## Bisection — framing correction

The original ticket framed this as a **1.28.0 regression**. Empirical bisection says otherwise:

- The strict throw at `run-compiled-rules.ts:113-119` was added in mmnto-ai/totem#1553 (2026-04-18).
- It shipped unchanged through 1.26.x and 1.27.0 — `git show "@mmnto/cli@1.26.1":packages/cli/src/commands/run-compiled-rules.ts` has identical bytes to the throw site on `main` before this PR.
- `#1828` (the only feature in the 1.27.0→1.28.0 delta) did not touch the lint code path.

What likely changed in the 1.28.0 environment for `totem-status` was an upstream shift that produced a non-empty diff at lint time — the throw was always gated by `getDiffForReview` returning a non-null result, so prior CI runs that returned early on an empty/clean-tree diff never surfaced the throw.

The behavior fix is the same regardless of framing; surfacing this here so future archaeology does not bisect onto `#1828`.

## Design choice

Status-Gemini suggested two paths:
1. Restore graceful skip-when-empty (default-permissive)
2. Add `lint.allowEmpty: true` opt-in (default-strict, opt-in to graceful)

This PR ships path 1 — the simpler default. Consumers that need a "rule count > 0" CI guardrail can check `compiled-rules.json` length directly in their pipeline; the runner deliberately does not opinionate on that policy (YAGNI on a `requireRules` opt-in flag).

## Bundled commits

- `51657f79` — the `#1831` fix (runner + test rewrite + changeset).
- `9e297427` — postmerge artifacts for `mmnto-ai/totem#1828` and `mmnto-ai/totem#1830` (9 lessons, manifest, exports). 0 of 14 attempted compiled to rules — 5 smoke-gate over-matching, 9 conceptual/architectural skips. Continued `mmnto-ai/totem#1793` corpus-health signal.
- `77fd4094` — pre-push lint warning fixes (changeset rephrase + a `totem-context` directive on the new test).

## Test plan

- [x] `pnpm --filter @mmnto/cli test` — all 1984 CLI tests pass; new `returns empty result and skips gracefully...` test verifies the graceful-skip path.
- [x] `pnpm exec totem lint` — PASS, 0 violations on full branch diff.
- [x] `pnpm exec totem review` — PASS, 0 findings.
- [x] Pre-push hook — `4014 tests passed` + lint PASS.
- [ ] CI green on PR.

## AI_PROMPT_BLOCK

This PR addresses mmnto-ai/totem#1831 by replacing the strict-on-empty throw with a graceful skip. The historical framing as a 1.28.0 regression is corrected via bisection in the PR body.


satur8d:
@gemini-code-assist Thanks for the review.

1. **`run-compiled-rules.ts:110` — missing `ruleCount: 0` on early-return — DECLINED.**

   The recommended field does not exist on the function's typed return interface. `RunCompiledRulesResult` at `packages/cli/src/commands/run-compiled-rules.ts:36-49` defines exactly five fields:

   ```ts
   export interface RunCompiledRulesResult {
     violations: Violation[];
     findings: TotemFinding[];
     rules: CompiledRule[];
     output: string;
     regexTimeouts: RuleTimeoutOutcome[];
   }
   ```

   There is no `ruleCount` field. The cross-surface reference to `compile-manifest.json`'s `rule_count` is a separate write-side telemetry surface (snake_case, manifest-only) — not a contract the runtime result type needs to mirror. Per the function's existing convention, the count is exposed to callers via `rules.length`. If the field were genuinely required by the interface, `tsc` would have rejected the omission at compile time; the diff's 1984 passing tests are empirical proof the type is satisfied.

   Decline + capture: `.gemini/styleguide.md` §6 has been updated with a new declined-pattern entry ("Inventing return-object fields by name-similarity to manifest or write-side surfaces"), and a `review-guidance` lesson `lesson-da5025f9.md` has been added so the rationale is in the knowledge base for future reviews on the same module.

Thanks!

PR #1833, opened 2026-05-06
Version Packages

satur8d:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   748b5e6: `totem lint` skips gracefully when `compiled-rules.json` is empty (mmnto-ai/totem#1831).

    Empty-corpus repos (e.g., aspirational pre-lessons setups like `totem-status`) used to hit a hard `NO_RULES` `TotemError` at `run-compiled-rules.ts:113-119` whenever a non-empty diff reached the lint runner. The runner now logs the empty-corpus state and returns an empty result so the caller exits cleanly. This matches the implicit no-op state of repos that have not yet run `totem compile` and have no `compiled-rules.json` on disk — both paths produce zero violations and a zero exit code.

    Consumers that need a "rule count > 0" CI guardrail can check `.totem/compiled-rules.json` rule count directly in their pipeline; the runner deliberately does not opinionate on that policy.

    Bisection note: the strict throw was added in mmnto-ai/totem#1553 (2026-04-18) and shipped unchanged through 1.26.x and 1.27.0. mmnto-ai/totem#1831 framed it as a 1.28.0 regression based on consumer-side observation; the actual 1.28.0 change was an environmental shift (a non-empty diff reached the runner where prior CI passes had returned early via `getDiffForReview`). The behavior change here is the same regardless of framing.

    -   @mmnto/totem@1.28.1

### Patch Changes

-   @mmnto/totem@1.28.1


satur8d:
@gemini-code-assist Thanks for the review.

1. **`packages/cli/CHANGELOG.md:9` — `totem compile` deprecated alias in changeset prose — ACCEPTED.** Replaced with `totem lesson compile` per `.gemini/styleguide.md` §6. Fix landed at `3871c266`.

Thanks!

PR #1868, opened 2026-05-10
feat(cli): totem init Ollama floor probe (#1851 PR-2)

### LINKED ISSUE #1851: DX: surface Ollama-as-floor proactively + lock fallback chain via regression test
## Summary

Two coupled asks closing the consumer-side discoverability gap on the existing `LazyEmbedder` fallback chain (shipped at `mmnto-ai/totem#522`):

1. **Proactive Ollama-as-floor discovery** — surface the embedder fallback expectation at install / init time via `totem init` prompt or `totem doctor` check, so consumers don't perceive the (well-formed) `TotemConfigError` as a crash and reach for the workaround.
2. **Empirical regression test** — assert the `LazyEmbedder` fallback chain produces the documented `TotemConfigError` (with the 3 numbered remediation steps) when both `@google/genai` SDK and Ollama are unavailable. Closes the "speculative residual gap" framing — without this, every future regression in the fallback chain produces another round of consumer-side workaround propagation.

## Context

`mmnto-ai/totem#517` (closed by `mmnto-ai/totem#522`, merged 2026-03-15) shipped the architecture: `LazyEmbedder` with full graceful-degradation chain (configured Gemini/OpenAI provider → Ollama local fallback → clear error with 3-step remediation), `peerDependenciesMeta` declaration on `@google/genai`, dynamic-import-only pattern in `packages/core/src/embedders/embedder.ts`. The architecture is sound.

But the architecture does not surface to consumers at the right time. Empirical observation: status-Gemini hit a perceived "crash" on `totem sync` against the freshly-bumped 1.30.0 cohort 2026-05-07 (`mmnto-ai/totem-status#8`). The "fix" was installing `@google/genai` directly — which short-circuits the fallback chain and propagates a Tenet 16 vendor-coupling workaround across consumers (`mmnto-ai/totem-strategy`, `mmnto-ai/totem-status`, `mmnto-ai/liquid-city` all carry the workaround as a result).

The most plausible interpretation: status-Gemini's environment had no Ollama running. The `LazyEmbedder` fallback reached its terminal state and threw `TotemConfigError('No embedding provider available...')` with the 3-numbered remediation steps. To a consumer, an error from `totem sync` with no obvious-fix pre-existing in their environment looks like a crash even when the error message is well-formed and actionable. The path of least resistance was `pnpm add @google/genai`, not `brew install ollama && ollama serve`.

Full reframe of the original observation lives at `mmnto-ai/totem-strategy:upstream-feedback/064-totem-sync-google-genai-peer-dep-gap.md` (postscript dated 2026-05-09, merged at `mmnto-ai/totem-strategy#260`). Strategy-Gemini synthesis call for this filing recommends the combined ask below.

## Proposed: ask 1 — proactive Ollama-as-floor discovery

The `LazyEmbedder` fallback expectation is currently invisible until an embedder operation triggers and the configured provider fails. By that point the consumer is mid-task and the error feels like an obstacle, not a design intention.

Two surfaces to consider:

- **`totem init` time** — when a consumer first scaffolds a totem-enabled repo, prompt: "Totem can use a cloud embedder (Gemini, OpenAI) or fall back to local Ollama. Ollama is the recommended floor — no API key, no quota, runs locally. Detected: [yes/no]. Install: https://ollama.com." This sets expectations before the consumer runs any embedder operation.
- **`totem doctor` time** — add a check that probes `http://localhost:11434/api/tags` (Ollama default) and reports whether the floor is satisfied. Surfaces the floor expectation diagnostically when the consumer is troubleshooting, regardless of whether they've configured a cloud provider.

`totem doctor` is the lighter-touch surface; `totem init` is the higher-leverage one. Both probably warranted; `doctor` might ship first since `init` flows tend to accumulate gates that require careful UX consideration.

## Proposed: ask 2 — empirical regression test for fallback chain

Add a test against `packages/core/src/embedders/embedder.ts` (or wherever the integration coverage lives) that asserts:

Given `provider: 'gemini'` configured, `@google/genai` SDK absent, and `isOllamaAvailable()` returning `false`, the first `embed()` call throws `TotemConfigError` with:
- Error message containing "No embedding provider available"
- Hint message containing all three numbered remediation steps (install SDK, install Ollama, set `provider: 'ollama'`)
- Error code `CONFIG_MISSING`

This locks the fallback contract. Without it, a future refactor could regress the fallback chain into an opaque crash, restarting the consumer-side workaround propagation cycle.

## Why both, why now

Current state: the architecture works, but the consumer-side workaround is being re-encoded in every new repo because the discoverability gap pushes consumers toward the path of least resistance. Each carrier (`totem-strategy`, `totem-status`, `liquid-city`) has a Tenet 16 vendor-coupling footprint that the architecture explicitly designed to avoid. Closing the discoverability gap (ask 1) removes the pressure to adopt the workaround. Locking the contract via test (ask 2) prevents the pressure from re-emerging via a future regression.

## Acceptance criteria

- `totem doctor` (and/or `totem init`) reports Ollama presence/absence with a clear note about the floor expectation
- Regression test exists and passes against the current `LazyEmbedder` implementation
- `mmnto-ai/totem-strategy:upstream-feedback/064`'s `filed-issue-url:` field updated to point at this issue when triaged

## Cross-references

- `mmnto-ai/totem#517` (the original embedder graceful-degradation feature, closed by `mmnto-ai/totem#522`)
- `mmnto-ai/totem#522` (the merged implementation: `LazyEmbedder` + Ollama fallback + `peerDependenciesMeta`)
- `mmnto-ai/totem-strategy:upstream-feedback/064-totem-sync-google-genai-peer-dep-gap.md` (full reframe with verification trail)
- `mmnto-ai/totem-strategy:design-tenets.md` § 16 (Tenet 16 — Model-Stack Agnosticism — the consumer-coupling problem this closes)
- `mmnto-ai/totem-status#8` (PR with the consumer-side workaround that motivated the original observation)

## Filing source

Filed by strategy-Claude (claude-0047, 2026-05-09) per strategy-Gemini synthesis call dispatched at `mmnto-ai/totem-substrate:.handoff/strategy-claude/inbox/2026-05-09T0845Z-strategy-gemini.md`.

satur8d:
## Summary

PR-2 of 2 closing `mmnto-ai/totem#1851` — the consumer-side discoverability gap on the `LazyEmbedder` fallback chain.

- Surfaces the Ollama floor expectation at `totem init` time via a one-line `log.info` emitted before embedding-tier branching, so cloud-key auto-detection (`OPENAI_API_KEY` / `GEMINI_API_KEY`) doesn't silently bury Ollama as the no-quota fallback option (Tenet 16).
- New exported helper `probeOllamaFloor()` from `@mmnto/cli`'s init module — wraps the existing public `isOllamaAvailable()` (shipped in PR-1, `mmnto-ai/totem#1860`). Returns `{ available, baseUrl, message }`, never throws, 3s `AbortSignal` timeout inherited from the underlying probe.
- Two message forms (covering all spec contract beats: floor framing, detection status, install URL when applicable):
  - **Detected:** `Ollama floor detected at http://localhost:11434 (recommended fallback — no API key, no quota, runs locally).`
  - **Absent:** `Ollama floor not detected (recommended fallback — no API key, no quota, runs locally). Install: https://ollama.com.`
- Skipped in `--bare` mode (no embedder configured) and on re-runs over an existing config (the floor was surfaced at first init).

## Why now

PR-1 (`mmnto-ai/totem#1860`) closed the troubleshoot-time half — `totem doctor` now reports Ollama presence. PR-2 closes the first-setup half. With both shipped, the consumer-side workaround pressure documented in `mmnto-ai/totem-strategy:upstream-feedback/064` (which propagated `pnpm add @google/genai` across `mmnto-ai/totem-strategy`, `mmnto-ai/totem-status`, `mmnto-ai/liquid-city`) is closed at both surfaces.

The architecture (`LazyEmbedder` graceful-degradation chain, `mmnto-ai/totem#522`) was already correct — this PR closes the UX-level discoverability gap so the architecture surfaces to consumers at the right time.

## Architectural mirror

Symmetric with the doctor-side `checkOllama` (`packages/cli/src/commands/doctor.ts:155`) shipped in PR-1: same probe primitive (`isOllamaAvailable`), same UX language ("recommended fallback", "no API key, no quota, runs locally"), different narrative position. Init catches the user at first-setup; doctor catches them at troubleshoot. The same probe runs from three call sites now (`LazyEmbedder` for runtime, `checkOllama` for doctor, `probeOllamaFloor` for init) — all built on `isOllamaAvailable`.

## Constraints honored

- **ADR-026** (Consumer Init Rewrite — Dynamic Agent Detection): one-line non-interactive log, not a yes/no questionnaire.
- **Init scaffolding lesson** (2026-03-05): probe is read-only — no file writes, no summary entry needed.
- **Tenet 16** (Model-Stack Agnosticism): proactively surfaces the local floor so cloud-key auto-detection doesn't bury it.

## Test plan

- [x] `probeOllamaFloor` returns `available=true` with detected message when fetch resolves 200
- [x] `probeOllamaFloor` returns `available=false` with install hint when fetch rejects (`ECONNREFUSED`)
- [x] `probeOllamaFloor` resilience contract — does not throw when probe times out (`AbortError`)
- [x] All 3 tests assert the spec contract beat `no API key, no quota, runs locally` is present
- [x] `pnpm exec vitest run` — 2041/2041 pass workspace-wide
- [x] `pnpm run build` — tsc clean
- [x] `pnpm run format` clean
- [x] `pnpm exec totem lint` — PASS (warnings are pre-existing false positives from over-broad lesson patterns flagging legitimate test mocking + genuinely-async test callbacks)
- [x] `pnpm exec totem review` — PASS, no findings
- [x] `pnpm exec totem verify-manifest` — PASS (475 rules, hashes match)

Closes #1851

### DIFF
--- a/.changeset/totem-init-ollama-floor-prompt.md
+++ b/.changeset/totem-init-ollama-floor-prompt.md
@@ -1,0 +1,27 @@
+---
+'@mmnto/totem': minor
+---
+
+`totem init` now probes the local Ollama daemon (`http://localhost:11434/api/tags`)
+during fresh project setup and emits a one-line floor-expectation message before the
+embedding-tier branch runs. When detected, the message reports the daemon URL; when
+absent, it includes the install hint (`https://ollama.com`). Skipped in `--bare`
+mode (no embedder configured) and on re-runs over an existing config (the floor was
+surfaced at first init).
+
+Closes the consumer-side discoverability gap that motivated the `LazyEmbedder`
+fallback chain in `mmnto-ai/totem#522`: cloud-key auto-detection silently picked
+Gemini/OpenAI without ever telling the user Ollama is the recommended local floor,
+so when the cloud provider failed at `totem sync` time, consumers reached for
+`pnpm add @google/genai` (a Tenet 16 vendor-coupling workaround that propagated
+across `mmnto-ai/totem-strategy`, `mmnto-ai/totem-status`, and `mmnto-ai/liquid-city`)
+instead of the documented Ollama install.
+
+New public helper: `probeOllamaFloor()` exported from `@mmnto/cli`'s init module —
+returns `{ available, baseUrl, message }`, never throws, uses the same 3-second
+`AbortSignal` timeout as `LazyEmbedder` and `totem doctor`. Mirrors the
+`checkOllama` doctor diagnostic shipped in PR-1 (`mmnto-ai/totem#1860`).
+
+Closes `mmnto-ai/totem#1851` (PR-2 of 2 — completes the original two-surface ask;
+PR-1 covered the `totem doctor` half plus the empirical regression test that locks
+the `LazyEmbedder` `TotemConfigError` fallback contract).
--- a/.totem/specs/1851.md
+++ b/.totem/specs/1851.md
@@ -1,82 +1,60 @@
 ### Problem Statement
 
-The `LazyEmbedder` fallback chain correctly degrades to a `TotemConfigError` when configured cloud SDKs and local Ollama are both missing, but consumers perceive this well-formed error as a crash and apply vendor-coupling workarounds (installing the cloud SDK directly). This requires proactively surfacing the Ollama-as-floor expectation via `totem doctor` and locking the fallback error contract with an empirical regression test to prevent future silent breakages.
+The `LazyEmbedder` gracefully degrades to local Ollama when cloud SDKs (e.g., `@google/genai`) are missing, but consumers perceive its terminal failure state—when Ollama is also absent—as an opaque crash, prompting them to manually install cloud SDKs and violate vendor agnosticism. We must surface Ollama as the expected baseline via `totem doctor` diagnostics and permanently lock the fallback error contract (`TotemConfigError` with `CONFIG_MISSING`) via a strict empirical regression test.
 
 ### Architectural Context
 
-This feature reinforces **Tenet 16 (Model-Stack Agnosticism)** by preventing consumer-side coupling to specific vendor SDKs like `@google/genai`. The existing fallback chain (shipped in `#522`) architecturally supports this agnosticism via dynamic imports, but requires the UX discoverability gap closed to be effective in practice. None of the retrieved Totem specific lessons mandate a new pattern here, but the architectural intent heavily prioritizes explicit graceful degradation without blocking.
+This aligns directly with **Tenet 16 — Model-Stack Agnosticism**. The architecture successfully avoids vendor-coupling natively, but the _discoverability gap_ creates consumer-side workarounds (manually installing `@google/genai`). No relevant specific `add_lesson` artifacts were found in the context for this issue, but the overarching theme is proactive graceful degradation.
 
 ### Files to Examine
 
-1. `packages/core/src/embedders/embedder.ts` — Contains `LazyEmbedder`, the target of the empirical regression test, and likely the `isOllamaAvailable` logic.
-2. `packages/core/src/embedders/embedder.spec.ts` (or `.test.ts`) — Where the new regression test asserting the `TotemConfigError` fallback chain will live.
-3. `packages/cli/src/commands/doctor.ts` — The primary surface for Ask 1 to inject the proactive Ollama probe.
-4. `packages/cli/src/commands/init.ts` — The secondary surface for Ask 1 to scaffold expectations at repo creation time.
+1. `packages/core/src/embedders/embedder.ts` — contains the `LazyEmbedder` implementation and its `initPromise` resolution logic.
+2. `packages/cli/src/commands/doctor.ts` — contains the `checkOllama` diagnostic check which needs to proactively surface the floor expectation message.
 
 ### Technical Approach & Contracts
 
-**Approach:**
+**1. Doctor Diagnostic Update:**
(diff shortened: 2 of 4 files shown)


satur8d:
## CR R1 disposition (commit 444cb853)

Both findings accepted.

**Major (init.ts:65 — non-throwing contract):** Wrapped the dynamic import + `isOllamaAvailable` call in try/catch. The pattern is `if (!(err instanceof Error)) throw err;` rather than empty catch — preserves the structural rethrow needed by the Tenet 4 fail-open ast-grep rule (`87aff037d7de47a7`) while still swallowing recoverable failures into floor-absent state. This also honestly separates "recoverable failure" from "unknown shape" so genuinely-unexpected throws still surface to the top-level handler instead of being silently buried.

**Nitpick (test):** Added the regression test mocking `isOllamaAvailable` to reject. Pairs with the Major to lock the never-throws contract against a future upstream loosening of the swallow-all-errors guarantee in the underlying probe.

Verified locally: 4/4 probeOllamaFloor tests pass, 135/135 init.test.ts pass workspace-wide, totem lint PASS, totem review PASS no findings.

satur8d:
@coderabbitai please confirm R1 dispositions on commit 444cb853.

PR #1971, opened 2026-05-19
feat(cli+core): totem mail cross-repo outbox poll subcommand (#1970)

satur8d:
## Summary

- Closes #1970. Option B locked per [ADR-107](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-107-ecl-canonical-over-a2a.md) (ECL canonical; no network-RPC alternative adopted).
- New `totem mail [--json] [--recursive] [--workspace <path>]` subcommand surfaces unread cross-repo mail addressed to this repo's agent(s) by polling sibling-repo outboxes per [ADR-106 § 3](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-106-ephemeral-coordination-layer.md). Replaces ad-hoc per-hook implementations with one cohort-portable surface.
- Canonical port of the strategy-side reference impl ([`mmnto-ai/totem-strategy#373`](https://github.com/mmnto-ai/totem-strategy/pull/373), merged at `ea0ee6d`). Once this lands and cohort-propagates, the strategy hook's inline `pollInboundOutboxes` is superseded by `exec totem mail --json`.
- New `resolveSelfAgents` helper in `@mmnto/totem` exposes the SELF_AGENT resolution logic (env > config.json > basename map) for reuse by hooks, MCP audits, and future surfaces.

## Surfaces

- `packages/core/src/orchestration-resolver.ts` — `resolveSelfAgents(repoRoot, env?)` + `SelfAgentResolution` type. 3-layer precedence chain (`TOTEM_SELF_AGENT` env > `.totem/orchestration/config.json` `host_agents` > basename cohort map). Path-traversal guard at every layer; Zod-validated config schema.
- `packages/cli/src/commands/mail.ts` — `pollMail()` programmatic entry returning structured `MailPollResult` + `mailCommand()` CLI wrapper. JSON output (`--json`) is the hook contract; text output is informational.
- `packages/cli/src/index.ts` — registration alongside the existing administrative subcommands.

## Behavior contract (locked per ADR-107 § Consequences)

| Aspect | Behavior |
|---|---|
| Match | Case-insensitive `to:` frontmatter vs SELF_AGENT set or literal `broadcast` (back-compat with existing `to: arhgap11-Gemini` capitalization) |
| Exclude | Files whose basename appears in any SELF_AGENT's `processed/` or `processed/_broadcast/` directory |
| Workspace | Parent dir of `repoRoot` by default; `TOTEM_WORKSPACE` env or `--workspace` flag override |
[...]

PR #1975, opened 2026-05-20
hooks: wire pollMail() into session-context.js for cross-repo mail surfacing

satur8d:
Closes the consumer-side wiring gap for totem-claude's session-start mail surfacing.

## Context

`pollMail()` shipped in `@mmnto/cli@1.44.0` (via `mmnto-ai/totem#1971`). `mmnto-ai/totem-strategy#377` wired it into strategy-claude's `.claude/hooks/SessionStart.cjs`. But the consumer-side wiring — for totem-claude, lc-claude, status-gemini, arhgap11-* — never propagated.

Result: cross-repo handoffs addressed to totem-claude were **silently invisible** at session-start, requiring user-relay as the workaround. The mail-channel was wired one-way: anyone could send to strategy-claude (whose hook polls), but strategy-claude (or anyone) couldn't send to totem-claude and have them see it auto-surfaced.

Empirical anchor: `mmnto-ai/totem-strategy#382` § Phase 2.C Phase 1 baseline shows 3 of today's 17 strategy-claude correction events were R-class (Reminder) corrections that user-relayed because consumer hooks weren't polling. Per `feedback_hook_enforcement_as_totem_justification`, wiring this hook should reduce R-class corrections at session-start.

## Change

`.claude/hooks/session-context.js` (the V2 hook) gains:

- **`pollInboundMail(gitRoot)` helper** — loads `pollMail` from `packages/cli/dist/commands/mail.js` (workspace-relative pattern matching the existing `loadResolvers` approach for `@mmnto/totem` resolvers; no node_modules round-trip in the monorepo).
- **`SELF_AGENT` constant** defaulting to `'totem-claude'` with the same env-override + path-traversal guard pattern as strategy-claude's hook (`TOTEM_HOOK_SELF_AGENT_OVERRIDE`).
- **Inbound-mail section** in `buildStaticContext()`, emitted BEFORE the active-proposal lookup so any unread handoff is the first inbound signal at session-start (mirrors strategy-claude hook order).

## Verification

Smoke-test from totem repo (full hook invocation):
```
node .claude/hooks/session-context.js
→ "── Inbound mail (cross-repo outbox poll, ADR-106 § 3) ──
   No unread mail addressed to totem-claude or broadcast."
```

Direct `pollMail()` test with proper Windows path:
- workspace resolution: ✅ `D:\dev` (parent of repo)
- self identification: ✅ `{"agents":["totem-claude"],"source":"env"}`
- outbox scanning: ✅ 5 outboxes scanned, 0 warnings
- error handling: ✅ soft-fails to scanError pointer on missing `packages/cli/dist`

Token budget unchanged (lines join + 10k char cap still in effect). Output protocol unchanged (additionalContext JSON payload extended, not restructured).

## Pre-push bypass justification

The repo has a pre-existing WWND warning at `docs/wiki/governing-ai-agents.md:58` documented as a known false-positive in totem-claude's `claude-0062` signoff ("Open at signoff" — *"Pre-existing WWND warning... known false-positive, bypass mechanism documented"*). Used `TOTEM_GATE_BYPASS_JUSTIFICATION` to push; this PR does not modify that file and the warning is unrelated to the wiring change.

## Test plan

- [x] Hook executes cleanly without errors
- [x] Mail section emits correct "No unread" message when no inbound queued
- [x] Workspace walker resolves to parent dir
- [x] Self-agent identification works via env override
- [x] Hook stays within 10k char budget
- [x] `pnpm -F @mmnto/totem build` succeeds with hook in place
- [ ] (Post-merge) Verify totem-claude's next session-start surfaces unread mail from `mmnto-ai/totem-strategy/.totem/orchestration/strategy-claude/outbox/`

## Cohort propagation

This is one of the four consumer repos that need wiring. After this lands, candidates for follow-on PRs (separate, not bundled):

- `mmnto-ai/liquid-city` — lc-claude
- `mmnto-ai/totem-status` — status-* agents
- `mmnto-ai/arhgap11` — arhgap11-* agents

Each follows the same shape (load `pollMail` workspace-relative, hardcode `SELF_AGENT`, emit inbound section in their existing hook). Not bundled here to keep blast radius bounded.

### DIFF
--- a/.claude/hooks/session-context.js
+++ b/.claude/hooks/session-context.js
@@ -51,6 +51,7 @@ async function pollInboundMail(gitRoot) {
       return {
         count: 0,
         files: [],
+        warnings: [],
         scanError: '@mmnto/cli not built at packages/cli/dist; run pnpm -F @mmnto/cli build',
       };
     }
@@ -65,6 +66,10 @@ async function pollInboundMail(gitRoot) {
     return {
       count: (result.mail || []).length,
       files: result.mail || [],
+      // Per-source repo poll failures (e.g. EACCES on a sibling's outbox,
+      // mid-rename race) surface here; pollMail does not throw on them.
+      // Surfaced in the session context per Tenet 13 (sensor visibility).
+      warnings: result.warnings || [],
       scanError: null,
       scanned: result.scanned,
       truncated: result.truncated,
@@ -73,6 +78,7 @@ async function pollInboundMail(gitRoot) {
     return {
       count: 0,
       files: [],
+      warnings: [],
       scanError: String(err && err.message ? err.message : err),
     };
   }
@@ -183,6 +189,12 @@ async function buildStaticContext(gitRoot, branch, ticket) {
       lines.push(`  [scan truncated at ${inbox.scanned} files]`);
     }
   }
+  // Surface per-source warnings (e.g. unreadable sibling outboxes) independently
+  // of the unread-count path — they can co-exist with both zero-mail and
+  // populated-mail states. Per Tenet 13: sensor visibility is the contract.
+  (inbox.warnings || []).forEach((w) => {
+    lines.push(`  Warning: ${w}`);
+  });
   lines.push('');
 
   // Active proposal matching ticket — proposals live in totem-strategy
--- a/.claude/hooks/session-context.js
+++ b/.claude/hooks/session-context.js
@@ -27,6 +27,57 @@ async function loadResolvers(gitRoot) {
   };
 }
 
+// Per-repo convention per ADR-106 § 3. Env override for hook validation
+// (e.g. simulating another agent's view); not used in production.
+// Reject path separators / traversal in override — SELF_AGENT feeds into
+// pollMail's filename-matching, where a malicious value could route to
+// unintended outboxes.
+const DEFAULT_SELF_AGENT = 'totem-claude';
+const _agentOverride = process.env.TOTEM_HOOK_SELF_AGENT_OVERRIDE;
+const SELF_AGENT =
+  _agentOverride && !/[\\/]/.test(_agentOverride) && !_agentOverride.includes('..')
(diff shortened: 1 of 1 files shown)


satur8d:
## Round disposition — `265275024c90f1c31ee3faf52856d71273c948a5` (1 accept pushed, 1 decline)

@coderabbitai @gemini-code-assist

| Finding | Source | Disposition | Notes |
|---|---|---|---|
| pollInboundMail() ignores `warnings` array from pollMail() — sensor-visibility gap per Tenet 13 | GCA HIGH (line 79) + companion at line 173 | **Accept** | `warnings: result.warnings || []` added across all three return paths (dist found, dist missing, outer catch). buildStaticContext() now emits each warning as `  Warning: <msg>` independently of unread-count branches; warnings can co-exist with both zero-mail and populated-mail states. |
| Throw on missing `packages/cli/dist/commands/mail.js` instead of returning scanError ("loud crash" pattern) | GCA HIGH (line 79, second concern) | **Decline — wrong class** | The cited rule applies to CI-only scripts. This is a session-start hook whose outer contract is `process.exit(0); // never crash the agent's session boot` (line 197). A throw inside pollInboundMail propagates through `main()` BEFORE `process.stdout.write(output)`, losing the entire `additionalContext` JSON payload — branch, ticket, MCP reminders, substrate journal, active proposal lookup, vector search. Soft-fail with scanError preserves all other session-start context on a dist-missing dev state. The current behavior matches the parallel pattern in `mmnto-ai/totem-strategy:.claude/hooks/SessionStart.cjs` (which throws inside try → outer catch → returns scanError; functionally equivalent shape). |

One accept pushed in `265275024c90f1c31ee3faf52856d71273c948a5`. Decline reasoned inline; the soft-fail-with-scanError is intentional architectural pairing with the hook's outer `process.exit(0)` contract.

Issue #1981, opened 2026-05-21
post-merge re-index gate skips non-lesson merges — ADR/proposal/journal changes leave LanceDB index stale (ADR-108 anchor)


satur8d:
## Empirical anchor

`mmnto-ai/totem-strategy#388` (ADR-108, P284 Phase 4 synthesis) merged at `b96e143` on `2026-05-21T01:15:23Z`. A follow-on commit `01bf28c` (Tenet 18 citation + ADR-006 cross-link) also landed on `main` post-merge. **~2 hours later**, three independent surfaces still show the index does not contain ADR-108 content:

| Surface | State |
|---|---|
| File on disk | `D:/Dev/totem-strategy/adr/adr-108-agent-state-continuity-architecture-synthesis.md` exists (61KB) |
| `git log` on `totem-strategy/main` | `01bf28c` and `b96e143` both present on `main` |
| `corpus/adr.parquet` | `SELECT id, title FROM 'corpus/adr.parquet' WHERE id LIKE '%108%'` → **0 rows** |
| MCP `search_knowledge` (boundary=strategy) | Three searches on ADR-108-specific terms (`audience layers L1 L2 L3 L4 ADR-108`, `AI as draft author deferred pending proof`, `ADR-108 Phase 4 agent state continuity`) → **0 ADR-108 hits**; top results are pre-ADR-108 (`adr-092`, `proposal-264`, etc.) |

## Root cause (in this repo)

`packages/cli/src/commands/install-hooks.ts:68` — the post-merge hook is gated on a hardcoded `.totem/lessons/` path:

\`\`\`sh
if [ -n \"\$TOTEM_CMD\" ] && git diff-tree -r --name-only ORIG_HEAD HEAD 2>/dev/null | grep -q '\.totem/lessons/'; then
  (\$TOTEM_CMD sync --incremental --quiet > .git/totem-sync.log 2>&1) &
fi
\`\`\`

ADR-108 lives at `adr/adr-108-*.md`. The diff for `b96e143` (and `01bf28c`) touches `adr/`, **not** `.totem/lessons/`. The hook gate returns false; `totem sync` does not fire; LanceDB does not re-embed the new content.

The same exclusion applies to every other targeted artifact class. In strategy repo's `totem.config.ts`:

\`\`\`ts
targets: [
  { glob: 'adr/**/*.md', ... },                  // ← skipped by hook gate
  { glob: 'governance-os-thesis/**/*.md', ... }, // ← skipped
[...]

PR #1939, opened 2026-05-16
feat(cli+core): compile_worker_fingerprint + verify-manifest drift gate (Proposal 278 § Action 3 Phase 1)

satur8d:
## Summary

Ships the implementation slice of [mmnto-ai/totem-strategy#335](https://github.com/mmnto-ai/totem-strategy/pull/335) (Proposal 278 — Compile-Worker Determinism Interim Policy). Phase 1 = anthropic provider only; shell-orchestrator capture is a Phase 2 follow-on that does not gate merge.

**Three implementation surfaces:**
- **Manifest schema** (`packages/core/src/compile-manifest.ts`): adds optional `compile_worker_fingerprint: z.string()` sibling to `model`. Pre-#1937 manifests parse unchanged.
- **Capture** (`packages/cli/src/commands/compile.ts`): both manifest-write sites compute `sha256(canonicalStringify({model, temperature?, seed?, promptTemplateContentHash}))` when `config.orchestrator.provider === 'anthropic'`. `canonicalStringify` drops undefined keys — fingerprint records *absence* (omits slot) when the configured model rejects a sampling parameter, never a placeholder. `--refresh-manifest` preserves the existing fingerprint (Q1 decision: refresh is provenance-preserving by design; `output_hash` and `compile_worker_fingerprint` are orthogonal axes).
- **Drift gate** (`packages/cli/src/commands/verify-manifest.ts`): after the existing hash checks, compares against origin/main's manifest. Fails when fingerprints differ AND `compile-templates.ts` not in branch diff. `--allow-compile-drift` override with two-stage articulation: CI requires `## Compile Drift Justification` PR-body heading (binding accountability); pre-push without an open PR requires `TOTEM_DRIFT_JUSTIFICATION` env var set non-empty (Q3 env-var fortification adopted from strategy-Claude T0855Z reply).

**Scope-limited to mmnto-ai/totem.** External `@mmnto/cli` consumers don't own `packages/cli/src/commands/compile-templates.ts`; firing the drift gate on every routine CLI upgrade would be hostile. Detected via `findMonorepoTemplate()` walking up from cwd looking for the template source; gate skips when not found. Fingerprint is still recorded in their manifest for observability — only the fail-loud surveillance is monorepo-scoped.

## Pre-implementation design alignment

[...]

PR #1922, opened 2026-05-15
feat(mcp+cli): Trap Ledger activity writers — MCP mcp_call + SessionStart session_start (A.3.a writers)

satur8d:
## Summary

Stack-follow-on to #1919 (A.3.a schema, merged at `1934f131`). Wires the two activity-event writers that turn the new schema types into real telemetry. Without these, the schema is inert — no `mcp_call` or `session_start` events ever get produced.

- **MCP `mcp_call` writer** (`packages/mcp/src/ledger-writer.ts`) — `logMcpCall(activityName)` helper wired into `search-knowledge.ts` handler entry. Reads `session_id` from `.totem/ledger/.session-id` if present (24h TTL), omits when missing.
- **SessionStart `session_start` writer** — `CLAUDE_SESSION_START` template extended to mint a UUID via `crypto.randomUUID()`, persist to `.totem/ledger/.session-id`, and append a `session_start` event to `events.ndjson` BEFORE the existing `totem describe` briefing. Stamps `agent_source: 'claude'` (hook knows its origin). Lightweight stderr breadcrumb on catch (CR R1 catch on the original #1920).
- **Core utilities** (`packages/core/src/session-id.ts`) — `mintSessionId` / `writeSessionId` / `readSessionId` with UUID-shape validation + 24h TTL fallback + fs error-class discrimination (ENOENT/EACCES/EPERM/EROFS swallow; unexpected classes rethrow per Tenet 4).

## What this unlocks

A.3.b's `totem doctor --compliance` can compute the ADR-029 metric: "% of sessions where `search_knowledge` was invoked at least once before the first file write." Today: schema present, no events. After this merge: real telemetry, queryable.

## Test plan

- [x] `packages/core/src/session-id.test.ts` — 13 tests (mint, read, write, TTL expiration via `utimesSync` backdating, fail-loud rethrow via `vi.mock('node:fs')`, defensive type guard).
- [x] `packages/mcp/src/ledger-writer.test.ts` — 5 tests (event emission, session_id presence/absence, getContext failure resilience, append-don't-overwrite).
- [x] `packages/mcp/src/tools/search-knowledge.test.ts` — 2 integration tests verifying handler emits `mcp_call` with `activity_name: 'search_knowledge'`.
- [x] `packages/cli/src/commands/init.test.ts` — 5 tests covering SessionStart template's mint/persist/emit/`agent_source`/stderr breadcrumb.
[...]

PR #1983, opened 2026-05-21
feat(cli+core): P281 per-lesson compile cache (incremental compilation)

satur8d:
# P281 Impl Contract Draft — Per-Lesson Hash Stability (Incremental Compile-Worker Cache)

**Author:** totem-claude (claude-0064)
**Date:** 2026-05-21
**Status:** Finalized for PR-open. Cross-stream reviewed by strategy-claude (`2026-05-21T0257Z` + `2026-05-21T0542Z`); 5 OQs dispositioned (no rejects); structural archive-addition retracted under subtraction discipline.
**Authorizes:** impl PR against `mmnto-ai/totem` implementing the P281 incremental compile-worker cache.
**Supersedes:** the prose framing of P281 § Implementation-contract details deferred (now resolved here).

> **Convention:** Per Option-1 disposition this session, P281 ships **without a separate ADR document**. The strategic doctrine is in `mmnto-ai/totem-strategy:proposals/accepted/281-per-lesson-hash-stability.md` (Accepted 2026-05-20). This document is the impl-side contract — it becomes the PR body when the impl lands. Matches the P278 → `mmnto-ai/totem#1939` precedent.

## Summary

Ships the implementation of `mmnto-ai/totem-strategy:proposals/accepted/281-per-lesson-hash-stability.md` (P281 Option C, Accepted 2026-05-20). A per-lesson cache keyed by `(lesson_source_hash, compile_worker_fingerprint)` short-circuits `totem lesson compile` for lessons whose source content is unchanged. Result: a +1 lesson PR produces a `compiled-rules.json` diff with 1 new row, not 4931 lines.

Cohort effect: resolves the productivity tax at source. Bot review surface becomes meaningful. `## Compile Drift Justification` headings become rare (reserved for `--force` / fingerprint-rotation events). Closes the falsifying-metric loop named in P281 § Falsifying Metric.

Sequencing note: 281-first per the 2026-05-19 re-grounded shape (H1 ruled out per `compile-templates.ts` recon; H2 dominant; P281 absorbs P280's primary failure class at source). P280 ships next as defense-in-depth for the rotation events the cache cannot suppress.

## Implementation surfaces

Three files in `packages/core/src/` + one wiring point in `packages/cli/src/commands/compile.ts`. Existing modules unchanged in shape; cache module is additive.

| File | Change | Lines (est.) |
|---|---|---|
| `packages/core/src/compile-cache.ts` (NEW) | Cache module: `CacheEntry` schema (Zod), `computeLessonSourceHash()`, `lookupCacheEntry()`, `writeCacheEntry()`, `migrateFromCompiledRules()`. | ~250 |
| `packages/core/src/compile-cache.test.ts` (NEW) | Partial-mutation invariant test; byte-for-byte preservation on cache hit; migration test; `--force` override test; fingerprint-bump invalidation test. | ~400 |
| `packages/cli/src/commands/compile.ts` (MODIFY at line 1624) | Wrap `compileLessonCore(lesson, ...)` call with cache lookup. Cache hit → return cached `CompileLessonResult` verbatim. Cache miss → invoke core compile + write cache entry. | ~40 |
| `packages/core/src/compile-lesson.ts` | **No change** to pipeline-orchestration logic. The cache wraps `compileLessonCore`; its internals stay untouched. Preserves `compile-noop-refresh.test.ts` + `compile-refresh-manifest.test.ts` byte-for-byte invariants. | 0 |

The cache module locus follows totem-Claude's 2026-05-17T0220Z impl-lane signal in `proposals/accepted/281-per-lesson-hash-stability.md` § Cross-stream sign-off state, point (1).

## Cache schema (with stableId reservation)

Per the P281 § Dependencies "Load-bearing forward-compat schema constraint" (and `mmnto-ai/totem-strategy#387` PR body section 3, merged 2026-05-20 at `8b1e3a5d`):

```typescript
// packages/core/src/compile-cache.ts

import { z } from 'zod';

import { CompileLessonResultSchema } from './compile-lesson.js';

/**
 * Cache entry for one lesson's compile output.
 *
 * `stableId` is reserved for P280 (Wind-Tunnel Decoupling, Accepted
 * 2026-05-20). P281 does not populate or query it. Reserving the slot
 * from day one avoids a full cache re-key event when P280 lands; with
 * the reservation, P280's integration is one new test, not a refactor.
 * Per mmnto-ai/totem-strategy#387 § Dependencies (load-bearing).
 */
export const CacheEntrySchema = z.object({
  sourceHash: z.string(),               // sha256 of lesson source content (normalized)
  stableId: z.string().optional(),      // reserved for P280; P281 ignores
  fingerprint: z.string(),              // compile_worker_fingerprint from P278
  output: CompileLessonResultSchema,    // verbatim CompileLessonResult
  compiledAt: z.string(),               // ISO 8601 timestamp of cached compile
});

export type CacheEntry = z.infer<typeof CacheEntrySchema>;
```

**Critical:** `stableId?: string` slot is in the FIRST commit. Retrofitting later eats the operational friction the constraint exists to prevent. CI-side schema test asserts the field is present and Optional.

## Cache key composition — layered

Per totem-Claude's 2026-05-17T0220Z signal, point (3) — **layered, not composite-hash**. Reasons:

1. A single composite hash conflates source-change vs. fingerprint-change invalidation classes. The `decision` enum in the telemetry event needs to distinguish them.
2. When P280 ships, its `stableId` extends the layered key as a non-breaking composite extension (look up by `stableId` first, fall back to `sourceHash`).
3. Layered key composition makes the `attribution` field (`h1_evidence` / `h2_evidence` / both / neither) populatable correctly at emission time.

**Key layers (in lookup order):**

1. `stableId` (when present in entry — for P280 forward-compat reads; P281 never writes this)
2. `sourceHash` — `sha256(normalize(lesson_source_content))`, where `normalize` collapses `\r\n` → `\n` ONLY. **No trailing-whitespace trim.** Same shape as `generateInputHash()` at `packages/core/src/compile-manifest.ts:60-80`, which is the canonical lessonHash-equivalent. Trimming here would desynchronize the cache key from the canonical hash: a whitespace-only edit would hit the cache and return a rule with a stale `lessonHash` while the manifest pipeline would compute a fresh `lessonHash` for the same edit. Per GCA R2 critical on `#1983` (which reversed an earlier R1 finding of "trim missing" — R1 cited an over-claim in this contract prose itself).

   **Composition (runtime + migration must use the same shape):** the `lesson_source_content` input is computed by `composeLessonSourceForHash(heading, body)` → `\`${heading}\n${body}\``. The runtime path (CLI compile loop) and the migration path (`migrateFromCompiledRules`) both route through this helper. Without the shared helper, the two paths would diverge — exactly the bug GCA R2 surfaced on `#1983` (migration hashed raw lessonSource including `## Lesson — ` framing; runtime hashed `${heading}\n${body}`; same lesson produced two different hashes; migration entries were permanently unreachable).
3. `fingerprint` — `compile_worker_fingerprint` from P278 (`packages/core/src/compile-worker-fingerprint.ts`)

**Lookup contract:**

```typescript
// Pseudocode
function lookupCacheEntry(
  lessonSource: string,
  fingerprint: string,
): CacheEntry | null {
  const sourceHash = computeLessonSourceHash(lessonSource);
  const entry = readCacheEntryByHash(sourceHash);
  if (!entry) return null;
  if (entry.fingerprint !== fingerprint) return null;  // fingerprint rotation invalidates
  return entry;
}
```

**Forbidden in cache key (per totem-Claude 2026-05-17T0220Z signal, point (6)):** `cli_version`. Including it would invalidate every cohort bump — exactly what this proposal prevents.

## Cache file format and storage path

**Path:** `.totem/cache/compile-lesson/<sourceHash-first-16-chars>.json`

**Rationale:**
- Co-located with other `.totem/` substrate; gitignored by default (cache is reproducible from source — verified at `.gitignore:5`).
- One file per cache entry. Reading hot entries doesn't pay for unrelated entries' bytes.
- ADR-083 (content-hash lock) precedent: per-hash file shape.

**Fan-out:** flat directory for v1. Current cohort scale is ~130 lessons; flat performs fine. **Follow-on ticket to be filed post-PR-open** with soft trigger 1000 lessons: when corpus grows past that, shard into prefix-keyed subdirectories (first 2 chars of hash → 256-way fan-out). Per strategy-claude's 0257Z disposition: "anticipated-future-work tickets get filed at insight time, not at trigger time" — filing now prevents lost-context cost when a future agent hits the threshold.

**Format:** JSON, schema-validated on read against `CacheEntrySchema`. Reject and treat as cache miss on schema mismatch (graceful — bad cache entry → recompile, not crash).

**Migration:** see § Migration sequence.

## Cache invalidation triggers

Three triggers covering the cases (per P281 § Scope, "In scope — proposal level", point 3):

| Trigger | Mechanism | Telemetry decision |
|---|---|---|
| Lesson source content change | `sourceHash` mismatch on lookup | `cache_miss_source_changed` |
| `compile_worker_fingerprint` change | `fingerprint` mismatch on lookup (lookup returns null) | `cache_miss_fingerprint_changed` |
| Explicit `--force` flag | CLI bypass: skip `lookupCacheEntry()` for the lesson(s) named | `cache_miss_force` |

**Out-of-scope for v1 (cross-lesson dependencies):** if cross-lesson deps surface empirically (cache produces wrong output for lesson N when lesson M changed), extend `CacheEntry` with `dependsOnHashes: string[]`. Phase 2; not in first impl.

**`--force` scope (per strategy-Gemini 0130Z calibration in P281 § Consequences):**

- `--force <lesson-id>` (default for ad-hoc): invalidate one lesson; surgical.
- `--force-all`: invalidate corpus-wide; reserved for compile-worker version bumps / ast-grep upgrades / large refactors. Requires `## Compile Cache Reset Justification` heading.

**Two distinct justification headings — different triggers, same shape:**

| Heading | Source | Triggered by |
|---|---|---|
| `## Compile Drift Justification` | `mmnto-ai/totem#1939` (P278 Phase 1, verify-manifest gate) | `compile_worker_fingerprint` rotation (model/template change produced new output despite same inputs) |
| `## Compile Cache Reset Justification` | This PR (P281 cache `--force-all`) | Explicit operator-side cache invalidation (`--force-all`) — bypasses cache, recompiles everything |

Both ride the existing `TOTEM_DRIFT_JUSTIFICATION` PR-body-scan discipline. Distinct headings keep the audit trail interpretable: a fingerprint-rotation event is structurally different from an operator-side cache flush, even though both produce a full corpus recompile.

## Migration sequence

**Step 1 (one-shot, on first run after impl ships):**

Existing `.totem/compiled-rules.json` is the seed. For each `lessonHash` in the manifest:
- Read the corresponding lesson file from `.totem/lessons/`
- Compute `sourceHash` from the file content (normalize line endings)
- Read the current `compile_worker_fingerprint` from `.totem/.totem-manifest.json`
- Write a `CacheEntry` at `.totem/cache/compile-lesson/<sourceHash-first-16-chars>.json` carrying the existing compile output

Cache hit rate after migration step: 100% (every lesson hashes the same input that produced the existing output).

**Step 2 (subsequent runs):** normal cache flow.

**No automatic eviction in v1.** Append-only substrate matches the codebase as ground truth (Tenet 17). Stale cache entries (lessons removed from `.totem/lessons/`) cost O(MB) disk. Eviction policy can land in a follow-on PR if disk pressure surfaces.

`totem cache --prune-orphans` (manual operator command) covers the "I want to clean this up" case. Out of scope for first impl; trivial follow-on.

## Telemetry — `compile_cache_decision` event

Per P281 § Telemetry. Emitted to the Trap Ledger (`events.ndjson`, A.3.a writers, `mmnto-ai/totem#1922`) per lesson per compile run.

```typescript
interface CompileCacheDecisionEvent {
  lesson_id: string;            // sourceHash (until P280's stable_id ships; then stable_id)
  decision: 'cache_hit' | 'cache_miss_source_changed' | 'cache_miss_fingerprint_changed' |
            'cache_miss_force' | 'cache_miss_no_prior_record';
  attribution: 'h1_evidence' | 'h2_evidence' | 'both' | 'neither';
  compile_worker_fingerprint: string;
  cli_version: string;          // observability only; NOT in cache key
}
```

`attribution` populated on every event (defense-in-depth for the H1-absence ruling, per P281 § Telemetry "Why `attribution` is named at proposal stage"). Initial impl emits `attribution: 'neither'` for `cache_hit` events; H1/H2 evidence categorization is a Phase 2 telemetry pass.

## Non-functional requirement (load-bearing)

**Compile-worker MUST NOT read sibling lessons.**

Per totem-Claude 2026-05-17T0220Z signal, point (6) — pin this as a non-functional requirement on the prompt template. Currently true: `packages/cli/src/commands/compile-templates.ts` has zero cross-lesson references. The requirement guards against future drift (new RAG path, new few-shot source, future template refactor) that would silently invalidate the H1-absence ruling and re-enable the rotation class.

**Enforcement (light):**
- Test asserts `compile-templates.ts` contains zero `for ... lesson` / `forEach(lesson` / `lessons.map` patterns **outside template literals**. The grep is scoped via a regex that strips backtick-delimited template strings before checking — strategy-claude's verification (`grep -E "for.*lesson|lessons\\.(map|forEach|filter)|\\.lessons\\["` against current `compile-templates.ts` returned only natural-language matches inside template-literal prompt prose). False positives on prose mentions of "lessons" would fail an unscoped grep.
- CI gate filed as a separate follow-on: a structural lint that walks the prompt template's AST and rejects multi-lesson references in code (not prose). Not blocking on this PR.

## Test strategy

| Test | Asserts |
|---|---|
| `cache_hit_byte_for_byte_preservation` | Cache hit produces a `CompileLessonResult` byte-identical to the cached output. No path through `compileLessonCore`. |
| `partial_mutation_invariant` | Corpus with 130 lessons; modify 1 lesson; verify only 1 `compiled-rules.json` row changes, the other 129 preserve their prior `lessonHash` exactly. This is the falsifying-metric test. |
| `fingerprint_change_invalidates_all` | Bump `compile_worker_fingerprint`; verify every lesson recompiles (`cache_miss_fingerprint_changed` for all). |
| `source_change_invalidates_one` | Modify 1 lesson's source; verify only that lesson recompiles. |
| `force_per_lesson` | `totem lesson compile --force <id>`; verify only that lesson bypasses cache. |
| `migration_seed_from_compiled_rules` | Start with no cache + existing `compiled-rules.json`; first run produces 100% cache hit on subsequent runs. |
| `stable_id_slot_reserved` | Cache entry schema accepts `stableId?: string`; field is Optional; absent field does not invalidate lookup. |
| `cli_version_not_in_key` | Bump `cli_version`; verify cache stays valid (no spurious invalidation). |
| `bad_cache_entry_graceful` | Inject a schema-invalid entry; verify lookup returns null (cache miss), no crash. |
| `compile_worker_no_sibling_reads` (light) | Heuristic check on `compile-templates.ts` content. |

Existing tests preserved unchanged:
- `compile-noop-refresh.test.ts` — `--refresh-manifest` byte-for-byte invariant
- `compile-refresh-manifest.test.ts` — manifest refresh semantics
- `compile-lesson.test.ts` — pipeline-orchestration behavior

## Performance expectations

| Scenario | Expected behavior |
|---|---|
| Steady-state PR (1 lesson changed in 130-lesson corpus) | ≥99% cache hit. 1 compile-worker call. `compiled-rules.json` diff ~5-20 lines. |
| New install (no cache, existing `compiled-rules.json`) | Migration seeds cache. Next run: 100% cache hit. |
| `compile_worker_fingerprint` bump (e.g., model upgrade) | 100% cache miss. Full recompile. Telemetry: 130x `cache_miss_fingerprint_changed`. |
| `--force-all` | Same as fingerprint bump; explicit operator intent. |
| Cache disabled (`TOTEM_DISABLE_COMPILE_CACHE=1`, env override for emergencies) | Behavior reverts to pre-#NNNN status quo. Emergency escape only — not a long-term flag. **Deprecation-watch ticket to be filed post-PR-open** so the flag has a documented end-of-life if cache stability holds across N=20 PR window. |

## Falsifying metric implementation

Per P281 § Falsifying Metric — ≥99% of compile runs touching K lessons (by source change) produce `compiled-rules.json` row changes for exactly K lessons.

The `compile_cache_decision` telemetry events drive measurement. `totem doctor --compliance` (per P281 § Telemetry, "composes with") aggregates across the first N=20 lesson-mutating PRs post-thaw. Threshold check is mechanical: across N=20 PRs, count (compile runs where actual changed rows == expected changed rows from source diff) / 20. Falsified if < 19/20 (95% lower bound on the 99% claim).

Window: event-count-bounded (N=20 PRs), not time-bounded. Same calibration as P280.

## Cross-references

- `mmnto-ai/totem-strategy:proposals/accepted/281-per-lesson-hash-stability.md` (canonical doctrine; Accepted 2026-05-20)
- `mmnto-ai/totem-strategy:proposals/accepted/280-wind-tunnel-decoupling.md` (sibling axis; ships next as defense-in-depth)
- `mmnto-ai/totem-strategy:proposals/accepted/278-compile-worker-determinism-interim-policy.md` (fingerprint mechanism; cache invalidation key)
- `mmnto-ai/totem-strategy#387` (P280+P281 Accepted PR, merged `8b1e3a5d`; load-bearing `stableId` reservation specified)
- `mmnto-ai/totem-strategy#388` (ADR-108 — P284 Phase 4 synthesis, impl-lane unblock signal)
- `mmnto-ai/totem-strategy:adr/adr-083-actor-aware-enforcement-content-hash.md` (content-hash cache precedent)
- `mmnto-ai/totem-strategy:adr/adr-103-deterministic-rule-compilation.md` (Convergent Spine; this work restores its end-to-end value-proof loop)
- `mmnto-ai/totem#1939` (verify-manifest drift gate, P278 Phase 1; precedent for impl-PR-body-as-contract; sibling drift-justification mechanism)
- `mmnto-ai/totem#1981` (post-merge re-index gate fix — adjacent value-proof loop concern; filed this session)
- `mmnto-ai/totem-strategy#284` (P284 parent, agent state continuity)

## Sequencing & next steps

1. ✅ This contract drafted (claude-0064, 2026-05-21).
2. ✅ Strategy-claude cross-stream review (`2026-05-21T0257Z-totem-claude.md`) — all 5 OQs dispositioned (no rejects).
3. ✅ Three-decision close (`2026-05-21T0542Z-totem-claude.md`) — scope narrowed to P281 PR-open; P280 contract revision deferred until P280 sequences.
4. ⏳ Open P281 impl PR Ready (not Draft) with this document as the PR body.
5. ⏳ Bot review walk (CR + GCA) — consolidated round comments per `feedback_bot_protocols_centralized` § 8.1.
6. ⏳ Merge. First end-to-end clean recompile (**first freeze-lift gate** per ADR-108 framing).
7. ⏳ Post-PR-open tail work (non-gating): file follow-on tickets (fan-out cutover; `TOTEM_DISABLE_COMPILE_CACHE` deprecation watch); update `mmnto-ai/totem#1981` with sharper diagnosis comment.
8. ⏳ P280 contract revision when P280 enters its lane (audit-freshness-rebase per strategy-claude 0542Z Decision 3).
9. ⏳ P280 impl PR. Second end-to-end clean recompile (**second freeze-lift gate**). Both gates close → `project_rule_compilation_freeze_2026_05_17` lifts → Convergent Spine unfreezes.

— totem-claude (2026-05-21)

### DIFF
--- a/.changeset/p281-compile-cache.md
+++ b/.changeset/p281-compile-cache.md
@@ -26,11 +26,11 @@ Layered (not composite). `stableId` first when present (P280 reservation; v1 nev
 
 ## Telemetry
 
-`compile_cache_decision` event per lesson per compile run. Best-effort fire-and-forget; ledger-write failures never block compile.
+`compile_cache_decision` event per lesson per compile run. Best-effort fire-and-forget; ledger-write failures are swallowed at the writer site so a degraded ledger does not block compile.
 
 ## Emergency escape
 
-`TOTEM_DISABLE_COMPILE_CACHE=1` reverts cache to no-op (lookup → no-prior-record, write → no-op). Emergency only — deprecation-watch tracked post-PR-open.
+`TOTEM_DISABLE_COMPILE_CACHE=1` reverts cache behavior to the pre-#NNNN status quo (lookup returns no-prior-record, write becomes a no-op). Emergency-only; deprecation-watch tracked post-PR-open.
 
 ## Falsifying metric
 
--- a/.changeset/p281-compile-cache.md
+++ b/.changeset/p281-compile-cache.md
@@ -1,0 +1,41 @@
+---
+'@mmnto/cli': minor
+'@mmnto/totem': minor
+---
+
+feat(cli+core): Proposal 281 per-lesson compile cache (incremental compilation) (#NNNN)
+
+Implements [Proposal 281 (Per-Lesson Hash Stability)](https://github.com/mmnto-ai/totem-strategy/blob/main/proposals/accepted/281-per-lesson-hash-stability.md), accepted via [`mmnto-ai/totem-strategy#387`](https://github.com/mmnto-ai/totem-strategy/pull/387) on 2026-05-20. Closes the first of two freeze-lift gates for `project_rule_compilation_freeze_2026_05_17` (Path A per [ADR-108](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-108-agent-state-continuity-architecture-synthesis.md)).
+
+A per-lesson cache keyed by `(sourceHash, compile_worker_fingerprint)` short-circuits `totem lesson compile` for lessons whose source content is unchanged. Result: a `+1` lesson PR produces a `compiled-rules.json` diff with 1 new row, not 4931 lines.
+
+## Surfaces
+
+- `packages/core/src/compile-cache.ts` — new module: `CacheEntrySchema` (with `stableId?: string` reserved for P280 per [`mmnto-ai/totem-strategy#387`](https://github.com/mmnto-ai/totem-strategy/pull/387) § Dependencies), `computeLessonSourceHash`, `lookupCacheEntry`, `writeCacheEntry`, `buildCacheEntry`, `migrateFromCompiledRules`, `cacheEntryPath`, `listCacheEntries`.
+- `packages/core/src/compile-cache.test.ts` — 21 tests covering the partial-mutation invariant (falsifying-metric test), byte-for-byte preservation on hit, fingerprint-rotation invalidation, source-change invalidation, `--force` bypass, `stableId` slot round-trip, graceful-miss on malformed cache files, `TOTEM_DISABLE_COMPILE_CACHE` env-var escape hatch, and migration idempotence.
+- `packages/core/src/ledger.ts` — `LedgerEventSchema.type` enum extended with `compile_cache_decision`. `ruleId` carries the lesson `sourceHash`; `activity_name` carries the decision enum value (`cache_hit` / `cache_miss_source_changed` / `cache_miss_fingerprint_changed` / `cache_miss_force` / `cache_miss_no_prior_record`).
+- `packages/cli/src/commands/compile.ts` — parallel compile path (~line 1614) wrapped with cache lookup. Cache hit returns the stored `CompileLessonResult` verbatim; cache miss invokes `compileLessonCore` and persists the result. `options.force` and `upgradeTargets` are honored as forced-recompile signals. Cache is bypassed when `compile_worker_fingerprint` is `undefined` (matching the `verify-manifest` provider gap discipline).
+
+## Cache key composition
+
+Layered (not composite). `stableId` first when present (P280 reservation; v1 never writes it) → `sourceHash` (SHA-256 of normalized lesson source, line endings collapsed to `\n`) → `fingerprint` (exact match required, else miss). `cli_version` is intentionally absent from the key — including it would invalidate every cohort bump.
+
+## Storage
+
+`.totem/cache/compile-lesson/<sourceHash-first-16-chars>.json`. Already gitignored (`.gitignore:5`). Flat directory for v1; fan-out follow-on tracked post-PR-open with soft trigger 1000 lessons.
+
+## Telemetry
+
+`compile_cache_decision` event per lesson per compile run. Best-effort fire-and-forget; ledger-write failures never block compile.
+
+## Emergency escape
+
+`TOTEM_DISABLE_COMPILE_CACHE=1` reverts cache to no-op (lookup → no-prior-record, write → no-op). Emergency only — deprecation-watch tracked post-PR-open.
+
+## Falsifying metric
+
+Per [Proposal 281 § Falsifying Metric](https://github.com/mmnto-ai/totem-strategy/blob/main/proposals/accepted/281-per-lesson-hash-stability.md): ≥99% of compile runs touching K lessons (by source change) produce `compiled-rules.json` row changes for exactly K lessons. Window: event-count-bounded (N=20 PRs post-thaw).
+
+## Sequencing
+
+P281 ships first (per [`mmnto-ai/totem-strategy#387`](https://github.com/mmnto-ai/totem-strategy/pull/387) re-grounded shape + ADR-108 Path A). P280 (`stableId` on lesson frontmatter) follows as defense-in-depth — the reserved `stableId` slot in `CacheEntry` makes P280's arrival additive, not breaking. The rule-compilation freeze lifts after both gates close + end-to-end clean recompile passes.
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -1622,15 +1622,25 @@ export async function compileCommand(
         // so caching would be unsafe. Matches the verify-manifest no-op
         // discipline on the same provider gap.
         const cacheFingerprint = computeFingerprintForManifest();
+        const cacheSessionId = readSessionId(totemDir);
         let cacheCliVersion: string | undefined;
         try {
           const { createRequire } = await import('node:module');
           const req = createRequire(import.meta.url);
           const pkg = req('../../package.json') as { version?: string };
           cacheCliVersion = pkg.version;
-          // totem-context: cli_version is a best-effort ledger enrichment; mirror of doctor-claim-discipline's pattern. Resolution failure must not block compile.
+          // totem-context: intentional cleanup — cli_version is a best-effort ledger enrichment; packaging-path resolution failure surfaces diagnostically via log.dim per CR R1 finding, but never blocks the compile path. Mirror of doctor-claim-discipline's pattern.
         } catch (err) {
-          void err;
+          // totem-context: intentional cleanup — cli_version is a best-effort ledger enrichment; packaging-path resolution failure surfaces diagnostically via log.dim per CR R1 finding, but never blocks the compile path.
+          // Surface at diagnostic level — silent swallow would mask packaging
+          // regressions that strip cli_version from every cache-telemetry event.
+          const errMsg =
+            // totem-context: String(err) is the canonical err-normalization idiom (10+ cohort precedents); the input-pattern lesson misfires on catch-block error extraction.
+            err instanceof Error ? err.message : String(err);
+          log.dim(
+            TAG,
+            `Unable to resolve CLI version for compile_cache_decision telemetry: ${errMsg}`,
+          );
         }
         const emitCacheDecisionEvent = (
           sourceHash: string,
@@ -1651,6 +1661,7 @@ export async function compileCommand(
                 justification: '',
                 source: 'lint',
                 activity_name: decision,
+                ...(cacheSessionId !== undefined ? { session_id: cacheSessionId } : {}),
                 ...(cacheCliVersion !== undefined ? { cli_version: cacheCliVersion } : {}),
               },
               (msg) => log.warn(TAG, msg),
@@ -1678,7 +1689,12 @@ export async function compileCommand(
               // recompile intent and must not short-circuit).
               const forceRecompile =
                 options.force === true || upgradeTargets?.has(lesson.hash) === true;
-              const sourceHash = computeLessonSourceHash(lesson.body);
+              // Cache key must cover heading + body (CR Major catch on
+              // `mmnto-ai/totem#1983`). Hashing body alone would let a
+              // heading-only edit falsely hit the cache while the rotation-prone
+              // `lessonHash` shifts to reflect the new heading — stale output
+              // would survive into compiled-rules.json.
+              const sourceHash = computeLessonSourceHash(`${lesson.heading}\n${lesson.body}`);
               if (cacheFingerprint !== undefined) {
                 const lookup = lookupCacheEntry(totemDir, sourceHash, cacheFingerprint, {
                   force: forceRecompile,
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -1645,12 +1645,11 @@ export async function compileCommand(
         }
         const emitCacheDecisionEvent = (
           sourceHash: string,
-          decision:
-            | 'cache_hit'
-            | 'cache_miss_source_changed'
-            | 'cache_miss_fingerprint_changed'
-            | 'cache_miss_force'
-            | 'cache_miss_no_prior_record',
+          // Use the canonical CacheDecision union exported from core
+          // (compile-cache.ts). Inline-import-type avoids a top-level
+          // type-import while keeping the wrapper site in lockstep with
+          // future additions to the decision enum.
+          decision: import('@mmnto/totem').CacheDecision,
         ): void => {
           try {
             appendLedgerEvent(
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -439,6 +439,7 @@ export async function compileCommand(
     resolveStage4Baseline,
     appendLedgerEvent,
     buildCacheEntry,
+    composeLessonSourceForHash,
     computeCompileWorkerFingerprint,
     computeLessonSourceHash,
     lookupCacheEntry,
@@ -1689,12 +1690,14 @@ export async function compileCommand(
               // recompile intent and must not short-circuit).
               const forceRecompile =
                 options.force === true || upgradeTargets?.has(lesson.hash) === true;
-              // Cache key must cover heading + body (CR Major catch on
-              // `mmnto-ai/totem#1983`). Hashing body alone would let a
-              // heading-only edit falsely hit the cache while the rotation-prone
-              // `lessonHash` shifts to reflect the new heading — stale output
-              // would survive into compiled-rules.json.
-              const sourceHash = computeLessonSourceHash(`${lesson.heading}\n${lesson.body}`);
+              // Cache key composition (CR Major on `mmnto-ai/totem#1983` R1)
+              // + same shape used by `migrateFromCompiledRules` to keep runtime
+              // and migration paths producing identical hashes for the same
+              // lesson (GCA R2 critical on the same PR). The shared helper is
+              // the canonical anchor — both paths route through it.
+              const sourceHash = computeLessonSourceHash(
+                composeLessonSourceForHash(lesson.heading, lesson.body),
+              );
               if (cacheFingerprint !== undefined) {
                 const lookup = lookupCacheEntry(totemDir, sourceHash, cacheFingerprint, {
                   force: forceRecompile,
--- a/packages/cli/src/commands/compile.ts
+++ b/packages/cli/src/commands/compile.ts
@@ -438,7 +438,10 @@ export async function compileCommand(
     readCompileManifest,
     resolveStage4Baseline,
     appendLedgerEvent,
+    buildCacheEntry,
     computeCompileWorkerFingerprint,
+    computeLessonSourceHash,
+    lookupCacheEntry,
     modelStripsTemperature,
     readPromptTemplateContentHash,
     readSessionId,
@@ -451,6 +454,7 @@ export async function compileCommand(
     STAGE4_MANIFEST_EXCLUSIONS,
     verifyAgainstCodebase,
     verifyRuleExamples,
+    writeCacheEntry,
     writeCompileManifest,
   } = await import('@mmnto/totem');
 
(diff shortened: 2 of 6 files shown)


satur8d:
## Round 1 — disposition

@coderabbitai @gemini-code-assist — thanks for the R1 walk. Six findings, all accepted, fixes pushed at `29bf62c0`.

| # | Bot | Sev | Location | Disposition |
|---|---|---|---|---|
| 1 | CR | 🟡 Minor | `compile.ts:1634` cli_version empty catch | **Accept** — wrapped in `log.dim()` with diagnostic message + dual `totem-context: intentional cleanup` directives (catch is best-effort enrichment by design; the directive is the right escape for Tenet 4) |
| 2 | CR | 🟠 Major | `compile.ts:1655` `compile_cache_decision` missing `session_id` | **Accept** — added `readSessionId(totemDir)` resolution + spread `{ session_id: ... }` into the event payload. Correlation back to `compile_run` now works |
| 3 | CR | 🟠 Major | `compile.ts:1687` cache key uses `lesson.body` only | **Accept** — composed key as `\`${lesson.heading}\n${lesson.body}\``. Heading-only edits now invalidate the cache correctly. Added regression test `distinguishes heading-only edits from body-only edits` |
| 4 | CR | 🟡 Minor | `compile-cache.test.ts:103` (5 sites) `os.tmpdir()` | **Accept** — workspace-scoped `TEST_TEMP_ROOT` under `<package>/.totem/temp/compile-cache-tests/`, anchored via `import.meta.url` so it resolves regardless of vitest invocation cwd. `.totem/temp/` already gitignored |
| 5 | GCA | high HIGH | `compile-cache.ts:101` missing `.trimEnd()` | **Accept** — added per contract. Inputs differing only in trailing whitespace now hash identically. Added regression test `normalizes trailing whitespace` with 4 assertions |
| 6 | CR | 🧹 Nit | `compile-cache.ts:154` bare catch missing `err` binding | **Accept** — `catch (err) { void err; }` per repo `err` (never `error`) convention |

### Bug-fix validation tests added (caught by local shield review)

Two tests added to lock in the bug-fix behavior:

- `distinguishes heading-only edits from body-only edits` — exercises the composition shape directly (catches regressions if the cache key ever loses the heading component)
- `normalizes trailing whitespace` — 4 assertions covering trailing `\n`, multiple newlines, trailing tabs/spaces, mixed whitespace

Total cache tests: 29 / 29 passing. Byte-for-byte invariant tests (`compile-noop-refresh`, `compile-refresh-manifest`) still pass unchanged.

### Local lint (strict): PASS (0 errors)

One remaining warning on the pre-existing `Promise.all(batch.map(...))` parallel-compile pattern from [Proposal 188 Phase 1](https://github.com/mmnto-ai/totem-strategy/blob/main/proposals/accepted/188-parallel-compile-phase-1.md) — load-bearing for concurrency. Not new in this PR; flagged because the diff range touched adjacent code. Not blocking.

Ready for R2 walk.

satur8d:
## Round 2 — disposition

@coderabbitai @gemini-code-assist — R2 walk complete. Seven findings; **2 accepted (both load-bearing), 5 declined (CR phantom replays of R1)**. Fixes pushed at `884d73dd`. PR body updated to reflect the contract correction.

### Accepted

| # | Bot | Sev | Location | Disposition |
|---|---|---|---|---|
| 1 | GCA | critical | `compile.ts:1697` runtime/migration hash mismatch | **Accept** — runtime hashed `${lesson.heading}\n${lesson.body}`; migration hashed raw `input.lessonSource` (with `## Lesson — ` framing). Same lesson → two hashes → migration entries permanently unreachable. Introduced `composeLessonSourceForHash(heading, body)` as the canonical composition anchor. Both call sites route through it. Migration's input shape changed from `{ lessonSource }` to `{ heading, body }` to enforce the call shape. Added regression test `runtime/migration hash consistency`. |
| 2 | GCA | critical | `compile-cache.ts:104` `.trimEnd()` breaks canonical lessonHash invariant | **Accept** — reverses GCA R1. The canonical `generateInputHash` (`compile-manifest.ts`) is trailing-whitespace sensitive (CRLF→LF only, no trim). My `.trimEnd()` — applied in R1 in response to GCA R1's "trim missing" finding — desynchronized the cache key from the manifest's lessonHash. A whitespace-only edit would cache-hit + return a rule with stale `lessonHash` while the manifest pipeline computed a fresh one. R2 is correct. Reverted to CRLF→LF only. Reversed the test assertions accordingly. **Note: GCA R1's framing cited an over-claim in my own contract prose ("trim trailing whitespace"). PR body now corrected to reflect the canonical no-trim shape.** |

### Declined (CR phantom replays)

CR R2 surfaced 5 findings that are **identical fixes already applied in R1 commit `29bf62c0`**. Per `feedback_cr_phantom_proposed_fix_equals_current`, declining with line-by-line evidence:

| # | CR Finding | Already-applied evidence (in `884d73dd`) |
|---|---|---|
| 1 | 🟡 Minor: cli_version empty catch | `compile.ts:1633-1647` — catch already wraps `log.dim(TAG, ...)` with dual `totem-context: intentional cleanup` directives |
| 2 | 🟠 Major: missing `session_id` on `compile_cache_decision` | `compile.ts:1625` `cacheSessionId = readSessionId(totemDir)`; `compile.ts:1664` spread `{ session_id: cacheSessionId }` into event |
| 3 | 🟠 Major: cache key uses `lesson.body` only | `compile.ts:1697` already composes `composeLessonSourceForHash(lesson.heading, lesson.body)` (improved further in R2 to use the shared helper) |
| 4 | 🟡 Minor: `os.tmpdir()` in tests | `compile-cache.test.ts:43-50` — workspace-scoped `TEST_TEMP_ROOT` anchored via `import.meta.url`; all 5 `mkdtempSync` sites route through `makeTestTmpDir()` |
| 5 | 🧹 Nit: bare catch missing `err` binding | `compile-cache.ts:157` — already `catch (err) { void err; ... }` per R1 |

CR appears to be re-walking the diff against `main` without recognizing the R1 commit applied the fixes. CR's own prompt-for-AI-agents footer directs "Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason" — that's what this disposition does.

### Bug-fix validation tests added

Three new tests lock in the R2 corrections:

- `runtime/migration hash consistency` — seed via migration, retrieve via runtime, assert cache_hit
- `composeLessonSourceForHash distinguishes heading-only edits from body-only edits` — defends the shared composition contract
- `preserves trailing-whitespace sensitivity` — reverses the R1-incorrect assertion; locks in canonical alignment with `generateInputHash`

Total cache tests: 33 / 33 passing. Byte-for-byte invariant tests (`compile-noop-refresh`, `compile-refresh-manifest`): 11 / 11 still passing.

### Local lint (strict): PASS (0 errors)

One remaining warning on `Promise.all(batch.map(...))` from Proposal 188 Phase 1 — load-bearing parallel-compile pattern, not new in this PR.

### Note on the GCA R1 → R2 reversal

Worth flagging for future review: GCA R1 cited the PR contract's "trim trailing whitespace" claim as authoritative for the `.trimEnd()` ask. The contract was wrong — over-claimed normalization scope. GCA R2 correctly anchored to the canonical `generateInputHash` shape and reversed. **Lesson:** prose-contract claims need to be sourced from canonical code, not the other way around. PR body now corrected. Banking as a discipline note for future contracts.

Ready for R3 walk.

satur8d:
## Round 3 — disposition

@coderabbitai — R3 walk complete. **2 NEW Majors, both accepted (both real bugs).** Fixes pushed at `70a11df5`.

### Accepted

| # | Bot | Sev | Location | Disposition |
|---|---|---|---|---|
| 1 | CR | 🟠 Major | `compile-cache.ts:37` CACHE_DIR duplicates `.totem` prefix | **Accept** — `path.join('.totem', 'cache', 'compile-lesson')` joined with `totemDir` (which already IS the `.totem` directory at runtime) landed cache files at `<totemDir>/.totem/cache/compile-lesson/...` instead of `<totemDir>/cache/compile-lesson/...`. Outside the documented path; tooling scanning `<totemDir>/cache` would miss them. Tests passed because they encoded the same wrong path internally (local shield review missed it). Dropped `.totem` from CACHE_DIR + fixed the path test + added regression test asserting no double-prefix. |
| 2 | CR | 🟠 Major | `compile-cache.ts:52` minimal schema accepts incomplete payloads | **Accept** — `{ status: 'compiled' }` (no `rule`) survived the status-enum schema, got cast to `CompileLessonResult`, would crash the cache-hit path on `result.rule` dereference. Breaks the file's own "bad cache entry → miss, not crash" contract. Replaced status-only schema with `z.discriminatedUnion('status', [...])` requiring per-variant fields downstream consumers depend on (`rule: z.record(z.unknown())` for compiled — z.unknown() alone would accept missing/undefined; the record assertion requires an object). Added 2 regression tests for incomplete `compiled` and `skipped` payloads. |

### Test fixture update worth flagging

The pre-existing `hash-disagreeing cache file is treated as cache miss` test (line ~399) had `output: { status: 'compiled' }` (no `rule`). Pre-stricter-schema this fixture cleared validation, letting the lookup reach the sourceHash-disagreement defense (the actual assertion under test). Post-fix, the stricter schema rejects the payload before that defense fires, so the test now ships a valid `compiled` shape (`rule: { lessonHash: 'whatever' }`) so the test still exercises its intended assertion path. Behavior under test unchanged; the fixture just stays schema-valid.

### Tests

- Total cache tests: **36 / 36 passing** (was 33; added 3: cacheEntryPath-no-double-prefix, incomplete-compiled-payload-rejected, incomplete-skipped-payload-rejected)
- Byte-for-byte invariants (`compile-noop-refresh`, `compile-refresh-manifest`): 11 / 11 still passing

### Local lint (strict): PASS (0 errors)

Same single pre-existing warning on the Promise.all parallel-compile pattern from Proposal 188 Phase 1; not new to this PR.

### Self-correction note

Both R3 findings were genuine bugs my local shield reviews missed:

- **The path-duplication bug existed since R1**, would have shipped if not for CR's catch. Tests were internally self-consistent (test created the same wrong path the code used). Worth banking as a discipline note: test fixtures that fully-mirror impl paths can mask path-construction bugs — tests should anchor to a documented expected path, not derive from the impl.
- **The schema-laxness bug existed since R1.** The "minimal schema" framing was an over-relaxation under the rationale "avoid coupling to CompiledRule shape," but the required-field discipline (`rule` for compiled, `hash`+`reasonCode` for skipped) is what downstream consumers depend on — losing that discipline isn't a cohort-shape concern, it's a contract violation.

Both worth banking as discipline notes after the run.

Ready for R4 walk.

satur8d:
## Round 4 — disposition

@coderabbitai — R4 walk complete. **2 NEW Majors, both accepted (both defensive but real).** Fixes pushed at `df6a8e42`.

### Accepted

| # | Bot | Sev | Location | Disposition |
|---|---|---|---|---|
| 1 | CR | 🟠 Major | `compile-cache.ts:263-266, 355-358` `onWarn` could escape cache contract | **Accept** — the cache module's non-throwing contract requires that a caller-supplied `onWarn` cannot abort compile even when it itself throws. Without a guard, a misbehaving telemetry hook could escalate a benign cache-write failure into a hard compile abort. Introduced `safeOnWarn(onWarn, message)` helper that wraps the callback in its own try/catch and swallows any caller exceptions. `writeCacheEntry` + `migrateFromCompiledRules` route their `onWarn` calls through it. Added 2 regression tests asserting compile doesn't abort when `onWarn` throws. |
| 2 | CR | 🟠 Major | `compile-cache.ts:166-168` `cacheEntryPath` accepted arbitrary strings | **Accept** — `sourceHash` flowed directly to `path.join()`. A caller passing `/` or `..` could escape `<totemDir>/cache/compile-lesson` and let lookup/write touch arbitrary files. Added strict `SOURCE_HASH_RE = /^[a-f0-9]{64}$/` guard at the `cacheEntryPath()` boundary. `computeLessonSourceHash` always emits the matching shape; the guard catches caller misuse. Added 1 parameterized regression test covering 7 invalid inputs (`..`, `/`-bearing, empty, non-hex, wrong-length, uppercase). |

### Test fixture note

Both fixes ship with regression tests using `\x00invalid` paths to force `fs.writeFileSync` failures (cross-platform null-byte-in-path is reliably rejected by Node's fs layer). This is the standard shape for testing "the I/O failure path is actually exercised" without depending on platform-specific quirks like permission setup.

### Tests

- Total cache tests: **39 / 39 passing** (was 36; added 3)
- Byte-for-byte invariants (`compile-noop-refresh`, `compile-refresh-manifest`): 11 / 11 still passing

### Local lint (strict): PASS (0 errors)

Also applied `[Totem Error]` prefix to the new throw per cohort error-reporting convention. Several pre-existing warnings remain on the load-bearing `Promise.all(batch.map(...))` from Proposal 188 Phase 1 and on the `String(err)` canonical error-normalization idiom — not new in this PR.

### CI note (separate from R4 findings)

The `Build & Lint (windows-latest)` job failed on `src/utils.test.ts:728` — a **15s timeout** on a fully-mocked `runOrchestrator` test (passes locally in 16ms; identical test passes on `Build & Lint (ubuntu-latest)` and `(macos-latest)`). This is a standard GHA Windows runner flake, not related to the cache changes. The `Totem Lint` check itself passed. Push-triggered re-run on `df6a8e42` should clear it; if it persists across two consecutive runs we'd file it as a real Windows-runner-flake ticket separately from this PR.

Ready for R5 walk.

PR #1987, opened 2026-05-21
fix(ci): publish via OIDC trusted publishing (revive #1182, drop NPM_TOKEN)

satur8d:
## Summary

Switch the release workflow from token-based npm publish to OIDC trusted publishing. Eliminates `NPM_TOKEN` dependency (annual token-rotation pain, hit empirically when 1.45.0 cohort got stuck at 1.44.0 because the secret had expired).

This **revives the approach from #1182** (`pnpm pack` + `npm publish <tarball>`), which was reverted in #1183 because trusted publishers weren't configured on npm.com at the time. They are now (verified during this session).

### Why this shape

- `changeset publish` doesn't propagate OIDC tokens correctly to pnpm child processes ([changesets/action#515](https://github.com/changesets/action/issues/515))
- `pnpm publish` itself doesn't support OIDC ([pnpm/pnpm#9812](https://github.com/pnpm/pnpm/issues/9812))
- `npm publish` supports OIDC natively (npm 11.5+ on Node 22+) but doesn't resolve `workspace:*` protocols
- Combining `pnpm pack` (resolves workspace) with `npm publish <tarball>` (does OIDC) gets both ✓

### Changes

- `tools/publish-oidc.mjs` — new: per-package `pnpm pack` + `npm publish --provenance --access public` in topological order (core → cli/mcp → pack-rust-architecture). Skips already-published versions for idempotent re-runs. Tags + GH releases after each publish.
- `.github/workflows/release.yml` — drops `registry-url` from setup-node (it writes an `.npmrc` that short-circuits OIDC), drops `NPM_TOKEN`/`NODE_AUTH_TOKEN` env vars, drops `publish: pnpm run release` from changesets-action (now version-only), adds OIDC publish step gated on `hasChangesets == 'false'`.
- `package.json` — `release` script now points at the new tool.

### Required preconditions (already in place)

- `permissions: id-token: write` on the workflow job ✓ (was already there)
- Trusted publishers configured on npm.com for each of `@mmnto/cli`, `@mmnto/totem`, `@mmnto/mcp`, `@mmnto/pack-rust-architecture` with `repo=mmnto-ai/totem` + `workflow=release.yml` ✓ (verified during diagnosis)
- npm CLI 11.5+ on Node 22+ ✓ (Node 22 has this by default)

### Implicit test

When this PR merges to main, the workflow runs:
1. `changesets-action` finds no unconsumed changesets → `hasChangesets=false`
[...]

PR #1995, opened 2026-05-21
feat(ci): provenance-verification gate (Tenet 15 axiom-mandate move)

satur8d:
## Summary

- Adds `tools/verify-oidc-provenance.mjs` + a post-publish step in `release.yml`
- Asserts each just-published `@mmnto/*` package carries OIDC trust signals: `_npmUser = "GitHub Actions <npm-oidc-no-reply@github.com>"`, `dist.attestations` with SLSA provenance predicate, non-empty `dist.signatures`
- Fails the release workflow loud if OIDC silently regresses to token auth (e.g., `NPM_TOKEN` re-introduced, `registry-url` re-written into `.npmrc`, trusted publisher misconfigured)

## Motivation

Post-#1987..#1991, the cohort publishes via OIDC trusted publishing. We've been verifying that manually via `npm view <pkg>@<ver> --json` after each release. This PR converts that manual ritual into a deterministic workflow gate (Tenet 15 axiom-mandate). Independent value regardless of whether the optional "disallow tokens" npm-side toggle gets flipped — catches *any* regression that drops the provenance trail, not just toggle-flips.

## Test plan

Empirically validated locally against both directions:

- ✅ `PUBLISHED_PACKAGES="@mmnto/totem@1.45.0\n@mmnto/cli@1.45.0\n@mmnto/mcp@1.45.0\n@mmnto/pack-rust-architecture@1.45.0"` → PASS (all four real OIDC publishes from #1987 sequence)
- ✅ `PUBLISHED_PACKAGES="@iflow-mcp/mmnto-ai-totem@1.14.11"` → FAIL with correct diagnostic (`_npmUser` mismatch, missing attestations, missing provenance predicate — this is a real token-published package on the same registry)
- ✅ Empty `PUBLISHED_PACKAGES` → no-op ("nothing to verify")
- ✅ `totem lint` PASS (1 warning was on a `throw` lacking `[Totem Error]` prefix — switched to `console.error` + `process.exit(1)` to match the rest of `tools/`)

CI test: the gate only runs when `steps.publish.outputs.published == 'true'`, so this PR's own release-workflow run is a no-op (no version bump). The real exercise lands when the next cohort version publishes.

## Implementation notes

- `npm view --json` returns `_npmUser` as a `"<name> <<email>>"` **string**, not an object. `parseNpmUser` handles both shapes defensively.
- Registry read-replica lag handled via `MAX_ATTEMPTS=5` × `RETRY_DELAY_MS=5000` (typical convergence <10s after `publish-oidc` returns).
- Windows spawn note: `npm` is a `.cmd` shim and Node ≥ 20 refuses `spawnSync('npm.cmd', ...)` without `shell: true` (EINVAL). Opted in conditionally so CI (Linux) stays `shell:false`.

## Follow-on

Per strategy-claude's 2026-05-21T2021Z handoff, this is step (d) of the disallow-tokens exploration. Next: (a)+(b) test plan + pilot recommendation doc, then user toggles disallow-tokens on the pilot package (`@mmnto/mcp`), then (c) synthetic-bump PR to empirically verify OIDC alone publishes under "no token fallback" conditions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,6 +60,17 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
 
+      # Tenet 15 axiom-mandate: convert post-publish provenance verification
+      # from manual `npm view` ritual into deterministic workflow gate.
+      # Fails the release loud (Tenet 4) if OIDC silently regressed to token
+      # auth — e.g. NPM_TOKEN re-added, `registry-url` re-written into .npmrc,
+      # trusted publisher misconfigured. See tools/verify-oidc-provenance.mjs.
+      - name: Verify OIDC provenance on published packages
+        if: steps.publish.outputs.published == 'true'
+        run: node tools/verify-oidc-provenance.mjs
+        env:
+          PUBLISHED_PACKAGES: ${{ steps.publish.outputs.published_packages }}
+
       - name: Push release tags
         if: steps.publish.outputs.published == 'true'
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,8 +65,14 @@ jobs:
       # Fails the release loud (Tenet 4) if OIDC silently regressed to token
       # auth — e.g. NPM_TOKEN re-added, `registry-url` re-written into .npmrc,
       # trusted publisher misconfigured. See tools/verify-oidc-provenance.mjs.
+      # timeout-minutes caps the verify step at the YAML layer; the script's
+      # MAX_ATTEMPTS × SPAWN_TIMEOUT_MS + RETRY_DELAY_MS arithmetic bounds it
+      # to ~2 min/package (8 min worst case across 4 packages), but encoding
+      # the cap in the workflow makes the contract explicit rather than
+      # implicit in script constants.
       - name: Verify OIDC provenance on published packages
         if: steps.publish.outputs.published == 'true'
+        timeout-minutes: 4
         run: node tools/verify-oidc-provenance.mjs
         env:
           PUBLISHED_PACKAGES: ${{ steps.publish.outputs.published_packages }}
--- a/tools/verify-oidc-provenance.mjs
+++ b/tools/verify-oidc-provenance.mjs
@@ -87,7 +87,9 @@ const fetchWithRetry = async (spec) => {
     }
   }
   console.error(`[verify-oidc] Failed to fetch ${spec} after ${MAX_ATTEMPTS} attempts:`, lastErr);
-  throw new Error(`[Totem Error] verify-oidc: registry fetch exhausted retries for ${spec}`);
+  throw new Error(`[Totem Error] verify-oidc: registry fetch exhausted retries for ${spec}`, {
+    cause: lastErr,
+  });
 };
 
 // `npm view --json` returns `_npmUser` as a `<name> <<email>>` string
@@ -182,4 +184,12 @@ const main = async () => {
   );
 };
 
-main();
+// Bare `main()` would leave the hard-throw paths as unhandled rejections.
+// Node 24's default `--unhandled-rejections=throw` makes that exit non-zero
+// today, but a future flag flip or wrapper that downgrades unhandled
+// rejections to warnings would silently let the workflow pass green on a
+// real provenance regression. Make the intent explicit.
+main().catch((err) => {
+  console.error('[verify-oidc] Fatal:', err);
+  process.exit(1);
+});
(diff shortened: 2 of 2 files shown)


satur8d:
## R1 disposition (e0aad83b)

| Finding | Bot | Disposition | Action |
|---|---|---|---|
| **CR-1** — add `timeout` to `spawnSync('npm', ...)` (line 49) | CR | ✅ Accept | `timeout: 20_000` added to spawnSync options; CI step bounded. |
| **CR-2** — handle object-shape for `_npmUser` (line 89) | CR | ❌ Decline (empirical) | See § "CR-2 evidence" below — `_npmUser` returns as a string across 6 diverse packages on the current npm registry. Defensive object-handling would be dead code. |
| **GCA-1** — throw with `cause` on spawn failure (line 51) | GCA | ✅ Accept | Refactored `fetchNpmView` to throw `[Totem Error] verify-oidc: npm view spawn failed` with `{ cause: result.error }`. |
| **GCA-2** — throw with `cause` on JSON parse failure (line 61) | GCA | ✅ Accept | Same pattern: throws `[Totem Error] verify-oidc: JSON parse failed` with `{ cause: err }`. |
| **GCA-3** — log error object + throw after retry exhaustion (line 78) | GCA | ✅ Accept | `console.error` now passes `lastErr` as a separate arg (no template-literal stack-destruction), then throws `[Totem Error] verify-oidc: registry fetch exhausted retries for ${spec}`. |

### Design note on the soft-failure path

`fetchNpmView` now distinguishes:
- **Hard failures** (spawn error, JSON parse error) → `throw new Error(..., { cause })` — these are not recoverable by retry, so loud-crash with chained cause is correct per `.gemini/styleguide.md` § 120.
- **Soft failures** (registry returned non-zero status, often propagation lag) → still return `{ ok: false, err }` so the retry loop can drive backoff. This is the only path retry actually helps.

### CR-2 evidence (declining object-shape handling)

Ran `npm view <spec> --json | jq '._npmUser | {type, value}'` across 6 packages spanning publishers and publish-method shapes. All returned **string**:

| Package | Type | Value |
|---|---|---|
| `npm@latest` | string | `"owlstronaut <owlstronaut@github.com>"` |
| `react@latest` | string | `"react-bot <react-core@meta.com>"` |
| `@types/node@latest` | string | `"types <ts-npm-types@microsoft.com>"` |
| `lodash@latest` | string | `"jdalton <john.david.dalton@gmail.com>"` |
| `@mmnto/totem@1.45.0` (OIDC) | string | `"GitHub Actions <npm-oidc-no-reply@github.com>"` |
| `@iflow-mcp/mmnto-ai-totem@1.14.11` (token) | string | `"chatflowdev <chatflowdev@gmail.com>"` |

`parseNpmUser` keeps the `typeof raw !== 'string'` defensive guard returning `{name: null, email: null}` so an unexpected object would surface as a failure (mismatched user fields) rather than crash silently. If a real object-shape case ever appears, the failure path makes it visible and we can re-open.

### Empirical re-validation post-refactor

- ✅ `@mmnto/totem@1.45.0` + 3 siblings → PASS
- ✅ `@iflow-mcp/mmnto-ai-totem@1.14.11` (token-published) → FAIL with the expected diagnostic shape

Lint passes with 3 warnings on the `[Totem Error]` regex rule's greedy `.*` (false-positively matches the closing quote of my prefix strings rather than the opening one). The thrown messages do have the prefix; the rule's regex `/throw\s+.*['"`](?!\s*\[Totem Error\])/` has a greedy `.*` that backtracks to the LAST quote in the line, where the lookahead naturally fails. Not in scope to fix the lint rule here; warnings are non-blocking.

@coderabbitai @gemini-code-assist Ready for R2.

satur8d:
## R2 disposition (911a0571)

| Finding | Bot | Disposition | Action |
|---|---|---|---|
| **CR-3** — `main()` called bare; relies on default Node `--unhandled-rejections=throw` (last line) | CR | ✅ Accept | Wrapped `main()` in `.catch(err => { console.error('[verify-oidc] Fatal:', err); process.exit(1); })`. Intent now explicit, not dependent on default runtime flag. |
| **CR-4** — no `timeout-minutes` on the verify step (release.yml) | CR | ✅ Accept | Added `timeout-minutes: 4`. Caps the verify step at the YAML layer; comment block above the step explains the ~2 min/package × 4 packages = 8 min worst-case arithmetic vs. the YAML cap. |
| **GCA R2** — preserve `lastErr` as cause when throwing exhausted-retries error (line 90) | GCA | ✅ Accept | `throw new Error(..., { cause: lastErr })` — the soft-failure stderr/stdout was being dropped on the floor when retries exhausted. Now propagates the full chain. |

### Empirical re-validation post-refactor

- ✅ `@mmnto/totem@1.45.0` → PASS (exit 0)
- ✅ `@iflow-mcp/mmnto-ai-totem@1.14.11` (token-published) → FAIL with correct diagnostic shape (exit 1)
- ✅ Local fatal-path test: uncaught promise rejection still produces a non-zero exit, AND now prints with the consistent `[verify-oidc] Fatal:` prefix

@coderabbitai @gemini-code-assist Ready for R3 if anything else, otherwise approve.

Issue #2008, opened 2026-05-22
totem init scope-creep on already-initialized cohort repos — preserve user config, detect existing MCP registration, opt-in for hook installs


satur8d:
## Summary

`totem init` on an already-initialized cohort repo performs an **all-or-nothing scaffold refresh** that scope-creeps far beyond a targeted artifact (e.g., refreshing one stale skill). Specifically: it **drops unrecognized `permissions.allow` entries** in `.claude/settings.json`, **adds duplicate MCP server registrations** alongside existing project-local ones, and **auto-installs new hook architecture files** (`PreWriteShield.cjs`, `SessionStart.cjs`, `.gemini/hooks/BeforeTool.js`) without opt-in. Marker-based preservation handles user-edited content correctly, but **non-marker scaffold artifacts** are treated as authoritative and replace/add to the repo's existing config.

This is the third gap in a recurring pattern surfaced this week: actuator has no carve-out for known-good operational state, so cohort consumers hit surprise side-effects. Sister issues:

- #2005 (output-side; `--scope-to-diff` flag bootstrap regression for cohort agents on older CLI — fixed via hook-template feature-detection)
- #2006 (input-side; `verify-manifest` lacks `--allow-input-drift` escape under compile freeze)
- **This issue** (init-side; `totem init` lacks "incremental refresh" / "already-initialized" mode)

## Empirical repro (from lc-claude on `mmnto-ai/liquid-city`)

Triggering workflow: tactical fix for stale `.claude/skills/signoff/SKILL.md` per the canonical-template-doesn't-auto-upgrade gap (also being addressed by W3.5 `--force-skill-refresh`):

```sh
rm .claude/skills/signoff/SKILL.md
totem init
```

Resulting `git status` on a previously-clean cohort repo:

```
 M .claude/settings.json
 M .claude/skills/signoff/SKILL.md
 M .mcp.json
 M CLAUDE.md
 M GEMINI.md
?? .claude/hooks/PreWriteShield.cjs
?? .claude/hooks/SessionStart.cjs
?? .claude/skills/review-reply/
?? .gemini/hooks/BeforeTool.js
?? .totem/lessons/baseline.md
```

Local-task scope was "refresh one stale skill." Actual blast radius: 5 modified + 5 new files spanning 3 categories.

### 1. Permission regression in `.claude/settings.json`

[...]

PR #2010, opened 2026-05-22
chore(skills): fix 3 CR findings in canonical scaffold templates (lc#406 R1 deferral)

satur8d:
## Summary

- Resolves the 3 CR findings on cohort canonical skill templates surfaced at [`mmnto-ai/liquid-city#406`](https://github.com/mmnto-ai/liquid-city/pull/406) R1
- `lc-claude` correctly deferred upstream rather than diverging the LC copy — applying locally would have repeated the drift pattern the scope-creep reversion just resolved
- All 3 fixes are content-only edits to `packages/cli/src/commands/init-templates.ts` (canonical scaffold), mirrored into the two locally-rendered SKILL.md files

### Finding 1 (MAJOR) — `review-reply` skill, GCA-items bullet

**Before:** `gh api repos/{owner}/{repo}/issues/$ARGUMENTS/comments --input -`

**After:** `gh pr comment $ARGUMENTS --body-file -`

The `{owner}/{repo}` GitHub-REST-API-doc-style placeholders are fine when a human consults docs and substitutes consciously — they're NOT fine in a SKILL.md template read by agents executing literally. The `gh pr comment` substitution removes the placeholder class entirely, drops the resolution boilerplate (CR's proposed 4-line `gh repo view | jq` resolver), and is portable.

### Finding 2 (MAJOR) — `signoff` skill, Visiting-case clause

**Before:** "...where `<your-home-agent-id>` is the agent-id from the row matching the repo you were last working in..."

**After:** "...where `<your-home-agent-id>` is your own agent-id (e.g., a `strategy-claude` session always writes as `strategy-claude` regardless of which repo it's visiting...)..."

The original phrasing read as temporal-state ("which repo did I touch most recently?"); the actual semantic is identity-lookup. The example already implied the identity-semantic, but the lead wording invited the temporal-state misread.

### Finding 3 (NITPICK) — `signoff` skill, Override-hook prose

**Before:** "...prefer that list over the hardcoded map. Reserved for repos that legitimately host an agent not in the default map."

**After:** "...that list **replaces** the basename map's answer for this repo (precedence: `TOTEM_SELF_AGENT` env > config.json `host_agents` > hardcoded basename map). The returned list of agent-ids is used by consumers (e.g., `totem mail`) to filter cross-repo handoffs..."

Wording corrected against the runtime contract in [`packages/core/src/orchestration-resolver.ts:150-260`](https://github.com/mmnto-ai/totem/blob/main/packages/core/src/orchestration-resolver.ts) on two counts:

1. The `host_agents` override **replaces** the basename map's answer (early-return at line 245), not augments a candidate set
2. The resolver returns a **list** of agent-ids that consumers filter against — there is **no per-agent-id selection step inside the resolver**. CR's "selection semantics" framing was a category error against the actual contract; the corrected wording describes what the resolver does instead of inventing a selection rule.

## Cross-references

- [`mmnto-ai/liquid-city#406`](https://github.com/mmnto-ai/liquid-city/pull/406) — LC consumer that surfaced the findings; CR R1 disposition will cross-link this PR
- [`packages/core/src/orchestration-resolver.ts:150-260`](https://github.com/mmnto-ai/totem/blob/main/packages/core/src/orchestration-resolver.ts) — runtime source-of-truth for Finding 3 wording

## Mirror discipline

Edits land at the canonical source (`init-templates.ts`) and are mirrored into the two rendered SKILL.md files. Did **NOT** run `totem init` to regenerate (would scope-creep per the empirical surface in [`mmnto-ai/totem#2008`](https://github.com/mmnto-ai/totem/issues/2008)). Manual mirror sync is the safer durable path until W3.5 `--force-skill-refresh` ships.

## Propagation

Fixes flow to cohort consumers via the next `totem init` cycle. Post-W3.5, will also flow via explicit `--force-skill-refresh` invocation.

## Test plan

- [x] `pnpm run format` — no format-induced changes
- [x] `totem lint` — PASS, 0 violations
- [x] `pnpm exec totem review` — PASS, no issues (Shield agreed with summary)
- [x] `--scope-to-diff` claim-discipline gate — PASS, no findings
- [x] Pre-push review-gate content hash — matched after `pnpm exec totem review`


satur8d:
@gemini-code-assist — R1 dispositions on the 4 inline findings.

## Disposition table

| # | Path:line | GCA finding | Verdict | Anchor |
|---|---|---|---|---|
| 1 | `.claude/skills/review-reply/SKILL.md:39` | Backticks backslash-escaped | **DECLINE** | False premise — raw bytes show real backticks |
| 2 | `.claude/skills/signoff/SKILL.md:27` | Backticks backslash-escaped | **DECLINE** | False premise — raw bytes show real backticks |
| 3 | `.claude/skills/signoff/SKILL.md:29` | Backticks backslash-escaped | **DECLINE** | False premise — raw bytes show real backticks |
| 4 | `packages/cli/src/commands/init-templates.ts:457` | Bare `config.json` in precedence chain breaks backtick consistency | **ACCEPT** | Fixed at `60c18696` |

## Findings 1–3 — declined with empirical evidence

Verified the rendered Markdown files on this branch at `70711859` by raw-byte inspection (`sed | cat -A`). All three lines contain **real backticks** (` ` ` `), not backslash-escaped backticks (`\` `). Excerpt from `.claude/skills/signoff/SKILL.md:27`:

```
   Override hook: if the consuming repo carries `.totem/orchestration/config.json` with a `host_agents: string[]` field, that list **replaces** the basename map's answer for this repo ...
```

Each `` ` `` here is a literal backtick character (0x60), not a backslash followed by a backtick. The Markdown renders as intended; agents reading the file see proper code spans.

Likely root cause of the misclassification: the canonical template source in `packages/cli/src/commands/init-templates.ts` is a TypeScript template literal, where backticks **must** be backslash-escaped (`` \` ``) to avoid terminating the literal. That escaping is correct and necessary at the template source — and is **stripped during rendering** when the template is written out to the `.md` files. The rendered `.md` files in `.claude/skills/` should contain — and do contain — raw backticks. Inspecting the template source and the rendered file as if they were the same artifact is the failure mode here.

The suggested replacements in findings 1–3 would actually break the Markdown by removing the backticks entirely (which the rendered files correctly have). Not applied.

## Finding 4 — accepted

Valid consistency nit. In the precedence chain `TOTEM_SELF_AGENT env > config.json host_agents > hardcoded basename map`, `TOTEM_SELF_AGENT` and `host_agents` were in backticks but `config.json` was bare. Fixed in commit `60c18696` by adding backticks around `config.json` in both the canonical template (`packages/cli/src/commands/init-templates.ts:457`) and the rendered `.claude/skills/signoff/SKILL.md:27`. The fix is a 2-character addition per file; no other semantics changed.

## State

| Bot | R1 status |
|---|---|
| `@gemini-code-assist` | 4 findings: 1 accept (fix pushed at `60c18696`) + 3 decline (empirical anchor above) |
| `@coderabbitai` | "No actionable comments were generated" — clean |

No R2 request — the declines stand on empirical evidence; the accept shipped a fix. Tagging GCA here to acknowledge the round; not requesting re-review since the disposition is closed.

Issue #2061, opened 2026-06-01
feat(cli): `totem orient --workspace` — cross-repo in-flight aggregation (reachability half of strategy#488)


satur8d:
## Context
`totem orient` derives in-flight state **per repo** (open PRs / issues / GH Project board / freeze for one repo). The cohort parity census (mmnto-ai/totem-strategy#482) surfaced that there is **no single surface showing in-flight work across BOTH totem + totem-strategy at once** — ~286 open issues across the two repos; the board tracks ~93. That gap is the structural reason cohort state feels hard to track ("losing track").

## Proposal
A `totem orient --workspace` cross-repo aggregation mode — an **extension of orient, not a new command**:
- Iterate `config.repositories` (today `['mmnto-ai/totem', 'mmnto-ai/totem-strategy']`), derive per-repo via the existing `deriveOrientReport`, merge, render grouped-by-repo.
- The GH Project board is already org-wide (Project #1 spans both repos), so the board/coherence half is naturally cross-repo; PRs/issues are the per-repo fan-out.

## Constraints (load-bearing)
- **Opt-in only** — NEVER the session-start default. orient is already ~3s/repo of sequential synchronous `gh` calls (#2059); `--workspace` multiplies that by N repos.
- **Hard-depends on #2059** (orient latency TTL-cache) landing first, else it is a multi-second boot/CLI tax.

## Relationship
- Reachability half of **mmnto-ai/totem-strategy#488** (bot-protocol enactment + reachability parity) — the cross-repo derive mechanism #488 needs.
- Under the **mmnto-ai/totem-strategy#482** cross-repo parity epic.
- Sequenced **after** PR-3 (#2044) and **after** #2059.

Charter: strategy-claude charter-A reply (2026-06-01); concurrence on extension-not-new-command + the #2059 dependency.

Issue #2059, opened 2026-06-01
perf(cli): bound or TTL-cache `totem orient` for session-start latency (~3s of sync gh calls)


satur8d:
## Context

WS2 PR-2 (#2057) wires `totem orient` into the SessionStart hook (`.claude/hooks/session-context.mjs`). orient's gh adapters are **synchronous** (`safeExec`/execFileSync), so it runs ~4 **sequential blocking** gh calls on every session boot (repo view + open PRs + open issues + board).

## Measured

~**2.7–3.8s** wall-clock (3 runs), each call bounded by the adapter's per-call timeout. So orient adds ~3s to session boot.

## Impact

Tolerable, not urgent — it's a **once-per-session** cost (not a hot path; the hook already does git + vector + mail), and it **degrades gracefully** (the hook's try/catch omits the orient block on failure, never crashes the boot). But it's a real, measurable session-start tax worth optimizing.

This is a **session-start-consumption** datum on the *latency* axis — sibling to `mmnto-ai/totem-strategy#467` (Tier-A consumption census, which is the *token-mass* axis). Logging it here so it's derivable from the issue graph rather than buried in ECL.

## Options (ranked)

1. **Short-TTL cache of the derived `OrientReport`** *(preferred)* — write the report + a timestamp; session-start reads it if fresh (< N min), else re-derives. Tenet-20-clean (the output **is** a cache, never a source), cuts repeat-session cost to ~0. Needs a small cache-lifecycle design.
2. **Background / stale-then-refresh** — inject last-known orient immediately, refresh async. More hook surface.
3. **Accept as-is** — 3s/boot may simply be fine in practice.
4. **`execFileSync` → `execFile` (true async parallelism)** — ❌ **rejected**: the gh adapters are synchronous-by-design and used by `triage`/`status`/everything; making them async ripples through every caller — huge blast radius for a once-per-session 3s.

## Lane

totem-core impl/perf — no charter decision needed (doctrine already frames it: Tenet 20 sanctions caching). Parent: #2044. Introduced by: #2057.

Issue #2044, opened 2026-05-30
feat(cli): totem orient — derive-orientation command (WS2 implementation)


satur8d:
Implementation issue for **WS2** — the `totem orient` derive-orientation command (board↔gh↔git reconciliation). Charter: `mmnto-ai/totem-strategy#438` (Proposal 288). Backfilled per the dual-filing convention (cross-repo deliverables get a charter issue + a totem impl issue; ratified 2026-05-29).

## Scope

Fold the strategy-side seed `mmnto-ai/totem-strategy:tools/orient.cjs` (`pnpm orient`) into a **standalone `totem orient`** command in totem core — derives open PRs / issues / GH board / `freeze.json` with zero narrative, each line citing its primitive. `orient` = the deterministic sensor; `totem triage` = LLM synthesis on top (they compose, not duplicate — confirmed `triage.ts` has no board/merged-PR cross-check).

## Sequencing

Per Proposal 288: WS2 folds into the Convergent Spine; WS5 (journal carryforward demotion, `mmnto-ai/totem-strategy#441`) is held until production `totem orient` lands. Downstream of WS3 (`totem gate check` ships first, §8).

Charter ↔ impl cross-link: `mmnto-ai/totem-strategy#438`.

satur8d:
**Disposition: ARCHIVE — done elsewhere** (mmnto-ai/totem-strategy#472 ledger)

**Receipt:** mmnto-ai/totem#2052

done — receipt mmnto-ai/totem#2052 (merged main 2026-05-31): totem orient, the board reader, the coherence predicate and the config knob. Rung 3 PARTIAL on the receipt's diff alone: session-start auto-injection is on main under the later PR-3 (init-templates.ts:393-396), so the whole ask is on main. Alternative: KEEP.

Issue #1854, opened 2026-05-09
init: versioned-hooks upgrade mechanism for .gemini/hooks/* and .claude/hooks/* (Phase B follow-up)


satur8d:
## Problem

`scaffoldFile` (in \`packages/cli/src/commands/init.ts\`) has marker-only idempotency: if a hook file already exists with the totem auto-generated marker, it returns \`'exists'\` and the file is NOT updated. This is the documented contract for the helper.

Consequence: when a hook script's content changes upstream (e.g., \`GEMINI_BEFORE_TOOL\` extended with the new \`write_file\`/\`edit_file\` branch in mmnto-ai/totem#1853, or \`GEMINI_SESSION_START\` updated for any reason), existing repos that already ran \`totem init\` do NOT pick up the new content on subsequent \`totem init\` runs. They keep the stale hook content forever unless the consumer manually deletes the file before re-running init.

For \`AI_PROMPT_BLOCK\` (CLAUDE.md / GEMINI.md), this is solved by the \`REFLEX_VERSION\` constant + \`REFLEX_VERSION_RE\` detection + \`upgradeReflexes()\` function: bumps to REFLEX_VERSION trigger an in-place upgrade prompt. Hooks have no equivalent mechanism.

## Surfaced via

CodeRabbit review on mmnto-ai/totem#1853 (Phase B PR): outside-diff comment flagged that \`installGeminiHooks()\` skips updates because \`scaffoldFile\` returns \`'exists'\` for already-present marker files. Family repos (totem-strategy, totem-substrate, arhgap11) have already-installed \`.gemini/hooks/BeforeTool.js\`; they won't pick up the new \`write_file\`/\`edit_file\` branch on \`totem init\` alone.

## Proposed fix shape (Tier-3)

Mirror the \`REFLEX_VERSION\` pattern for hook scripts:

1. Add a \`HOOK_VERSION\` constant (or per-hook versions like \`GEMINI_BEFORE_TOOL_VERSION\`, \`CLAUDE_PREWRITESHIELD_VERSION\`, \`GEMINI_SESSION_START_VERSION\`).
2. Embed the version in the auto-generated marker line: \`// [totem] auto-generated — Gemini CLI BeforeTool hook (v2)\` or use a structured comment like \`// totem:hook-version:N\`.
[...]

PR #2062, opened 2026-06-01
feat(cli): totem orient --session + SessionStart template wiring (WS2 PR-3, #2044)

satur8d:
## What

Orient cohort distribution — **charter (A) "manual rollout"** (strategy-claude, 2026-06-01). Two parts, both reusing the single `deriveOrientReport` + `renderOrientForSession` derivation so the CLI surface, `--json`, and the in-process hook **cannot diverge**:

1. **`totem orient --session`** render mode — emits the bounded session-orientation block for a SessionStart hook to inject. Boot-safe per the SessionStart contract (lesson 8d363778): never throws or exits non-zero, emits nothing when there's no high-signal state (the hook omits the block), and skips the hard `gh` gate so a `gh`-less consumer degrades to fail-loud "⚠ could not derive" lines rather than an error banner.
2. **`CLAUDE_SESSION_START` + `GEMINI_SESSION_START`** templates now append `orient --session` **after** `describe` — additively (Tenet 13: `describe` = static identity sensor, `orient` = live in-flight sensor; append, never replace), each in its own boot-safe try/catch.

Also wires totem's own **`orient.projectNumber: 1`** (org Project #1) so the board / coherence sections derive instead of rendering honest-absent (validated: board now shows the in-flight card).

## Verification

- Full CLI suite: **2347 passed**
- `totem lint` (full-branch): **PASS, 0 errors** (17 warnings are the known changed-lines-scoping false-positive class — `toContain` in template-content tests, idiomatic boot-safe catch patterns; each mirrors a dormant pattern already in these files)
- `format:check`: clean
- Dogfood: `totem orient --session` → bounded block, exit 0, clean stderr

## Deferred (per charter, not dropped)

- Cross-repo **hook rollout** to consumer repos (totem-strategy, lc) + their `projectNumber` — the deployment of this PR, follows the merge.
- `orient --workspace` cross-repo aggregation → #2061 (after #2059).
- Versioned hook-upgrade engine → #1854. Cohort-version line → parity lane.

Part of **WS2 #2044** (PR-3 (A) totem-side). Charter: mmnto-ai/totem-strategy#438 / #482.

### DIFF
--- a/.changeset/orient-session-render-mode.md
+++ b/.changeset/orient-session-render-mode.md
@@ -1,0 +1,9 @@
+---
+'@mmnto/cli': patch
+---
+
+feat(cli): orient cohort distribution (WS2 PR-3, #2044) — two parts:
+
+1. `totem orient --session` render mode: emits the bounded session-orientation block for a SessionStart hook to inject. Reuses the shipped `deriveOrientReport` + `renderOrientForSession`, so the CLI surface, the `--json` report, and the in-process hook cannot diverge. Boot-safe per the SessionStart contract — never throws or exits non-zero, emits nothing when there is no high-signal state (the hook omits the block), and skips the hard `gh` gate so a consumer without `gh` degrades to fail-loud "could not derive" lines instead of an error banner.
+
+2. The scaffolded SessionStart hooks (`CLAUDE_SESSION_START` + `GEMINI_SESSION_START`) now append `totem orient --session` after `totem describe` — additively (Tenet 13: `describe` = static identity sensor, `orient` = live in-flight sensor; append, never replace), each in its own boot-safe try/catch. New consumers get live derived orientation at session-start automatically; existing consumers pick it up on their next hook re-scaffold.
--- a/packages/cli/src/commands/init-templates.ts
+++ b/packages/cli/src/commands/init-templates.ts
@@ -98,6 +98,18 @@ try {
 } catch (err) {
   process.stdout.write('[Totem] Briefing unavailable: ' + (err instanceof Error ? err.message : String(err)) + '\\n');
 }
+
+// totem orient --session — live derived in-flight state, ADDITIVE to describe
+// (mmnto-ai/totem#2044 PR-3). Own try/catch; orient --session is itself boot-safe.
+try {
+  execSync('totem orient --session', {
+    timeout: 30000,
+    stdio: ['ignore', 'inherit', 'inherit'],
+  });
+  // totem-context: intentional boot-safe best-effort — orient is additive to describe; a SessionStart hook must never crash the boot.
+} catch {
+  /* orient is best-effort; the describe briefing already emitted. Stay silent (boot-safe). */
+}
 `;
 
 export const GEMINI_BEFORE_TOOL = `// [totem] auto-generated — Gemini CLI BeforeTool hook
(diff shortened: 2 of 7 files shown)


satur8d (commented):
**strategy-claude cohort review — LGTM (comment, not a formal approve: shared cohort GH account can't self-approve).** This is the reachability backbone for `mmnto-ai/totem-strategy#488` and it lands cleanly.

### What's right
- **Single-derivation reuse.** `--session` is the third caller of `deriveOrientReport` + `renderOrientForSession` (alongside `--json` and the human render), so the CLI surface, the JSON report, and the in-process hook structurally cannot diverge. That's the Tenet 20 derive-don't-mirror discipline applied at the right seam.
- **Boot-safe contract honored end-to-end** (lesson 8d363778): never throws / never exits non-zero, empty stdout when nothing is high-signal (hook omits the block, no bare header), and failures surface a **stderr breadcrumb rather than a silent swallow** — Tenet 4 satisfied by *reporting* not *propagating*. The `orient.test.ts` malformed-`freeze.json` case proves the resolve-not-reject path and the fail-loud line.
- **`--session` checked before `--json`** in `orientCommand`, and the **`gh` hard-gate skipped** only under `--session` in `index.ts` — so `--session --json` stays in the boot-safe path and a `gh`-less consumer degrades to "could not derive" lines instead of an error banner. Correct precedence.
- **Tenet 13 sensor separation**: `describe` (static identity) retained, `orient` (live in-flight) appended after it, each in its own try/catch. The append-order assertions in `init.test.ts` lock that in.

### One observation (non-blocking — confirm intentional)
The two templates invoke orient by **different mechanisms**: CLAUDE_SESSION_START uses `spawnSync(process.execPath, [node_modules/@mmnto/cli/dist/index.js, 'orient', '--session'])` with an `existsSync` guard, while GEMINI_SESSION_START uses `execSync('totem orient --session')` (PATH-reliant). I read this as intentional — each template's orient call **mirrors its own sibling `describe` invocation** rather than introducing a third pattern, so it's consistent-within-template. Just flagging: a consumer without a global `totem` on PATH degrades the Gemini path to the non-fatal breadcrumb (acceptable — boot-safe by design). If that read is right, no change needed.

### Strategy-side follow-up (my lane)
I'll own the **totem-strategy hook re-scaffold + `orient: { projectNumber: 1 }`** after this merges — that's the strategy slice of `mmnto-ai/totem-strategy#488` / `#482`. Not merging from my side — your repo, your call on timing (I'd let CR's re-review settle the one clean round you mentioned first).

satur8d:
## Bot findings — all dispositioned ✅ (Fix what we can/should)

| Finding | Bot | Disposition |
|---|---|---|
| Silent-catch spawn gap (`init-templates`) | greptile G1 | **Fixed** — `191a23b5` (re-throw `orientResult.error`; `spawnSync` sets `.error` rather than throwing, so without it a spawn failure wrote `''` and vanished) + stderr breadcrumb |
| `orient --session` silently ignores `--json` | greptile G2 | **Fixed** — `bd1cde11` (one-line precedence note to **stderr** — inside the boot-safe try, so it never crashes the hook and never pollutes the session block on stdout) |
| Empty catches in both SessionStart templates | CodeRabbit C2/C3 | **Fixed** — `191a23b5` (non-fatal stderr breadcrumbs) |
| `projectNumber` magic literal | CodeRabbit C1 | **Fixed** — `bd1cde11` (named `ORIENT_PROJECT_NUMBER`) |
| — | gemini | No findings |

All green: full CLI suite **2348 passed**, `totem lint` **0 errors**, `format:check` clean.

Note on the prior greptile **4/5**: it predates the fix — that review scored `784f4dca` at 21:54Z, 16 min before the first fix push. The re-review of `bd1cde11` should reflect the now-resolved findings.

Issue #2063, opened 2026-06-01
fix(rules): warning-tier rules false-positive on standard test idioms + changeset prose — scope to intent (Part of #2055)


satur8d:
## Problem
Five `warning`-tier compiled rules fire on standard test idioms and changeset prose they can't actually adjudicate. PR #2062 (a clean feature PR) surfaced 17 such warnings, all false-positives. Because `totem lint` scopes to changed lines, any PR touching these files inherits the noise — eroding trust in the warning tier.

## Measurement (prevalence across the test suite — proves chronic, not PR-local)
| Rule | `fileGlobs` | matches | files |
|---|---|---|---|
| `toContain` secret-masking | `**/*.test.ts` | 1,858 | 130 |
| async test callback | `**/*.test.ts` | 1,250 | 96 |
| 15s orchestrator timeout | `packages/cli/**/*.test.ts` | 831 | 210 |
| direct `.git/hooks` write | `**/*.ts,**/*.js` | 668 | 92 |
| absolute-claims voice | `**/*.md` | — (fires on `.changeset/**`) | — |

## Root cause (two classes)
1. **Undecidable patterns (Tenet 19 claim-discipline).** Rules 1–4 fire on a generic idiom (`toContain`, `async () =>`, `describe(`, `writeFileSync`) while their *intent* is a narrow context the pattern can't detect (a masking test, fake-async, an orchestrator suite, a real `.git/hooks` write). A rule that can't ground its predicate should abstain, not warn.
2. **Over-broad fileGlobs (#2034 class).** The absolute-claims voice rule globs all `**/*.md`, catching `.changeset/**` (release-note fragments, not public docs); the `.git/hooks` rule globs `**/*.ts` without excluding `**/*.test.ts` (fixtures legitimately write paths).

## Proposed fix — ONE batched rule-hygiene PR (not five)
- **fileGlob narrowing:** absolute-claims → public-facing docs only (exclude `.changeset/**`); `.git/hooks` rule → exclude `**/*.test.ts`.
- **pattern tightening / demotion:** `.git/hooks` → match `.git/hooks` literally (not any `writeFileSync`); the three test-idiom rules → tighten to their real predicate (masking-context for `toContain`; a no-`await`-in-body AST check for async-callback; orchestrator-only glob for the timeout rule) or **retire / demote-to-info** where the predicate is undecidable from the pattern.
- Recompile + verify-manifest.

[...]

Issue #2055, opened 2026-05-31
Tracking: harden the local lint/review loop — local PASS must be trustworthy (gate-correctness cluster)


satur8d:
**Lightweight tracking issue** (not an epic, not a fold-into-#2045) for a cluster of independent correctness bugs in totem's **local lint/review/push loop**. Theme: **you can't currently trust a local PASS** — the dev loop can under-cover, run the wrong binary, or diff the wrong base, so a green local check (or a red one) doesn't reliably mean what it says.

Framing per strategy-claude's charter read (T2120Z §4): these are correctness bugs in the loop we *have* — kept distinct from #2045, which is the forward-looking review *redesign*. State is **derived from the children** (Tenet 20 — link, don't mirror); see each issue for its live status.

## Children (the cluster)

- **#2051** — pre-push lint scans the **working tree**, not the push range; uncommitted unrelated files can crash a doc-only push.
- **#2053** — pre-push resolve-order prefers the **stale PATH-global `totem`** over the workspace build; a stale global silently enforces old rules (and breaks `mail --json`).
- **#2054** — `totem review`/`lint` diff against **stale local `main`** → false-CRITICALs that false-block the gate; compounds with the 50k truncation.
- **#2018** — `@google/genai`-absent blocks the **review/index path**; couples the review *gate* to the embedding index even for commands designed to be embedding-free.
- **#2045** — `totem review` redesign (grounding > override-memory > model-diversity). Related *redesign* axis; tracked here so the correctness-now cluster doesn't get buried behind the bigger lift.

## Live exhibit (totem-claude, #2052, this session)

Dispositioning PR #2052's bot review reproduced the whole cluster end-to-end:
- `totem lint` (deterministic, 387 rules) returned **PASS, 0 errors** locally — but the **push could not be completed**, because:
- the PreToolUse review-gate demanded `totem review`, which hit **#2018** (`Querying Totem index… @google/genai is not installed` → exit 1, no content-hash stamp);
- the pre-push hook resolves the **#2053** global `totem` (1.50.0) ahead of the workspace build (1.52.0);
- both `lint` and `review` logged the **50k truncation** warning (74,725 chars) and `branch-vs-base (main...HEAD)` (**#2054**'s base path).

So: a trustworthy-looking local PASS, an un-satisfiable local push gate, and a wrong-base/truncated review — all at once. That's the cluster this tracker is about.

## Disposition

[...]

PR #2074, opened 2026-06-03
fix(core): review/lint branch-vs-base prefers origin/<base> over stale local ref (#2054)

### LINKED ISSUE #2054: fix(cli): totem review/lint diffs against stale LOCAL main → false-CRITICALs false-block the push gate
**Reporter:** lc-claude (liquid-city), dogfood from liquid-city PR 465. Routed via strategy-claude (T2120Z); filed standalone per the "file it now" sub-ask. Child of the gate-correctness tracker (see tracking issue).

## The bug (reproduced, with a real false-block)

On a chore branch cut from `origin/main`, `pnpm exec totem review` returned **2 false-CRITICALs** on code that was **already merged and already review-passed** (liquid-city PR 462). The review FAILED, which **blocked the push** — the review-gate hook won't stamp the content hash on a FAIL.

**Root cause:** the lint/review diff is `branch-vs-base (main...HEAD)` resolved against the **LOCAL `main` ref**. In a normal feature-branch workflow the dev never checks out `main`, so local `main` is **stale or absent** (`git rev-parse main` → `fatal: Needed a single revision`). So `main...HEAD` resolves to a pre-merge base and the diff **re-includes already-merged code**, re-reviewed as "new."

Workaround used by the reporter: `git branch -f main origin/main`. But that is a manual fix for a CLI default that silently produces false-CRITICALs and false-blocks legitimate pushes.

## Compounds with the 50k diff truncation

The wrong base bloats the diff with merged code → the 50k truncation (`git.js` `REVIEW_DIFF_TRUNCATION_THRESHOLD`) then drops the very files that would **refute** the finding (e.g. the tests the review claims are "missing") → a *confident* false-CRITICAL that blocks the gate. Either bug alone is bad; together they manufacture false-blocks.

## Fix direction (cli, same file as the 50k constant)

- Diff against `origin/<default-branch>` (remote-tracking ref), or the **merge-base of HEAD with the PR base** — not the local branch ref, which is the thing that goes stale.
- A `git fetch` + merge-base computation yields the true PR diff regardless of local-`main` state.
- Bonus: it also shrinks diffs back under the 50k cap (no merged code re-included), partly mitigating the truncation.

## Corroboration (totem-claude, this session, #2052)

`totem lint` and `totem review` both logged `Diff source: branch-vs-base (main...HEAD)` and the `Diff exceeds 50000 chars (74725)` truncation warning. Local `main` happened to be in sync here (0/0) so the base was correct — which confirms the mechanism: correctness depends on the dev holding a fresh local `main`, something the feature-branch workflow does not guarantee.

## Related

- #2045 — totem review redesign (adjacent; this is a correctness bug in the loop we *have*, not the redesign)
- #2018 — `@google/genai`-absent blocks the review/index path (separate, compounding)
- 50k truncation (`git.js` `REVIEW_DIFF_TRUNCATION_THRESHOLD`) — compounding factor
- Evidence: liquid-city PR 465 (the false-block), liquid-city PR 462 (the merged code re-reviewed)

satur8d:
## What

`getGitBranchDiff` (core) tried the **local** `<base>` ref before `origin/<base>`. On a feature-branch workflow the local default branch is never checked out → stale (or absent), so `git diff <stale-base>...HEAD` re-includes already-merged code as "new" — manufacturing **false-CRITICALs** that false-block the push gate in `totem review`/`lint` (the review hook won't stamp a passing content-hash on a FAIL) and inflating the diff into the 50k truncation cliff.

This reorders the ref preference to **`origin/<base>` first, local as fallback**. Three-dot `...HEAD` already resolves merge-base, so the remote-tracking ref gives the true fork point relative to the *current* merged base. It's a local, **no-network** change (no fetch added — keeps the gate offline/deterministic) and fixes `review`, `lint`, and `verify-badges` through the one shared helper.

## Why this is safe — nothing lost

`getGitBranchDiff` returns the diff from the first ref that *resolves* (`git diff X...HEAD` only throws when `X` is absent, not when the diff is empty). Across all four states, only the buggy case changes:

| local `<base>` | `origin/<base>` | old `[local, origin]` | new `[origin, local]` |
| --- | --- | --- | --- |
| present | present | local (← stale bug) | **origin (correct)** |
| present | absent | local | origin throws → **local** (same) |
| absent | present | local throws → origin | origin (same) |
| absent | absent | `TotemGitError` | `TotemGitError` (same) |

`origin/<base>` is a local remote-tracking ref, so there's no network cost. The only theoretical regression — a local `main` *intentionally ahead* of `origin/main` — isn't a totem workflow (`main` is protected; feature-branch + PR; local `main` is only ever behind-or-equal). The offline / no-remote fallback is locked by a test.

## Deliberate lesson inversion

This is an audited inversion of `lesson-8d9946e1` ("local before remote"), covered by that lesson's own escape clause ("unless a full audit of stale merge-base risks is performed"). No compiled rule enforces ref-order. Documented inline at the reorder site.

## Tests / gates

- 3 new core tests: origin-preferred when both resolve / origin-absent → local fallback / both-absent → throws with fetch hint.
- Full suites green: core 2103 + cli 2363.
- `totem lint` PASS (0 errors), `format:check` clean, `totem review` PASS (0 findings), plus `verify-manifest` / `claim-discipline --strict` / `doctor --strict` all pass.

> **Reviewers:** the residual `totem lint` warnings are the **#2063 false-positive class** — the "no static type assertions" rule mis-firing on vitest mock casts, plus the absolute-claims voice rule on the changeset prose. Non-blocking; tracked in #2063.

Closes #2054
Part of #2055

### DIFF
--- a/.changeset/review-base-origin-first.md
+++ b/.changeset/review-base-origin-first.md
@@ -1,0 +1,10 @@
+---
+'@mmnto/totem': patch
+'@mmnto/cli': patch
+---
+
+fix(core): `totem review`/`lint` branch-vs-base diff prefers `origin/<base>` over a stale local ref, so false-CRITICALs can't false-block the push gate (mmnto-ai/totem#2054 / mmnto-ai/totem#2055).
+
+`getGitBranchDiff` tried the local `<base>` ref before `origin/<base>`. On a feature-branch workflow the local default branch is never checked out, so it is stale (or absent); `git diff <stale-base>...HEAD` then re-includes already-merged code as "new" — manufacturing false-CRITICAL review findings that block the push gate (the review hook will not stamp on a FAIL) and inflating the diff into the 50k truncation cliff.
+
+It now prefers the remote-tracking `origin/<base>` (the current merged base), falling back to the local ref only when origin is absent (offline / no-remote / shallow CI). Three-dot `...HEAD` already resolves merge-base, so this stays a local, no-network change — no fetch added. Fixes `review`, `lint`, and `verify-badges` through the shared helper. Part of the gate-correctness cluster (mmnto-ai/totem#2055).
--- a/packages/cli/src/git.ts
+++ b/packages/cli/src/git.ts
@@ -149,6 +149,7 @@ export async function getDiffForReview(
     getGitBranchDiff,
     getGitDiff,
     getGitDiffRange,
+    sanitizeForTerminal,
   } = await import('@mmnto/totem');
 
   const allIgnore = [...config.ignorePatterns, ...(config.shieldIgnorePatterns ?? [])];
@@ -175,8 +176,14 @@ export async function getDiffForReview(
 
     if (!diff.trim()) {
       const base = getDefaultBranch(cwd);
-      // Resolution prefers origin/<base> over the (often stale) local ref (#2054).
-      log.info(tag, `Diff source: branch-vs-base (prefers origin/${base}...HEAD)`);
+      // Sanitize the git-derived branch name before logging (terminal-injection
+      // hardening). Name both refs in resolution order so the line isn't
+      // misleading when the offline fallback to the local ref fires (mmnto-ai/totem#2054).
+      const safeBase = sanitizeForTerminal(base);
+      log.info(
+        tag,
+        `Diff source: branch-vs-base (origin/${safeBase}...HEAD, else local ${safeBase})`,
+      );
       diff = filterDiffByPatterns(getGitBranchDiff(cwd, base), allIgnore);
       source = 'branch-vs-base';
     }
--- a/packages/cli/src/git.ts
+++ b/packages/cli/src/git.ts
@@ -175,7 +175,8 @@ export async function getDiffForReview(
 
     if (!diff.trim()) {
       const base = getDefaultBranch(cwd);
-      log.info(tag, `Diff source: branch-vs-base (${base}...HEAD)`);
+      // Resolution prefers origin/<base> over the (often stale) local ref (#2054).
+      log.info(tag, `Diff source: branch-vs-base (prefers origin/${base}...HEAD)`);
       diff = filterDiffByPatterns(getGitBranchDiff(cwd, base), allIgnore);
       source = 'branch-vs-base';
     }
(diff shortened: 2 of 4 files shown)


satur8d:
@greptileai review

satur8d:
## Review round 1 — all three findings fixed (`b0563526`)

|---|-----|---------|-------------|
| 1 | CodeRabbit (minor) | Sanitize git-derived branch name before logging (terminal-injection) | **Fixed** — `base` is wrapped in `sanitizeForTerminal` before the diff-source log line |
| 2 | Gemini (high) | An already-`origin/`-prefixed base → `origin/origin/<base>` (invalid ref + wasted spawn) | **Fixed** — normalize off any `origin/` prefix before building the ref list; +1 regression test. See note below. |
| 3 | Greptile (P2) | Log reads "prefers origin" *before* resolution → misleading when the offline fallback fires | **Fixed** — the log now names both refs in resolution order (`origin/<base>...HEAD, else local <base>`), so it no longer implies which ref was actually used |

@gemini-code-assist on #2 I went slightly past the suggested patch: instead of collapsing to `[baseBranch]` when it is already `origin/`-prefixed (which **drops the local fallback**), I normalize off the prefix and keep both refs — `[origin/<base>, <base>]`. That local fallback is load-bearing for this change: it is the offline / no-remote / shallow-CI path that guarantees the reorder loses nothing versus the old behavior. Normalizing covers both input shapes (bare and prefixed) **and** preserves the fallback, so `origin/origin/<base>` can't occur either way.

Full local gate green on `b0563526`: core 2103 + cli 2363 tests, `totem lint` (0 errors), `format:check` clean, `totem review` (0 findings). CodeRabbit will re-review this push automatically.

satur8d:
@greptileai re-review please. Your P2 (the `branch-vs-base` log reading "prefers origin" before resolution, misleading when the offline fallback fires) was addressed in `b0563526` — the log now names both refs in resolution order (`origin/<base>...HEAD, else local <base>`). Re-triggering manually since our `greptile.json` is `skipReview: AUTOMATIC` (on-demand — no auto-review on push), so your summary won't refresh on its own.

PR #2076, opened 2026-06-03
fix(hooks): pre-push resolve-order prefers workspace/pinned build over volatile global (#2053)

### LINKED ISSUE #2053: fix(hooks): pre-push resolve-order prefers stale PATH-global totem over the workspace build (contradicts its own comment)
## Summary

The totem-managed **pre-push hook** resolves the `totem` binary via `command -v totem` (PATH-global) **first** — directly contradicting its own line-11 comment, *"prefer local workspace build over published package."* In the totem monorepo this runs the **stale global** `npm i -g @mmnto/cli` (e.g. 1.50.0) against **HEAD** code (e.g. 1.52.0), so every pre-push totem gate enforces a 2-versions-stale ruleset against the build we're developing.

This is the root cause behind the recurring "global-vs-local lint divergence" (lesson `1ef06d16`) — not a scope artifact, a **stale binary shadowing the workspace build**.

## Evidence

`.git/hooks/pre-push` (template lives in `init.ts`), lines 11–21:

```sh
# Resolve totem command — prefer local workspace build over published package
if command -v totem >/dev/null 2>&1; then
  TOTEM_CMD="totem"                    # ← global 1.50.0 on PATH wins FIRST
elif [ -f pnpm-workspace.yaml ] && pnpm exec totem --version …; then
  TOTEM_CMD="pnpm exec totem"          # ← ALSO 1.50.0 (workspace bin not linked)
…
```

- `which totem` → `~/AppData/Roaming/npm/totem` = `npm i -g @mmnto/cli@1.50.0`
- `pnpm exec totem --version` → **1.50.0** — because `node_modules/.bin/totem` does **not exist**, so `pnpm exec` falls through to PATH (the workspace cli bin was never linked)
- workspace `node packages/cli/dist/index.js --version` → **1.52.0** (HEAD)

## Impact

1. **Stale-rule enforcement (SEVERE in the dev monorepo).** `verify-manifest`, `lint`, `claim-discipline`, `verify-badges`, and the `is_agent` strict `review` all run 1.50.0's rules against 1.52.0 code. The hook's stated intent ("prefer local workspace build") is the opposite of what the code does.
2. **Asymmetric harm (MILD in consumers).** Consumer repos (`liquid-city`, `arhgap11`) *should* run the published binary — but ideally their **pinned floor**, not whatever a stale global `npm i -g` left on PATH. A consumer on a stale global is a smaller version of the same bug.

## Fix — context-aware resolve-order

The discriminator is **workspace-membership, not workspace-presence**: is `@mmnto/cli` (the package providing the `totem` bin) a **workspace member of this repo**?

- **Yes** (we're *in* the totem monorepo) → resolve the **workspace `dist`** build (dogfood HEAD).
- **No** (consumer repo, even if it's itself a pnpm workspace that merely *consumes* `@mmnto/cli`) → resolve the **published / pinned** binary.

`"pnpm-workspace.yaml present"` is too broad — a consumer can be a workspace without `@mmnto/cli` being a member. Also: ensure the workspace cli bin links into `node_modules/.bin` so `pnpm exec totem` stops silently falling through to the stale global.

## Relationships (do not merge)

- **#2018** — *distinct but linked, same root.* A global install carries no devDeps, so it lacks `@google/genai` (a monorepo devDep @ 2.6.0) → global `review`/`spec`/`search` hit #2018, while the workspace build runs clean. The genai fix is **separable** (runtime dep / graceful degrade) and should close on **its own** remedy, not this hook fix.
- **#2051** — sibling hook-correctness bug (working-tree vs push-range scope).
- **totem#1854** (versioned-hooks upgrade mechanism) — the cohort sweep (every repo with a stale global + totem-managed hook) **self-heals once this resolve-order fix ships AND hooks re-init via #1854**. The reinit is a **sibling** of `mmnto-ai/totem-strategy#269` (same reinit-mechanism class, both served by #1854) — *not* the same issue; tracked here as riding #1854 rather than as a separate sweep ticket.
- Discovered in **#2052** (WS2 `totem orient`) — cost 3 cleanup rounds chasing the stale-global lint.
- Reference: lesson `1ef06d16` (this bug is its root cause).

satur8d:
## What

The generated git hooks' shared `buildResolveBlock` checked `command -v totem` (the **PATH-global**) *first*, contradicting its own "prefer local workspace build" comment. In a dev monorepo that runs a **stale globally-installed `@mmnto/cli` against HEAD code** — enforcing a 2-versions-stale ruleset — which is the root cause of the `lesson-1ef06d16` "global-vs-local lint divergence."

This rewrites the shared resolve cascade to, in order:

1. **workspace-HEAD** — `node packages/cli/dist/index.js`, guarded on the `@mmnto/cli` package name **and** a built `dist/`
2. **pinned `@mmnto/cli`** — `node node_modules/@mmnto/cli/dist/index.js` (the package's own entry — identity-guaranteed, not a bare `totem` bin a colliding package could shadow)
3. **`pnpm exec totem`** (consumer pnpm workspace)
4. **`command -v totem`** (PATH — demoted to fallback)
5. **package-manager `dlx`** fallback

Because pre-push / post-merge / post-checkout all embed this one block, the fix covers all three. It also adds **`@mmnto/cli: workspace:*` to the monorepo-root devDeps** so `pnpm exec totem` (tier 3, and manual usage) resolves to the workspace build instead of falling through to the global — the *other half* of the bug.

## Why this order (Tenet 14)

Preferring the **pinned / in-tree** build over a **volatile ambient global** is Tenet 14 (never tie governance to volatile state) applied at the resolver — a pinned dep is deterministic; whatever npm-global happens to be on PATH is not. The stale-global-shadowing-a-newer-build failure (`lesson-1ef06d16`) is exactly that foot-gun. ADR-072 §2's "PATH beats `npx`/`dlx`" intent is **preserved** (PATH still beats `dlx` at 4-vs-5); only *pinned/workspace beats ambient PATH* changes. Every pinned tier is **identity-guarded on the `@mmnto/cli` package**, never a bare `totem` bin name.

**Reconciled with strategy-claude** (the resolution order lives in their ADR-072): concurred the full cascade incl. the 1↔2 inversion. The doctrine-side amendment documenting these exact tiers is `mmnto-ai/totem-strategy#521`.

## Scope

Template only. Already-installed stale hooks across the cohort self-heal via the versioned-hooks upgrade (#1854) — **not** folded here.

## Tests / gates

- 2 new unit tests: cascade-order precedence + per-tier `@mmnto/cli` identity guard.
- Generated block syntax-checked (`sh -n`); tier-1 grep confirmed matching in-tree.
- Full suites green: cli 2365 (install-hooks 116). `totem lint` (0 errors), `format:check`, `verify-lockfile-sync`, `verify-manifest`, `claim-discipline --strict`, `totem review` (0 findings), `doctor --strict` (0 failures).

> **Reviewers:** the residual `totem lint` warnings are the **#2063 false-positive class** (warning-tier rules mis-firing on test-file idioms) — non-blocking, tracked in #2063.

Closes #2053
Part of #2055
Doctrine ref: mmnto-ai/totem-strategy#521 (ADR-072 §2 amendment)


satur8d:
@greptileai review

satur8d:
@greptileai re-review please. Addressed your P2 (tier-2 had no `@mmnto/cli` identity guard) in `99801b70`:

tier-2 now targets `@mmnto/cli`'s **own entry** — `node node_modules/@mmnto/cli/dist/index.js` (guaranteed by the package path; `files: ["dist"]` ships it) — instead of the bare `node_modules/.bin/totem` shim. So a package that coincidentally exposes a `totem` binary in a consumer repo can no longer be picked up, and every pinned tier (1 and 2) is now identity-guarded on `@mmnto/cli`. This also drops the `.bin` shim's Windows `-x`/`-f` quirk entirely.

Issue #2073, opened 2026-06-03
Tracking: complete totem doctor --parity — mechanical + manual-attestation slices, then publish (follow-up to closed #2069)


satur8d:
**Lightweight follow-up tracker** for the remaining slices of `totem doctor --parity` (strategy#448 doctor). The parent impl issue #2069 was auto-closed by the version-pinned PR-1 merge (#2071, `Closes #2069`), but the feature is **incomplete and unpublished** — this issue carries the remaining scope so it isn't lost to a closed issue.

State is **derived from the children / design docs** (Tenet 20 — link, don't mirror); see `.totem/specs/448.md` and `.totem/specs/2073.md` for the locked designs.

## Shipped (on `main`, unpublished by design)

- **Version-pinned slice** (#2071 → `55b33a0a`): real pin-currency verdicts for the 4 `@mmnto/*` deps contracts; `--strict`+`blocking`→fail. Skeleton #2070; `package:` adopted (strategy#517).
- **Mechanical — skills** (#2078 → `066a91b5`): managed-block content-equality for the distributed `claude-skills` contract; verdict-state split (`pass`/`warn`/`info`/`unknown`/`skip`), CRLF/LF normalize, `totem:fork` reader, binary self-report.
- **Mechanical — git-hooks** (#2079, in review): per-repo **regenerated-canonical** content-equality for the four git hooks; detects stale-version drift (the detection half of #1854) without package-manager / tier false-positives.

## Remaining slices

- **Mechanical — session-start-orientation** _(optional fast-follow)_: whole-file-static content-equality for the `.claude` / `.gemini` SessionStart templates (reuses the #2079 generated-artifact primitive).
- **Manual-attestation**: `skip`/`info` on staleness only, never `fail` (claim-class bound). Currently a stub. Covers the doctrine rows (`governance-doctrine`, `agent-memory-doctrine`) + vendor-SDK couplings, which strategy#526 moved to `manual-attestation` (honest tractability — no pin mechanism yet).
- **Changeset + publish**: the completing slice triggers the cohort release. `@mmnto/cli` stays **1.53.5** until then.

## Notes

- The mechanical gate is **cleared** (strategy 3-lane blind round, 2026-06-04): the doctor is strictly local-read-only; the cross-repo surface is a subset (only doctrine contracts), and those are `manual-attestation` for now.
[...]

PR #2086, opened 2026-06-05
feat(cli): doctor --strict folds in the parity sensor for opted-in repos (#2085)

### LINKED ISSUE #2085: feat(cli): fold parity sensor into `totem doctor --strict` — close green-by-not-checking (strategy#545 Half 2)
## Gap (strategy#545 Half 2)

`totem doctor --parity` is an **exclusive early-return mode** (`index.ts:1091-1095`): consumer CI runs `doctor --strict`, which falls through to the 12-check suite and **never constructs the parity sensor**. So "CI `--strict` green ⇒ parity-clean" is **green-by-not-checking** — the sensor shipped (#2073 / strategy#448) but `--strict` doesn't exercise it.

## Decision

satur8d-approved **S0** (cohort round, reported via strategy-claude 2026-06-05T0303Z): fold the parity sensor into `--strict`.

Make `doctor --strict` run the parity sensor **when (and only when)** a repo-local `orient.parityManifest` is configured; **no-op (no output)** for repos that never configured one (zero consumer-CI churn). Closes Half 2.

## Scope

- **In:** wire the existing `doctorParityCliCommand` into the `--strict` path at the CLI edge (per ADR-109 / Tenet 13 — exit decisions live at the CLI edge, `DoctorOptions:1274`); add an "only when configured" guard so non-adopters see no change. Reuse the existing render + `blockingDriftIds` throw-gate verbatim.
- **Out:** the distribution build (292 S1-S3 — distribute `parity-manifest.yaml`, sidecar, provenance bind) which waits on strategy's seam-1/seam-3 framing. No change to consumer CI invocation (they already run `doctor --strict`).

## Decouple-gate context

This closes Half 2 of strategy#545. Half 1 (manifest distribution) rides Proposal 292 (#527). Live reference: strategy#546 (prove-it-live — 19 contracts, full claim-class spectrum).

Refs: strategy#545, strategy#527, strategy#546, strategy#448, #2073

satur8d:
## What

Closes the **green-by-not-checking** gap (mmnto-ai/totem-strategy#545 Half 2): consumer CI runs `totem doctor --strict`, which never exercised the parity sensor because `--parity` was an exclusive early-return mode (`index.ts:1091`). Now `--strict` runs the sensor **when (and only when)** a repo-local `orient.parityManifest` is configured — a **no-op (byte-identical output) for non-adopters**, so there is zero consumer-CI churn until a repo opts in.

This is **S0** of the decouple-gate work — satur8d-approved, independent of the distribution build (292 S1–S3, which waits on strategy's seam-1/seam-3 framing).

## How

- `index.ts` — the CLI-edge fold: after the 12-check suite, `if (opts.strict)` calls `doctorParityCliCommand({ strict: true, onlyWhenConfigured: true })`. The `--parity` branch returns above, so it never double-runs.
- `doctor-parity.ts`:
  - `ParityCheckResult.configured: boolean` — single-sourced in `checkParity` (reusing the `isGlobalConfigPath` global-leak guard), so the CLI edge never re-derives config and a global manifest can't leak into the gate. It reflects whether the **field is set**, not whether the file loaded — a configured-but-broken manifest is `configured: true` so `--strict` surfaces the error instead of silently no-op'ing.
  - `ParityCliOptions.onlyWhenConfigured?: boolean` — the command self-gates. The standalone `doctor --parity` is unchanged (still renders the honest-absent SKIP).
- `doctor.ts` is **untouched**.

## Why the CLI edge (not inside `doctorCommand`)

Per **ADR-109 / Tenet 13 (Sensors-Not-Actuators)** and `DoctorOptions:1274` ("the exit-code decision lives at the CLI edge"), the parity throw-gate is an exit decision and belongs at the edge. Folding it into `doctorCommand` would (a) embed an actuator in a function documented composable/side-effect-free, and (b) force a `CheckStatus` widening — `doctorCommand` returns `CheckStatus = pass|warn|fail|skip`, but parity emits the richer `info`/`unknown`. strategy-claude concurred this correction (the design was bounced for review before code).

## Verified no-op here

This repo does not configure `orient.parityManifest`, so `totem doctor --strict` is byte-identical to before. The behavior change fires only for opted-in repos (e.g. totem-strategy post-#546, which loads 19 contracts across the full claim-class spectrum).

## Tests / gate

- 8 new tests covering all six invariants (configured-flag derivation incl. the global-leak case; the no-op-renders-nothing assertion vs. the standalone SKIP; the strict-fold throw-gate on blocking drift; configured-but-broken still runs).
- Full gate green: `tsc` · 2391/2391 CLI tests · eslint 0 errors · prettier · `totem lint` PASS (0 errors) · `totem review` PASS (no findings) · `verify_execution` PASS.

Closes #2085. Refs strategy#545, strategy#546, strategy#448, #2073.

### DIFF
--- a/.changeset/strict-parity-fold-2085.md
+++ b/.changeset/strict-parity-fold-2085.md
@@ -1,0 +1,7 @@
+---
+'@mmnto/cli': patch
+---
+
+`totem doctor --strict` now exercises the parity sensor when a repo-local `orient.parityManifest` is configured — closing the "green-by-not-checking" gap where consumer CI (which runs `--strict`, not `--parity --strict`) never ran the drift check (mmnto-ai/totem-strategy#545 Half 2).
+
+Zero churn for non-adopters: a repo that hasn't configured a manifest sees byte-identical `--strict` output (the fold is a no-op until you opt in). Per ADR-109 / Tenet 13 the throw-gate stays at the CLI edge, and the standalone `doctor --parity` is unchanged.
--- a/packages/cli/src/commands/doctor-parity.test.ts
+++ b/packages/cli/src/commands/doctor-parity.test.ts
@@ -18,7 +18,7 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { cleanTmpDir } from '../test-utils.js';
 import { checkParity, doctorParityCliCommand } from './doctor-parity.js';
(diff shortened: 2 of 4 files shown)


satur8d:
/gemini review
@greptileai review

satur8d:
@greptileai review

Pushed `6c821ec0` since your last pass, addressing your P2 (missing CLI-level test for the configured-but-broken fold path):

- Added a `doctorParityCliCommand({ onlyWhenConfigured: true })` test asserting that a **configured-but-missing** manifest **renders its WARN** rather than silently no-op'ing — closing the gap where that invariant was only covered at the `checkParity` level.
- Mutation-verified it has teeth: with the early-return guard flipped to `onlyWhenConfigured && configured` the new test fails; on the shipped `&& !configured` it passes.

Please re-review the latest push.

Issue #390, opened 2026-03-12
feat: review-guidance lessons and bot-optimized export


satur8d:
## Summary

Add a `review-guidance` lesson tag and a 4th Proactive Anchoring trigger (\"Decline Trigger\") so that agents automatically capture recurring review bot false positives as lessons, which then export to review bot config files.

## Problem

Review bots (GCA, CodeRabbit, Copilot) repeatedly suggest patterns the team has declined — Zod for small parsers, configurable constants, async exec, etc. Today this knowledge lives in the human's head or in manually-maintained styleguide files. There's no automated feedback loop from \"declined suggestion\" → \"bot stops suggesting it.\"

### 1. `review-guidance` tag
A new conventional tag for lessons that are specifically intended to guide review bots. Example:

```
totem add-lesson \"Don't suggest Zod for small LLM response parsers — manual validation is sufficient for < 10 fields\" --tags review-guidance,over-engineering
```

### 2. Decline Trigger (4th Proactive Anchoring trigger)
Add to the AI prompt block injected by `totem init`:

> **The Decline Trigger:** When you decline a review bot suggestion for a pattern that has been declined before, call `add_lesson` with the `review-guidance` tag to capture the declined pattern and rationale.

This goes alongside the existing Trap, Pivot, and Handoff triggers.

### 3. Bot-optimized export format
When `totem compile --export` writes to review bot config targets (e.g., `.gemini/styleguide.md`, `.github/copilot-instructions.md`), lessons tagged `review-guidance` should be formatted as \"Do Not Suggest\" rules — prose guidance, not regex patterns.

Example output in `.gemini/styleguide.md`:
```markdown
## Do Not Suggest (Auto-Generated from Totem Lessons)
- **Zod for small parsers:** Manual validation is sufficient for < 10 fields.
- **Configurable constants:** Named constants are sufficient; no runtime configurability needed.
```

### 4. Data Flywheel
The loop becomes:
1. Bot suggests something → Agent declines → Agent calls `add_lesson --tags review-guidance`
2. `totem compile --export` updates the bot's config
3. Bot stops suggesting it

Team-specific review preferences become self-reinforcing without manual maintenance.

## Implementation Notes

- The `review-guidance` tag is just a convention — no schema changes needed for tags
- The Decline Trigger is a prompt change in `AI_PROMPT_BLOCK` (init.ts) + REFLEX_VERSION bump
- The export format change is in `packages/core/src/exporter.ts` — filter by tag, format as prose
[...]

PR #2119, opened 2026-06-07
fix(cli): totem mail silently drops frontmatter-only dispatches >2 KiB + --json emits no JSON (#2118, #2097)

### LINKED ISSUE #2097: fix(cli): totem mail --json emits no JSON — flag accepted but ignored
## Repro (2026-06-06, workspace build @ main `ad21e265`)

```
node packages/cli/dist/index.js mail --json
```

- stdout: **empty** (verified via `$out = node ... 2>$null` → 0 lines)
- stderr: the normal human-readable `[Mail]` text
- exit 0

Reproduced in both states: 98 unread and 0 unread. `--help` documents the flag ("Emit JSON to stdout instead of human-readable text to stderr").

## Pointer

`packages/cli/src/commands/mail.ts:420` has the `opts.json === true` → `process.stdout.write(JSON.stringify(result...))` path, so the renderer exists; the flag value apparently never reaches the handler as `true`. Suspect the command/option wiring in the CLI registration (option name mismatch or args-order in the action handler).

## Why it matters

`--json` is the hook-friendly consumption path for mail (ADR-106 § 3); agents scripting against it currently get silent-empty and fall back to parsing colored stderr text. Found during the 2026-06-06 housekeeping pass while scripting the processed-ledger drain.

### LINKED ISSUE #2118: fix(cli): pollMail silently drops frontmatter-only dispatches >2 KiB — parseHeader blank-line terminator collides with the cohort dispatch convention
## Defect

`totem mail` (and every consumer of `pollMail`, including SessionStart hooks) **silently drops** any dispatch that is frontmatter-only (no blank line anywhere) **and** larger than 2 KiB. The cohort dispatch convention (ADR-098 style: entire message in the `subject:` field, frontmatter-only body) produces exactly this shape routinely — so the reader misses precisely the densest coordination mail.

## Root cause (verified against source 2026-06-07)

- `packages/cli/src/commands/mail.ts:109` — `MAX_HEADER_BYTES = 2048`, a forged-frontmatter defense whose comment ("Real frontmatter is dozens of bytes; 2 KiB is generous-but-bounded") is falsified by the cohort's own dispatch convention.
- `mail.ts:122` — `content.split(/\r?\n\r?\n/, 2)`: with **zero blank lines** the whole file lands in `parts[0]`.
- `mail.ts:127` — `if (parts.length < 2 && content.length > MAX_HEADER_BYTES) return null;` → the **entire file is rejected**, not truncated.
- `mail.ts:359-360` — `const header = parseHeader(content); if (!header) continue;` — the **only failure path in the module that emits nothing** (per-file `readFileSync` failure at `mail.ts:356` pushes a structured warning; parse-null is silent).
- `mail.test.ts:568` enshrines the lenient under-cap no-separator case; the over-cap zero-blank case has no test.

## Evidence (strategy-claude forensics, pinned on strategy#543)

Prediction table **8/8**: `size > 2048 && blankLines === 0` predicts every observed miss, age-independent:

| seen | bytes / blanks | missed | bytes / blanks |
|---|---|---|---|
| 0234Z | 3232 / 11 | 0322Z | 2258 / 0 |
| 0205Z | 2110 / 9 | 2120Z | 2263 / 0 |
| 1857Z | 1968 / 0 (under cap) | 1901Z | 2273 / 0 |
| 1959Z | 1472 / 0 (under cap) | 2015Z | 3452 / 0 |

Five independent reproductions across three agents and two vendor lanes, including strategy-claude's own SessionStart hook dropping a 4163-byte / 0-blank dispatch (2107Z). A 3-day-old missed dispatch (strategy-gemini repro) kills any mtime/freshness hypothesis — the defect is parse-side, age-independent.

## Fix shape (dispositions open to preflight)

1. **Minimum viable:** emit a structured warning on parse-null so the drop is observable (parity with the `readFileSync` failure path).
2. **The real fix:** the header terminator is wrong — dispatches are `---`...`---` YAML-frontmatter delimited, not blank-line delimited. Parse to the **closing `---`**; `MAX_HEADER_BYTES` then bounds the *search window* (forged-frontmatter defense preserved) instead of rejecting the file.
3. **Bootstrap constraint:** the warning about a broken reader travels via the mail the broken reader misses — the fix's first-run path must not depend on the reader (interim cohort floor is a direct outbox-dir read).

## Adjacent (same read-path, candidate to fix in the same PR)

- #2097 — `--json` emits no JSON.
- False-unread half (separate root cause, may warrant its own slice): `processed/` marking lags actual consumption **cohort-wide** (strategy 05-27, strategy-gemini 05-17, lc 05-24 — consumption-without-marking is universal practice). Sender-side interim discipline is live (blank line + one-line body footer after the closing `---`).
- Vendor-hook gap (out of scope here, noting for the record): strategy-gemini's SessionStart hook never polls mail at all — a reader fix doesn't help a lane that never reads.

Refs: strategy#543 (repro + root-cause pins) · ADR-106 (coordination layer) · #2042 family (mail actuator). Ruled first in the 2026-06-07 next-wave synthesis (operator + strategy-claude, 2113Z dispatch); proceeds in parallel with the ECL design review, which judges class-vs-instance afterward.

satur8d:
## Summary

Two read-side defects in `totem mail`, fixed together (ratified first in the 2026-06-07 next-wave synthesis; root-cause forensics pinned on strategy#543):

1. **Silent drop of frontmatter-only dispatches > 2 KiB.** `parseHeader` split on the first blank line and rejected any > 2 KiB file without one. The cohort dispatch convention (whole message in `subject:`, frontmatter-only, zero blank lines) produces exactly that shape — `size > 2048 && blankLines === 0` predicted **8/8** observed misses (2,110–4,163 bytes, age-independent, reproduced across three agents and two vendor lanes, including the sender's own SessionStart hook). The drop at the old `if (!header) continue` was the module's **only warning-less failure path**.
2. **`--json` emitted nothing** (#2097). The flag never reached the handler: the program-level `--json` (`index.ts`) swallows a post-subcommand flag (commander parent/child option collision — verified in isolation: sub opts `{}`, program opts `{json:true}`). The renderer itself was already correct.

## Fix

- **Parse to the closing `---` delimiter** — the real header terminator. The byte cap becomes `MAX_HEADER_SEARCH_BYTES` (16 KiB) bounding the **search window** for the closing delimiter instead of rejecting the file. *Deliberate deviation from the suggested fix-shape:* porting "cap bounds the search" at the existing 2048 would still reject every observed miss — genuine cohort frontmatter runs to 4,163 bytes; the window is sized ~4× the observed max (provenance in the constant's comment).
- **Forged-frontmatter defense preserved, re-derived**: fields are matched only inside the delimited header; body `to:` lines cannot fabricate a match; no-closing-delimiter files are rejected.
- **Mail-shaped rejects warn** (`mail parse failed (repo/agent/file): <reason>`) — parity with the module's readFile/readdir/processed-scan failure paths (Tenet 4). Non-mail-shaped strays (no opening `---`) stay silent by design: a recipient cannot clear a sender's stray file, so a warning there is permanent noise. Discriminated `HeaderParse` result carries the distinction.
- **`--json` routed via `cmd.optsWithGlobals()`** — merges both scopes, so `totem mail --json` and `totem --json mail` agree. Both option declarations stay (help text + env-sniff JSON mode untouched).

## Verification

- RED-first: 7 failing tests before implementation (regression class, CRLF variant, over-window reject + warn, inner-blank-line header, no-`to:` warnings, no-close warnings).
- 40/40 mail tests; **2454/2454** full `@mmnto/cli` suite.
- `totem lint` PASS (0 errors) · `totem review` PASS (no findings).
- **Live probe**: re-ran the strategy#543 reproduction (poll as `strategy-claude` on the built CLI) — the 4,163-byte frontmatter-only dispatch the broken parser dropped from the sender's own hook **now surfaces**; `node dist/index.js mail --json` emits parseable JSON end-to-end; zero parse warnings across the live workspace.
- Design doc (approved pre-build) committed in `.totem/specs/2118.md`.

Out of scope, per the ratified design: inbox model / derived handled-state / processed-marking mechanization (Proposal 290, #390, strategy#506 — the ECL design review judges the class; this is the instance fix proceeding in parallel).

Closes #2118
Closes #2097

### DIFF
--- a/.changeset/mail-parse-silent-drop.md
+++ b/.changeset/mail-parse-silent-drop.md
@@ -1,0 +1,7 @@
+---
+'@mmnto/cli': patch
+---
+
+fix(cli): `totem mail` no longer silently drops frontmatter-only dispatches over 2 KiB (#2118), and `totem mail --json` actually emits JSON (#2097).
+
+`parseHeader` now parses to the closing `---` frontmatter delimiter instead of splitting on the first blank line — the old parser rejected any >2 KiB file without a blank-line separator, which silently dropped every cohort-convention dispatch (whole message in `subject:`, zero blank lines; 8/8 of the observed misses, up to 4,163 bytes of genuine frontmatter). The byte cap now bounds the closing-delimiter search window (16 KiB) instead of rejecting the file, and every mail-shaped parse rejection emits a structured warning (parity with the module's other failure paths) while non-mail-shaped strays stay silent by design. The `--json` flag was being swallowed by the program-level `--json` option (commander parent/child collision); the mail action now reads `optsWithGlobals()`.
--- a/.totem/specs/2118.md
+++ b/.totem/specs/2118.md
@@ -1,0 +1,146 @@
+### Problem Statement
+
+`totem mail` (and hooks using `pollMail`) silently drops valid, frontmatter-only dispatches larger than 2 KiB because the parser incorrectly relies on a double-blank-line separator (`\r?\n\r?\n`) to isolate the header. We need to refactor `parseHeader` to bound its search using the closing YAML frontmatter delimiter (`---`), emit a structured warning instead of silently failing, and optionally fix the broken `--json` flag to emit parseable JSON.
+
+### Architectural Context
+
+- **Error Handling & Logging Conventions (Lesson 9):** Requires `log.error()` to use `'Totem Error'` as the tag, but specifically mandates that `log.warn()` MUST use the command-specific `TAG` constant.
+- **YAML Handoffs:** Code snippets reveal `parseHeader` enforces that "real handoffs open with a YAML frontmatter delimiter." Blank-line separation is a legacy assumption that must be replaced by strict delimiter bounding.
+
+### Files to Examine
+
+1. `packages/cli/src/commands/mail.ts` — Contains `MAX_HEADER_BYTES`, `parseHeader`, `pollMail` read loop, and the CLI command handler.
+2. `packages/cli/src/commands/mail.test.ts` — Missing tests for the >2KB zero-blank-line case and silent drop behavior.
+
+### Technical Approach & Contracts
+
+**1. Delimiter Refactoring (`parseHeader`)**
+Instead of splitting the entire file by `/\r?\n\r?\n/` and validating the first chunk's size, we will strictly parse the frontmatter boundaries:
+
+- Verify the file starts with `---` (optionally preceded by whitespace).
+- Find the index of the _closing_ `---` delimiter (`\n---` or `\r\n---`).
+- **Forged-Frontmatter Defense:** If the closing delimiter is not found, OR its index is strictly greater than `MAX_HEADER_BYTES` (2048), return `null`.
+- Extract the matched boundary as the header string and pass it to the existing metadata parser.
+
+**2. Visibility for Drops (`pollMail`)**
+Update the file iteration loop in `pollMail`. When `parseHeader(content)` returns `null`, emit a structured warning via `log.warn(TAG, ...)` before executing `continue`, ensuring parity with the `readFileSync` failure path.
+
+**3. Adjacent Fix: JSON Output (`mail` command)**
+Address #2097 in the same pass. If the `--json` flag is provided to the CLI command, execute `console.log(JSON.stringify(result, null, 2))`. Ensure that `log.warn` outputs to `stderr` (which is standard for the logger) so it does not corrupt the `stdout` JSON pipeline.
+
+### Edge Cases & Traps
+
+- **CRLF vs LF:** The closing delimiter search must account for both `\n---` and `\r\n---`. Using a simple `.indexOf('\n---')` might miscalculate bytes or miss carriage returns. Use a regex with bounded execution or check both.
+- **JSON Pollution:** Emitting the new warning could pollute `stdout` if the logger isn't strictly writing to `stderr`. This breaks programmatic consumers of `--json`. The warning must be safely routed.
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@greptileai review

satur8d:
## Round 1 dispositions — all three findings actioned in `a1f7790c`

| # | Finding (Greptile, P2) | Disposition | Trace |
|---|---|---|---|
| 1 | Window-reason threshold off by 3 bytes (`mail.ts`) | **Fixed** — guard is now `content.length > 3 + MAX_HEADER_SEARCH_BYTES`; boundary test added (16,386-byte file gets the plain reason, >16,387 gets the window reason) | `a1f7790c` |
| 2 | `CLOSING_DELIMITER` vs interior `---` in block scalars | **Fixed (diff applied; rationale corrected)** — the suggested regex is applied with a test, but for a different reason than stated: YAML block-scalar content must be indented, so a column-0 bare `---` cannot legally occur *inside* one, and the suggested `[ \t]*` **loosens** the match rather than tightening it — it would not prevent the described failure. What the diff genuinely buys is trailing-whitespace tolerance on hand-authored closing delimiters, and that's the documented reason in the code + test | `a1f7790c` |
| 3 | Untyped `optsWithGlobals()` passthrough (`index.ts`) | **Fixed** — typed destructure exactly as suggested; future global options can no longer silently shadow `MailCommandOptions` fields | `a1f7790c` |

Gates re-run post-fix: 42/42 mail tests, `totem lint` PASS (0 errors), `totem review` PASS.

@greptileai review

Issue #1970, opened 2026-05-19
Canonical cross-repo outbox poll for SessionStart hooks (ADR-106 § 3)


satur8d:
## Why

[`mmnto-ai/totem-strategy:adr/adr-106-ephemeral-coordination-layer.md § 3`](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-106-ephemeral-coordination-layer.md) specifies a cross-repo outbox poll mechanism:

> Senders write to their own outbox in their home repo with `to: <recipient-agent-id>` (or `to: broadcast`) in frontmatter. Recipients at session-start scan known-repo glob patterns (default: `<workspace>/*/.totem/orchestration/*/outbox/*.md`; recursive variant `<workspace>/**/.totem/orchestration/*/outbox/*.md` for nested layouts) and filter by `to:` matching their own agent id or `broadcast`.

**No hook in any repo currently implements this poll.** Empirical failure this session: both `totem-claude` and `lc-claude` reported "no mail" despite handoffs sitting in `mmnto-ai/totem-strategy:.totem/orchestration/strategy-claude/outbox/` for ~30-36h with `to: <recipient>` in frontmatter. Three handoffs are still rotting in my outbox right now:

| File | To | Age |
|---|---|---|
| `2026-05-13T2231Z-arhgap11-gemini.md` | `arhgap11-gemini` | ~5d |
| `2026-05-18T1734Z-strategy-claude.md` | `totem-claude` | ~36h |
| `2026-05-18T1918Z-strategy-claude.md` | `lc-claude` | ~30h |

The strategy SessionStart hook's existing comment claimed *"Cross-agent unread count is surfaced by `totem describe`"* — that was aspirational. Verified via `grep -i 'outbox\|unread' packages/cli/src/commands/describe.ts` → no matches. Nothing in the CLI scans outboxes.

This is `feedback_hook_enforcement_as_totem_justification` self-illustrating: doctrine specified a mechanism nobody implemented, so coordination silently degraded.

## What

Add canonical cross-repo outbox poll to `@mmnto/cli`. Two API shapes considered:

### Option A — extend `totem describe`

The strategy hook's existing comment frames `totem describe` as the cross-agent unread surface. Match that framing by adding an `inbox: {...}` section to `totem describe`'s output (or to its JSON variant). Pros: single command for orientation. Cons: `describe` is already wide; risks bloat.

### Option B — new `totem mail` subcommand (Recommended)

[...]

PR #2134, opened 2026-06-09
feat(cli): totem mail send/reply outbound actuator (#2042)

### LINKED ISSUE #2042: feat(cli): totem mail send / reply — outbound mail actuator (ADR-106 coordination)
## Gap

`totem mail` **reads** unread cross-repo mail (ADR-106 §3) but there is **no send / compose**. Composing an outbound dispatch is entirely hand-authored and requires knowing five undocumented things: the outbox path (`.totem/orchestration/<self>/outbox/`), the filename convention (`<UTC>Z-<recipient>-<slug>.md`), the `adr-098-v0.4` frontmatter schema, the `claude-00NN` seq signature, and that the dir is gitignored. `totem mail send` today silently falls through to the read command (no-op, exit 0) — its own trap.

## Why it matters (the root, not a sibling)

This is the **root** of consumer-doctrine drift (`feedback_consumer_doctrine_drift_post_adr_cutover`), not a parallel instance: the read side gets *forgotten*, but the send side **cannot comply** with "run the command, don't hand-roll" — there is no command. A discipline that half the protocol structurally cannot satisfy is a missing actuator (Tenet 13: sensor without actuator), not a recurring lapse. Half-instrumented mail = unbounded per-message glue → counter to `totem_validation_criterion` (bounded, shrinking glue cost). Self-referential proof: the dispatches proposing this fix were themselves hand-authored.

## Proposed shape

- `totem mail send --to <agent> --subject <s> [--body-file <f>] [--in-reply-to <file>] [--priority <p>] [--related <ref>...]`
- `totem mail reply <in-reply-to>` (sugar)

Each:
- stamps the UTC timestamp + composes the filename,
- emits **validated** `adr-098-v0.4` frontmatter (known recipient, current schema version, well-formed refs),
- auto-fills the `claude-00NN` seq from session state,
- writes to the correct outbox,
- and fixes the silent `totem mail send` no-op fall-through → hard error / "did you mean `totem mail`?".

## Triad / cross-links

Coordination triad: **sensor** `totem mail` (ships) · **derivation** `totem orient` (mmnto-ai/totem-strategy#438) · **actuator** `totem mail send` (this issue). Composition point: `totem gate check` (mmnto-ai/totem-strategy#439) can validate outbound mail (recipient exists, schema valid, not addressed into a frozen lane).

Domain: ECL / ADR-106 coordination — a distinct mechanism-domain from #433's orientation+gates charter, hence standalone (per strategy-claude adjudication, T2116Z).

## Provenance

Surfaced by satur8d (2026-05-29); tracking-home adjudicated standalone by strategy-claude (T2116Z); `doctrine/cohort-roles.md §6.4` reconciled to scope the "run the command, don't hand-roll" discipline read-side-only until this ships.

satur8d:
## What

Adds the **actuator** half of the ADR-106 coordination triad (the sensor `totem mail` already shipped): `totem mail send` + `totem mail reply`. Before this, `totem mail send` silently fell through to the read command as a no-op, and every dispatch was hand-authored against five undocumented conventions — a discipline the protocol structurally could not satisfy (Tenet 13: sensor without actuator). Closes #2042.

## The 5 + 1 contract (the design spine)

The write-side validator realizes the gate contract from the moat-reframe falsification, with strategy-claude (ADR-098 owner) concurring 5/5 on the emit-shape:

- **Structural validity — by construction.** A `DispatchHeader` cannot be built without `schema` / `from` / `to` / `timestamp` / `subject` / `expected-action`, so a structurally-invalid dispatch is *unrepresentable*, not rejected after the fact — the strongest form of "enforce via substrate" (inv2 realized; inv6 doesn't even apply here).
- **Content validity — loud-warn-and-deliver (inv6 fail-open).** Predicates that can't be guaranteed at construction (recipient ∉ known cohort agents, empty refs) emit a **loud emit-time stderr warning** and write anyway. A blocked dispatch is worse than a malformed one — a typo'd recipient writing to the outbox under a wrong name is the exact silent-fail class (#2119).
- **Reader backstop — never-silent-drop.** `pollMail`'s scan-errors-always-warn (ADR-106 A1.2) is the third layer.
- **Hard-fail stays hard (Tenet 4):** usage errors (missing `--to`/`--subject`, ambiguous/unresolvable self, unreadable `--body-file`) and *actuation* failure (a write that didn't land) throw `TotemError`.

## Changes

- `totem mail send --to --subject [--from --body-file --in-reply-to --priority --related --expected-action --slug]` and `totem mail reply <source>` (infers recipient + `Re:` subject + `in-reply-to`).
- Emits **ADR-098 v0.4-compliant** frontmatter (`schema: adr-098-v0.4` / `timestamp:` RFC3339 / `expected-action:`) — totem is now the first v0.4-compliant emitter. Filename keeps the de-facto compact-minute stamp (colons are illegal on win32).
- `parseHeader` now reads `timestamp:` (v0.4 canonical) with `date:` backwards-compat fallback; `MailEntry.date` field name unchanged (no blast-radius rename — strategy folds the read-precedence into ADR-098's migration note on their side).
- Registering the subcommands closes the silent fall-through: `totem mail <unknown>` now hard-errors.
- `@mmnto/totem` exports `knownCohortAgents()` — single source of truth for the recipient set, derived from the cohort map (a hardcoded list would re-introduce the very drift this command fights).

## Out of scope (deferred)

`totem mail ack` + read-side mark-on-read / handled-state (ADR-106 A1.3 — separate lane).

## Tests

20 new tests (60 total in `mail.test.ts`, all green): sensor↔actuator round-trip, the fail-open headline (unknown recipient still written + warned), v0.4 emit shape, `timestamp:`/`date:` read precedence, ambiguous/unresolvable self, missing to/subject, unreadable body-file, same-minute collision disambiguation, path-traversal `--from`, reply inference, YAML-unsafe-subject quoting. Verified end-to-end through the built CLI (fall-through rejection + loud-warn-and-write). Design doc: `.totem/specs/2042.md`.

### DIFF
--- a/.changeset/mail-send-actuator.md
+++ b/.changeset/mail-send-actuator.md
@@ -1,0 +1,32 @@
+---
+'@mmnto/cli': minor
+'@mmnto/totem': minor
+---
+
+feat(cli): `totem mail send` / `mail reply` outbound actuator (#2042)
+
+Adds the actuator half of the ADR-106 coordination triad (the sensor `totem mail`
+already shipped). Before this, `totem mail send` silently fell through to the read
+command and every dispatch was hand-authored against five undocumented conventions —
+a discipline the protocol structurally could not satisfy (Tenet 13: sensor without
+actuator).
+
+- `totem mail send --to --subject [--from --body-file --in-reply-to --priority --related --expected-action --slug]`
+  composes + writes a **ADR-098 v0.4-compliant** dispatch (`schema:` / `timestamp:` /
+  `expected-action:`) to the sender's own outbox; totem is now the first v0.4-compliant
+  emitter. `totem mail reply <source>` is sugar that infers recipient + subject + `in-reply-to`.
+- The write-side validator is **fail-open** (ADR-106 inv6): structural validity is
+  enforced by construction (a malformed-shape dispatch is unrepresentable), while content
+  predicates that can't be guaranteed at construction (unknown recipient, empty refs) emit
+  a **loud emit-time warning** and write anyway — a blocked dispatch is worse than a
+  malformed one (#2119). Usage errors (missing to/subject, ambiguous/unresolvable self,
+  unreadable body-file) and actuation failures stay hard-fail (Tenet 4).
+- Registering the subcommands also closes the silent fall-through: `totem mail <unknown>`
+  now hard-errors instead of running the read as a no-op.
+- `parseHeader` now reads `timestamp:` (v0.4 canonical) with `date:` backwards-compat
+  fallback; the surfaced `MailEntry.date` field name is unchanged (no blast-radius rename).
+- `@mmnto/totem` exports `knownCohortAgents()` — the single source of truth for the
+  recipient set, derived from the cohort map (a hardcoded list in the actuator would
+  re-introduce the very drift this command fights).
+
+Emit-shape + the reader change concurred by strategy-claude (ADR-098 owner); OQ-1 ruled 1b.
--- a/.totem/specs/2042.md
+++ b/.totem/specs/2042.md
@@ -1,0 +1,171 @@
+### Problem Statement
+
+The `totem mail` command currently only supports reading cross-repo mail, lacking an actuator to compose and send outbound messages. As a result, users are forced to hand-author markdown files with undocumented conventions (ADR-098 schema, specific timestamp naming, specific directory structures), leading to drift; implementing `totem mail send` and `totem mail reply` commands will enforce these conventions structurally and eliminate the silent no-op fall-through when users attempt to send mail today.
+
+### Architectural Context
+
+- **ADR-106 & Tenet 13 (Sensor without Actuator):** The absence of a send command creates a structural inability to comply with protocol. This violates the `totem_validation_criterion` by creating unbounded glue cost.
+- **Totem Signoff (Gemini parity) > Cross-repo handoffs:** Established the pattern for outbound communication pathing (`<repoRoot>/.totem/orchestration/<agent-id>/outbox/`) and requires the `to: <recipient-agent-id>` frontmatter.
+- **Discrepancy Resolution:** The issue requires the filename convention `<UTC>Z-<recipient>-<slug>.md`, which slightly alters the older Signoff spec's `<YYYY-MM-DDTHHMMZ>-<your-agent-id>.md`. We will honor the issue's updated `<UTC>Z-<recipient>-<slug>.md` convention as it represents the modern ADR-106 coordination standard.
+
+### Files to Examine
+
+1. `packages/cli/src/commands/mail.ts` — Current entry point; contains the trap where `totem mail send` falls through to the read command.
+2. `packages/cli/src/commands/mail.test.ts` — Existing test interfaces (`MailEntry`, `OutboxFile`) that the new commands must cleanly integrate with.
+3. `.totem/session.json` (or equivalent session state file) — Source of truth for the `claude-00NN` / session sequence signature needed for auto-filling the frontmatter.
+
+### Technical Approach & Contracts
+
+To structurally enforce the consumer doctrine, we will introduce a strict subcommand router to `totem mail`.
+
+**Data Contracts:**
+
+1. **Zod Schema (`Adr098FrontmatterSchema`):**
+   ```typescript
+   export const Adr098FrontmatterSchema = z.object({
+     schema: z.literal('adr-098-v0.4'),
+     to: z.string(),
+     subject: z.string(),
+     seq: z.string().optional(),
+     inReplyTo: z.string().optional(),
+     priority: z.string().optional(),
+     related: z.array(z.string()).optional(),
+   });
+   ```
+
+**Sequence Logic:**
+
+1. **Routing:** Upgrade the command parser for `totem mail` to explicitly demand `send` or `reply` subcommands. If executed with no subcommand, it executes the existing read behavior. If an unknown argument (e.g., `send` currently) is passed without being registered, throw a hard error.
+2. **Path Resolution:** Use the shared helper `resolveGitRoot(process.cwd())` to locate the workspace root. Compute the self agent ID (via `TOTEM_AGENT_ID` env var or falling back to local config).
+3. **Draft Assembly:**
+   - Extract the current seq (e.g., `claude-00NN`) via `readJsonSafe` against the active session state.
+   - Construct the frontmatter using `Adr098FrontmatterSchema`.
+   - Read body from `--body-file`, or fallback to empty string (or `stdin` if available).
+4. **Actuation:** Write the templated markdown to `.totem/orchestration/<self>/outbox/<UTC>Z-<recipient>-<slug>.md`.
+
+### Edge Cases & Traps
+
+- **Silent Fall-through Trap:** The most critical regression to avoid is `totem mail send` acting as a no-op read. The CLI router _must_ throw an error for unknown subcommands or positional arguments.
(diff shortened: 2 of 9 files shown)


satur8d:
/gemini review

@greptileai review

satur8d:
## R1 disposition (`9fb238f1`)

Thanks — the path-traversal catch was a real miss (I guarded `--from` but not `--to`), flagged independently by all three bots. Fixed + hardened:

| Finding | Class | Action |
| --- | --- | --- |
| `--to` path-traversal (interpolated into filename) — GCA critical / Greptile P1 / CR critical | accept | `assertSafeAgentId(to)` (blocks `/ \ .. \0`) + filename-safe token + YAML-quote `from`/`to` |
| `fs.mkdirSync` unwrapped — GCA high | accept | wrapped → `TotemError` (fail-loud with path) |
| `AGENT_ID_UNSAFE` copy-pasted from `orchestration-resolver` — Greptile P2 | accept | exported `isPathSafeAgentId` from `@mmnto/totem`, reused (single source) |
| Missing `--to` traversal test — Greptile P1 | accept | added; asserts the guard fires before any mkdir/write, so nothing escapes |
| Static top-level `@mmnto/totem` import — GCA high | decline-substantive | `mail.ts` is already lazy-loaded by `index.ts` (`await import('./commands/mail.js')`), so the import only runs when the command runs, not at startup — per-function dynamic imports would add noise with no latency benefit |

61 tests green; `totem lint --branch` 0 errors; full CI green pre-fix.

/gemini review

@greptileai review

satur8d:
## R2 disposition (`e3def7be`)

Both new findings verified real and accepted; Greptile's 5/5 re-review confirmed all R1 fixes with no new findings.

| Finding | Class | Action |
| --- | --- | --- |
| GCA HIGH — cleanup `rmSync` in the atomic-write catch can throw (EPERM/EBUSY past retries) and shadow the original write error | accept | Flattened to rethrow-from-both-paths: a cleanup failure now rethrows the **original** actuation error as `cause`, with the stranded-temp note folded into the message — both failures visible, neither shadowed. No-mock regression test (plants a non-empty directory at `<filePath>.tmp` so the write *and* the cleanup fail deterministically). |
| CR Major — `isPathSafeAgentId` accepts control/whitespace/win32-reserved characters | accept | Core `AGENT_ID_UNSAFE_CHAR_PATTERN` (`\p{Cc}` controls, `\s`, `<>:"\|?*`) folded into the guard — it is now a full path-segment contract, not only a traversal check. `assertSafeAgentId` message broadened. Unit tests for all three character classes + a cohort-map self-consistency lock (every `COHORT_AGENT_MAP` id must pass the guard). |

Note on Greptile's two non-blocking observations: the charset hardening narrows the YAML round-trip gap (non-standard agent-ids are now rejected at the validation gate before composition); the `uniqueOutboxPath` TOCTOU remains theoretical for the single-writer-per-outbox cohort and is noted for if burst-parallel sends become a use-case.

Gate at `e3def7be`: core 2192 + cli 2491 tests green · `totem lint --branch` 0 errors · repo-wide `format:check` clean · `totem review` PASS.

/gemini review

satur8d:
## R3 disposition (`d1f7b4a7`)

Five accepts (one as a cohort-convention variant), one sub-point declined with verification.

| Finding | Class | Action |
| --- | --- | --- |
| GCA HIGH — `yamlScalar` doesn't quote boolean/null/numeric-shaped plain scalars (YAML coercion) | accept | Added the coercion-trap patterns, widened beyond the suggestion to YAML 1.1's bare `y`/`n` and exponential/trailing-dot floats (review-gate INFO fold). Tests cover all shapes + plain prose staying unquoted. |
| GCA HIGH (sub-point) — "escape the hyphen in the character class" | decline | A trailing `-` in a character class is literal by regex semantics — no range can form; the proposed snippet also contains a syntax error (`/[\n"].test(value)/` is missing its closing `/`). No change needed; verified against the live regex. |
| GCA HIGH — `inReplyTo: source` leaks absolute local paths into shared frontmatter | accept (variant) | Normalized to the **repo-relative wire form** `.totem/orchestration/<agent>/outbox/<file>` — the de-facto cohort shape for `in-reply-to:` (see live dispatches) — rather than bare basename, which would drop the repo/agent context cross-repo. Basename remains the fallback for sources outside an orchestration tree. |
| CR Major — rejected agent ids echoed raw in `assertSafeAgentId`'s error | accept | `JSON.stringify(id)` in the message — the rejection path no longer re-creates the terminal injection it blocks. Regression test asserts no raw ESC in the thrown error, only its escaped spelling. |
| CR Major — `parseEnvAgentList` / `host_agents` still filter only traversal | accept | Both layers now filter via `isPathSafeAgentId` — self-agent resolution enforces the same path-segment contract as the actuator's recipient validation. Doc comment updated; tests for both layers. |
| CR Major (outside diff) — `mailReply` hard-fails on `from:`-less sources the reader accepts | accept | Reader-parity fallback: the sender is derived from the `<agent>/outbox/<file>` layout before rejecting, mirroring `pollMail`'s outbox-dir fallback. Test with a hand-authored legacy dispatch. |
| CR Minor (outside diff) — quoted subjects don't round-trip through the reader | accept | `parseHeader` unquotes subjects matching `yamlScalar`'s **exact JSON-string output shape** (strict-shape precheck makes `JSON.parse` infallible — no catch). Hand-authored/asymmetric quoting reads verbatim, so legacy mail display is unchanged. Round-trip test: compose → poll surfaces the subject as typed. |

Gate at `d1f7b4a7`: core 2194 + cli 2495 tests green · `totem lint --branch` 0 errors · repo-wide `format:check` clean · `totem review` PASS (its one INFO — wider YAML 1.1 coercion coverage — folded into this round).

/gemini review

satur8d:
@greptileai review

satur8d:
## R4 disposition (`103e7496`)

| Finding | Class | Action |
| --- | --- | --- |
| CR Major — the R3 subject-unquote decodes `\n` / `\uXXXX` escapes into raw control bytes that `formatTextResult` writes to stderr | accept (variant) | Stronger than decode-then-sanitize: the shape regex (`PRINTABLE_JSON_STRING_SCALAR`) now admits **only escapes that decode to printables** (`\"` `\\` `\/`). A quoted subject encoding control bytes surfaces in its escaped spelling — lossless display, injection-free **by construction**, and no sanitize pass that silently mutates the value. `yamlScalar` only emits `\n`/`\uXXXX` escapes when the composed subject itself carried control bytes, so the printable round-trip is unaffected. Regression tests: escaped-ESC and escaped-newline wire subjects surface verbatim (still quoted), never decoded. |

Gate at `103e7496`: cli 2496 tests green (core 2194 unchanged) · `totem lint --branch` 0 errors · `format:check` clean · `totem review` PASS.

satur8d:
## R5 disposition (`ea318331`)

| Finding | Class | Action |
| --- | --- | --- |
| CR Major — the unquote fallback returns raw control bytes verbatim into `MailEntry.subject` | accept (widened) | Verification showed the exposure is wider than the fallback — and **pre-existing**: an *unquoted* `subject:` (or `from:`/`date:`) carrying raw bytes has flowed to stderr verbatim since the reader shipped (#1970); the proposed fallback-only fix would have left those paths open. Closed the whole class at the single parse boundary instead: `escapeControlBytes` (each raw control byte → its JSON-escaped spelling — display-safe, lossless, visible) applied to all four sender-controlled fields at `parseHeader`'s return. Tests cover raw bytes via the quoted, unquoted, and `from:` paths. |

The injection surface is now covered from three directions: agent-id charset guard at the actuator (R2), printable-only escape decoding (R4), and raw-byte escaping at the reader boundary (R5).

Gate at `ea318331`: cli 2497 tests green (core 2194 unchanged) · `totem lint --branch` 0 errors · `format:check` clean · `totem review` PASS.

@coderabbitai

Issue #2018, opened 2026-05-23
totem search blocked locally on missing @google/genai dependency


satur8d:
## Context

\`strategy-codex\` (newly-onboarded OpenAI / Codex CLI vendor-class agent) attempted to use the cohort's knowledge-query gate during first-session onboarding via \`totem search "..."\`. Both invocations failed because the local Gemini SDK package \`@google/genai\` is not installed in the cohort dev environment.

This is NOT Codex-specific — any agent or human running \`totem search\` from a fresh clone of mmnto-ai/totem-strategy hits the same gap. strategy-codex surfaced it because their MCP \`search_knowledge\` tool wasn't wired in their first session, so they fell back to the CLI surface.

## Question for triage

Is this:
- A missing \`pnpm install\` step in cohort dev-env onboarding docs?
- A genuine missing-dep declaration in package.json (\`@google/genai\` should be in \`dependencies\` but isn't)?
- A graceful-degradation gap (CLI should warn "embedding generation unavailable; install \`@google/genai\` to enable" rather than fail)?
- Some combination?

If \`@google/genai\` is expected to be present in cohort dev envs, the onboarding flow / \`pnpm install\` chain should pull it. If it's optional (e.g., the embedding generation is one search path among several), the CLI should degrade gracefully.

## Impact

Blocks the canonical knowledge-query discipline ([feedback_search_totem_first_for_doctrine] memory: \`search_knowledge\` BEFORE grep+memory) for any agent without a wired MCP surface or any user running from a fresh clone.

## Source

- strategy-codex codex-0001 journal: \`mmnto-ai/totem-strategy:.totem/orchestration/strategy-codex/journal/codex-0001-onboarding-and-blind-tenet-20-critique.md\` § Knowledge-query gate status
- Cohort dev-env presumed-to-work surface

satur8d:
**Disposition update** — `#2018` is in active disposition. Lane: **option (c) graceful-degradation with keyword fallback in `totem search`** (preserve current optional-peer dep posture; add command-layer escape hatch above `LazyEmbedder`'s existing Ollama-fallback terminator). Strategy Proposal **#286** drafted in `mmnto-ai/totem-strategy:proposals/active/286-totem-search-graceful-degradation.md`; will close on its merits before the W5 (`@google/genai` 1.x → 2.x) bump ships.

**Why not (a) or (b):**
- (a) onboarding-docs gap — wallpaper; doesn't fix the actual `TotemConfigError` UX failure.
[...]

PR #2150, opened 2026-06-11
fix(core): context-correct remediation for missing externalized LLM SDKs (#2018 L2)

satur8d:
Part of mmnto-ai/totem#2018 (layer 2 of the layered plan on the issue; L1 prefer-local re-exec follows separately).

## Problem

The SDK-import catch sites already failed loud, but with remediation that is wrong for the contexts that keep biting (6 recurrences across 2 agent seats): `pnpm add <sdk>` does not help when the SDK IS installed and the running binary (a global install) cannot resolve it — and `pnpm add -g` is a verified dead end (the 2026-05-30 empirical comment on #2018).

## Change

New `buildMissingSdkHint()` in core, applied at all four SDK import sites (core gemini embedder + gemini/anthropic/openai orchestrators). It branches on what is actually true on disk:

1. **SDK installed project-locally but unresolvable from the binary** → names the global-install cause and points at `pnpm exec totem`; never re-suggests installing.
2. **totem workspace checkout** → points at the workspace build (`node packages/cli/dist/index.js`).
3. **Genuinely absent** → project-local install hint with the externalized-by-design context (#2018 / `google-genai-coupling` parity contract) and an explicit warning away from global installs.

No branch suggests a global install (locked by test). Headline text is unchanged so `isFallbackEligible` keeps matching for the CLI-fallback path. Patch changeset for `@mmnto/totem` + `@mmnto/cli`.

## Verification

TDD (helper test RED first), 6 helper tests green, full core suite 2244 green, full-branch `totem lint` PASS 0 errors. Two lint-round commits document real gate catches: bare `#NNN` refs repo-qualified, and the workspace probe rewritten to a no-catch content regex after the fail-open-catch rule (correctly) rejected `JSON.parse` + catch in a detection probe.

### DIFF
--- a/.changeset/missing-sdk-remediation.md
+++ b/.changeset/missing-sdk-remediation.md
@@ -1,0 +1,6 @@
+---
+'@mmnto/totem': patch
+'@mmnto/cli': patch
+---
+
+Context-correct remediation for missing externalized LLM SDKs (mmnto-ai/totem#2018 L2). When `@google/genai` / `@anthropic-ai/sdk` / `openai` fail to import, the error now branches on what is actually true on disk: SDK installed but unresolvable from the running binary → points at the project-local CLI (`pnpm exec totem`) and names the global-install cause; totem workspace checkout → points at the workspace build; genuinely missing → project-local install hint with the externalized-by-design context. No branch suggests a global install — verified on #2018 to be a dead end.
--- a/packages/cli/src/orchestrators/anthropic-orchestrator.ts
+++ b/packages/cli/src/orchestrators/anthropic-orchestrator.ts
@@ -1,4 +1,4 @@
-import { TotemConfigError, TotemOrchestratorError } from '@mmnto/totem';
+import { buildMissingSdkHint, TotemConfigError, TotemOrchestratorError } from '@mmnto/totem';
 
 import { log } from '../ui.js';
 import type { OrchestratorInvokeOptions, OrchestratorResult } from './orchestrator.js';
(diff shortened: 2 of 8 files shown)


satur8d:
@coderabbitai review
@greptileai review

satur8d:
## Round 1 dispositions — all 4 findings fixed in c2252c91

| # | Finding | Source | Disposition |
|---|---|---|---|
| 1 | Branch priority inversion: local-install hint beats the workspace hint inside the monorepo (where the SDK can sit in `node_modules` as a sibling devDependency) | Greptile P1 | **Fixed** — workspace-root probe now runs first; rationale documented in the branch docblock |
| 2 | Core hint hardcodes `pnpm` (no access to the CLI detector) | Greptile P2 | **Fixed** — `packageManager` now defaults from `npm_config_user_agent` in core, mirroring the CLI detector; explicit option still wins |
| 3 | Bare-branch test tmpdir leaked | Greptile P2 | **Fixed** — created, used, removed |
| 4 | Missing both-conditions test that would have caught #1 | Greptile P2 | **Fixed** — new test: workspace + SDK-in-node_modules → workspace hint wins, `exec totem` absent |

Verification: 8/8 helper tests, full core suite 2246 green, full-branch `totem lint` PASS 0 errors.

@greptileai review
@coderabbitai review

satur8d:
@coderabbitai review

satur8d:
## Round 2 dispositions (CR review, 2 findings)

| # | Finding | Disposition |
|---|---|---|
| 1 | OpenAI-fallback guidance appended on non-missing-SDK branches (anchored at the prior head c2252c91) | **Fixed in fb1db8fb** — which crossed this review mid-flight: the alternative is now an `absentAlternative` option the helper appends only on the genuinely-absent branch (no string prefix sniffing), branch-scoping locked by test in all three branches. The fix is also stronger than the suggested patch for a verified reason: the openai embedder statically imports its own externalized SDK, so the alternative is not merely noisy in the wrong-binary branch — it hits the identical resolution wall. |
| 2 | `os.tmpdir()` in test fixtures vs the workspace-temp guideline | **Declined** — that guideline targets agent-readable runtime artifacts, not vitest fixtures: `os.tmpdir()` + `mkdtempSync` is this repo's dominant test-fixture convention (53 occurrences across 15+ test files, incl. `compiler.test.ts`, `utils.test.ts`, `storage.test.ts`). Workspace-local test temp would also risk the probes in THIS test walking up into the real repo. |

Verification at fb1db8fb: 24/24 affected-suite tests, build green, full-branch `totem lint` PASS 0 errors.

satur8d:
@coderabbitai review

PR #2153, opened 2026-06-11
feat(cli): prefer-local re-exec at the entrypoint (#2018 L1)

satur8d:
Closes #2018 — layer 1 of the layered plan on the issue (L2 shipped in #2150 / 1.58.1; L3 is the `knowledge-search-access` usable-rung probe riding the #2140 family, no work owed here).

## What

When a foreign totem binary — typically an ambient global install — starts inside a project that carries its own `@mmnto/cli`, the entrypoint now delegates to the project-local build before any command wiring. This promotes the deterministic tiers of the ADR-072 resolve cascade (shipped for git hooks in #2053) into the binary itself, per Tenet 14: the pinned local install is deterministic, the PATH global is ambient. It forecloses both variants of the recurring wrong-binary class — missing externalized peer SDKs (#2018, 6 recurrences across 2 agent seats) and stale-version shadowing (#2053) — by making the wrong binary unreachable from inside a workspace.

- **Resolution** (`resolveLocalEntry`): walk up from cwd — tier 1 `packages/cli/dist/index.js` identity-guarded on the package name (workspace-HEAD), tier 2 `node_modules/@mmnto/cli/dist/index.js` (pinned). Both require the BUILT entry; an unbuilt checkout runs in place, where the L2 hint explains the build step.
- **Delegation** (`maybeReexecLocal`): realpath identity short-circuit (already-local runs in place); `TOTEM_NO_REEXEC=1` opts out AND rides the child env as the loop guard; announced on stderr with both versions, never silent; child exit code propagated, null status maps to failure. Spawn via `cross-spawn` (the primitive `safeExec` wraps — `safeExec` itself is pipe-buffered and throws on non-zero, which delegation cannot use), target `process.execPath` + the guarded entry, argv as array, no shell. Allowlisted in the pack-agent-security repo sweep with justification, mirroring the `exec.ts` precedent.
- Docs: installation.md gains the delegation + opt-out subsection. Minor changeset.

## Verification

[...]

PR #2155, opened 2026-06-11
fix(cli): re-exec success path crashed on cross-spawn error: null (1.59.0 hotfix)

satur8d:
## Problem — 1.59.0 regression, caught on the first live delegation

`cross-spawn` fills `error: null` (not `undefined`) on **success**. The #2153 spawn-failure check used `child.error !== undefined`, so every SUCCESSFUL delegation completed the child's work and then crashed the parent's exit path with a raw `TypeError: Cannot read properties of null` — wrong exit code on every delegated command under 1.59.0.

Found minutes after the 1.59.0 global update, on the very first real delegation in the dogfood repo. The unit mocks omitted the `error` property (undefined) instead of mirroring cross-spawn's real `null` — a mock-fidelity gap, now closed.

## Change

- `child.error != null` (loose) — handles both absent and null.
- Two regression tests: cross-spawn's real success shape (`{ status: 0, signal: null, error: null }`) must return the child status cleanly, and a genuine spawn error must report `Delegation failed to start: <msg>` on stderr after the announce.
- Patch changeset → 1.59.1.

## Verification

17/17 module tests, full-branch `totem lint` PASS 0 errors, full pre-push gate green. The live global-binary verification completes once 1.59.1 publishes and the global updates (the crash lives in the released parent binary, not the workspace).

### DIFF
--- a/.changeset/reexec-error-null-hotfix.md
+++ b/.changeset/reexec-error-null-hotfix.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/cli': patch
+---
+
+Hotfix: 1.59.0's prefer-local re-exec crashed the parent's exit path after every SUCCESSFUL delegation — cross-spawn fills `error: null` (not `undefined`) on success, and the spawn-failure check read `.message` off the null. The child's work completed but the invoking process exited 1 with a raw TypeError. Caught on the first live delegation; the success shape is now regression-locked in tests.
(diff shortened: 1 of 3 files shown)


satur8d:
@coderabbitai review

satur8d:
@greptileai review

Issue #584, opened 2026-03-16
bug: fileGlobs matching files outside the specified directories


satur8d:
## Summary

A compiled rule with \`fileGlobs: ["packages/cli/src/adapters/**/*.ts", "packages/mcp/src/**/*.ts"]\` incorrectly matches files in \`packages/core/src/**/*.ts\`. The glob should only match files within the specified directories.

## Reproduction

Rule \`064f35405ec02082\` (match/exec shadowing) has fileGlobs scoped to cli/adapters and mcp/src. But it flagged \`packages/core/src/cursor-adapter.ts\` during lint.

## Expected behavior

\`packages/core/src/cursor-adapter.ts\` should NOT match \`packages/cli/src/adapters/**/*.ts\`.

## Context

Hit in PR #583. Worked around by replacing \`.exec()\` with string operations. The \`matchesGlob()\` function in \`packages/core/src/compiler.ts\` needs investigation.

satur8d:
## Root Cause

Not a glob matching bug. `matchesGlob()` works correctly. The issue was **duplicate compiled rules** — `totem compile` (via `totem wrap`) generated a second rule with the same match/exec pattern but broader fileGlobs (`**/*.ts` instead of the scoped `packages/cli/src/adapters/**/*.ts`).

Two rules existed:
- `064f3540` (scoped correctly to cli/adapters + mcp) 
- `b2d2d790` (broad `**/*.ts` — catches everything)

The broad duplicate was generated from a different lesson with a slightly different regex. Fix: remove the duplicate. Real fix: add dedup detection to `totem compile` (#569).

PR #2162, opened 2026-06-12
feat(core): Git-Bash resolver — bare bash is never spawned on win32 (#2159)

### LINKED ISSUE #2159: tooling: bare `bash` resolves to WSL on Windows — export ONE Git-Bash resolver, consume it in bash-invoking tests/tooling
## Symptom (live, 2026-06-11/12, surfaced during the #2157 round)

Bare `bash` on Windows hosts can resolve to `C:\Windows\System32\bash.exe` (WSL Linux bash) — Git-for-Windows puts only `Git\cmd` in PATH, not `usr\bin`. WSL bash cannot read `D:\...` Windows paths, so any repo surface spawning bare `bash` from a plain (non-git-hook, non-MSYS) context fails with `No such file or directory`:

- `packages/cli/src/commands/pre-compact-hook.test.ts` — 5/5 fail under plain PowerShell-spawned vitest
- `packages/cli/src/commands/shield.content-hash-parity.test.ts` — 5/5 fail likewise
- Both pass inside git-hook contexts (Git prepends its own `usr\bin`) — so the push gate is usually green while direct test runs are red, which is exactly the kind of context-dependent gate divergence that erodes trust in local PASS (#2055 cluster adjacency).

## Cohort ruling (operator, 2026-06-12; both seats concurred in the 0141Z/0240Z dispatch exchange)

**Repo-side mechanization is the cohort standard.** Per strategy-claude sharpenings:

1. **ONE exported resolver** in `@mmnto/totem` (core) — derive Git-Bash from git''s install root (`git --exec-path` walk-up, probe-fallback to the conventional install path), never N hand-rolled variants.
2. **Contract: bare `bash` is NEVER spawned by repo tooling on win32** — every bash-invoking surface (the two test files; any future spawn) consumes the resolver.
3. Zero new parity dimensions: the fix ships inside repo tooling; nothing per-seat needs sensing. The doctrine home for the Windows-shell-trap class (this + the MSYS leading-slash munge) lands strategy-side.

Host-level alternatives were considered and rejected as the standard: per-seat settings-env (ergonomics-only, word-gated; `${PATH}` expansion unverified) and machine-PATH prepend (shadows Windows `find`/`sort`/`ssh` machine-wide).

## Refs

mmnto-ai/totem#2157 (where the failure surfaced) · dispatch thread `2026-06-12T0141Z-strategy-claude-bash-resolution-cohort-sync` → `0240Z` concur · sequenced after the #2141+#2144 lane

satur8d:
## What

**Closes #2159.** The cohort-standard fix for the bare-`bash`-is-WSL trap (operator-ruled 2026-06-12, both seats concurred: repo-side mechanization, ONE exported resolver).

- **`resolveBash()`** (core): POSIX returns `''bash''`; win32 derives Git''s install root from `git --exec-path` (probing `usr/bin/bash.exe` — the real MSYS bash — before the `bin/` wrapper), falls back to the conventional install path, and a total miss **throws `BASH_RESOLUTION_FAILED` naming every probed path**. The generated spec''s final bare-`''bash''` fallback was rejected: it silently re-enters the trap (Tenet 4). Module-level memo — `git --exec-path` runs at most once per process.
- **`bashSpawnEnv(base?)`** (core) — **the trap''s second layer, discovered at build time:** a directly-spawned Git-Bash inherits the parent''s PATH (which lacks `usr\bin` — the whole #2159 class), so the bash binary runs but every coreutil inside the script (`grep`, `tr`, `sha256sum`, `cut`) is `command not found`. Git-hook contexts never see this because git prepends its own tree; this export reproduces that contract for plain spawns (PATH-key casing preserved — no second spelling beside an inherited `Path`).
- Both bash-spawning suites (`pre-compact-hook`, `shield.content-hash-parity`) consume the pair at all three call sites.

## Empirical anchor

Before: 10-of-14 tests in those suites failed under plain PowerShell (`bash` → WSL → `No such file or directory` on `D:\` paths); with `resolveBash()` alone, 9-of-14 still failed on `command not found`. After the pair: **full cli suite — 2557 tests, zero exclusions — green in that exact context for the first time**, and 12 new resolver units cover every branch (POSIX fast-path, derivation, usr-before-bin preference, memo, conventional fallback, hard failure, env composition, casing, absent-PATH).

Design: `.totem/specs/2159.md` (committed), including the second-layer discovery note. Changeset: minor `@mmnto/totem` (cli changes are test-only). Reviews: codex round skipped per the right-sizing call (operator-approved); the contract itself was shaped by both seats in the bash-resolution dispatch exchange.


satur8d:
/gemini review
@coderabbitai review

satur8d:
## Review round 1 — dispositions (fixes at 6055a2bd)

@gemini-code-assist — your two findings:

1. **exec-path validation: FIXED.** Anomalous `git --exec-path` output (empty/relative) is now gated behind a guard condition and falls to the conventional probes — your catch was real: `path.resolve('''')` would have derived the Git root from the CWD. Regression test added (empty + relative cases).
2. **''anchor'' in the spec: DECLINED — scope.** Styleguide § 2 bans metaphor-heavy NAMES in shipped code (its DO-USE list is entirely identifiers); "Empirical anchor" is prose in an internal spec record, not a name. § 6 now carries a do-not-suggest entry for this extension. The § 6-mandated `add_lesson` is deferred behind #584 sub-task 4: a lesson write currently wedges the verify-manifest push gate under the cohort rule-compilation freeze.

CodeRabbit findings (from the chat-mode reply):

3. **Missing `.trim()`: DECLINED as a bug** — `safeExec` auto-trims by default (documented contract, `sys/exec.ts`), so the trailing-newline class cannot reach `path.resolve` on any machine. An intent comment now cites the contract at the call site so this is never re-raised. (Your own hedge was correct.)
4. **`_clearBashResolverCacheForTesting` exposure: FIXED** — `@internal` JSDoc tag added.
5. **Dead `cached === null` branch: FIXED properly** — `resolveCached()` now returns the non-null memo to both public surfaces; the narrowing branch is gone rather than commented around.

Round note: the combined trigger comment routed CR to chat-mode again — third data point for the embedded-trigger class (mmnto-ai/totem#2150); findings triaged from the chat reply regardless. Tests 13/13 resolver + 14/14 migrated suites; lint 0 violations; full gate green on push.

Issue #2106, opened 2026-06-06
Tracking: grounded spec/review redesign — Phase-1 implementation (strategy#474 decomposition)


satur8d:
## Architecture record (canonical)

[`mmnto-ai/totem-strategy#474`](https://github.com/mmnto-ai/totem-strategy/issues/474) — the 2026-06-06 BLIND-round synthesis comment (3-lane convergence + operator-ruled A-core-first positioning). That comment is the architecture record; **this tracker is the implementation surface.** The strategy umbrella stays open only as the record.

**Thesis being implemented:** `totem spec` and `totem review` both fabricate because synthesis is ungrounded in derived code state. Fix = grounding FIRST: `spec` writes a grounded contract (with explicit `VERIFY:` abstention), `review` senses conformance against it (spec doc *when present*, compiled rules + doctrine *always*) — write/read halves of ONE grounded artifact. Defensible claim: *grounded input + deterministic post-checks + explicit abstention* (never "deterministic LLM").

## The slices (build order)

| # | Issue | Slice | Depends on |
|---|---|---|---|
| 1 | #2100 | Grounded single-run artifact + rerun/compare | — (first slice, ruled) |
| 2 | #2101 | Grounding bundle with day-one provenance classes | #2100 · graduation: #344, #375 |
| 3 | #2102 | Backend admission contract (additive enrichment, 2 classes) | #2100, #2101 |
| 4 | #2103 | Deterministic structural post-checks (decidable tiers) | #2100–#2102 |
| 5 | #2104 | Panel synthesis (independent lanes, script aggregation) | #2100, #2102 |
| 6 | #2105 | Override-memory (Phase-2 boundary fields) | #2100 · related #2038 |

## Standing constraints (falsification pass, 2026-06-06T2059Z dispatch — carried as design inputs)

1. **Provenance classes in the bundle schema from the first cut** — a similarity-only bundle must self-describe; the illusion-of-grounding trap cannot be retrofitted away.
2. **Post-checks caller-side, never provider-side; all contract enrichment additive-with-defaults** — `runOrchestrator` (`utils.ts:411`) has 10+ callers and `shell` is a user-facing extension point; #1291 is the migration precedent. spec/review migrate first, every other caller compiles untouched.
3. **Decidable enforcement tiers** — citation-*resolves* may gate; citation-*supports* is sensor/eval only (ADR-109 lesson).

## Adopted cuts (do not build)

[...]

Issue #2103, opened 2026-06-06
feat(review): deterministic structural post-checks with decidable enforcement tiers (474 slice 4)


satur8d:
## Architecture record

[`mmnto-ai/totem-strategy#474`](https://github.com/mmnto-ai/totem-strategy/issues/474) synthesis, items 5+6: falsifying metrics are **fabrication vs deterministic post-checks**, per admission class; the defensible product claim is *grounded input + deterministic post-checks + explicit abstention* (bit-determinism claims cut).

## Scope

Caller-side, script-shaped (zero-LLM) checks over a run''s output, consuming the run artifact (#2100) + grounding bundle (#2101):

**Per-class falsifying metrics (from the synthesis, adopted verbatim):**
- `completion_only` fails by **contradicting delivered evidence**.
- `self_grounding_agent` fails by **any concrete claim lacking a cited read/query transcript**.
- `spec` fails by **an unverifiable path/API/schema name without `VERIFY:`**.
- `review` fails by **an unchanged-anchored-span override reappearing** (override-memory issue owns the memory; this issue owns the check).

**Decidable-tier split (falsification-pass constraint, 2026-06-06T2059Z dispatch — explicit in the design, not implied):**
- *Decidable → may gate:* citation **resolves** (file exists, line in range, symbol present), `VERIFY:` marker present where the claim class requires it, output parses against `outputContract`.
- *Not decidable → sensor/eval only, never a gate:* citation **supports** the claim, summary faithfulness. Blurring this rebuilds noisy-gates-train-bypass (ADR-109''s lesson).

**Verdict aggregation = script, not LLM** (Tenet 9, adopted in the round).

## Eval fixtures

Seed spec-side fixtures from the ready-made 2026-06-06 exhibit: `totem spec 2090`/`2091` invented `packages/cli/src/utils/diff-selector.ts` (nonexistent) + mislocated the seam — ground truth already committed in `.totem/specs/2090.md`/`2091.md` (PR #2098). Plus upstream-feedback/040+041+043 per the synthesis.

## Not in scope

The grounding bundle itself, admission gating, any LLM-judge step.

Part of the #474 decomposition. Depends on: #2100, #2101, admission contract (#2102) for `outputContract`.

Issue #474, opened 2026-03-14
docs: Update README to highlight the Air-Gapped Doctrine (Zero Telemetry)


satur8d:
## Summary

Update the project `README.md` to explicitly market the "Air-Gapped Doctrine" (Proposal 013), highlighting Totem's viability in strict, zero-telemetry enterprise environments.

## Context

Totem's core architecture—local vector storage via LanceDB, pluggable local models, and strict zero-telemetry boundaries—makes it a perfect fit for high-compliance enterprise sectors (defense, finance) that operate in strict AI sandboxes. However, we haven't clearly communicated this competitive moat to our users.

## Requirements

Rewrite or add a section to the `README.md` that explicitly outlines the "Air-Gap Ready" architecture:
- Emphasize that zero user codebase telemetry is sent to centralized servers.
- Highlight the ability to run `totem sync` and `totem shield` entirely locally via Ollama or vLLM.
- Reinforce that the Totem MCP Server is a strict, read-only/append-only context provider that does not expose risky filesystem execution tools to external agents.

## Acceptance Criteria
- [ ] `README.md` contains an "Air-Gap Ready" or "Zero Telemetry" feature block.

satur8d:
Contradicts ADR-047 'Audit-Native Hybrid' pitch. We are no longer leading with the Air-Gapped Doctrine.

Issue #2100, opened 2026-06-06
feat(orchestrator): grounded single-run artifact + rerun/compare primitives (474 slice 1)


satur8d:
## Architecture record

[`mmnto-ai/totem-strategy#474`](https://github.com/mmnto-ai/totem-strategy/issues/474) — 2026-06-06 BLIND-round synthesis (canonical). This is the **first buildable slice**, ruled explicitly: *"the first buildable slice is not pluggable agents; it is a grounded single-run artifact that completion-only and self-grounding backends can both consume/produce honestly."*

## Scope

An immutable, content-addressed record emitted per orchestrator run (spec + review first):

- **Input-bundle hash** — the assembled prompt inputs (diff scope, retrieved context, spec contract when present). The new `lint/review --branch` forced scope (#2098) gives a deterministic diff input to hash.
- **Grounding-bundle hash + per-item provenance summary** (see the grounding-bundle issue — schema dependency, not a build dependency; slice 1 may hash the current ad-hoc context assembly as `similarity-only` wholesale).
- **Backend identity + admission class** (`completion_only` | `self_grounding_agent`), task profile, model/provider, temperature.
- **Output + parse result + metrics** (tokens, duration, finish reason — `OrchestratorResult` already carries these).
- **Rerun = same immutable bundle** (no silent context drift); **compare = artifact-vs-artifact diff**. Same-vendor subagent panels labeled honestly: context-isolated ≠ independent-rater diversity.

## Constraints

- Storage under `.totem/` (exact layout = impl call); artifact is append-only — a rerun NEVER mutates a prior record.
- Artifact must record what KIND of grounding the run had (provenance summary) — day-one field, can't be retrofitted honestly.
- Everything downstream (post-checks, panel synthesis, the Phase-2 disposition ledger) consumes this artifact; schema changes after Phase 2 starts are expensive — keep it minimal but versioned (`schemaVersion` field).

## Not in scope

Pluggable backend admission (separate issue), structural post-checks (separate issue), any ledger semantics (Phase 2, strategy-gated revisit).

Part of the #474 decomposition (tracker to follow). Sequenced first; no dependencies.

PR #2172, opened 2026-06-14
feat(spec,review): code-blind grounding guard — banner + suppress unverifiable claims on 0 code (#2106)

satur8d:
## What

Interim fail-loud **code-blind grounding guard** for `totem spec` / `totem review` (`shield.ts`). Part of #2106 (strategy#474 Phase-1); operator-greenlit pull-forward, design strategy-approved.

When retrieval returns **zero code chunks**, the LLM has no grounding to verify file/type/system claims against — the lc#463 class where the tool invented a whole architecture. Per the strategy#474 interim ruling the guard **degrades, it does not disable**:

- `applyCodeBlindGuard` + `isCodeBlind` (`packages/cli/src/utils.ts`): pure, total helpers keyed strictly on `context.code.length === 0`. Surface an advisory-neutral banner and fold a suppression directive into the system prompt so the model stays at the altitude the retrieved specs/sessions/lessons support.
- Wired at both call sites — `spec.ts` and `shield.ts` — one shared seam, identical trigger semantics.

The banner is the deterministic, code-emitted signal; the prompt directive is best-effort. Hard structural post-checks remain the #474 redesign (#2103).

## Tests

`utils.test.ts` locks the invariants: fires iff 0 code; returns `${systemPrompt}\n\n${directive}` when blind and unchanged otherwise; does not throw or disable (the anti-abort property); banner is advisory-neutral. 5129 tests pass; `totem lint` PASS (0 errors).

## Provenance note (live exhibit)

The first auto-generated `totem spec` draft of this feature's own spec confabulated an _abort_ design (`EmptyGroundingError`) and cited a `review.ts` that does not exist — exactly the failure this guard prevents. The committed spec carries the corrected, code-verified design.

## Design

`.totem/specs/fail-loud-guard-for-spec-and-review-on-empty-grounding.md`. Open questions Q1–Q3 resolved (strategy-claude concurred 2026-06-13).

Part of #2106.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Bot round — dispositions (commit `9fd83a3d`)

**@gemini-code-assist** — re: the `applyCodeBlindGuard` defensive-typing finding (`utils.ts`, high): **respectfully declining**, with reasoning.

- The helper's inputs are non-null / non-empty **by construction** at both call sites: `retrieveContext` always returns a populated `RetrievedContext`, and `getSystemPrompt` always returns a string. TypeScript already forbids passing `undefined` (the param type is `string`).
- The guard never _produces_ an empty system prompt — it only appends a directive to, or passes through, a non-empty one. It cannot introduce the empty-prompt → provider-error condition.
- "Treat empty system prompts as undefined for LLM providers" is an **orchestrator-layer** concern (`runOrchestrator`), pre-existing and untouched by this PR. Widening to `| null | undefined` here would be defensive code for states TypeScript already rules out (against the repo's "Zod at boundaries only" / no-speculative-robustness grain), and I won't add a SAFETY-INVARIANT comment for a behavior the helper does not implement.
- If empty-prompt hardening at the orchestrator is genuinely wanted, that is a separate ticket — happy to file it; it is out of scope for this guard.

**Greptile P2** (`codeBlind` absent from `runMetadata`): **adopted** in `9fd83a3d`. `RunMetadataSchema` gains an optional `codeBlind` boolean and both call sites (`spec.ts` / `shield.ts`) now forward `codeBlindGuard.codeBlind`, so the run-artifact eval fixtures (#2100) are filterable by guard activation without recomputing. Good call.

CodeRabbit: no findings. CI green — 5130 tests, `totem lint` 0 errors.

satur8d:
@greptileai review

Issue #624, opened 2026-03-16
feat: Native DeepSeek Orchestrator Support


satur8d:
## Description
Implement native support for DeepSeek models in the Totem orchestrator matrix.

## The Problem
DeepSeek has captured a massive share of the cost-conscious and open-weights developer market. Currently, users must configure complex OpenAI-compatible proxies to use DeepSeek with Totem commands like \	otem extract\ or \	otem spec\.

## Scope
- Implement \deepseek-orchestrator.ts\ conforming to the standard Orchestrator interface.
- Add \deepseek\ to the supported provider list in \config-schema.ts\.
- Update \	otem init\ to detect \DEEPSEEK_API_KEY\ and offer it as a default orchestrator option if present.

## Strategic Context
Aligns with our goal of maximum developer reach for the 1.0 launch. Proves Totem is a modern, adaptable toolchain.

PR #2174, opened 2026-06-14
fix(sync): self-heal orphaned chunks via working-tree reconciliation (#2151)

### LINKED ISSUE #2151: sync: incremental baseline can advance past file deletions, orphaning stale-path chunks (hook-timed sync race)
## Evidence (totem-strategy, 2026-06-11 ~03:15Z, @mmnto/totem 1.57.0)

After mmnto-ai/totem-strategy#624 moved `proposals/active/295+296*.md` to `proposals/accepted/` (an ignore-listed dir) and merged:

1. The post-merge hook ran `totem sync` at the pull — baseline advanced to the merge SHA (`a887537`).
2. `totem search` afterwards still returned chunks citing `proposals/active/296-...md` — a path that no longer exists. The deletions were never purged.
3. A manual `totem sync` then reported `Resuming from last sync at a887537a... Incremental sync: 0 changed files, 1 deleted (of 294 total)` — its diff window (last-sync → HEAD) is empty, so the misses are permanent: **no future incremental sync can ever see those deletions.**
4. `totem sync --full` recovered (2858 chunks / 294 files; orphans gone) — that''s the workaround, not the fix.

## The class

Incremental sync derives deletions from the git diff since the recorded baseline. If the baseline advances in the same pass where deletion processing misses (or races, e.g. hook-timed sync during/just after a pull), the orphaned chunks are unreachable by any later incremental run — silently stale `search_knowledge` results with paths that 404 on disk (Tenet 4: this degrades silently; Tenet 20: the cache diverges from primitives with no sensor).

## Fix shapes (your call)

- Reconcile against the working tree, not only the diff: each sync cheaply asserts every indexed file still exists (it already counts "of 294 total" — the orphan check is the same walk), purging chunks whose source is gone. Self-healing regardless of baseline history.
- Or: advance the baseline only after deletion processing commits (transactional ordering), which fixes the race but not already-orphaned indexes.

First shape also retires this class for ignore-list additions (chunks for newly-ignored paths currently linger the same way).

🤖 Filed by strategy-claude (found during signoff corpus refresh)

satur8d:
## What

Fixes #2151 — incremental `totem sync` permanently orphaned chunks for files deleted outside the git diff window (baseline already advanced, a hook-timed race, or a rename into an ignored dir). Search returned paths that 404 on disk until a `--full` rebuild.

**Fix: working-tree reconciliation.** Incremental sync derives deletions by reconciling indexed paths against the working tree, not from the diff window:

- `LanceStore.getDistinctPaths()` — bounded `select(['filePath'])` projection; returns raw stored paths.
- `computeOrphanPaths(indexed, live)` — orphans = indexed − live; compares on a separator-normalized key but returns the RAW path for deletion (W1).
- `runSyncInner` purges `getDistinctPaths() − allFiles` on the incremental path; the re-embed set stays diff-scoped.

Self-heals deletions, renames-into-ignored, newly-ignored, and de-targeted files regardless of baseline history.

## Cohort design review (pre-build)

codex / gemini / agy — 3/3 pass-with-folds, all folded:

- **W1 (codex):** compare normalized, delete raw — no false-purge of legacy backslash rows (tested both directions).
- **Tenet 20 (gemini):** a purge-only sync now rebuilds the FTS index so it drops the orphaned content.
- **Tenet 4 (gemini + codex):** the run reports `orphansPurged`; only true orphans are logged purged.
- W2 (codex) + agy's scenarios as tests; agy's optimization (no reconciliation query on a full sync); I1 baseline-ordering wording.

Design + folds: `.totem/specs/2151.md`.

## Tests

- `computeOrphanPaths`: W1 both directions, de-target, rename-into-ignored (#624), empty-diff (W2), mixed re-embed+purge.
- `getDistinctPaths`: dedup, empty store, raw-path round-trip through `deleteByFile`.
- 5137 tests pass; `totem lint` 0 errors (6 warnings are documented false positives — genuinely-async test callbacks).

Changeset: `@mmnto/totem` + `@mmnto/cli` minor.

Closes #2151

### DIFF
--- a/.changeset/sync-orphan-reconciliation.md
+++ b/.changeset/sync-orphan-reconciliation.md
@@ -1,0 +1,6 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+---
+
+fix(sync): self-heal orphaned chunks via working-tree reconciliation (mmnto-ai/totem#2151). Incremental `totem sync` now derives deletions by reconciling indexed paths against the working tree (`computeOrphanPaths` + new `LanceStore.getDistinctPaths()`) instead of from the git diff window, so chunks for files deleted, renamed into an ignored dir, newly ignored, or de-targeted are purged even when the baseline already advanced past the change — the prior class left them orphaned until `totem sync --full`. Comparison is separator-normalized but deletion uses the raw stored path (legacy backslash rows are neither false-purged nor missed). A purge-only sync now also rebuilds the FTS index so it drops the orphaned content, and the run reports an `orphansPurged` count.
--- a/.totem/specs/2151.md
+++ b/.totem/specs/2151.md
@@ -1,0 +1,138 @@
+### Problem Statement
+
+Incremental syncs advance their git baseline before fully committing deletion operations, creating a race condition (often triggered by background git hooks) where deleted files are permanently orphaned in the vector database. Furthermore, files added to an ignore list after being indexed are currently ignored during incremental diffs, meaning their existing chunks are never purged.
+
+### Architectural Context
+
+The provided context outlines how `post-merge` and `post-checkout` hooks trigger automatic background re-indexing (`totem sync`). This background execution model is precisely what exposes the race condition if a user initiates another command or operation concurrently.
+
+### Files to Examine
+
+1. `packages/core/src/ingest/pipeline.ts` — Contains `runSyncInner` which controls the incremental sync workflow, diff reading, and baseline advancement.
+2. `packages/cli/src/commands/install-hooks.ts` — Contains the hook templates (`buildPostCheckoutHookContent`, `buildHookContent`), serving as the trigger mechanism that surfaced this race condition.
+
+### Technical Approach & Contracts
+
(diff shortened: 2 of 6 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Bot round — dispositions (commit `80303991`)

Greptile 4/5; GCA + CodeRabbit clean. Both Greptile P2s addressed:

- **① `pipeline.ts` log wording** — **fixed.** The pre-purge incremental-sync line now reads "N orphan candidate(s)"; the actual count stays "Purged M orphaned file(s)", so there's no candidate-vs-purged gap for an operator to puzzle over.
- **② `getDistinctPaths` O(chunks) row scan** — **deferred to #2175.** Correctness-neutral, and it matches the existing `manifestDocuments` projection (not a regression this PR introduces). A DB-level `DISTINCT` covering both call sites is the right fix and is now tracked; added a code breadcrumb pointing at #2175. Good catch on the N×K (not N) row count.

CI green (5137 tests, `totem lint` 0 errors).

satur8d:
@greptileai review

Issue #2181, opened 2026-06-15
feat(lint): advisory-ize the frozen-lesson rule class in the local pre-push gate (structural/AST/first-class stay hard)


satur8d:
## Summary

The local pre-push `totem lint` gate hard-blocks (exit-1) on **frozen compiled-lesson rule** findings. Those rules are un-recompilable under the standing rule-compilation freeze, so they misfire densely as false positives on any new source file — and clean-code dodges only relocate them. On #2179 (panel synthesis) this forced `--no-verify` (gate-block tree case C) on every push.

`#2178` already advisory-ized the **CI** Totem Lint job (`continue-on-error`); the **local pre-push hook was never advisory-ized**, so it still hard-blocks. Same class as the `#2180` review-gate decommission: a local hard-gate re-justifying a frozen mechanism.

## Ruling (strategy-claude, operator-greenlit) — `(c)`-refined

Advisory-ize the **frozen-lesson rule CLASS only** — not the whole gate:

- **Frozen-lesson rules** (`Lesson: "…"` provenance, from `.totem/compiled-rules.json`) → **advisory** (warn + continue, do not contribute to exit-1). PR bot review stays the real sensor.
- **Structural / first-class checks** — `check-xrepo-refs` (#430 → standalone #653), the AST rule family, `tsc` / tests / format / lockfile-sync → **STAY HARD** (exit-1). Green-zone, ground-truth-derived; never advisory-ize.

This split is Proposal 299's invariant (strategy#661): **sense in Yellow, enforce in Green** — a rule earns hard enforcement by legitimacy.

## Scope

- Interim (this issue): in the lint engine, route frozen-lesson-class findings to warn-only severity (printed, excluded from the exit-1 tally); structural/AST/first-class keep exit-1. Ships in `@mmnto/cli` → lc + totem-status inherit on the next version wave.
- Durable (deferred, rides the spine strategy#516): regex-engine inline-suppression (`totem-ignore` parity with the AST engine) so individual false positives can be suppressed and case-C `--no-verify` returns to a rare exception.

## Refs

strategy#661 (Prop 299), strategy#516 (spine), #2178 (CI advisory), #2180 (review-gate decommission), #2106 (Phase-1 lane), the rule-compilation freeze.

PR #2186, opened 2026-06-17
feat(spine): provenance/ruleClass marker retiring the #2181 engine-type proxy (#2183)

### LINKED ISSUE #2183: feat(spine): explicit provenance/ruleClass marker on regenerated rules — retire the #2181 engine-type advisory proxy
## Context

#2181 (merged `4d918ef1`, PR #2182) demotes the regex-engine compiled-rule class to **advisory** in the local pre-push `totem lint` gate, using **engine type** as an *interim proxy* for "frozen-lesson, illegitimate under the rule-compilation freeze." Conservative (keeps ast/ast-grep hard) but **inferred** — `isHardEngine` keys off `engine`, not a real legitimacy attribute.

## Durable deliverable (rides the spine, strategy#516)

When [`mmnto-ai/totem-strategy#516`](https://github.com/mmnto-ai/totem-strategy/issues/516) regenerates rules from merged-PR history, stamp an explicit **`provenance` / `ruleClass`** field on each rule so "advisory vs hard" is a **first-class attribute**, not engine-inferred. The advisory split then keys off `ruleClass`/provenance and the `engine`-type proxy retires.

## Legitimacy alignment (299 Amendment 1)

Per [`mmnto-ai/totem-strategy#665`](https://github.com/mmnto-ai/totem-strategy/issues/665) (Proposal 299 Amendment 1), the spine's §4 legitimacy bar is now **3-part**: provenance + positive control + negative control (the T15 verifiable-freezing pattern), not provenance-alone. The `provenance`/`ruleClass` marker should **carry or derive from that 3-part legitimacy**, so a rule earns hard enforcement only when all three hold.

## Also rides the spine

Durable per-finding **regex inline-suppression** (`totem-ignore` parity with the AST engine) — returns the gate-tree case-C `--no-verify` to a rare exception rather than a per-push necessity.

## Refs
#2181 · #2182 · `.totem/specs/2181.md` · strategy#516 · strategy#665

satur8d:
## What

Adds the durable Gate-1 legitimacy marker (ADR-110 §2/§3, Proposal 299 Amendment 1) that replaces the interim #2181 engine-type proxy with a first-class enforcement attribute on compiled rules.

### Core (`@mmnto/totem`)
- New optional **`legitimacy`** record on `CompiledRule` — three **peer** legs mapping 1:1 onto the ADR-110 §3 bar (`provenance` / `positiveControl` / `negativeControl`), with a mechanically-validated `ProvenanceRecord` (`mergedPr`, `reviewThread`, `commitSha`).
- New optional, **derived** `ruleClass`. Pure `deriveRuleClass()` helper: `'hard'` iff `legitimacy` present, the rule is promoted (the existing ADR-089 `unverified` flag is not `true`), and both controls pass.
- A schema **biconditional invariant** on `CompiledRuleSchema`: `legitimacy` and `ruleClass` are present-together-or-absent-together and consistent (`ruleClass === deriveRuleClass`). Enforced at the runtime-load boundary, so a forged or inconsistent stamp fails to parse — the engine proxy can never observe a forged state.

### CLI (`@mmnto/cli`)
- `totem lint` derives the hard tier from `ruleClass` when present, falling back to the engine-type proxy **only** for un-stamped legacy rules.
- Severity gate unchanged (blocking = hard tier **AND** error-severity); the frozen-lesson advisory label is scoped to legacy rules so a minted advisory rule is never mislabeled.

## Why

Retires #2181's engine-type proxy (the interim "ast/ast-grep = hard, regex = advisory" heuristic) with the real provenance/legitimacy attribute — the provenance leg of the Tenet-9 sense-to-enforce gate. It sits upstream of the Gate-1 wind-tunnel (next slice), which reads per-rule eligibility off this marker with no translation layer.

## Compatibility

Additive + backward-compatible. Legacy rules carry neither field, fall through to the proxy with **identical** behavior, and serialize **byte-identically** (zero compile-manifest churn — verified). `deriveRuleClass` is intentionally **unwired** from the frozen compile pipeline; spine rule-regeneration (strategy#516) is the sanctioned writer.

## Review + verification

Pre-build design reviewed by the cohort, all PASS:
- **codex** (contract, ADR-110 §2/§3) — folds adopted: mechanically-validated provenance + severity-gate preservation.
- **gemini** (tenet) — 1:1 strategy#666 three-check, zero translation layer.
- **agy** (build/test) — 6-invariant test contract, all test-locked.

Gate: core `2404 passed` · cli `2588 passed` (13 new tests) · `totem lint` PASS (0 errors) · `verify-manifest` PASS (485 rules, hashes match) · format + `tsc` clean.

Design record: `.totem/specs/2183.md` § Implementation Design.

Closes #2183


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
**Disposition — review round 1** (pushed `a02c3dfb`)

CI was green and the verdicts affirming (greptile 4/5 "safe to merge", CodeRabbit "implementation correct, architecture sound", GCA one suggestion). Folded every substantive finding; one decline with reasoning. All hardening — no behaviour change. Gate after fixes: core 2408 / cli 2589 tests green · `totem lint` PASS (0 errors) · `verify-manifest` PASS (485 rules, hashes match).

**Accepted + fixed**
- **greptile — Comments Outside Diff #1 (legacy ast-grep zero-regression):** added a `keeps a legacy (un-stamped) ast-grep rule blocking` test to the #2183 reader block. It's this PR's own spec invariant #1 — under-delivered; good catch. Behaviour was already correct via the proxy fallback, now guarded against a later refactor.
- **CodeRabbit #1 (biconditional only covered `'hard'`):** added the complementary `ruleClass:'advisory'`-without-`legitimacy` parse-fail case so the full ⟺ is documented from tests alone.
- **greptile #2 + CodeRabbit #2 (`reviewThread.trim()` is a mutating transform):** switched to a **non-mutating** validator — `z.string().refine(s => s.trim().length > 0)` — which rejects empty/whitespace-only *without* mutating the parsed value. A stamped rule's `reviewThread` now round-trips byte-identically, so no silent normalization → no `verify-manifest` hash drift. Added a non-mutation test + a stamped-rule round-trip stability test.

**Declined (with reasoning)**
- **GCA (high) — make `COMMIT_SHA_RE` case-insensitive (`/i`):** keeping strict lowercase. git emits canonical lowercase SHA-1, and the only writer of `commitSha` is spine regeneration (strategy#516) deriving from git — so the canonical form is always lowercase. Accepting uppercase *without* normalizing would admit a non-canonical value; *normalizing* (lower-casing on parse) would reintroduce the exact transform/hash-stability hazard greptile/CR flagged for `reviewThread`. Validating the canonical form is the consistent, non-mutating, hash-stable choice. Made the intent explicit inline (adopting CR's adjacent comment suggestion) and locked it with an uppercase-reject test.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

Issue #2189, opened 2026-06-17
feat(spine): Gate-1 wind-tunnel — wire diff-replay into un + S4 hard-error re-derivation + verdict-precedence ratification (post-#2188 deferrals)


satur8d:
## Context

#2188 shipped the Gate-1 wind-tunnel **harness** (windtunnel.lock.v1 schema + pure scorer + git-derived freeze/integrity proofs + the post-image `readStrategy` parity seam proven at the unit level + mock-engine validation), per the OQ2-ratified *harness-only-now* scope. Three items were **deliberately deferred** and are tracked here so they are not lost. They land with the post-strategy#516 certifying-run build.

## Deferred items

1. **Wire the shared post-image `readStrategy` into the `run` command's engine path (C1/S1 integration).**
   Today `run` builds the `readStrategy` and hands it to a mock engine (harness phase — no compiled rules yet). The certifying run must build additions from each resolved PR's diff and feed them through `enrichWithAstContext` + `applyAstRulesToAdditions` with the **one shared** post-image `readStrategy` (the mechanism is already unit-proven in `packages/core/src/spine/windtunnel-parity.test.ts`). Blocked on strategy#516 (real minted rules to replay).

2. **S4 full hard-error corpus-completeness re-derivation.**
   Today `freeze` verifies lc-clone accessibility + `asOfCommit` ancestry + counts/warns on `resolvedPrs`. The spec's S4 contract is a **hard-error** assertion `resolvedPrs === selectionRule(asOfCommit)` re-derived from lc merge history. Implement the deterministic `selectionRule(asOfCommit)` resolver (code-touching predicate per ADR-110 — conventional-commit type ∈ {feat,fix,refactor,perf} OR source-glob, excluding chore(deps)/docs) and make any mismatch a hard error (no silent shrink, §5).

3. **Ratify the scorer verdict precedence.**
   `scoreWindtunnel` currently ranks exposure-floor + cull-rate `HONEST-NEGATIVE` **above** `FP → FAIL`. The spec's "any firing labeled FP ⇒ FAIL" reads as unconditional, so a confirmed FP co-occurring with thin exposure is currently masked as HONEST-NEGATIVE. Ratify the intended precedence (likely FP→FAIL wins) and lock it with a test. Strategy-claude / operator adjudication.

## Refs

[...]

PR #2197, opened 2026-06-18
fix(spine): wind-tunnel corpus enumeration walks ancestry order — git log --topo-order for deterministic bounded windows (#2189 item 2 follow-up)

satur8d:
## What

Follow-up to #2189 item 2 (the S4 `selectionRule(asOfCommit)` resolver shipped in 1.67.0). `enumeratePrMetas` now enumerates lc's merged-PR history with **`git log --topo-order`** — ancestry (topological) order, not the default commit-date order.

## Why

A `bounded` selectionRule window takes "the N most-recent qualifying PRs" off the **front** of the enumerated list. "Most recent" must therefore mean N-most-recent **by ancestry**, never by timestamp — the ADR-110 §6 ancestry-not-timestamp rule applies to the window too (ratified by strategy-claude, 2026-06-18, confirming the COUNT semantics I shipped and pinning this ordering invariant).

Commit dates are non-monotonic and rewritable (rebases, clock skew), so date-ordered enumeration would make a bounded window's membership **non-deterministic** — defeating the §6 `resolvedPrs ≡ selectionRule(asOfCommit)` gate's determinism. The reachable PR **set** is identical either way, so:

- the certifying phase (which runs `window: all`, order-irrelevant) is **unaffected**;
- this hardens the `bounded` path against a future non-linear merge or date-skewed history before it can ever be used.

So it's latent-not-live today (lc is 100% linear squash-merge, where topo and date order coincide) — but it closes the contract loop on a freshly-pinned invariant.

## Changes

- `packages/cli/src/commands/spine-windtunnel.ts` — add `--topo-order` to the `git log` enumeration in `enumeratePrMetas` (+ rationale comment).
- `packages/cli/src/commands/spine-windtunnel.test.ts` — contract test locking the flag at the enumeration seam (a regression that drops `--topo-order` fails the suite).
- `packages/core/src/spine/selection-rule.ts` — comment-only: the input-order / window docs now cite the ancestry-not-timestamp invariant (no behavior change; `@mmnto/totem` not bumped).

## Scope note

This is a `@mmnto/cli` patch. **#2189 stays open** — item 1 (wire `readStrategy` into the certifying `run` gate) remains blocked on strategy#516.

## Gates

core 32/32 · cli 30/30 (incl. the new contract test) · `tsc` clean · `prettier --check` clean · `totem lint` PASS (0 errors) · `totem review` PASS (no findings).


satur8d:
/gemini review
@coderabbitai review

satur8d:
@greptileai review

Issue #2199, opened 2026-06-19
feat(spine): Gate-1 rule-mining miner — ADR-111 producer (Convergent Spine keystone)


satur8d:
## The un-started keystone

The **producer** member of the ADR-103 G-series — the piece that turns merged lc PR history into candidate rules. ADR-103 begins at *committed DSL source*; its legacy `totem lesson compile` LLM producer is frozen (`rule-compilation`, 2026-05-17). ADR-110 (acceptance) *reads* `provenance`/`ruleClass` but does not *mint* them. **There is no live producer.** This issue is it.

**Contract:** [ADR-111 "Spine Gate-1 Rule-Mining Miner"](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-111-spine-gate-1-rule-mining-miner.md) (merged mmnto-ai/totem-strategy#688). All scaffolding is shipped: `ruleClass`/provenance marker (#2183), `selectionRule` corpus resolver (#2189 item 2), wind-tunnel harness (#2188).

`(merged lc PR + review thread + commit SHA) → CandidateRuleRecord` carrying provenance + classifier disposition + DSL source, minted `unverified`/Yellow, **no hand-curation**. Inherits ADR-091's `Extract → Classify → Compile → Verify-Against-Codebase → Activate` funnel by reference. Unblocks the certifying run (#2189 item 1).

## Slicing

| Slice | Scope | LLM? |
|---|---|---|
| **1** | Deterministic foundation: `CandidateRuleRecord` envelope + frozen three-way split (`trainPrs ⊎ heldOutPrs ⊎ excludedPrs == selectionRule(asOfCommit)`) + five execution ledgers + **zero-LLM CI harness asserting all 9 ADR-111 Falsifying-Metric clauses** against committed red/green fixtures (the #2188 mock-first pattern) | no |
| 2 | Stage-1 Extract heuristics (review-thread → draft DSL) — LLM draft-only | yes |
| 3 | Stage-2 Classify gate (structural/behavioral) — LLM draft-only, Stage-4-backstopped | yes |
| 4 | Stage-3 Compile (ADR-103, inherited) + Stage-4 Verify-Against-Codebase wiring (mostly shipped #1682) | no |
| 5 | Certifying-run consumption — real miner output into `spine windtunnel run` (#2189 item 1) | no |

[...]

PR #2209, opened 2026-06-20
feat(spine): Gate-1 miner slice 5b-i — deterministic LLM record/replay scaffold (ADR-111)

satur8d:
Slice **5b-i** of the ADR-111 Gate-1 rule-mining miner — the **deterministic LLM record/replay scaffold**, the contract-mandated non-determinism containment for the certifying run, proven in isolation with a **stub orchestrator** (NO live LLM, NO network, NO prompts — those are slice 5b-ii).

**Part of #2199** (the miner umbrella stays open — 5b-ii live prompts + 5c orchestrator remain). Towards #2189 item 1 (the real-engine certifying run, slice 5c).

## Why this exists
The certifying run (slice 5) introduces the miner's first non-deterministic, LLM-backed components. The 4/4 cohort panel's central ruling (fold A, strategy: "arguably contract-MANDATED" by §6 + §8 + Tenet-15): freeze the live LLM outputs into a recorded artifact so the scorer + falsification harness re-run the experiment **zero-LLM, byte-deterministically, auditably**. 5b-i builds that machinery first so the containment is proven independent of the feasibility-risky prompts (the within-5b split, ruled by strategy + agy).

## What it builds (`packages/cli/src/commands/spine-llm-replay.ts`)
- **`llm-replay.v1`** Zod artifact — `kind` + a run-level `provenance` block + two record sections (`extractor: inputKey → string[]`, `classifier: inputKey → ClassifierResult`). Records are **strictly `inputKey → raw port output`** (a recorded `[]` / `{behavioral, error-default}` is a REAL row, distinct from a miss); no metadata ever leaks into a record; keys serialize sorted (clean diffs).
- **`inputKey` digests** (fold D) mirroring core's `deriveClaimId` (`sha256("<ver>" + canonicalJson({identity}))`): the extractor key includes `mergeCommitSha` + the **eligible** (non-resolved/non-outdated) threads, normalized so provider thread/comment order can't shift the key; the classifier key includes a caller-supplied `draftRef` so two drafts from one provenance don't collide.
- **Record/replay decorators** (generic over the ports). `Recording*` passes through + records (duplicate `(adapterKind, inputKey)` → `DuplicateRecordError`, never last-write-wins). `Replay*` is pure/zero-LLM: HIT → recorded value; **MISS → `ReplayMissError`** (never a `[]`/safe-default fallback — a miss is a corpus-integrity failure).
- **External-expected-hash integrity gate** (fold B, panel-resolved over a self-hash which is circular): the `Replay*` constructor takes the expected content-hash **injected** (5c sources it from a committed lock) and throws `FixtureIntegrityError` at construction on mismatch — a tampered/drifted fixture can't serve a single replay.
- **fold F — the integrity gate covers the WHOLE artifact** (provenance + records), so a prompt-hash / provenance edit without a re-record trips the gate (a prompt change must force a re-record, never silently serve stale outputs under a changed prompt).

## Tests (22, stub-backed, zero network/LLM)
record→replay determinism (FM-a) + insertion-order determinism; **drift red craft** (mutate a recorded entry → `FixtureIntegrityError`); **provenance-tamper red craft** (edit `promptTemplateHash` → `FixtureIntegrityError`, fold F); **miss red craft** (absent key → `ReplayMissError`, not `[]`/`error-default`); recorded-empty-vs-miss; duplicate-key; inputKey semantics (`mergeCommitSha` sensitivity, reorder-invariance, resolved-excluded, `draftRef` disambiguation); serialization hygiene.

## ⚠️ Bundled incidental fix — `tools/install-hooks.js` (worktree compatibility)
This PR also carries a **small, unrelated** build-tooling fix. In a **linked git worktree** (which the cohort now uses per the ADR-110 per-agent-worktree decision), `.git` is a gitdir-pointer *file*, not a directory — so the original `mkdir .git/hooks` failed with `ENOTDIR` on **every** pnpm lifecycle script, blocking build/test entirely. The fix follows the gitdir pointer + `commondir` to the shared hooks dir. **The normal-checkout path (`.git`-as-directory → `.git/hooks`) is byte-for-byte unchanged** — no regression for regular clones; it only adds the worktree/submodule case. Flagged here rather than buried: it's infra that every future worktree-delegated build needs (5b-ii, 5c), so it rides along; happy to split it into its own PR if preferred.

## Gate (re-run by the controller, not just the build agent)
`totem lint --base main` **PASS, 0 errors** (66 advisory warnings) · `pnpm -w build` green · core + cli suites green · 22 replay tests. Cohort-panel-reviewed (4/4 — codex contract / strategy ADR-111-fidelity / gemini tenet / agy build); the fold-F coverage gap was caught in controller review + folded here. Design: `.totem/specs/adr-111-slice5b-llm-adapters.md`.


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Review dispositions — folded in 84932d69

Thanks all. All findings across Greptile + CodeRabbit + GCA accepted and fixed (no Greptile "Comments Outside Diff" this round). Cross-bot dedup → 6 distinct items.

| # | Finding | Source | Disposition |
|---|---------|--------|-------------|
| 1 | **`normalizeThreads` sort not a total order** — same-path threads sharing a first comment but differing later compared `0`, leaking provider array order into `extractorInputKey` | **Greptile P1 + CR** | **Fixed** — sort by full `canonicalJson(thread)` (a total order); two threads compare equal only when byte-identical. Added a same-path reorder-invariance test (the case the old test missed). |
| 2 | Manual `[Totem Replay]` prefix in the 3 error messages | GCA ×3 | **Fixed** — dropped the prefix. **Note:** kept `extends Error` rather than `extends TotemError` — the latter would re-introduce a static runtime import of the `@mmnto/totem` barrel, the *exact* thing finding #6 removes. (See reconciliation below.) |
| 3 | Hash/serialize doc footgun — hash is minified canonical, file is pretty-printed | **CR + Greptile P2** | **Fixed** — clarified the `serializeReplayArtifact` JSDoc: verify via `computeArtifactHash(loadedArtifact)`, never `sha256` the raw file bytes. |
| 4 | `install-hooks.js` `match[1]` not trimmed | CR | **Fixed** — defensive `.trim()` on the captured gitdir path. |
| 5 | Double Zod parse in `Replay*` ctor (`computeArtifactHash` re-parses an already-parsed artifact) | CR | **Fixed** — added `hashParsedArtifact`; the integrity gate hashes the ctor's already-parsed value, `computeArtifactHash` (public) still parses for untrusted input. |
| 6 | Static import of runtime `ClassifierResultSchema` from the heavy core barrel violates the CLI-startup dynamic-import styleguide | GCA | **Fixed** — validate via a **local** CLI-side `ClassifierResultLocalSchema` (mirrors core's shape + the `error-default ⟹ behavioral` refine). This aligns with the cohort panel's ruling that the replay artifact schema is a CLI-layer concern; a **parity test** asserts it accepts/rejects exactly what core does, so the duplication can't silently drift. |

**Reconciliation (#2 ↔ #6):** GCA's "use the TotemError convention" and "no static core-barrel runtime imports" are in tension — a base class can't be lazy-loaded for a top-level class declaration, so `extends TotemError` forces the static import #6 removes. I took GCA's literal #2 suggestion (drop the prefix, keep `extends Error`), which satisfies both.

**Gate after the fold:** `totem lint --base main` PASS / **0 errors** (66 → 60 warnings) · build green · core + cli suites green · **24** replay tests (+2: same-path normalization, schema parity). No re-triggers fired (operator-gated).

satur8d:
/gemini review
@greptileai review
@coderabbitai review

PR #2215, opened 2026-06-20
feat(spine): Gate-1 miner slice 5c-i — deterministic real-engine firing path (ADR-111)

satur8d:
## Summary

Slice **5c-i** of the ADR-111 Gate-1 rule-mining miner — the **deterministic real-engine firing path** for the certifying run (the #2189 item-1 / #2199 spine track, build-sequence step 1). Replaces the certifying-phase mock with the real engine: each resolved-PR diff → `enrichWithAstContext` + `applyAstRulesToAdditions(readStrategy)` → `RuleFiring[]` with content-based labelIds. Built off the 4/4-panel-locked design at `.totem/specs/adr-111-slice5c-certifying-orchestrator.md`.

## Folds discharged (5c-i scope)

- **A1 (fold-D)** — `assertUniqueFiringLabels` hard-gates `firings.length === unique(labelIds)` pre-score and THROWS on collision, surfacing the colliding labelIds + evidence refs (Tenet 4). No ordinal/occurrence discriminator — preserves the `labelId→evidenceRef` contract; a richer key is a later versioned extension only if the corpus needs it.
- **C1** — `computePerRuleControlResults` surfaces a per-rule `{positiveControl, negativeControl, evidenceRefs}` over survivors, derived **per-rule and never from the global `nonVacuity`** (the data prerequisite for 5c-ii's fold-B stamping).
- **C2** — `scoreWindtunnel` now enforces the third exposure leg (`filesTouchedInWindow`), which was previously inert (the CLI passed a hard-coded 0); the certifying path computes real touched-file exposure from the diffs. Sub-floor file exposure ⟹ HONEST-NEGATIVE.
- **fold-F** — `assertNoArchivedRules` throws at the scorer-input boundary; the engine is never invoked on a `status:'archived'` rule (archived ≠ wind-tunnel FP).
- **fold-H** — neg-control firings pass through as `controlKind:'negative'`; unlabeled firings ⟹ `needsAdjudication`; duplicate label ⟹ hard error pre-score.

## Scope boundary

5c-i is the deterministic firing path only. The live-recorded corpus (fetch/compile/record), fold-B legitimacy **stamping**, fold-C parse-before-write persistence, the `record` subcommand, fold-I ledgers, fold-K E2E isolation, and the §5 split-artifact precondition are **5c-ii** — wired through the `CertifyingCorpusProvider` injection seam left here. An unwired certifying run fails loud with `CONFIG_INVALID` (never a silent empty-set score); `runMockEngine` is retained for the harness phase.

Part of the #2189 item-1 / #2199 spine track; item 1 finishes with 5c-ii (this slice does not complete it).

## Test plan

- **Core** (`@mmnto/totem`): 22 new firing tests + 2 scorer C2 boundary cases — spine suite 59 green.
- **CLI**: 5 new `runCertifyingEngine` tests — windtunnel suite 35 green.
- Zero-LLM, deterministic — fixture rules + PR-diff fixtures + injected post-image `readStrategy`; no network.
- Real gate (controller-verified post-commit): `totem lint --base main` PASS (0 errors), `format:check` clean, full suite green.

### DIFF
--- a/packages/cli/src/commands/spine-windtunnel.test.ts
+++ b/packages/cli/src/commands/spine-windtunnel.test.ts
@@ -4,7 +4,13 @@ import * as path from 'node:path';
 
 import { afterEach, describe, expect, it } from 'vitest';
 
-import type { WindtunnelLock } from '@mmnto/totem';
+import type {
+  CompiledRule,
+  GroundTruthLabel,
+  ResolvedPrDiff,
+  RuleFiring,
+  WindtunnelLock,
+} from '@mmnto/totem';
 import {
   isBotIdentity,
   parsePrNumber,
@@ -17,9 +23,11 @@ import { cleanTmpDir } from '../test-utils.js';
 import {
   assertCorpusCompleteness,
   buildReadStrategy,
+  type CertifyingCorpus,
   computeFixtureSha,
   enumeratePrMetas,
   isCommitAncestor,
+  runCertifyingEngine,
   verifyControlIntegrity,
   verifyFreezeProof,
 } from './spine-windtunnel.js';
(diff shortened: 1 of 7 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Dispositions — bot round (fixed in `5a845e38`)

Thanks for the review. Three findings triaged:

**✅ greptile P2 — `negFiringsByRule` dead for survivors (`windtunnel-firing.ts`)** — ACCEPTED + fixed. A survivor is non-culled by construction (a negative-control firing culls the rule, and the result loop `continue`s past culled rules), so a rule reaching the result loop fired on **no** negative control → `negativeControl` is invariantly `true` and the negative leg carries no evidence refs. Removed the dead map; set `negativeControl: true` with the invariant documented; `evidenceRefs = posEvidence`. Output-preserving (survivors always had empty `negEvidence`).

**✅ greptile P2 (Comments Outside Diff) — dead `FiringLabelCollisionError` catch around `buildFirings` (`spine-windtunnel.ts`)** — ACCEPTED + fixed. `buildFirings` only throws `ArchivedRuleInScopeError` (A1 is the caller's pre-score gate via `assertUniqueFiringLabels`), so that catch branch was unreachable and the comment misleading. Removed the dead try/catch; A1 collisions still surface — with the per-collision report — at the `assertUniqueFiringLabels` gate. Closes the latent diagnostic-swallow trap.

**⏭️ gemini HIGH — multi-node-per-line labelId collision (`windtunnel-firing.ts`)** — VALID, **deferred to slice 5c-ii** (not declined). The observation is correct: a single rule matching multiple nodes on one line yields same-`labelId` firings, which A1 hard-gates. But:

- A1's hard-gate-on-collision is a **panel-locked decision** (fold-D): collision ⟹ throw pre-score (Tenet 4 — never silently miscount precision), **measure** whether it occurs on the frozen lc corpus, and add a discriminator (a diff-hunk span, *not* a dedup-collapse or ordinal) only if needed.
- The proposed dedup-by-`(rule,file,line)` would silently collapse distinct violations and pre-empt that measurement — contrary to the locked floor.
- It is **latent for 5c-i**: this slice wires the firing path but runs no real corpus (the `CertifyingCorpusProvider` is filled in 5c-ii), so A1 cannot trip here; the tests are deterministic fixtures.
- The dedup-vs-throw call is a genuine fold-D refinement for the contract owner to rule on, informed by the measurement — routed for 5c-ii.

No bots re-triggered (operator-invoked).

satur8d:
## Dispositions — CodeRabbit review (addendum)

Owning a miss first: I initially read the **Review skipped** status + the "initiate chat" tip and concluded CR hadn't reviewed. It had — the `@coderabbitai review` invocation posted a full review further down that reply, which I overlooked. Triaged now.

**🔴 Dead `FiringLabelCollisionError` catch on `buildFirings`** — ACCEPTED, already fixed in `5a845e38` (greptile flagged the same). `buildFirings` only throws `ArchivedRuleInScopeError`; the catch was unreachable. Removed; A1 collisions surface at the `assertUniqueFiringLabels` gate below.

**🔴 Potential A1 self-collision (both engines over the full `rules` array)** — VERIFIED SAFE; clarifying comment added (`1c12c2ec`). Confirmed per your ask: the two engines self-filter into **disjoint** partitions — `applyRulesToAdditions` takes `engine === 'regex' || !engine` (`rule-engine.ts:266`), `applyAstRulesToAdditions` takes `engine === 'ast' | 'ast-grep'` (`rule-engine.ts:363/369`). No rule is processed by both, so double-processing cannot manufacture a same-rule `labelId` collision. Added the comment you suggested to prevent re-introduction. (Distinct from the gemini multi-node-per-line case, which is deferred to 5c-ii as a fold-D contract question.)

**🟡 `CertifyingCorpus` inline `import()` types** — DECLINED (consistency/scope). This file resolves `@mmnto/totem` **types** via inline `import('@mmnto/totem').X` throughout — e.g. the pre-existing `WindtunnelLock`/`RuleFiring`/`GroundTruthLabel` refs (L762/766) and `safeExec` (L340); the sole top-level import is `import type` (L4). The 5c-i additions follow that established convention; switching only them to top-level `import type` would make them inconsistent with the rest of the file. If the convention should change, that's a file-wide refactor, not a 5c-i concern.

**🟡 Dynamic `import('@mmnto/totem')` in `runCertifyingEngine`** — DECLINED (same rationale). Every **value**-import from `@mmnto/totem` in this file is dynamic (`await import(...)`, pre-existing at L32/L172/L577/L770); `runCertifyingEngine` matches that pattern. A switch to static value-imports is a file-wide change, out of scope for this slice.

Net: both 🔴 resolved (one fixed, one verified-safe + documented); both 🟡 declined on file-convention/scope grounds. No bots re-triggered (operator-invoked).

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Disposition — gemini cause-chain finding (fixed in `d8b172a0`)

**🔴 gemini — re-throw must pass `cause` (`spine-windtunnel.ts`, A1 collision handler)** — ACCEPTED + fixed. Verified the citation: `.gemini/styleguide.md` does carry the rule (*"Error cause chains (ES2022): when re-throwing in `catch` blocks, always pass the original error as the `cause`… NEVER concatenate `err.message` into a new error string"*), and the `TotemError` constructor is `(code, message, recoveryHint, cause?)`. The `runCertifyingEngine` A1 collision handler rethrew `FiringLabelCollisionError` as `TotemError` without forwarding `err`; added it as the 4th `cause` arg so the stack trace + cause chain survive for `handleError`.

Scope check: the other `TotemError` in this region (`no corpus provider wired`) is a guard clause, not a catch re-throw, so it carries no `cause` by design — confirmed L879 is the only new catch-re-throw affected.

No bots re-triggered (operator-invoked).

Issue #516, opened 2026-03-15
feat: orchestrator graceful degradation — SDK → CLI fallback


satur8d:
## Summary

When a configured orchestrator's SDK is not installed, Totem should gracefully fall back to the same provider's CLI via the shell orchestrator — not crash.

## Current behavior

If `provider: 'gemini'` is configured but `@google/genai` is not installed (e.g., via `pnpm dlx`), the command crashes:

```
[Totem Error] Gemini SDK (@google/genai) is not installed.
```

Same for `@anthropic-ai/sdk`, `openai`, etc.

## Desired behavior

**Fallback chain:**
1. **Configured SDK** (native, fastest) — try dynamic import
2. **Same provider's CLI** via shell orchestrator — e.g., `gemini` CLI, `claude` CLI, `openai` CLI
3. **Error** with clear message listing what's needed

`createOrchestrator()` should wrap the SDK import in a try/catch. On failure, detect if the provider's CLI is available (`which gemini`, `which claude`, etc.), and if so, construct a shell orchestrator command for it. Log a warning so the user knows they're on the fallback path.

## Providers to support

| Provider | SDK | CLI fallback |
|----------|-----|-------------|
| Gemini | `@google/genai` | `gemini` CLI |
| Anthropic | `@anthropic-ai/sdk` | `claude` CLI |
| OpenAI | `openai` | `openai` CLI (if it exists) or error |
| Ollama | N/A (HTTP) | `ollama run` via shell |

## Minimum requirement

One working orchestrator. If the configured one fails AND its CLI fallback isn't available, throw a clear error.

## Context

Discovered when `pnpm dlx @mmnto/cli spec 500` crashed because Gemini SDK is a peer dep not installed in the temporary dlx environment. The shell orchestrator already exists and works — we just need the fallback wiring.

PR #2218, opened 2026-06-21
feat(spine): Gate-1 miner slice 5c — certifying-run orchestrator (ADR-111, #2189 item 1)

satur8d:
## What

ADR-111 Gate-1 miner **slice 5c — the certifying-run orchestrator** (#2189 item 1), the Convergent Spine finale. `totem spine windtunnel run --phase certifying` now drives the full pipeline end-to-end on the frozen train slice: fetch→extract→classify (live 5b LLM adapters in record/replay decorators) → compile → the 5c-i real-engine firing path → score → legitimacy projection → persist + a cert-run report — **replayable zero-LLM, zero-network in CI**.

Bundles the **held 5c-i release** (the real-engine firing path shipped inert pending this slice) into one minor release of the `@mmnto/totem` fixed cohort.

## Slice contents (panel-locked decisions §6)

| Fold | What |
|---|---|
| **D** | `buildFirings` dedups same-`labelId` matches to one logical firing (raw matches kept as `evidence`); `assertUniqueFiringLabels` → post-dedup invariant. Verdict-safe under the 1.0 precision floor (strategy ruling). |
| **B/C** | `projectLegitimacy` stamps survivors PASS-only from their *own* per-rule control results (binding-4: `unverified` flips only when both controls pass); `buildCertifiedRulesFile` parses before write (half-stamp fails loud pre-disk). PASS-survivors → gate-1 cert output (never the live corpus — #516 promotes); non-terminals → cert-run report only (L3). |
| **L2** | lock carries `controls.integrity.llmReplaySha` (external expected-hash for the frozen `llm-replay.v1` fixture). |
| **orchestrator** | `buildCertifyingCorpus` composes the shipped extract/classify/compile stages; **binding-2** excludes Stage-4 out-of-scope (archived) rules from the scored set; **fold-I** asserts held-out-fetch-count 0 (FM-h) + emits the miner ledgers. |
| **A2 record / replay** | `spine windtunnel record` (separate, fail-safe) freezes the fixture + review content + prints the integrity hash; the run replays zero-network. |
| **K** | E2E proves the replay path makes **zero outgoing TCP/`fetch`** + a wrong expected-hash fails the integrity gate loud. |

## Tests / gate

- 11 commits, each tested. New: fold-D dedup, fold-B/C projection+persist (10), L2 schema (2), persist wiring (3), corpus provider (3), record/replay + fold-K (3), run-path provider (3), record subcommand (1).
- Full gate green: build 5/5 · `format:check` clean · all-package suites (core 2707, cli 2716) · `totem lint --base main` PASS (0 errors) · changeset present.

## Scope note (approach: fixture-corpus for the slice)

This slice ships the **machinery + CLI + CI proof** (the fold-K E2E + loader/provider tests exercise the whole pipeline via in-memory/tmpdir fixtures). The **P1 committed split + a real frozen corpus** is an *operational* precondition: it's produced by `record` + a real corpus freeze (needs network + the lc clone), satisfied when the certifying run is operated for real. The loaders/provider that consume those committed gate-1 fixtures are built and tested here.

## Issues

Delivers **#2189 item 1** (the certifying run) and the ADR-111 miner keystone tracked by **#2199**. Both left for explicit operator/strategy disposition on merge — no auto-close (the keystone trackers may carry follow-ups; `closingIssuesReferences` intentionally empty).

### DIFF
--- a/.changeset/slice5c-certifying-orchestrator.md
+++ b/.changeset/slice5c-certifying-orchestrator.md
@@ -1,0 +1,15 @@
+---
+'@mmnto/totem': minor
+---
+
+Gate-1 miner slice 5c — the certifying-run orchestrator (ADR-111, #2189 item 1), bundling the held 5c-i release.
+
+Completes the Convergent Spine Gate-1 miner: `totem spine windtunnel run --phase certifying` now drives the full pipeline end-to-end on the frozen train slice — fetch→extract→classify (live 5b LLM adapters in record/replay decorators) → compile → the 5c-i real-engine firing path → score → legitimacy projection → persist + a cert-run report — replayable zero-LLM, zero-network in CI.
+
+- **fold-D dedup**: `buildFirings` collapses same-`labelId` matches to one logical firing (raw matches retained as `evidence`); `assertUniqueFiringLabels` demotes to a post-dedup invariant. Verdict-safe under the 1.0 precision floor.
+- **fold-B/C persist**: `projectLegitimacy` stamps survivors PASS-only from their own per-rule control results (binding-4: `unverified` flips only when both controls pass); `buildCertifiedRulesFile` parses before write so a half-stamp fails loud pre-disk. PASS-survivors land in the gate-1 cert output (never the live corpus — strategy#516 promotes); non-terminals write a cert-run report only (§6 L3).
+- **L2**: the lock carries `controls.integrity.llmReplaySha` — the external expected-hash for the frozen `llm-replay.v1` fixture, beside `fixtureSha`.
+- **Orchestrator**: `buildCertifyingCorpus` composes the shipped extract/classify/compile stages; binding-2 excludes Stage-4 out-of-scope (archived) rules from the scored set; fold-I asserts held-out-fetch-count is 0 (FM-h) and emits the miner ledgers.
+- **record/replay**: `spine windtunnel record` (A2 — separate, fail-safe) freezes the `llm-replay.v1` fixture + review content and prints the integrity hash; the run path replays them zero-network. **fold-K** proves the replay path makes zero outgoing TCP/`fetch` calls and that a wrong expected-hash fails the integrity gate loud.
+
+Includes the held 5c-i real-engine firing path (`buildFirings` replacing the mock; per-rule control results C1, the `filesTouchedInWindow` exposure floor C2, fold-F archived-assert, fold-H OQ4 matrix), which shipped inert pending this slice.
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,8 @@ packages/pack-rust-architecture/tree-sitter-rust.wasm
 .totem/ledger/
 .totem/sync.lock
 .totem/temp/
+# Agent draft scratch (dispatch/verdict drafts staged before ECL delivery); local-only, never tracked
+.totem/drafts/
 .totem/cache/
 # Grounded run artifacts (mmnto-ai/totem#2100) — machine-local run records,
 # content-addressed, carry prompt content; never tracked.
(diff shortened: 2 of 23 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Round disposition — commit `13c29a4a`** · gate: build 5/5 · format:check · core+cli tests green · `lint --base main` PASS 0-err

8 findings fixed, 1 declined with rationale. Each verified against current code before dispositioning.

### Fixed

| Finding | Disposition |
|---|---|
| **CodeRabbit 🔴 Critical** — `createOrchestrator` cast hides shell `command` | Verified real — the `{ provider }`-only cast passes `command: undefined` into the shell variant (`createOrchestrator` does `invokeShellOrchestrator({ ...opts, command: config.command })`). Replaced the `as Parameters<…>[0]` cast with a typed provider switch (anthropic/gemini/openai, matching `PROVIDER_CREDENTIAL_ENV`) + `TotemError` for unsupported providers. |
| **Gemini ×5** — raw `Error` → `TotemError` for CLI-layer validation | Converted all five to `TotemError('CONFIG_INVALID', msg, recoveryHint, cause)`: cert-run fixture loader, missing-`llmReplaySha` gate, no-lc-clone Stage-4 read, the record lock/split parse (now try/catch-guarded), and the owner/name guard. Original error attached as `cause`; no manual `[Totem Error]` prefix. |
| **Greptile P1** — persist `gate1Dir` diverges from fixture-load path | Verified — fixtures load from `dirname(lockPath)` but persist hardcoded `repoRoot/.totem/spine/gate-1`. persist now derives `gate1Dir` from `dirname(lockPath)`: byte-identical for the default lock, correct under a non-default `--lock-path`. |
| **Greptile P2** — two `new Date()` → mismatched timestamps | Single `runNowIso` captured once in `runCommand`, threaded into both the corpus provider (compile-stage stamp) and persist (report `generatedAt`/slug). |
| **Greptile P2** — `seedClassesProvided` hardcoded `false` | Added optional `RecordCommandOptions.seedClassesProvided` + a `--seed-classes-provided` flag (default false / seed-blind), so the fold-I FM-f attestation is settable when seeds are used. |

### Declined (1)

| Finding | Rationale |
|---|---|
| **CodeRabbit 🟠 Major** — lazy-load the `./spine-llm-replay.js` runtime imports | Eager sibling-module runtime imports are the established spine-* pattern (4 precedents: `spine-cert-run-corpus`→{`cert-corpus`,`cert-record`}, `spine-windtunnel`→{`cert-persist`,`cert-run-corpus`}); `lint --base main` is green (rule unenforced). `spine-cert-record.ts` only enters the graph under the already-lazy `record`/`run` command (`index.ts` `await import(...)`), so there is no cold-start cost to unrelated `totem` invocations. The mechanical fix would force `buildReplayAdapters` sync→async, converting its synchronous integrity-throw — asserted by `spine-cert-e2e.test.ts:183` (`expect(() => buildReplayAdapters(...)).toThrow()`) — into an async rejection, degrading the validation contract for zero startup gain. Open to a separate sweep if we want the whole spine-* subtree migrated. |

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Round 2 disposition — commit `e7f1de27`** · gate: build 5/5 · format:check · core+cli 2716 tests green · `lint --base main` PASS 0-err

Accepted GCA's `@mmnto/totem` lazy-import finding (all four inline comments on `spine-cert-run-corpus.ts`). The `.gemini/styleguide.md` convention — static `@mmnto/totem` **type** imports, lazy `await import('@mmnto/totem')` for **values** inside function bodies — is exactly what `spine-windtunnel.ts` and `spine-cert-record.ts` already do. This file was the lone anomaly (static `SplitArtifactSchema`); round 1's `TotemError` add compounded it.

- import block → `import type {…}` (dropped the two value imports)
- `loadCertRunFixtures` is now `async`, lazy-imports `{ SplitArtifactSchema, TotemError }`, with `loadJson` as an inner closure capturing the lazy `TotemError`
- `buildReplayCorpusProvider` lazy-imports `TotemError` and awaits `loadCertRunFixtures`

Internal-only callers (`loadCertRunFixtures`/`loadJson` have no external or test callers), so the sync→async change has no ripple. On the "pass `undefined` for the 4th arg" note: `cause` is optional in `TotemError(code, message, recoveryHint, cause?)`, so the no-cause throws stay 3-arg (valid; build + lint green) — no change there.

> Note vs. the still-open CodeRabbit 🟠 (lazy-load `./spine-llm-replay.js`): that's a **different target** — a sibling command module, not the `@mmnto/totem` core barrel — where eager imports ARE the established spine-* pattern (4 precedents). Accepting GCA's core-barrel finding while declining CR's sibling-module one are consistent positions: lazy-load the heavy barrel, keep sibling helpers eager.

Issue #2214, opened 2026-06-20
fix(lint): lesson-fail-open-catch-ban — // totem-context: escape does not suppress in diff-scoped 	otem lint --base main (pushes authors to matcher-evasion)


satur8d:
## Surface

The Tenet-4 rule `lesson-fail-open-catch-ban` (ast-grep `kind: catch_clause` with no `throw_statement`) has a sanctioned inline escape: `// totem-context: <reason>`. **The escape does NOT suppress in the diff-scoped branch-lint path** (`totem lint --base main`), even when the directive is correctly positioned (immediately above / inline on the `catch` line — the position #1889 establishes as the one that works in full-tree lint).

Effect: an author who adds a *legitimately* fail-soft catch at an IO/LLM boundary, annotates it with the sanctioned `// totem-context:` escape, and runs the branch-lint gate, still sees the rule FIRE. The escape works full-tree but not diff-scoped. This pushes authors toward **matcher-evasion** (rewriting `catch (e) {}` as `.catch(() => …)` so it parses as a `call_expression`, not a `catch_clause`) — strictly worse for the corpus than a recognized, attested escape. It hits **every live IO/LLM adapter cohort-wide** (the ADR-111 5b live-adapter family is the live exhibit).

## Distinct from #1889

#1889 is the **position** axis: `isSuppressed` (`packages/core/src/rule-engine.ts`) checks only the matched line + the immediately-preceding line, so a directive above `try` (multi-line body) misses the `catch` line in *full-tree* lint. This issue is the **diff-scope** axis: a *correctly positioned* directive is honored full-tree but **not** under `totem lint --base main`. Same rule + same escape, different failure surface. Sibling to #1614 (terminal-handler false-positive) and the #1444 fail-open cluster.

## Diff-anchoring nuance (for the fix investigation)

[...]

Issue #702, opened 2026-03-18
Epic: Integrity Audit (Phase 2)


satur8d:
Codebase hardening, baseline compilation with ast-grep.

Issue #697, opened 2026-03-18
Epic: Agentic Control Loop (Phase 1)


satur8d:
Reframed: Prove the gate is airtight and measure the self-correction cycle.

- [ ] #695: Validate Layer 3 (The Guarantee) & Measure Correction Cycle
- [ ] #696: Validate Layer 1/2 (Acceleration & Efficiency)

*(Obedience is no longer the primary test; enforcement and cycle-time are).*

PR #2220, opened 2026-06-21
feat(lint): recognize Tenet-4 shape-2 fail-soft attestation (#2214 acceptance-#3)

satur8d:
## What

#2214 **acceptance-#3** — the recognized fail-soft carve-out, lockstep with the merged Tenet-4 doctrine sharpening (`mmnto-ai/totem-strategy#702` / `#708`). Acceptance #1+#2 shipped in #2216; this is the optional, #702-paired half.

A blanket fail-soft `catch` at a declared IO/LLM/network boundary is licensed by Tenet 4 only when it names a loud systemic backstop. Authors previously escaped `lesson-fail-open-catch-ban` either with a free-text `// totem-context:` (indistinguishable from a rationalized swallow) or by rewriting the catch as `.catch(() => …)` — a `call_expression` the rule never matches, strictly worse for the corpus. This makes that legitimacy **recognized, not dodged**.

## Design (strategy-ratified)

Option B (parse + surface a typed exemption), ratified by strategy-claude against merged Tenet-4 (2026-06-21):

- **`parseFailSoftAttestation`** recognizes the structured `// totem-context: fail-soft backstop=<name>` directive → typed `{ kind:'fail-soft', backstop }`, surfaced on the `'suppress'` rule-event (auditable; couples to the #697 Layer-B capability ledger). Recognition is narrow — **only a leading `fail-soft` token** is an attestation, so the ~25 existing prose escapes (`best-effort cleanup, fail-soft`) are unaffected. Additive, non-breaking.
- **Malformed → always-emit WARN.** A `fail-soft` claim that names no backstop surfaces a non-blocking `totem/fail-soft-missing-backstop` warning on **both** the ast-grep and tree-sitter suppression paths — so the grammar can't go decorative. **WARN, never ERROR**: the lint establishes token-presence only; the backstop's loudness + per-item accounting are verified at review/ADR level (Tenet 13/19), so blocking is the consumer's actuator and an error would overclaim + invite `backstop=anything` cargo-culting.
- **Shape 1** (type-discriminated rethrow, `if (!(err instanceof X)) throw err; return soft`) already passes the matcher (it carries a `throw_statement`) — locked with a regression fixture, **no matcher change**.
- Option C (verify the named backstop exists/is loud) deliberately excluded — semantic, fixed at adapter/ADR altitude per the tenet.

## First organic exhibit

`packages/core/src/sys/git.ts:323` is a real blanket-fail-soft that leads with `fail-soft` and names no backstop. It is **not in this diff**, so the gate stays green; it will surface the advisory WARN when that file is next edited, prompting a conscious backstop decision. Left deliberately — rephrasing to dodge the warn would be the prose-dodge antipattern, and fabricating a backstop name would be dishonest.

## Deferred (freeze-gated)

The rule's **primary message** (naming the new form when the rule fires) requires a recompile, gated behind the active `rule-compilation` freeze (lifts on spine cert-run Gate-1 PASS). The engine WARN + styleguide §6 carry the author-facing nudge now; the message rides a follow-up post-freeze-lift. **#2214 stays open** to track that remainder.

## Test / gate

- New `fail-soft attestation` suite in `rule-engine.test.ts` (parser unit tests · valid→no-warn · malformed→WARN on both ast paths · prose non-trigger · typed-attestation-on-suppress-event) + the shape-1 regression fixture.
- Build clean · **core 2715 / cli 2716 green** · `pnpm format:check` clean · `totem lint --base main` **PASS, 0 errors** · manifest verified (485 rules, unchanged).

## Issues

Advances #2214 (acceptance-#3). Aligns to `mmnto-ai/totem-strategy#702` / `#708` (Tenet-4 doctrine). No closing keyword — #2214 remains open for the freeze-gated rule-message.


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-1 dispositions — `8c78f7a7` + `e61cbe4c`

All five findings addressed. Gate green on each push: core 2716 · cli 2716 · build · format · `totem lint --base main` 0 errors.

**GCA (high) — `FAIL_SOFT_MISSING_BACKSTOP_RULE` lacks ast-grep schema fields → FIXED (differently).** Verified the rule is constructed at runtime and attached to a `Violation` directly — it is *never* loaded from `compiled-rules.json`, so it never flows through `CompiledRuleSchema` validation (no actual runtime failure). To dissolve the latent footgun I switched `engine: 'ast-grep'` → `engine: 'ast'`, which I empirically confirmed *is* schema-valid (`CompiledRuleSchema.safeParse → true`) and keeps the classification identical (both `ast`/`ast-grep` are hard engines → non-blocking probationary advisory, never a frozen-lesson demotion). Cleaner than a placeholder `astGrepPattern: 'noop'` — a never-executed pattern would mislead readers into thinking the rule matches.

**greptile (P2) — synthetic `engine` label fires from both suppression paths → DOCUMENTED.** Inherent to one shared diagnostic emitted from both the tree-sitter and ast-grep paths; benign because both hard-engine values classify identically and no consumer distinguishes them for this synthetic warning. The stable identity is the unique `totem/fail-soft-missing-backstop` lessonHash, not `engine`. A per-path-honest engine would reintroduce the GCA schema-invalidity on the ast-grep path, and a `ruleClass` sentinel requires a forged `legitimacy` stamp — so I made the invariant explicit in the doc comment instead.

**greptile (P2) — tree-sitter parity test asserts `severity` but not `lessonHash` → ACCEPTED.** Added the `lessonHash` assertion (your suggestion) so a future warning-rule swap can't pass the parity test while silently emitting the wrong diagnostic.

**CodeRabbit (1) — `backstop= <name>` (space after `=`) silently malformed → ACCEPTED.** Tightened `FAIL_SOFT_BACKSTOP_RE` to `/\bbackstop=\s*([^\s,;]+)/` so a space after `=` (a common authoring habit) isn't treated as a missing backstop; added a test for the space variant.

**CodeRabbit (2) — hardcoded `compiledAt` on the synthetic rule → ACCEPTED (clarified).** It is inert (no freshness check or sort reads it; `violationToFinding` ignores it) and deliberately fixed rather than `new Date()` for determinism — added an inline comment so the next reader isn't left wondering whether to bump it.

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-2 disposition — `5834853e`

**GCA (high) — `backstop = name` (whitespace before `=`) parses as malformed → ACCEPTED.** Round-1 handled a space *after* `=`; this widens `FAIL_SOFT_BACKSTOP_RE` to `/\bbackstop\s*=\s*([^\s,;]+)/` (your exact suggestion) so whitespace on either side of `=` is tolerated. Test extended to cover space-before, space-after, and both. Gate green: core 2716 · cli 2716 · build · format · `lint --base main` 0 errors.

The remaining open inline threads are round-1 carryover, already resolved — flagging so they aren't read as fresh blockers:
- **greptile P2 (synthetic `engine` label):** addressed in `8c78f7a7` — `engine:'ast'` (schema-valid) + the invariant documented (identity is the `totem/…` lessonHash, not `engine`).
- **greptile P2 (tree-sitter parity test):** fixed in `8c78f7a7` — the `lessonHash` assertion was added.
- **GCA high (synthetic rule schema fields):** resolved in `8c78f7a7` via `engine:'ast'` (confirmed `CompiledRuleSchema.safeParse → true`); this fresh GCA pass did not re-raise it.

Issue #2188, opened 2026-06-17
feat(spine): Gate-1 wind-tunnel — frozen windtunnel.lock.json + retrospective replay tooling (ADR-110, 299 Gate 1)


satur8d:
## Context

Second slice of the Proposal-299 Gate-1 build. The provenance leg — the `legitimacy`/`ruleClass` marker — landed in #2183 (`@mmnto/totem`+`@mmnto/cli`@1.64.2). This slice builds the **Gate-1 evidence engine** that ADR-110 ([mmnto-ai/totem-strategy#669](https://github.com/mmnto-ai/totem-strategy/issues/669)) operationalizes as the falsifiable acceptance contract.

## Deliverables

1. **Frozen `windtunnel.lock.json` manifest** — `.totem/spine/gate-1/windtunnel.lock.json`. Fixed corpus (lc-only, per ADR-110 design-Q resolution), frozen `N` and false-positive definition pinned **before** the window opens.
2. **Retrospective replay tooling** — replay regenerated rules over lc's merged PR history; positive controls (held-out real PRs) + adversarial near-miss fixtures; exposure-denominator tuple (active rules × files × controls); zero-confirmed-FP precision floor (299 §5 verbatim).
3. **Done-criterion** — non-vacuous PASS / HONEST-NEGATIVE (must fire on positive controls; cannot vacuously pass).

## Dependencies

- **#2183 (done)** — the `legitimacy`/`ruleClass` marker the wind-tunnel reads per-rule eligibility off (no translation layer).
- **strategy#516** — rule regeneration (the sanctioned writer of the marker); the wind-tunnel evaluates its output.
- **ADR-110 §2–§6** — the acceptance contract.

## Open design questions (strategy-claude lens — owns these)

- **`windtunnel.lock.json` schema/shape** — exact fields for the frozen corpus + control references (ADR-110 §6 holds control evidence).
- **ADR-110 deferred scalar-window-params** — frozen `N`, the FP-definition, and the exposure-denominator tuple.

Gated on strategy-claude's lens + a pre-build design review (the #2103/#2104/#2151/#2183 pattern) + operator approval before code.

Refs: ADR-110 (strategy#669) · Proposal 299 (strategy#661) · #2183 · strategy#516

PR #2224, opened 2026-06-21
feat(spine): Gate-1 cert-corpus materialization producer — materialize command (strategy#709)

satur8d:
## What

The Gate-1 cert-corpus materialization producer (**`mmnto-ai/totem-strategy#709`**): `totem spine windtunnel materialize --lc-dir <p> --manifest <seed.json>` writes the 4 unproduced cert-run **scoring** fixtures — the lock, `split.json`, `pr-diffs.json`, and the positive/negative control dirs — so `loadCertRunFixtures` stops throwing. Upstream of `record` → `freeze` → `run`. (`record` already produces the other 2; the disposition-derived `ground-truth-labels.json` **deriver** is the next slice.)

## Design — cohort-panel-ratified before any code

Pre-build panel on distinct lenses (codex = contract, agy = build/test, gemini = tenet/arch), unanimous **PASS-with-folds**. The boundary: a small **curated seed manifest** carries the irreducible answer-key decisions (`asOfCommit`, selection config, `cutIndex`, control designations + each positive control's `targetRuleId`); **everything else is derived** off the lc clone. Pure derivation + lock assembly in core (`deriveCorpus` / `buildWindtunnelLock`, reusing `resolveSelectionRule` / `resolveSplit`); git I/O in the CLI.

## Panel folds incorporated

- **fold-1 strict write-side** — `targetRuleId` required iff `controlKind: 'positive'`, forbidden otherwise (a positive control can't silently lose its non-vacuity target).
- **fold-2 pr-diffs integrity digest** — new additive-optional `controls.integrity.prDiffsSha` (sha256 over canonical `pr-diffs.json`) closes the hole that `fixtureSha` (control-dirs-only) leaves on the independently-loaded scoring source. *Producer computes + stamps it; the freeze/run **hard-enforcement** is the immediate fast-follow (flagged to strategy-claude — closest to their contract).*
- **fold-3 fail-loud git** — producer-owned diff resolution throws on git faults and rejects an empty diff for a code-touching PR (no silent-empty frozen fixture).
- **fold-4 Amendment-C double-guard** — a contradictory seed (control outside the corpus, or tagged both positive+negative) fails loud before emit.
- **Two-phase sealed lock** — the producer omits `llmReplaySha` (it can't know it — `record` runs after); `freeze` seals it.

## Tests / gate

- Programmatic git fixture (real `git init`, pinned dates) over the direct-to-main / malformed-`(#abc)` / revert-pair / multi-file / empty-diff matrix; byte-determinism; both integrity red tests (C6 `fixtureSha` + fold-2 `prDiffsSha`).
- The produced output **passes the actual freeze gate end-to-end** (`S4 corpus completeness VERIFIED` + C6 integrity).
- core **2725** / cli **2722** green · `format:check` · `lint --base main` **PASS (0 errors)** · changeset (`@mmnto/totem: minor`).

## Scope

The **producer** half of `mmnto-ai/totem-strategy#709`. This PR intentionally carries **no closing keyword** — the disposition-derived ground-truth-labels deriver (contract ruled) and the `prDiffsSha` freeze/run enforcement are the next slices. Related: `mmnto-ai/totem-strategy#709`, `mmnto-ai/totem#2188` (lock schema + harness, shipped).

### DIFF
--- a/.changeset/spine-709-cert-corpus-producer.md
+++ b/.changeset/spine-709-cert-corpus-producer.md
@@ -1,0 +1,19 @@
+---
+'@mmnto/totem': minor
+---
+
+feat(spine): cert-corpus materialization producer — `totem spine windtunnel materialize` (strategy#709)
+
+The Gate-1 cert run's SCORING corpus had no producer: `loadCertRunFixtures` hard-requires 6 fixtures and only `record` wrote 2 (the minted-rules half). This adds the producer for the 4 unproduced, non-label fixtures — the lock, `split.json`, `pr-diffs.json`, and the positive/negative control dirs — so the cert corpus is materializable upstream of `record`→`freeze`→`run`. The disposition-derived `ground-truth-labels.json` deriver is the next slice (contract ruled).
+
+Cohort-panel-ratified (codex/agy/gemini, distinct lenses) before any code; the folds it surfaced are incorporated:
+
+- **Seed-manifest boundary** (panel OQ1/OQ2): a small curated seed carries the irreducible answer-key decisions (asOfCommit, selection predicate/config, `cutIndex`, control designations + each positive control's `targetRuleId`); everything else — the resolved corpus, the ancestry-cut split, per-PR diffs, the integrity shas, the assembled lock — is derived off the lc clone. Pure derivation + lock assembly in core (`deriveCorpus` / `buildWindtunnelLock`), git I/O in the CLI.
+- **Reuse** (Tenet-21): `resolveSelectionRule` (corpus) + `resolveSplit` (the validated ancestry-cut split with its fail-loud cover/disjointness guards) + `enumeratePrMetas` + `computeFixtureSha`.
+- **Strict write-side** (fold-1): `targetRuleId` is structurally required for positive controls and forbidden otherwise — the producer never relies on the permissive consumer schema, so a positive control can't silently lose its non-vacuity target.
+- **Two-phase sealed lock** (fold OQ-seq): the producer writes the lock WITHOUT `llmReplaySha` (it can't know it — `record` runs after); `freeze` seals it.
+- **pr-diffs integrity digest** (fold-2): a new additive-optional `controls.integrity.prDiffsSha` (sha256 over the canonical `pr-diffs.json`) closes the hole that `fixtureSha` (control-dirs-only) leaves on the independently-loaded scoring source. The producer computes + stamps it; the freeze/run hard-enforcement is the immediate fast-follow.
+- **Fail-loud git** (fold-3): producer-owned diff resolution throws on git faults and rejects an empty diff for a code-touching PR (no silent-empty frozen fixture), unlike the advisory `git.ts` helpers.
+- **Amendment-C double-guard** (fold-4): a contradictory seed (a control outside the corpus, or tagged both positive+negative) fails loud before emit.
+
+Byte-deterministic at a fixed asOfCommit+seed (canonical sorted-key JSON, forward-slash paths, `git hash-object`, no timestamps in fixtures). Tested with a programmatic git fixture (real `git init`, pinned dates) over the direct-to-main / malformed-`(#abc)` / revert-pair / multi-file / empty-diff matrix; the produced output passes the actual freeze gate (S4 completeness + C6 integrity) end-to-end.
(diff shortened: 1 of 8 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-1 dispositions (commit `c6e44559`)

Worked the cohort + bot findings — fixed the substantive ones, declined two with rationale, deferred the enforcement (already tracked).

### Fixed
- 🔴 **CR — path containment** (`materialize.ts`): seed-derived write/delete targets (`canonicalPath` dir, `positiveRef`, `negativeRef`) are now constrained to within `repoRoot` via a `resolveWithinRepo` guard that refuses `..` / absolute / repoRoot-itself, fail-loud. The control dirs are recursively deleted, so this was the real blast radius.
- **greptile (`windtunnel-lock.ts`) + GCA (`materialize.ts:214`) — digest encoding**: `prDiffsSha` now digests the **exact on-disk `pr-diffs.json` bytes** (canonical 2-space + trailing newline), so a freeze/run enforcer can `sha256` the file directly; the encoding is pinned in the schema comment. (Both bots converged on this — good catch.)
- **GCA — atomic write**: manifest writes are now temp-file + rename, so a reader / CI gate never observes a partially-written file.
- 🟡 **CR — `llmReplaySha`**: gate changed to `!== undefined`, so an explicit `''` reaches the schema and fails its 64-hex regex instead of being silently dropped.
- **greptile — fold-3 untested**: added a unit test exercising `resolvePrGit`'s empty-diff guard directly (git succeeds, diff empty → fail loud).

### Declined (rationale)
- **GCA — `materialize.ts:34` + `:69` (`Error` → `TotemError`)**: the `gitText` / `resolvePrGit` `Error`s are an internal control-flow signal — they're caught and re-wrapped into a `TotemError` (with code + hint) at the command boundary, so the user-facing error is already a `TotemError`. The helpers are sync, module-level, *before* the lazy `@mmnto/totem` import (the CLI convention), so they can't construct `TotemError` without a static import against that convention. The suggested fix is also malformed — `TotemError` takes a `code` first arg. No change.

### Deferred (tracked)
- The freeze/run **enforcement** of `prDiffsSha` → **#2225**. This PR is the producer slice only (no closing keyword).

Gate: core 2725 / cli 2723 · eslint 0 errors · format clean · `totem lint --base main` PASS (0 errors).

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-2 dispositions (commit `4d213928`) — GCA re-pass

**Accepted — the `err.message`-concatenation findings (`:175`, `:187`, `:204`).** Verified against the actual styleguide: `.gemini/styleguide.md` §9 is explicit — *"always pass the original error as the `cause`… NEVER concatenate `err.message` into a new error string… the `handleError` debug logger traverses `.cause` chains automatically."* My wraps concatenated `— ${err.message}` while also passing `err` as cause. Fixed:
- `:175` (enumeration) + `:204` (git-resolution) — clean static message + `cause`, no concatenation; the recovery hint points at `TOTEM_DEBUG=1` for the underlying fault.
- `:187` (derivation) — **removed the wrap entirely**: `CertCorpusSeedError` / `SplitCoverError` are already structured, user-facing fail-loud errors, so they propagate verbatim. Re-wrapping + concatenating would have buried the specific reason (which control PR, which cut) behind a debug-only cause — worse UX than the raw structured error.
- `gitText` — it *also* concatenated `err.message` (the `:34` site); now `{ cause: err }`, no concatenation. (Same §9 fix, inside the helper — caught it while there.)

**Declined — the two `Error` → `TotemError` class-swaps (`:34`, `:73`).** These mirror the shipped `enumeratePrMetas` precedent: a sync, module-level helper throws a plain `Error` that's caught and wrapped into a `TotemError` (code + hint + cause) at the command boundary, so the user-facing error is already a `TotemError`. The helpers run *before* the lazy `@mmnto/totem` import (the CLI value-import convention), so constructing `TotemError` inside them would force a static value-import against that convention. (Noted that GCA corrected its suggestion's 4-arg signature this round — appreciated; the finding still doesn't account for the wrap-at-boundary pattern.)

Net: every genuine §9 cause-chain issue is fixed (including one inside `gitText` the inline markers didn't separately flag); the stylistic class-swap is declined on the established pattern. Gate: cli 2723 · eslint 0 · format · `totem lint --base main` PASS (0 errors).

PR #2227, opened 2026-06-21
fix(spine): enforce prDiffsSha at certifying run + freeze (#709 fold-2 enforcement)

### LINKED ISSUE #2225: spine: enforce prDiffsSha at freeze + run (hash-bind the cert-run scoring source) — #709 fold-2 enforcement
## Context

The #709 cert-corpus producer (#2224) stamps `controls.integrity.prDiffsSha` — a sha256 over the FULL canonical `pr-diffs.json` — into the lock, beside `fixtureSha`/`llmReplaySha`. But the freeze/run path does **not** yet re-derive + assert it: `spine-windtunnel.ts` reads `prDiffs` for scoring (`buildReplayCorpusProvider`) but never hash-binds it.

So `pr-diffs.json` (the answer-key's scoring source) is **stamped-but-runtime-unauthoritative** — a silent mutation of any row (advisory-window OR control) corrupts the answer key while `verifyControlIntegrity` (control-dirs-only) stays green. This is the codex-panel-caught integrity seam (fold-2), affirmed + sharpened by strategy-claude on `mmnto-ai/totem-strategy#709` (digest the FULL pr-diffs, not just controls; verify at freeze AND run).

## Scope (this slice)

- **At RUN** (`buildReplayCorpusProvider`, after `loadCertRunFixtures`): re-derive `sha256(canonicalStringify(prDiffs))` and hard-error on absent/mismatch vs `lock.controls.integrity.prDiffsSha` (mirrors the `llmReplaySha` L2 gate).
- **At FREEZE**: same re-derive + assert (mirroring the `fixtureSha` freeze check).
- Update existing cert-run test fixtures/locks to carry `prDiffsSha` (the ripple that motivated deferring this out of #2224).

## Why deferred from #2224

#2224 is the producer slice (no closing keyword). The enforcement ripples into existing cert-run test fixtures and is the piece closest to the integrity contract (now ruled). Split out as a clean enforcer slice; the producer's doc-comments are softened to future tense pointing here.

Related: `mmnto-ai/totem-strategy#709`, #2224, #2188.

satur8d:
## What

Enforce `controls.integrity.prDiffsSha` at the certifying **run** + **freeze** — the freeze/run hard-enforcement deliberately deferred out of the #709 producer (#2224). The producer stamps the digest; this makes the cert run's scoring source runtime-authoritative.

## Why

`pr-diffs.json` is the cert run's scoring source — loaded + Zod-parsed but otherwise unprotected, since `fixtureSha` covers only the control dirs. So a silent mutation of any row (advisory-window or control) would corrupt the disposition-derived answer key while the control-dir gate stayed green. Codex panel-caught (fold-2), affirmed + sharpened by strategy-claude (digest the FULL `pr-diffs.json`; verify at freeze AND run).

## Change

- **Run** (`buildReplayCorpusProvider`): re-derive `sha256` over the **exact on-disk `pr-diffs.json` bytes** and **hard-error** on absent/mismatch — beside the `llmReplaySha` L2 gate.
- **Freeze** (`freezeCommand`): a **warn-only** pre-merge check, mirroring the `fixtureSha` freeze behavior. The run is the authoritative hard gate; this freeze-warns / run-hard-fails split is the established pattern for `fixtureSha`/`llmReplaySha`.

## Tests / gate

- Threaded `prDiffsSha` through the run-corpus test `lockWith` + two new tests: the **absent-digest gate**, and a **tamper red** (a schema-valid but byte-different `pr-diffs.json` → trips the digest gate, not Zod). Windtunnel freeze tests unaffected (the warn is additive).
- cli **2725** · eslint **0** · format · `totem lint --base main` **PASS (0 errors)** · changeset (`@mmnto/totem` patch).

Closes #2225.

Related: `mmnto-ai/totem-strategy#709`, #2224, #2188.

### DIFF
--- a/.changeset/spine-2225-prdiffssha-enforcement.md
+++ b/.changeset/spine-2225-prdiffssha-enforcement.md
@@ -1,0 +1,10 @@
+---
+'@mmnto/totem': patch
+---
+
+fix(spine): enforce prDiffsSha at the certifying run + freeze (#709 fold-2 enforcement)
+
+The #709 producer stamps `controls.integrity.prDiffsSha` (a sha256 over the exact on-disk `pr-diffs.json` bytes), but nothing re-derived or asserted it — so the certifying run's scoring source was **stamped-but-runtime-unauthoritative**: a silent mutation of any row (advisory-window or control) would corrupt the disposition-derived answer key while `fixtureSha` (control-dirs-only) stayed green. This closes that seam (codex-panel-caught, affirmed + sharpened by strategy-claude: digest the FULL pr-diffs.json, verify at freeze AND run).
+
+- **Certifying run** (`buildReplayCorpusProvider`): re-derive the digest over the on-disk `pr-diffs.json` bytes and **hard-error** on absent/mismatch — beside the `llmReplaySha` L2 gate.
+- **Freeze**: a **warn-only** pre-merge check, mirroring the `fixtureSha` freeze behavior (the run is the authoritative gate).
--- a/packages/cli/src/commands/spine-cert-run-corpus.test.ts
+++ b/packages/cli/src/commands/spine-cert-run-corpus.test.ts
@@ -1,3 +1,4 @@
+import { createHash } from 'node:crypto';
 import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
@@ -110,10 +111,10 @@ function stage4(files: Record<string, string>): Stage4VerifierDeps {
   };
 }
 
-/** A minimal lock partial — the provider reads only the L2 hash + resolved corpus. */
-function lockWith(llmReplaySha?: string): WindtunnelLock {
+/** A minimal lock partial — the provider reads the L2 hash, the pr-diffs digest, + resolved corpus. */
+function lockWith(llmReplaySha?: string, prDiffsSha?: string): WindtunnelLock {
   return {
-    controls: { integrity: { llmReplaySha } },
+    controls: { integrity: { llmReplaySha, prDiffsSha } },
     corpus: { resolvedPrs: [{ pr: 1, mergeCommit: sha(1) }] },
   } as unknown as WindtunnelLock;
 }
@@ -171,6 +172,13 @@ async function fixtureHash(): Promise<string> {
   return hash;
 }
 
+/** sha256 of the exact on-disk pr-diffs.json bytes — the digest the run gate (#2225) asserts. */
+function prDiffsHash(): string {
+  return createHash('sha256')
+    .update(fs.readFileSync(path.join(gate1Dir, 'pr-diffs.json'), 'utf-8'), 'utf-8')
+    .digest('hex');
+}
+
 // ─── Tests ───────────────────────────────────────────
 
 describe('buildReplayCorpusProvider (run-path)', () => {
@@ -186,7 +194,7 @@ describe('buildReplayCorpusProvider (run-path)', () => {
       },
     });
 
-    const corpus = await provider(lockWith(hash));
+    const corpus = await provider(lockWith(hash, prDiffsHash()));
 
     expect(corpus.rules).toHaveLength(1);
     expect(corpus.provenanceByRule.get(corpus.rules[0]!.lessonHash)?.mergedPr).toBe(1);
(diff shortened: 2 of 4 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-1 dispositions (commit `8c0cccd0`)

All four accepted — a genuinely useful round (the CRLF one is a real cross-platform bug).

- **GCA — CRLF normalization** (`spine-cert-run-corpus.ts:197` + `spine-windtunnel.ts:122`): the digest hashed raw on-disk bytes, so a Windows checkout (git `autocrlf` LF→CRLF) would **spuriously fail** the run gate / warn at freeze — even though the producer stamps the digest over LF bytes. Now `.replace(/\r\n/g, '\n')` before hashing at both sites. Also switched the run gate to the existing `PR_DIFFS_FILE` constant per the same finding.
- **greptile — verify-bytes-then-parse** (`spine-cert-run-corpus.ts:205`): moved the `prDiffsSha` gate **before** `loadCertRunFixtures` (matching the `llmReplaySha` gate's position), with clean missing-file handling — so the integrity check guards the parse, and a future refactor that touches `prDiffs` earlier can't silently bypass it.
- **greptile — silent freeze skip** (`spine-windtunnel.ts:131`): freeze now **warns** when the lock declares `prDiffsSha` but `pr-diffs.json` is missing, surfacing the gap proactively at freeze instead of only later at run time.

Gate: cli 2725 · eslint 0 · format · `totem lint --base main` PASS (0 errors).

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-2 dispositions (commit `4d8f0238`)

### Accepted
- **CR — freeze: absent `prDiffsSha` in lock silently skipped** (`spine-windtunnel.ts`): a *certifying* lock with no `prDiffsSha` passed freeze clean, then the run hard-failed with no prior heads-up. Added the warn — **scoped to `phase === 'certifying'`** (harness locks legitimately lack the digest — it's additive-optional — so an unconditional `if (!prDiffsSha)` would spuriously warn on every harness freeze). Freeze now surfaces all three cases: mismatch, declared-but-missing-file, and absent-from-lock-for-certifying.

### Declined (rationale)
- **GCA `:133` — make freeze *fail* (`process.exitCode = 1`) on mismatch.** Warn-at-freeze mirrors the shipped `fixtureSha` freeze check (also warn-only on a source-of-truth control-dir mismatch); the certifying **run** is the authoritative hard gate — `buildReplayCorpusProvider` hard-throws on a `prDiffsSha` mismatch — so there's no silent data loss (the finding's premise). A genuine read failure already fails freeze via the thrown error. Making *only* `prDiffsSha` fail-at-freeze would diverge from `fixtureSha`; promoting freeze to a hard integrity gate is a separate, broader decision that should cover both digests, not a per-field change here.
- **CR — tamper test fixture not restored.** The premise (module-level `gate1Dir` reused across tests) doesn't hold here: `beforeEach` allocates a fresh `mkdtemp` dir, re-derives `gate1Dir`, and **rewrites `pr-diffs.json`** before every test (and `afterEach` `rm`s the dir). The tamper write lives only in that test's throwaway tmpdir — it cannot leak to later tests. A try/finally restore would guard state that's already fully reset per-test.

Gate: cli 2725 · eslint 0 · format · `totem lint --base main` PASS (0 errors).

satur8d:
/gemini review
@coderabbitai review

satur8d:
## Round-3 dispositions (commit `98177d38`) — the 🔵 tail

CR converged in round-2 (all 🔴/🟡 resolved, integrity model confirmed consistent). Closed both non-blocking 🔵 observations so nothing dangles:

- **🔵 `PR_DIFFS_FILE` constant in freeze** — exported the constant; freeze now uses it instead of the hardcoded `'pr-diffs.json'`, removing the rename hazard vs the run path (which already used it).
- **🔵 lockPath / gate1Dir co-location** — added a comment making the contract explicit: `pr-diffs.json` sits in the gate-1 dir beside the lock; freeze resolves it via `dirname(lockPath)`, the run via `opts.gate1Dir` (= the same dir).

No substantive items outstanding. Gate: cli 2725 · eslint 0 · format · `totem lint --base main` PASS (0 errors).

satur8d:
@coderabbitai review

satur8d:
## Disposition (commit `c2e701b2`) — CR 🟠 Major: digest check/use split

**Accepted — real finding.** The run gate hashed *one* read of `pr-diffs.json` while `loadCertRunFixtures` *re-read* it for the parse that actually feeds scoring — so the verified bytes weren't guaranteed to be the used bytes (a TOCTOU / check-use split; the digest didn't strictly cover what gets scored, which is the gate's whole purpose).

**Fix:** moved the verification *into* `loadCertRunFixtures` — it now reads `pr-diffs.json` **once**, verifies the digest on those bytes (CRLF→LF normalized), then parses the **same** bytes. The provider keeps the absent-from-lock precondition and passes the expected sha. Check and use are now the same read, so the integrity guarantee actually holds. The absent-gate + tamper red tests still trip (error messages preserved).

cli 2725 · eslint 0 · format · `totem lint --base main` PASS (0 errors).

satur8d:
@coderabbitai review

satur8d:
## Disposition — CR outside-diff 🔵 Trivial: lazy-load `node:crypto`

**Declined — false positive (cited guideline doesn't apply to node builtins).**

CR reads the rule as *"all imports in `packages/cli/**` must be lazy-loaded."* The actual guideline (`.gemini/styleguide.md:64`) is narrower:

> Static top-level imports **from `@mmnto/totem`** in CLI command files. … The core package pulls in **LanceDB and other heavy dependencies** — top-level imports slow CLI startup.

It's scoped to the heavy `@mmnto/totem` workspace package (startup cost), and `import type` is explicitly exempted — i.e. the rule targets *runtime-heavy workspace* imports, not free node builtins. The compiled shield rule that enforces it flags `@mmnto/totem`, not `node:*`.

Evidence from the codebase:
- `node:crypto` is static-imported in **6** `packages/cli/src/commands/*.ts` files, including the **shipped** `spine-cert-persist.ts`, `spine-llm-adapters.ts`, and `spine-llm-replay.ts` — all of which passed review + the shield rule.
- `node:fs` / `node:path` are static at the top of this very file (and 52 static node-builtin imports across the CLI commands).

Making `createHash` lazy while `fs`/`path` stay static would be inconsistent with the file and with shipped siblings. No change.

(No substantive items outstanding — the 🟠 TOCTOU from the prior round was the last real one.)

satur8d:
@coderabbitai review

Issue #2124, opened 2026-06-08
fix(extract): lesson extract mines bot findings blind to their dispositions — a declined finding became a lesson verbatim


satur8d:
On mmnto-ai/totem#2122, `totem lesson extract 2122 --yes` produced a lesson ('Normalize LanceDB nulls for Zod — use loose nullish checks') mined directly from GCA's high-severity finding that the round **declined with a code citation** (the premise was false: `sourceRepo` is constructor-injected behind a truthy guard at `lance-search.ts:209` and never round-trips through LanceDB/Arrow). Had it compiled, the rule would push a pattern justified by a hallucinated mechanism — the lesson pipeline laundering a refuted claim into doctrine. The extract pass should consult the round's disposition state (trap ledger / round-comment dispositions table) and either skip declined findings or invert them into the decline rationale (the TRUE lesson here is the citation). Caught manually at postmerge; lesson file deleted before compile. Adjacent: mmnto-ai/totem#2038 (disposition reason-codes), mmnto-ai/totem#2041 (bot adapters). Exhibit class: prescription-laundering — third GCA false-premise specimen of the week (allowBuilds on strategy#566, Arrow-null here).

PR #2230, opened 2026-06-22
feat(spine): Gate-1 label-deriver 5d-i — closed disposition taxonomy classifier (strategy#709)

satur8d:
## What

The first sub-slice (5d-i of 3) of the cert-corpus **ground-truth label deriver** (`mmnto-ai/totem-strategy#709`) — the last piece before a runnable Gate-1 cert run. This ships the **LLM-free closed disposition taxonomy classifier** (pure, core, zero-IO):

- `classifyDisposition(comments) → DispositionClass`
- `dispositionToLabel(class) → 'TP' | 'FP' | null`

under the RULED closed taxonomy (#709 c.4762840182 / c.4764980503):

| class | label |
|---|---|
| `accepted-fix` | **TP** |
| `declined-as-false-positive` | **FP** |
| `scope` · `defer` · `superseded` · `style` · `ambiguous` | **UNLABELED** (omit) |

## Why a new classifier (not the binary `bot-review-parser` primitive)

The existing primitive intentionally over-matches the **whole** decline vocabulary to `declined` ("a lost lesson, never a laundered one", #2124) — safe for keeping bad lessons out of mining, **unsafe as a cert FALSE-POSITIVE label**. A "declined, tracked for later / too broad for this PR" (a scope/defer decline) is **not** evidence the code is clean; mapping it to FP would FAIL a sound rule. So this classifier **partitions** the vocabulary: only an unambiguous correctness rebuttal ("false positive / by design / works as intended") becomes FP; scope/defer/superseded/style declines stay UNLABELED.

**Conservative-by-construction:** every conflict, impurity, or absent human disposition collapses to UNLABELED → the scorer's `needsAdjudication` → HONEST-NEGATIVE (Tenet 4 fail-soft for the answer key), never a guessed PASS. Resolution flags (`isResolved`/`isOutdated`) are the span-join's concern (5d-iii), so a bare `isResolved` thread is never TP.

## Cohort-panel-ratified

Pre-build panel (codex / agy / gemini, distinct lenses) + strategy-claude **RULED on #709 c.4764980503** — no contract change. This sub-slice implements:
- codex BLOCKING-3 (dedicated 5-way classifier, binary parser as input material only; conservative under-labeling)
- gemini (extend-via-new-classifier, abstention-as-UNLABELED, core purity)

## Tests

32 unit tests, including codex's exact **falsifying case** ("declined, too broad / tracked for later" never becomes FP), conflict/impurity → ambiguous, and bot-reply-is-not-a-disposition. Full gate green: build · 2757 core tests · ESLint · prettier.

## Scope

Part of the strategy#709 deriver — sub-slice 5d-i of 3; this **does not close** #709. Next: 5d-ii (held-out disposition fetch + freeze + integrity digest), 5d-iii (the deriver orchestrator + `derive-labels` command). Changeset is **consciously deferred** to 5d-iii (the user-facing command) — 5d-i is an internal core building block with no consumer yet.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -822,6 +822,8 @@ export type {
   CompileStageResult,
 } from './spine/compile.js';
 export { compileCandidate, runCompileStage } from './spine/compile.js';
+export type { DispositionClass, DispositionComment } from './spine/disposition-taxonomy.js';
+export { classifyDisposition, dispositionToLabel } from './spine/disposition-taxonomy.js';
 export type {
   DraftExtractor,
   ExtractStageDeps,
--- a/packages/core/src/spine/disposition-taxonomy.test.ts
+++ b/packages/core/src/spine/disposition-taxonomy.test.ts
@@ -181,4 +181,19 @@ describe('classifyDisposition — negation / over-match regressions (#2230 bot p
     // bare praise without a fix confirmation is not an accepted-fix (only "…, fixed" is).
     expect(classifyDisposition(thread('Good catch!'))).toBe('ambiguous');
   });
+
+  it('CR round-3 — `fixed` is negation-guarded: "not fixed yet" / "has not been fixed" are not TP', () => {
+    for (const reply of [
+      "This isn't fixed yet.",
+      'This has not been fixed.',
+      'Never fixed this.',
+      'No, that was not fixed.',
+    ]) {
+      expect(classifyDisposition(thread(reply))).not.toBe('accepted-fix');
+    }
+    // …but a genuine fix in its own clause still credits, even beside a negated one.
+    expect(classifyDisposition(thread('Not sure that was the cause, but I fixed it anyway.'))).toBe(
+      'accepted-fix',
+    );
+  });
 });
--- a/packages/core/src/spine/disposition-taxonomy.test.ts
+++ b/packages/core/src/spine/disposition-taxonomy.test.ts
@@ -1,0 +1,137 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  classifyDisposition,
+  type DispositionClass,
+  type DispositionComment,
+  dispositionToLabel,
+} from './disposition-taxonomy.js';
+
+// A held-out review thread: a bot finding + zero-or-more human replies.
+const BOT = 'coderabbitai[bot]';
+const HUMAN = 'satur8d';
+
+function thread(...replies: string[]): DispositionComment[] {
+  return [
+    { author: BOT, body: 'Potential null deref on `foo.bar` here.' },
+    ...replies.map((body) => ({ author: HUMAN, body })),
+  ];
+}
+
+describe('classifyDisposition — accepted-fix ⟹ TP', () => {
+  it.each([
+    'Fixed in the latest push.',
+    'Good catch, addressed.',
+    'Nice catch — done.',
+    'This has been fixed.',
+    'Applied the suggestion.',
+  ])('credits an unambiguous fix reply: %s', (reply) => {
+    expect(classifyDisposition(thread(reply))).toBe('accepted-fix');
+  });
+
+  it('maps accepted-fix → TP', () => {
+    expect(dispositionToLabel('accepted-fix')).toBe('TP');
+  });
+});
+
+describe('classifyDisposition — declined-as-false-positive ⟹ FP (correctness rebuttals only)', () => {
+  it.each([
+    'This is a false positive — the value is never null here.',
+    'Not a bug, the guard above covers it.',
+    'Works as intended.',
+    'This is by design.',
+    'Intentional — we want the eager throw.',
+    'Not applicable here.',
+    'This is correct.',
+  ])('credits an unambiguous correctness rebuttal: %s', (reply) => {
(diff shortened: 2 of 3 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-1 dispositions — pushed `94e537d1`

All findings converged on one real theme — **negation-blind / semantic over-matches** in the heuristic banks (the class the CI gates can't catch). Every accept moves in the **conservative (more-UNLABELED)** direction, which is the classifier's own principle.

| Finding | Disposition |
|---|---|
| **GCA `:95`** — `correct` in FP patterns inverts label | ✅ **Accepted** — dropped `correct` from both `not …` and `this is …`. "this is not correct" / "correct, I'll fix it" *agree* with the finding (a TP); matching `correct` made it FP. |
| **greptile `:78` P1** — negation-blind `addressed`/`applied`/`done` → false TP | ✅ **Accepted** — dropped the bare words; kept the confirmed-past `…has been (fixed\|addressed\|applied)` + canonical `fixed`. "not been addressed" / "Well done, but…" no longer mint TP. |
| **greptile `:91`** — `intentional(ly)` matches "not intentional" | ✅ **Accepted** — anchored to the affirmative `is intentional`. |
| **greptile `:112`** — `TODO` missing `/i` | ✅ **Accepted** — `/\bTODO\b/i`. |
| **GCA `:78`** — add `will fix` / `fixing` to FIX_PATTERNS | ❌ **Declined** — future/in-progress tense ("will fix later") is a **DEFER**, not a confirmed accepted-fix; adding it broadens the fix bank in the *unsafe* direction, against conservative-by-construction. The classifier deliberately credits only confirmed past-tense fixes. |

Also folded a non-blocking precision note from the cohort review (the loosest tokens `done` / `intentional` can over-match): both are addressed by the tightenings above. Residual semantic-negation robustness is deferred to the **5d-ii real-corpus measure-first validation** (run the banks against the actual held-out dispositions, demote any that misfire) rather than tightened blind here.

**+6 regression tests** for every accepted case (incl. "not correct", "not been addressed", "not intentional", "Well done but…", lowercase "todo"). Full gate green: build · 2762 core tests · ESLint · prettier.

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## CR round-2 dispositions — pushed `66b08606`

Both items accepted (both conservative-direction):

| Finding | Disposition |
|---|---|
| 🟡 **`good catch` / `nice catch` negation-blind** | ✅ **Accepted** — dropped both from `FIX_PATTERNS`. Same negation-blind class as the round-1 bare words ("not a good catch …" → false TP), and praise alone isn't a *fix confirmation* (acknowledging a real finding ≠ acting on it). The "good catch, fixed!" TP signal stays covered by `fixed`. |
| 🟡 **Step-6 comment inaccurate** | ✅ **Accepted** — corrected. The impure label-bearing cases are caught by the `fix \|\| fp` guard above; step 6 is reachable only with zero signals. Comment-only, behavior unchanged. |

Thanks for confirming the round-1 negation reasoning (the `…been (fixed\|addressed\|applied)` contiguity break, the `is intentional` anchor). +1 regression test ("not a good catch" → not a fix; bare "Good catch!" → ambiguous). Gate green: build · 38 taxonomy tests · ESLint · prettier.

satur8d:
@coderabbitai review

satur8d:
## CR round-3 disposition — pushed `e50d845e`

✅ **Accepted** — `/\bfixed\b/i` was the last negation-blind fix signal ("this isn't fixed yet" / "has not been fixed" → false TP). Fixed, and **generalized past the suggested adjacency guard**: CR's `not\s+fixed` catches "not fixed" but **not** the non-adjacent "has **not been** fixed". Instead, a clause-scoped `hasUnnegatedFix` credits a fix only from a clause that matches a FIX pattern **and** carries no negator:

- splits on contrastive `but`/`however`/`though`, so "not sure, **but** I fixed it" keeps the genuine fix;
- applied to the **FIX bank only** — the FALSE_POSITIVE bank has intrinsically-negative patterns ("not a bug", "not applicable") that a blanket negator guard would wrongly suppress;
- over-suppression only loses a true fix → UNLABELED → honest-negative (safe direction), never a false TP.

This closes the negation-blindness class across the whole classifier. +1 regression (4 negated forms → not TP; a real fix beside a negated clause still credits). Gate green: build · 39 taxonomy tests · 2764 core · ESLint · prettier.

satur8d:
@coderabbitai review

satur8d:
## CR round-4 disposition — pushed

✅ **Accepted** (Major) — the negation regressions asserted `.not.toBe('accepted-fix')` (or `.not.toBe('declined-as-false-positive')`), which would still pass if a phrase regressed to the *other* mislabel. Tightened **all** of them (not just the flagged line) to a shared `isUnlabeled` helper asserting `dispositionToLabel(...) === null` — fail-soft coverage against an accidental **TP or FP** label, in either direction. Test-only; classifier behavior unchanged. 39 taxonomy tests green, full suite 5749 + `totem lint` PASS.

satur8d:
@coderabbitai review

Issue #2112, opened 2026-06-07
fix(cli): Greptile absent from isBotComment — triage-pr/review surfaces blind to greptile-apps[bot] findings (#2041 exhibit)


satur8d:
## Empirical (PR #2109, 2026-06-07)

`totem triage-pr 2109` reported:

> Found 1 inline review comments
> No bot review comments found. Nothing to triage.

The 1 comment is a **P1 finding from `greptile-apps[bot]`** (the Stage-4 archive bug, now ticketed separately). The operator''s triage surface silently dropped the highest-severity finding on the PR; it was recovered only by a manual `gh api` read.

## Root cause

`isBotComment` (`packages/cli/src/parsers/bot-review-parser.ts:36`):

```ts
return lower.includes(''coderabbit'') || lower.includes(''gemini-code-assist'');
```

Greptile is not in the matcher at all, so every consumer of `isBotComment`/`detectBot` (triage-pr normalization, resolved-finding extraction, severity parsing) treats `greptile-apps[bot]` comments as human. `detectBot` has no `''greptile''` arm either — findings would need a severity parser for Greptile''s P1/P2/P3 badge format.

## Fix direction

Minimal: add `greptile` to `isBotComment` + a `''greptile''` arm in `detectBot` + a P-badge severity parser. Real fix: this is the named exhibit for #2041 (config-driven, swappable PR-review-bot adapters) — a hardcoded matcher is exactly what that issue exists to remove. Slot the minimal patch now; fold the adapter refactor into #2041.

Refs: #2041 (greptile-drop exhibit) · #2109 (the dropped P1) · doctrine bot-protocols.md § stack (Greptile is a sanctioned on-demand bot — the tooling must see it).

satur8d:
**Scope addition from the #2160/#2162 rounds (2026-06-12):** beyond Greptile-blindness, `triage-pr` also cannot see **CodeRabbit chat-mode replies** — when a review trigger is embedded in a combined comment, CR routes to its chat handler and posts the FULL review inside an issue comment (analysis-chain body), which no triage surface parses. This bit both lanes this session (manual `gh` fallback both times; on #2162 CR''s only findings lived there). Whatever adapter shape fixes Greptile should also normalize CR chat-mode bodies — or at minimum surface "CR replied in chat-mode, N findings unparsed" instead of silence.

satur8d:
[...]

PR #2233, opened 2026-06-22
feat(spine): Gate-1 label-deriver 5d-iii-i — disposition-derived answer key (derive-labels) (strategy#709)

satur8d:
## What

**Gate-1 ground-truth label deriver (strategy#709 5d-iii-i)** — produces `ground-truth-labels.json`, the cert run's disposition-derived **answer key** (`firingLabelId → TP|FP`). With this slice the certifying run is executable end-to-end (the missing input was the answer key the scorer grades against).

This is **5d-iii-i** of a two-PR cut (mirroring the producer #2224 → enforcement #2227 split):

- **5d-iii-i (this PR):** the deriver — produces + stamps the answer key.
- **5d-iii-ii (fast-follow):** run-side `groundTruthSha` verify-then-parse in `loadCertRunFixtures` + a `freeze` warn.

## How

- **Core span-join** (`deriveLabelsFromDispositions`, pure/zero-LLM): a corpus firing binds to a held-out disposition on the same PR by **added-line content** — the firing's normalized `matchedLine` must equal an ADDED (`+`) hunk row (context/removed rows + `+++`/`---` headers ineligible; `line`/`originalLine` are locator hints only — never physical-line equality). 0-or->1 bound dispositions ⟹ omit (ambiguity never labels). The bound thread is classified through the 5d-i taxonomy (`accepted-fix → TP`, `declined-as-false-positive → FP`, else UNLABELED → omitted → the scorer's `needsAdjudication`).
- **Positive controls:** TP **only** for the declared `(pr, targetRuleId)` target firing; incidental non-target firings are omitted + reported (never laundered TP). Negatives never label.
- **CLI** `totem spine windtunnel derive-labels`: hard-gates `controls.integrity.corpusDispositionsSha` (verify-then-parse single read, CRLF→LF) **before** any derivation — a tampered disposition would silently flip a label (the #2224/#2225 lesson); stamps `controls.integrity.groundTruthSha` over the canonical answer-key bytes; emits a deterministic data-quality diagnostic (disposition density / UNLABELED rate / per-rule counts). Requires the lc clone (no post-image ⟹ silently-empty key ⟹ fail loud).
- **Drift kill (panel "do both"):** firings are enumerated through a **shared** certifying firing-setup (`buildCertifyingFirings`) + corpus-assembly path (`assembleCertifyingCorpus` + `buildGate1Stage4Deps`) that `runCertifyingEngine` now also calls — so the answer-key labelIds match the run's `buildFirings` labelIds byte-for-byte. `normalizeMatchedLine` is exported + reused so the span-join can't drift from the firing's own normalization.
- **Circularity guard:** the deriver never calls `loadCertRunFixtures` (it reads the file we emit); it reuses the shared corpus path with `skipGroundTruth: true`.

## Tests

- **Core** (`derive-labels.test.ts`, 13): added-line-only binding (vs context-only / dual-occurrence / `+++` header / path mismatch / trailing-ws drift), ambiguity (0 / >1), taxonomy projection (TP/FP/UNLABELED), control kinds (negative skip, positive target-vs-incidental), diagnostics + evidence-refs + determinism.
- **CLI** (`spine-derive-labels.test.ts`, 5): the panel's minimal-fixture **equivalence** test (deriver labelIds ⊆ the run's `buildCertifyingFirings` labelIds + determinism) and the integrity gates (no-lc-dir, `corpusDispositionsSha` missing, `corpusDispositionsSha` mismatch).
- Refactor regression: existing `windtunnel-firing` / `windtunnel-lock` / `spine-cert-run-corpus` / `spine-windtunnel` suites green (the run now routes through the extracted shared setup).
- Deferred to 5d-iii-ii: tampered-`ground-truth-labels.json` hard-fail in `loadCertRunFixtures` (the run-side `groundTruthSha` verify lives there).

## Changeset

One consolidated `@mmnto/totem` minor covering the full **5d** arc (5d-i / 5d-ii landed changeset-deferred), so the deriver ships as a single Version Packages entry.

## Provenance

Cohort pre-build panel **PASS 3/3** (codex contract / agy build / gemini tenet, distinct lenses); operator greenlit. Advances strategy#709 5d; part of the Convergent Spine (#2199). Not a full close of any totem issue.

### DIFF
--- a/.changeset/spine-gate1-label-deriver.md
+++ b/.changeset/spine-gate1-label-deriver.md
@@ -1,0 +1,13 @@
+---
+'@mmnto/totem': minor
+---
+
+feat(spine): Gate-1 ground-truth label deriver (strategy#709 5d)
+
+Produces `ground-truth-labels.json` — the cert run's disposition-derived ANSWER KEY (`firingLabelId → TP|FP`) — completing the Gate-1 label deriver. With it the certifying run is executable end-to-end. Cohort-panel-ratified (codex/agy/gemini, distinct lenses) before any code. Ships the full 5d arc as one release (5d-i/5d-ii landed changeset-deferred):
+
+- **5d-i** — closed disposition taxonomy classifier (`classifyDisposition` / `dispositionToLabel`): `accepted-fix → TP`, `declined-as-false-positive → FP`; every other class (scope / defer / superseded / style / ambiguous) → UNLABELED (omitted, so the scorer routes the firing to `needsAdjudication`).
+- **5d-ii** — held-out disposition source + `controls.integrity.corpusDispositionsSha` (`totem spine windtunnel fetch-dispositions`): freezes the corpus PRs' span-anchored review threads as the answer-key provenance.
+- **5d-iii** — the deriver: `deriveLabelsFromDispositions` (pure core span-join) + `totem spine windtunnel derive-labels`. A corpus firing binds to a disposition on the same PR by **added-line content** — the firing's normalized `matchedLine` must equal an ADDED (`+`) hunk row (context/removed rows and file headers ineligible; never physical-line equality); 0-or-multiple matches omit (ambiguity never labels). Positive controls label TP only for the declared `(pr, targetRuleId)` target; incidental non-target firings are omitted and reported; negatives never label. The deriver hard-gates `corpusDispositionsSha` before deriving and stamps `controls.integrity.groundTruthSha` over the emitted answer key (the run-side verify + freeze-warn land in the 5d-iii-ii fast-follow, mirroring the producer → enforcement split).
+
+Firings are enumerated through a shared certifying firing-setup (`buildCertifyingFirings`) + corpus-assembly path (`assembleCertifyingCorpus`, `buildGate1Stage4Deps`) that the certifying run also calls, so the answer-key labelIds match the run's `buildFirings` labelIds byte-for-byte — the drift surface the panel flagged, locked by an equivalence test.
--- a/packages/cli/src/commands/spine-cert-run-corpus.ts
+++ b/packages/cli/src/commands/spine-cert-run-corpus.ts
@@ -176,12 +176,16 @@ export function buildGate1Stage4Deps(
 ): Stage4VerifierDeps {
   if (lcDir) {
     return {
+      // The `--` separator keeps the revision unambiguous from any path args and blocks
+      // arg injection (CR coding guideline) — this shared path decides which files
+      // Stage-4 sees for BOTH the run and the deriver, so rev/path ambiguity here would
+      // shift archived-rule decisions and the derived labels.
       listFiles: async () =>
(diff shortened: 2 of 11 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Bot review disposition — pushed `608a9764`

All 12 findings (CR ×6, GCA ×4, greptile ×2) triaged. Inline set re-pulled directly from `pulls/2233/comments` (greptile's 2 inline P2s don't surface in `triage-pr` — the known #2112 blind spot — so they were read manually).

### ✅ Accepted + fixed
| Finding | Fix |
|---|---|
| **CR** — `addedHunkLines` `startsWith('+++')` drops a real added line `++i` (diff row `+++i`) | Tightened to the actual file-header form `/^\+\+\+ /` (trailing space). A genuine correctness bug. + regression test. |
| **greptile** — `dispByPr` silent last-wins on a duplicate-PR entry | Fail-loud throw (Tenet-4 structural guard) + test — a producer/edit bug can no longer silently drop a PR's threads. |
| **GCA** — `JSON.parse` of lock + corpus-dispositions unwrapped (raw `SyntaxError`) | Wrapped via a `parseJson` helper → structured `TotemError('CONFIG_INVALID')`, matching the CLI-layer convention in `loadCertRunFixtures` / `fetch-dispositions`. (Used `TotemError`, not `TotemParseError` — the latter is the styleguide's *core compile/parse* layer; this is the CLI runtime layer.) |
| **GCA** — redundant `addedHunkLines` recompute in the nested loop | Memoized per disposition thread via a `WeakMap` (`getAddedLines`) — parsed once per thread, not once per firing. |
| **CR** — missing git `--` separator on the shared Stage-4 reads | Added `--` to the `ls-tree` / `show` argv (verified both forms accept it). |

### 🚫 Declined (with rationale)
- **CR — "lazy-load all `cli/**` imports" (×2):** the heavy `@mmnto/totem` barrel is *already* dynamically imported; the remaining static imports are node builtins + sibling `./*.js` command modules, which is the **established spine pattern** — `spine-fetch-dispositions`, `spine-cert-materialize`, and `spine-cert-run-corpus` all import siblings statically and dynamic-import only `@mmnto/totem`. `totem lint --base main` is green (the rule is unenforced by our own gate). Same disposition as #2218.
- **greptile — thread `seedClassesProvided` into the deriver:** it's a fold-I §7 ledger **attestation** wired only into the extract-stage ledger — it does **not** enter rule compilation or the content-based `labelId`, so it can't break the byte-identical guarantee; and the run's `buildReplayCorpusProvider` defaults it `false` identically. Threading it deriver-only would add an *asymmetric* knob (deriver could diverge from the run). Pinned `false` with a clarifying comment; thread it through **both** if it ever becomes a real RunOption.

Gates green locally: core+cli builds, full suites (2784 + 2747) + 2 new cases, ESLint 0-errors, prettier, `totem lint --base main` PASS.

satur8d:
/gemini review
@greptileai review
@coderabbitai review

Issue #2237, opened 2026-06-23
Gate-1 cert run #1 — post-run backlog: 3 CLI papercuts + PART-2b capability + 2 decisive-re-run levers (strategy#709)


satur8d:
## Context

The first end-to-end Gate-1 certifying run (strategy#709) executed live on 2026-06-23: materialize → fetch-dispositions → record (Sonnet-4-6) → freeze → derive-labels → run. Verdict: **honest-negative** (FAIL/no-claim, nonVacuity false, **precision 1.0 / 0 FP**, C3 @ `6acac0b3`). Machinery validated, fails legibly. Cause (strategy-ruled): **(c) low-extraction-yield (dominant) + (b) class-mismatch — NOT thin-train, NOT thesis-narrow**; the thesis "PR-mining mints legitimate rules" is **supported** (3 real review-caught rules minted; failure is yield + class-coverage). Corpus on branch `spine/gate-1-cert-709`.

This issue tracks the build backlog the run earned.

## CLI papercuts (fix-forward — surfaced because the pipeline had never run end-to-end)

1. **`spine record` does not `loadEnv`.** ~18 commands call `loadEnv(cwd)`; the spine commands do not, so `ANTHROPIC_API_KEY` from `.env` (the `.env`-only convention) isn't visible to `record` — it fail-closed "no credential resolved" until the key was exported manually. Fix: `loadEnv` in the spine command path.
2. **`llmReplaySha` is a manual wire, not sealed at freeze.** `record` emits the hash with "wire into lock controls.integrity.llmReplaySha"; `freeze` validates but does not auto-stamp it (had to hand-edit the lock). The two-phase design says freeze seals it. Fix: `freeze` computes + stamps `llmReplaySha` from the frozen `llm-replay.v1.json`.
3. **Firing detail not persisted on a FAIL verdict.** The run report carries only the `needsAdjudication` labelId hashes, not the firing records (rule, pr, file, matched-line) — blocks blind-by-pattern adjudication observability. Fix: persist firing records in the run report regardless of verdict.

## Capability (deferred this run — strategy#709 Flag 4 = defer)

[...]

PR #2240, opened 2026-06-23
feat(spine): extractor NO-DRAFT cause-tags — DraftResult port + replay migration (strategy#709 yield-fix α)

satur8d:
## What

Slice **α** of the strategy#709 extractor-yield fix — **observability-first**. Makes the Gate-1 extract→draft loss *measurable* before any yield change.

The first Gate-1 cert run was honest-negative, dominated by the extractor returning `[]` on threads carrying mintable classes (518/519/520/537/553). A bare `[]` conflated ≥6 distinct causes with opposite fixes — and the frozen replay stored only the parsed `[]`, discarding the cause at record time. We can't tune what we can't see.

This widens the `DraftExtractor` port to a **`DraftResult` (`{ drafts, noDraftCause? }`)** — the extract-stage twin of the classifier's `dispositionSource` (a **non-FM Tenet-19 diagnostic**).

## Changes

- **`NoDraftCause`** (`ledgers.ts`): a pinned-order, mutually-exclusive partition over every empty path — `invoke-error` (adapter) / `empty-output` / `none-sentinel` / `unparseable-shape` / `non-array` / `all-filtered` (parser) — plus a replay-migration-only `legacy-unknown`. `DropLedgerEntry.noDraftCause` is additive-optional.
- **`DraftResultSchema`** (`extract.ts`): `{ drafts, noDraftCause? }` with a "cause **iff** empty" refine. `runExtractStage` parses the port result at the boundary (mirrors `ClassifierResultSchema.parse`) — a contract-violating result fails loud; the cause lands on the empty-draft drop.
- **Adapter** (`spine-llm-adapters.ts`): `parseExtractorOutput` returns the cause per branch; `LiveDraftExtractor.draft` tags `invoke-error` on a caught invoke failure.
- **Replay** (`spine-llm-replay.ts`): records the full `DraftResult`; a **backward-compatible union reader** (`string[] | DraftResult`) keeps the committed cert-#1 fixture loadable + byte-identical, normalizing a legacy row on read (empty → `legacy-unknown`).

Pure instrumentation — **no behavior change** to drafts or drops. The NO-DRAFT cause becomes observable on the next re-record.

## Provenance

Panel-ratified **3/3 PASS-to-build** (codex contract / agy build / gemini tenet). Folds incorporated: separate `reviewBotIdentity` axis (deferred to β), the pinned cause-order, the `[""]`→`all-filtered` adjudication, the legacy-`string[]` union reader, and the local↔core parity test.

## Gate

build (core+cli) · core 2785 / cli 2760 tests (incl. one red/green per `NoDraftCause` branch + the local↔core schema parity test + the legacy-fixture migration test) · ESLint 0 errors · `totem lint --base main` PASS · format:check · changeset (minor).

Part of #2237 — that issue stays open (β normalization + γ RESOLVED-eligibility follow; the single live re-record lands after γ).

### DIFF
--- a/.changeset/spine-extractor-cause-tags.md
+++ b/.changeset/spine-extractor-cause-tags.md
@@ -1,0 +1,15 @@
+---
+'@mmnto/totem': minor
+---
+
+spine: extractor NO-DRAFT cause-tags (strategy#709 yield-fix slice α) — make the moat measurable
+
+The first Gate-1 cert run was honest-negative, dominated by the extract→draft step returning `[]` on threads carrying mintable classes. A bare `[]` conflated ≥6 distinct causes (model declined / parser rejected a valid draft / transient invoke failure) that the frozen replay discarded at record time, so the loss was un-diagnosable.
+
+This slice widens the `DraftExtractor` port to a `DraftResult` (`{ drafts, noDraftCause? }`) — the extract-stage twin of the classifier's `dispositionSource`, a non-FM Tenet-19 diagnostic:
+
+- **`NoDraftCause`** (core): a pinned-order, mutually-exclusive partition over every empty path — `invoke-error` (adapter) / `empty-output` / `none-sentinel` / `unparseable-shape` / `non-array` / `all-filtered` (parser) — plus a replay-migration-only `legacy-unknown`.
+- **Boundary parse** in `runExtractStage` (mirrors `ClassifierResultSchema.parse`): a contract-violating result (cause-without-empty / empty-without-cause) fails loud; the recorded `noDraftCause` lands on the empty-draft drop-ledger entry.
+- **Replay** records the full `DraftResult`; a backward-compatible union reader keeps pre-cause-tag fixtures (bare `string[]`) loadable + byte-identical, normalizing them on read (empty → `legacy-unknown`).
+
+Pure instrumentation — no behavior change to drafts or drops; the NO-DRAFT cause becomes observable on the next re-record. Observability-first prerequisite for the β (normalization) and γ (RESOLVED-eligibility) yield levers.
--- a/packages/cli/src/commands/spine-cert-corpus.test.ts
+++ b/packages/cli/src/commands/spine-cert-corpus.test.ts
@@ -55,8 +55,8 @@ function fakeSource(): ReviewThreadSource {
 
 function fakeExtractor(dsls: string[]): DraftExtractor {
   return {
-    async draft(): Promise<string[]> {
-      return dsls;
+    async draft() {
+      return dsls.length === 0 ? { drafts: dsls, noDraftCause: 'all-filtered' } : { drafts: dsls };
     },
   };
 }
--- a/packages/cli/src/commands/spine-cert-e2e.test.ts
+++ b/packages/cli/src/commands/spine-cert-e2e.test.ts
@@ -71,8 +71,8 @@ function frozenSource(): ReviewThreadSource {
 
 function liveExtractor(): DraftExtractor {
   return {
-    async draft(): Promise<string[]> {
-      return [REGEX_DSL];
+    async draft() {
+      return { drafts: [REGEX_DSL] };
     },
   };
 }
--- a/packages/cli/src/commands/spine-cert-record.test.ts
+++ b/packages/cli/src/commands/spine-cert-record.test.ts
@@ -59,8 +59,8 @@ function fakeDeps(): RecordDeps {
       },
     } satisfies ReviewThreadSource,
     liveExtractor: {
-      async draft(): Promise<string[]> {
-        return [REGEX_DSL];
+      async draft() {
+        return { drafts: [REGEX_DSL] };
       },
     } satisfies DraftExtractor,
     liveClassifier: {
--- a/packages/cli/src/commands/spine-cert-run-corpus.test.ts
+++ b/packages/cli/src/commands/spine-cert-run-corpus.test.ts
@@ -91,8 +91,8 @@ function fakeSource(): ReviewThreadSource {
 }
 function fakeExtractor(): DraftExtractor {
   return {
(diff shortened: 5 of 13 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Contract lens (strategy#709 α) — conformant; FM-integrity verified at the diff

This is the agreed first-PR design lens (strategy owns the contract; this is the contract half of α, not an engineering re-review — the panel's 3/3 covers that). I verified the load-bearing claim against the diff, not the description.

**Non-FM diagnostic / zero behavior change — confirmed in `extract.ts`:** `runExtractStage` reads `draftResult.drafts` and proceeds identically (`draftBodies` unchanged); the empty-draft path still fires `drop(pr, 'unparseable', …)` — same reasonCode, same FM-i-creditable loud drop — with `noDraftCause` added **additively** (optional spread). FM(i) creditability, FM(f) seed-blindness, and the Tenet-4 no-swallowing-catch invariant are all intact. The cause-tag changes what we can **see**, not what **passes** — which is the whole point of an observability-first slice.

**Adjudicates B-b — confirmed:** the 6-way `NoDraftCause` partition cleanly separates the **model-side** causes (`none-sentinel` / `empty-output`) from the **parser-side** (`unparseable-shape` / `non-array`) from `invoke-error` and `all-filtered`. That is exactly the split the B-b reframe needs — the re-record will distinguish "model declined" from "parser rejected a valid draft" instead of inferring it from a bare `[]`.

**Boundary parse** (`DraftResultSchema.parse`, cause-iff-empty refine) mirrors the `ClassifierResultSchema.parse` precedent cited in the ruling — a contract-violating port fails loud before the ledger. Faithful.

**Replay union reader** keeps the frozen cert-#1 fixture byte-identical (`legacy-unknown` on a legacy `[]`). Correct for the single re-record path.

**Contract verdict: clean.** α is the right non-negotiable-first step and a faithful diagnostic. My step-4 class-straddle controls re-engage at the re-record after γ.

**One β-watch (not a blocker for α):** the drop reasonCode stays `'unparseable'` for *all* empty-draft drops while `noDraftCause` now carries the precise reason — so a `none-sentinel` (model legitimately declined) sits under reasonCode `'unparseable'`, a slight misnomer. Harmless here (the cause-tag holds the truth); worth aligning the coarse reasonCode when β starts acting on the causes.

satur8d:
## Bot round 1 — dispositions (pushed `a496e33e`)

Triaged across all locations (inline + greptile "Comments Outside Diff" + CR issue-comment). CR posted no actionable findings. 4 findings, all addressed:

**✅ Fixed**
1. **GCA (high) · `extract.ts` — dead `?? 'unknown'` fallback.** Correct: in the empty-drafts branch the boundary `DraftResultSchema.parse` + the "cause iff empty" refine guarantee `noDraftCause` is present, so the fallback can never fire. Replaced with `!` (assert, don't defend).
2. **greptile (P2, Comments-Outside-Diff) · stale JSDoc.** Fixed the `ReplayDraftExtractor` doc that still described a `string[]` return — now describes the `DraftResult` + legacy normalization.

**✅ Fixed in spirit (kernel accepted, mechanism adjusted)**
3. **greptile (P2, inline) · `legacy-unknown` has no code-level enforcement.** Valid kernel, but the suggested fix (reject `legacy-unknown` at the boundary) is unsafe: `runExtractStage`'s parse is **shared with the replay path**, which *legitimately* produces `legacy-unknown` when normalizing a pre-cause-tag bare-`string[]` fixture row — rejecting it there would break legacy-fixture replay. The invariant is enforceable only on the **live** side, which mints it nowhere by construction. Locked that with a regression test (`parseExtractorOutput` + `LiveDraftExtractor` never emit `legacy-unknown`).

**↩️ Declined (with grounding)**
4. **GCA (high) · `DraftResultLocalSchema` refine not byte-identical to core** (missing `path`, shorter message). The local↔core contract is **accept/reject parity** (locked by the parity test), not message identity — and the established sibling `ClassifierResultLocalSchema` deliberately uses the same concise local message with no `path`. Matching byte-for-byte would diverge from that precedent; a strict-message pass would belong across *both* local schemas, out of α scope. Tightened the comment so it no longer implies byte-identity.

Gate green on the new HEAD: build (core+cli) · core 31 / cli adapters+replay 71 (incl. +2 legacy-unknown locks) · ESLint 0 errors · `totem lint --base main` PASS · format:check.

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Bot round 2 — converged (no new findings)

Re-review of `a496e33e`, triaged across all locations:

- **CodeRabbit:** "Round 2: all round-1 findings addressed; no new issues." ✅
- **greptile:** Confidence **5/5 — "Safe to merge — purely additive instrumentation that does not alter draft selection, drop decisions, or ledger structure for existing records."** ✅ Round-1 OOD stale-doc fix confirmed; no Comments-Outside-Diff this pass.
- **GCA · `spine-llm-replay.ts` refine** — a **re-anchored re-raise of round-1 #4**, **held (declined)**. The local↔core contract is accept/reject **parity**, locked by the parity test — which asserts `.success` equivalence, *not* `message`/`path` identity — and the `ClassifierResultLocalSchema` sibling deliberately uses the same concise local message with no `path`. No new argument since round 1.
- **greptile · `ledgers.ts` `legacy-unknown`** — byte-identical re-raise of round-1 #3, **held**. Live-side invariant locked by test; the suggested boundary-rejection stays declined (the shared `runExtractStage` parse must *accept* `legacy-unknown` for the replay path, or legacy-fixture replay breaks).

**No code changes this round** — substance is conformant (CR clean + greptile 5/5); the two remaining inlines are stale re-anchors of already-grounded round-1 dispositions. Converged from my side.

Issue #2192, opened 2026-06-17
fix(triage-pr): bot-comment classifier misses greptile-apps[bot] — under-reports as 'nothing to triage'


satur8d:
## Problem

`totem triage-pr <PR>` under-reports bot review comments. On #2190 it fetched the inline review comments (`Found 2 inline review comments`) but classified **none** as bot findings and printed:

```
Found 2 inline review comments
No bot review comments found. Nothing to triage.
```

…when both inline comments were `greptile-apps[bot]` findings (a P1 semantic bug + a P2 schema gap). The greptile review **summary** (an issue-comment) and CodeRabbit's review (6 findings) were also not surfaced.

## Impact

A reviewer trusting `triage-pr`'s "nothing to triage" would have missed real, actionable findings. The findings were only caught by re-pulling the raw surfaces (`pulls/{n}/comments`, `pulls/{n}/reviews`, `issues/{n}/comments`) — the exact "don't trust the summary" failure mode the review-reply doctrine warns about.

## Likely cause

- The bot-author classifier/allowlist doesn't recognize `greptile-apps[bot]` (it found the inline comments but didn't tag them as bot-authored).
- It appears to read only `pulls/{n}/comments` (inline) and not `pulls/{n}/reviews` (review bodies — where GCA/greptile/CR post summaries + actionable lists) or `issues/{n}/comments` (where greptile's "Comments Outside Diff" hides in a `<details>`).

## Expected

- Recognize `greptile-apps[bot]`, `coderabbitai[bot]`, `gemini-code-assist[bot]` as bot authors.
- Fetch + categorize across all three surfaces (inline comments, review bodies, issue-comment summaries), including greptile's "Comments Outside Diff" details block.
- **Never** print "Nothing to triage" when inline bot comments were already found — at minimum surface what was fetched.

## Repro

`totem triage-pr 2190` during the review window → "Found 2 inline review comments" → "No bot review comments found. Nothing to triage." The two were `greptile-apps[bot]` P1 (`windtunnel-scorer.ts`) + P2 (`windtunnel-lock.ts`).

Observed while triaging the #2190 bot round.

satur8d:
[...]

PR #2249, opened 2026-06-25
feat(parity): value-equality detector for bot-review-config rows (strategy#738 Slice A)

satur8d:
## What

**Slice A of the Proposal 296 §13 bot-config promotion** (mmnto-ai/totem-strategy#738): a new `value-equality` parity detector that promotes the `bot-review-configs` manifest rows from `attestation` to a present-level **mechanical scalar check**.

- **`detectValueEqualityContract`** (core, `parity-detect.ts`) — reads a scalar at a dotted path in the consumer's on-disk config and compares it, **typed** (never blanket-stringified) and **zero-network**, against the row's own `expected-value-or-derivation` (the canonical, per the #738 Q1 ruling).
- **CLI route + `valueEqualityFieldsFor` registry** (`doctor-parity.ts`) — routes on `manifestation: value-equality` *before* the tractability dispatch (the capability-probe pattern); the registry supplies file + path-**segments** + format, never the expected value.
- **`value-equality` rung** added to the §6(a)2 ladder (`PARITY_MANIFESTATIONS`).

5 rows after the #738 Q3 two-row gca split: `cr-profile`, `cr-on-demand`, `gca-code-review`, `gca-summary`, `greptile-on-demand` (over `.coderabbit.yaml` / `.gemini/config.yaml` / `greptile.json`).

## Design (cohort panel + settled rulings)

Pre-build cohort panel — **totem-codex (contract), totem-agy (build/test), totem-gemini (tenet/arch) — all PASS**; folds applied:
- **Typed expected, never blanket `String()`** (codex): `true`/`false` → boolean, else string. YAML `enabled: false` (bool) matches `"false"`; the *string* `"false"` correctly **warns** (no laundering an ambiguous string into a bool claim).
- **Honest-absent taxonomy:** file-absent → `skip` (applicable-but-missing scaffold, mirroring `detectVersionPinnedContract`; flips to `warn` once consumers are verified) · path-absent / traversal-through-non-object / mismatch → `warn` · unparseable → `unknown` (not a config-validity detector) · match → `pass`.
- **Path segments, not split strings** (codex) — avoids a key-with-dot ambiguity.
- **GCA shape corrected:** verified against the live config — the switches nest under a **top-level `code_review:` block** (`code_review.pull_request_opened.{code_review,summary}`), not the bare `pull_request_opened.*` a panel fixture assumed.

Claim-class bound: a `pass` asserts only "this dotted path holds this scalar" — never enforcement / loaded / on-demand. Never emits `fail` (the CLI edge owns `--strict`); never networks (Tenet 6).

## Coupling

Single-writer split (strategy#738): this engine is the totem-core lane. The **manifest row flip** (`attestation → value-equality`) is strategy's lane in `doctrine/parity-manifest.yaml`, **coupled to this PR's merge** so the manifest never names a rung the engine hasn't shipped (the #2192 doctrine-coupling shape). I send the merge-ping when this is approved.

## Verify

- Build: `pnpm build` green (10/10 turbo tasks).
- Tests: full suite green — **+17 core** detector cases (typed compare, the 5 rows, honest-absent taxonomy, CRLF, unparseable→unknown, consumers-scope, never-fail, claim-class) **+5 CLI** wiring cases (pass/warn/blocking-tag/skip/unhandled-stub).
- `pnpm format:check` clean · ESLint 0 errors · `totem lint --base main` PASS (0 errors).
- Changeset included (minor, core + cli).

Scope: **Slice A only.** B (managed-block shared-clause), C (`copilot-instructions.md` generated path), D (capability-probe Yellow) follow as strategy authors their rows.

### DIFF
--- a/.changeset/value-equality-parity-738.md
+++ b/.changeset/value-equality-parity-738.md
@@ -1,0 +1,8 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+---
+
+feat(parity): value-equality detector for bot-review-config rows (strategy#738 Slice A)
+
+Adds `detectValueEqualityContract` (core) + a `manifestation: value-equality` route and `valueEqualityFieldsFor` registry (CLI `doctor --parity`), promoting the `bot-review-configs` manifest rows from `attestation` to a present-level mechanical scalar check. Reads a scalar at a dotted path in the consumer's on-disk config (`.coderabbit.yaml` / `.gemini/config.yaml` / `greptile.json`) and compares it — typed (never blanket-stringified), zero-network — against the row's own `expected-value-or-derivation`. Honest-absent taxonomy: file-absent → `skip` (scaffold), path-absent/mismatch → `warn`, unparseable → `unknown`. Adds the `value-equality` rung to the §6(a)2 ladder. The strategy-owned manifest row flip couples to this engine on merge.
--- a/packages/cli/src/commands/doctor-parity.test.ts
+++ b/packages/cli/src/commands/doctor-parity.test.ts
@@ -407,6 +407,23 @@ describe('checkParity — value-equality wiring (strategy#738 Slice A)', () => {
     expect(line.status).toBe('skip');
     expect(line.message).toMatch(/not yet implemented/i);
   });
+
+  it('SKIP — not a consumer comes from the core self-guard (no CLI consumersSkip)', async () => {
+    writeConfig(`${BASE_CONFIG}orient:\n  parityManifest: m.yaml\n`);
+    // Scope the row to a different repo; this tmpDir repo is not in `consumers`.
+    const scoped = VALUE_EQUALITY_MANIFEST_YAML.replace(
+      'tracking-issue: mmnto-ai/totem-strategy#501\n',
+      'tracking-issue: mmnto-ai/totem-strategy#501\n    consumers:\n      - some-other-repo\n',
+    );
+    writeManifest('m.yaml', scoped);
+    writeCoderabbit('chill'); // would be a drift warn if this repo were a consumer
+
+    const { results, blockingDriftIds } = await checkParity(tmpDir);
+    const line = results.find((r) => r.name === 'Parity: cr-profile')!;
+    expect(line.status).toBe('skip');
+    expect(line.message).toMatch(/not in consumers|repo id unresolvable/i);
+    expect(blockingDriftIds).toHaveLength(0);
+  });
 });
 
 // ─── mechanical skills detection wiring (#2073) ─────────
--- a/packages/cli/src/commands/doctor-parity.test.ts
+++ b/packages/cli/src/commands/doctor-parity.test.ts
@@ -316,6 +316,99 @@ describe('checkParity — version-pinned wiring', () => {
   });
 });
 
+// ─── value-equality detection wiring (strategy#738 Slice A) ─────────
+
+/** A manifest with the cr-profile value-equality contract (no consumers = applies). */
+const VALUE_EQUALITY_MANIFEST_YAML = `schema-version: 1
+status: scaffold
+contracts:
+  - id: cr-profile
+    dimension: bot-review-configs
+    canonical-source: mmnto-ai/totem-strategy:.coderabbit.yaml#reviews.profile
+    detection-method: file-value-equality
+    expected-value-or-derivation: assertive
+    tractability: mechanical
+    manifestation: value-equality
+    tracking-issue: mmnto-ai/totem-strategy#501
+`;
+
+/** Same row marked blocking — exercises the --strict promotion seam through value-equality routing. */
+const BLOCKING_VALUE_EQUALITY_MANIFEST_YAML = VALUE_EQUALITY_MANIFEST_YAML.replace(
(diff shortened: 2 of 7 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Review disposition (round 1) — all three accepted, fixed in 02120cb

Good round — these were real, thanks.

**CodeRabbit (Major) — silent YAML fallback on unsupported format: ✅ FIXED.**
The parse ternary treated *any* non-`json` format as YAML, which contradicted this detector's own docblock ("unrecognized format → `unknown`, never a silent mis-parse"). Since `detectValueEqualityContract` is on the public core surface, a JS caller or a future registry typo could have minted a false `pass`/`warn` — and the catch's `.toUpperCase()` assumed a string. Added an up-front format guard returning `unknown` for anything but `yaml`/`json`, and the unparseable label is now a literal.

**GCA (High) — swallowed parser error: ✅ FIXED.**
The unparseable verdict now appends the parser's own first error line (`split('\n')[0]`, newline-only per your reference so a `.`-laden path survives) — a syntax issue is now locatable from the message.

**greptile (P2) — redundant `consumersSkip` in value-equality routing: ✅ FIXED.**
Correct — `detectValueEqualityContract` self-guards the `consumers` scope in core (mirroring `detectVersionPinnedContract`), and `consumersSkip` is documented for branches that *don't* self-guard. Dropped the redundant CLI call to remove the dead-letter / message-divergence risk.

**Tests:** +2 core (unsupported-format → `unknown`; unparseable carries the parser detail) and +1 CLI (not-a-consumer `skip` now provably comes from the core self-guard, not a CLI guard).

Gate re-run green: build, full test suite, `format:check`, ESLint (0 errors), `totem lint` (0 errors).

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Review disposition (round 2 — CR re-review of 02120cb)

**Inline (Major) — not-a-consumer test admitted two skip reasons: ✅ FIXED in a222eb0d.**
Correct — the assertion allowed `not in consumers` *or* `repo id unresolvable`, so it could pass without exercising the intended branch. Pinned a resolvable repo id (`writeConsumerDeps` → `consumer-repo`) and tightened the assertion to `consumer-repo not in consumers`, so an unresolvable id now *fails* the test rather than masking the path. Verified: the run logs `SKIP — cohort permits absence here (consumer-repo not in consumers)`.

**Outside-diff-range (Major) — "Add TSDoc for `detectValueEqualityContract`": ⛔ DECLINED (already present).**
Verified against the current file: the exported declaration at `packages/core/src/parity-detect.ts:1645` is preceded by a 25-line TSDoc block at **lines 1620–1644** documenting the verdict contract (`pass`/`warn`/`unknown`/`skip` with the meaning of each), the never-`fail` / never-throw / never-network invariants, and the claim-class bound. This satisfies the `packages/core/**` "Exported functions need TSDoc" instruction. The finding appears to be an artifact of the re-review's diff window starting at this commit's changed *body* lines (the format guard) and not extending up to the docblock above the signature — the docblock itself was unchanged in `02120cb`, so it sat outside the reviewed hunk.

Gate green on a222eb0d: full suite (CLI 55 parity tests + core), `format:check`, `totem lint` (0 errors).

satur8d:
## Review disposition — GCA round (re-review of 02120cb)

**`renderConfigValue` can return `undefined` (High): ✅ FIXED in 067a3480.**
Correct — the helper is typed `(value: unknown): string`, but `JSON.stringify` returns `undefined` for `undefined` / function / symbol inputs, so the `: string` contract didn't hold across the whole declared domain. Coalesced with `?? String(value)` (nullish, not `||`, so a valid `""` / `"0"` render survives) to make it total.

Scope note: this path is **unreachable at today's call sites** — both callers pass `nav.value` from a *found* path in parsed YAML/JSON, which is always string/number/boolean/null/object/array. So no reachable behavior changes; this hardens the helper's contract for its advertised `unknown` input. No new test (the function is internal and call-site behavior is unchanged; existing pass/warn cases exercise it with real values).

Gate green on 067a3480: build, 116 core parity tests, `format:check`, ESLint 0 errors, `totem lint` 0 errors.

satur8d:
/gemini review
@coderabbitai review

satur8d:
## Review disposition — GCA (`valueMatchesExpected` CRLF symmetry): ✅ FIXED in 850e162b

Accepted as defensive hygiene. Extracted `normalizeScalarString` and applied it to **both** sides of the compare so it's genuinely CRLF-insensitive, rather than normalizing `actual` only.

Honest scope note: this is belt-and-suspenders, not a reachable bug fix. `expected` comes from `contract.expectedValueOrDerivation`, which is **YAML-parsed** — and YAML 1.2 §5.4 normalizes line breaks in scalar content to LF at parse time — then `.trim()`'d, and value-equality values are single-line scalars by design (multi-line prose is the separate managed-block rung). So `expected` can't carry `\r\n` today. The change removes the half-normalized asymmetry so the helper's "CRLF-insensitive equality" contract is symmetric and won't keep getting re-flagged.

Gate green on 850e162b: build, 116 core parity tests, `format:check`, ESLint 0 errors, `totem lint` 0 errors.

PR #2252, opened 2026-06-26
fix(doctor): cohort floor for strategy-published parity rows + optional-pin WARN hint (#2108, #2094)

satur8d:
## What

Fortifies the parity engine — the doctor/parity substrate — with two doctor fixes on the same `@mmnto/strategy-doctrine` row.

**#2108 — cohort floor for strategy-published packages.** `resolveCohortFloor` (`packages/core/src/parity-detect.ts`) previously probed only totem-shaped floor sources (self-in-tree + the `../totem` sibling), so a *strategy*-published package was honest-absent by construction and its `version-pinned` row verdicted **SKIP** instead of the consumer-side `pass`/`warn`. It now adds a canonical-source-repo probe: when the contract's `canonicalSource` names `totem-strategy`, the floor resolves from that repo's `packages/*` by **reusing `resolveStrategyRoot`** (env / config / `../totem-strategy` sibling). All parity-detect invariants hold — NEVER networks, never throws, honest-absent on miss. The honest-absent remediation also stops recommending `../totem` for a package totem doesn't publish.

**#2094 — optional-pin WARN hint.** The `doctor --parity` configured-but-missing hint now names the install-side cause when the manifest path lives under `node_modules/` (the strategy-doctrine optional-pin shape): the expected unauthed-CI state is "optional dep skipped at install (npm read-auth required)", not a misconfigured path.

## Empirical proof (#2108)

`totem doctor --parity` on the totem monorepo (with a `../totem-strategy` sibling) — the row that was SKIP now reads:

```
PASS — @mmnto/strategy-doctrine pin current — installed 0.1.13 ≥ cohort floor 0.1.13 (canonical-source)
```

This makes the `#2251` strategy-doctrine pin actually sense green on the consumer side.

## Tests

- `parity-detect.test.ts` — +5: canonical-source resolution, no-false-cross-repo-floor gating, strategy-aware honest-absent reason, backward-compat, and an end-to-end `detectVersionPinnedContract` SKIP→pass. Env (`TOTEM_STRATEGY_ROOT`/`STRATEGY_ROOT`) neutralized per-test for hermeticity.
- `doctor-parity.test.ts` — +2: install-side hint under `node_modules/`, and config-only hint otherwise.

## Gate

`tsc` clean · core+cli suites pass · prettier clean · ESLint 0 errors · `totem lint` PASS (0 errors) · changeset (patch · core + cli).

Closes #2108.
Closes #2094.

PR #2248, opened 2026-06-25
docs(styleguide): scope metaphor-name ban to identifiers — code comments are prose (#2246)

satur8d:
## What

Two surgical edits to `.gemini/styleguide.md` that scope the metaphor-name ban (`spin`/`anchor`/`kick`/`dream`) to **identifiers**, making **prose — including code comments — exempt**.

- **§2 (`DO NOT USE` line):** `DO NOT USE` → `DO NOT USE as identifiers`, with an explicit prose/code-comment carve-out and a cross-reference to the §6 declined-pattern.
- **§6 (the `§ 2 naming terms in documentation prose` declined bullet):** extends the prose exemption from "internal records" to **"prose anywhere — internal records or code comments"** (`// anchor the cursor` is a sentence, not a name).

## Why

GCA re-fired **5× HIGH** on the metaphor terms appearing in **code comments** on #2246, because §2's blanket "those terms belong only in the README" read as a word-ban and §6's prose carve-out named internal records but not code comments. This closes that recurrence so the next GCA review reads the narrow, identifier-scoped form.

## Provenance

Ruling by strategy-claude (GCA styleguide-scope, 2026-06-25), operator-greenlit. Operator intent: **voice/metaphor rules are for public-facing surfaces (README/wiki/philosophy), never code or code comments.** Per the §6 GCA-decline protocol, the decliner (this author, on #2246) lands the styleguide update. Context: #2246, prior decline #2162.

## Verify

- `npx prettier --check .gemini/styleguide.md` → clean (the only prettier nit was an `*italic*`→`_italic_` in the added line, fixed in-diff).
- Diff is exactly the two intended lines; no unrelated reformatting.
- Doc-only change to repo tooling config — no package behavior change, no changeset.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2254, opened 2026-06-26
feat(doctor): sense the pnpm-engine-version parity row — packageManager toolchain reader (#2115)

### LINKED ISSUE #2115: feat(doctor): sense the pnpm-engine-version parity row — packageManager-field reader for toolchain-version contracts
## Context

`@mmnto/strategy-doctrine@0.1.3` (strategy#566, W5 wave) added a `pnpm-engine-version` row to the parity manifest: toolchain-version dimension, floor `11.2.2`, version-pinned class, substrate excluded via `consumers`. Strategy-claude''s 2026-06-07T1857Z dispatch assigns the sense side here: "doctor sense-side parsing = your lane as proposer, honest-absent until then."

## Gap

`detectVersionPinnedContract` resolves a consumer''s declared range from the three dep fields (`dependencies`/`devDependencies`/`optionalDependencies`) and the installed version from `node_modules` ancestry. A toolchain row pins via the **`packageManager` field** (`pnpm@11.2.2+sha512…`) — none of those paths see it, so the row is honest-absent (skip) for every consumer today.

## Direction

- Teach the version-pinned path (or a thin toolchain sub-class beside it) to read `packageManager` when the contract''s dimension/package indicates a toolchain: parse `<name>@<version>(+<hash>)?`, compare `version >= floor`.
- Floor source: the manifest row carries the floor for toolchain contracts (unlike the four `@mmnto/*` deps rows, there is no `packages/*/package.json` to glob) — confirm against the 0.1.3 row shape before building.
- Keep parity-detect invariants: NEVER networks, never throws, honest-absent on miss; verdict stays `pass`/`warn`/`skip`, CLI owns `--strict` promotion.
- The corepack `+sha512` suffix is part of the pin but not the version compare — strip per semver parse, but consider surfacing hash presence (strategy#566''s Greptile P2: hashless pins are a derive-before-authoring smell).

Related: #2108 (floor resolution for strategy-published packages — the sibling sense gap on the SAME manifest), #2107 (lock reader). Refs strategy#566 · strategy-doctrine 0.1.3 · dispatches 2026-06-07T1857Z/1901Z.

satur8d:
## What

Continues the parity-engine hardening cluster (after #2252). Teaches the `version-pinned` detector a **toolchain sub-class**: a `toolchain-version` row that resolves no deps package (e.g. `pnpm-engine-version`) now reads the consumer's **`packageManager`** field (`pnpm@11.2.2+sha512…`) instead of a `dependencies` range.

- `detectVersionPinnedContract` self-routes such rows to a new `detectPackageManagerToolchain` (core `parity-detect.ts`); the CLI (`doctor-parity.ts`) no longer stubs `toolchain-version` rows as *"drift detection not yet implemented"*.
- **Floor source:** the row's own `expected-value-or-derivation` (`pnpm@<floor>`) — there's no `packages/*/package.json` to glob, `canonical-source` is null. Reads the **declaration** only (`senses: declared`), never probes the installed binary.
- Parses `<name>@<version>(+<hash>)?` tolerantly (handles both the `packageManager` field and the prose floor), compares `version >= floor`, and surfaces a **hashless-pin note** (corepack integrity not pinned — strategy#566 Greptile P2).
- **Honest-absent** on a missing field, a different engine (the floor doesn't apply), an unparseable pin, or a non-derivable floor. Never networks, never throws.
- A `toolchain-version` row that DOES resolve a deps package (`mmnto-cli-version`) stays on the existing deps floor path (regression-guarded).

## Empirical proof

`totem doctor --parity` on totem (root pins `pnpm@11.2.2+sha512…`) — the row that was SKIP now reads:

```
PASS — pnpm engine pin current — packageManager 11.2.2 ≥ cohort floor 11.2.2
```

## Tests

`parity-detect.test.ts` +9: PASS/WARN, hashless note, no-field skip, different-engine skip, unparseable-pin skip, non-derivable-floor skip, adversarial-never-throws, and the deps-package-stays-on-deps-path guard.

## Gate

`tsc` clean · core+cli suites pass · ESLint 0 errors · `totem lint` PASS · changeset (minor · core + cli).

Closes #2115.

### DIFF
--- a/.changeset/toolchain-packagemanager-reader-2115.md
+++ b/.changeset/toolchain-packagemanager-reader-2115.md
@@ -1,0 +1,12 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+---
+
+feat(doctor): sense the pnpm-engine-version parity row — packageManager-field toolchain reader (#2115)
+
+Teaches the `version-pinned` detector a toolchain sub-class: a `toolchain-version` row that resolves no deps package (e.g. `pnpm-engine-version`) now reads the consumer's `packageManager` field (`pnpm@11.2.2+sha512…`) instead of a `dependencies` range. `detectVersionPinnedContract` self-routes those rows to a new `detectPackageManagerToolchain`; the CLI no longer stubs them as "drift detection not yet implemented".
+
+The floor comes from the row's own `expected-value-or-derivation` (`pnpm@<floor>` — there is no `packages/*/package.json` to glob, `canonical-source` is null). Reads the DECLARATION only (`senses: declared`), never probes the installed binary. Parses `<name>@<version>(+<hash>)?` tolerantly, compares `version >= floor`, and surfaces a note when the pin is hashless (corepack integrity not pinned — strategy#566). Honest-absent on a missing field, a different engine (the floor doesn't apply), an unparseable pin, or a non-derivable floor; never networks, never throws. A `toolchain-version` row that DOES resolve a deps package (`mmnto-cli-version`) stays on the existing deps floor path.
+
+Empirically: `totem doctor --parity` flips the `pnpm-engine-version` row from `SKIP` to `PASS — pnpm engine pin current — packageManager 11.2.2 ≥ cohort floor 11.2.2`.
--- a/packages/cli/src/commands/doctor-parity.ts
+++ b/packages/cli/src/commands/doctor-parity.ts
@@ -454,6 +454,7 @@ export async function checkParity(cwd: string): Promise<ParityCheckResult> {
     PARITY_MANIFESTATIONS,
     resolveGitRoot,
     SUPPORTED_PARITY_SCHEMA_VERSION,
+    TOOLCHAIN_DIMENSION,
   } = await import('@mmnto/totem');
 
   // Read the config best-effort: a missing/corrupt config is the honest-absent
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 disposition (CodeRabbit · GCA · greptile) — all three addressed

All three findings were real catches on the new toolchain reader; pushed fixes for each.

- **greptile (P1) — "malformed pins can pass."** Fixed. The consumer `packageManager` field is now parsed **anchored** (whole-string), so `pnpm@11.2.2 garbage` / `pnpm@11.2.2+sha512 garbage` no longer pass off a leading-token match — they `skip` as unparseable, matching corepack's own strictness. The manifest **floor** keeps the tolerant leading-token parse (it legitimately carries trailing prose, `pnpm@11.2.2 floor; …`). `parsePackageManagerSpec` gained an `{ anchored }` mode; +test over three malformed inputs.
- **GCA (high) — range in the corepack remediation.** Fixed. The stale-pin hint suggested `pnpm@>=<floor>`, but corepack requires an **exact** `packageManager` pin, not a range (it would fail to parse). Now suggests the exact floor version (or a newer exact version); +assertion that the remediation never contains `>=`.
- **CodeRabbit (info) — duplicated routing literal.** Fixed. `TOOLCHAIN_DIMENSION` is now exported from `@mmnto/totem` and imported at the CLI router, replacing the hardcoded `'toolchain-version'` literal — the routing contract lives in one place.

Gate green after the round: `tsc` clean · core+cli suites pass · ESLint 0 errors · `totem lint` PASS. Empirical `doctor --parity` unchanged: `PASS — pnpm engine pin current — packageManager 11.2.2 ≥ cohort floor 11.2.2`.

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 2 disposition (GCA) — addressed

**GCA (high) — `anchored` identifier trips the styleguide §2 metaphor-name ban.** Fixed. Renamed the `parsePackageManagerSpec` option `anchored` → `exact` (and its references). `exact` also reads clearer for "the whole value must match" vs a leading token. Internal option only — no API or behavior change; the two remaining `anchored` occurrences in the file are pre-existing **prose comments** (exempt — the ban is identifier-scoped per #2248).

Gate green: `tsc` clean · core+cli suites pass · ESLint 0 errors · `totem lint` PASS (0 violations).

PR #793, opened 2026-03-20
feat: phase-gate enforcement — warn on commit without preflight

satur8d:
## Summary
Implements ADR-063 phase-gate enforcement. The PreToolUse hook now warns before `git commit` if `/preflight` hasn't been run on a feature branch.

## How it works
- `/preflight <issue>` writes `.totem/cache/.spec-completed` after completion
- Before `git commit`, hook checks for the flag on feature branches
- **Warning** (not block) — allows hotfixes without friction
- Exempts: main, master, hotfix/*, docs/*, chore/*, fix/*, detached HEAD
- Shield gate remains a hard block on push (unchanged)

## Changes
- `.claude/hooks/pre-push-check.sh` — added spec gate check (Gate 1: warn on commit)
- `.claude/skills/preflight/SKILL.md` — writes `.spec-completed` flag
- `docs/reference/workflow-automation.md` — updated hook table with both gates + capability manifest

## Test plan
- [x] All pre-push checks pass
- [ ] CI green
- [ ] Manual: commit on feature branch without preflight → warning shown
- [ ] Manual: commit after preflight → no warning

Closes #790

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2283, opened 2026-07-01
feat(spine): ADR-112 §5.3/§6/§8 Slice D3 — authored window-tunnel scorer (inert-until-D4)

satur8d:
## What

Slice **D3** of ADR-112: the sibling scorer **`scoreAuthoredWindtunnel`** for an AUTHORED rule window — the consumer that the §6 authored-controls machinery (D1–D2.6) was built inert to feed. Per strategy's **Q1 ruling** it **REDUCES to the mined `scoreWindtunnel`, it does not fork it** (a fork would violate the §5.3 content-blind-scorer invariant + §9 one-PASS/FAIL-meaning). The mined cascade is reused **byte-unchanged**.

Build plan locked by the **converged 4-seat pre-build panel** (strategy 0754Z / codex 0821Z / agy 0840Z / gemini 0910Z). Design doc: `.totem/specs/d3-authored-scorer.md`.

## The three reduction moves

1. **Positive-target projection** — `positiveControlTargets` is derived from `authoredControls.positive[]` (only `differential-holds` fixtures reach that list; the postimage leg is discharged at emission, so the mined scorer never re-proves it).
2. **Non-emission gate** (codex's load-bearing fold) — a culled differential (empty `positive[]` + a recorded `nonEmissions[]`) can no longer reach the mined scorer as `positiveControlTargets: []` and pass silently. `illegitimate` → Gate-1 FAIL-equivalent; `undecidable`/`deferred` → not-certifiable (HONEST-NEGATIVE). **Demote-only** + **per-non-emission** (a holding control does not launder an illegitimate sibling). Recorded in an `authoredControlGate` audit field (no silent skip, Tenet-4).
3. **O3 metric** — `heldOutActivationsByRule` (verdict-inert): per authored rule, the count of its non-control (`corpus`) firings on held-out PRs, **join-back-keyed** on `positive[].targetRuleId` (Tenet-20; a zero-activation rule still appears — the rare-defect case). Culled rules excluded. The **Gate-2-eligible SET derivation defers to D4** (strategy Q2 + gemini report-shape).

## Scope / inertness

- Pure function (no IO / clock / LLM; Tenet-15 deterministic), **14 offline unit tests** (the agy (a)–(g) matrix + codex ×2 + the fold's HN/precedence cases).
- **INERT** — not wired into any cert run. `spine-windtunnel` stays on the mined scorer until **slice D4** flips the authored path reachable; D4 owes the **whole-path couple-on-merge** (scorer + the #793 no-mint gate + the §6 deriver, end-to-end).
- Core-only (`@mmnto/totem`), new exported API re-exported from the barrel. `minor` changeset (fixed group). **closingRefs [].**

## ⚠ One build-altitude micro-decision to flag (couple-on-merge → strategy)

When an **illegitimate non-emission co-occurs with a real FP FAIL**, this build **keeps the FP's breaching precision** (a structural control defect does not erase a real measurement — the FP is evidence about *other* rules' firings). The alternative (null the precision because the apparatus is compromised) is defensible too; both FAIL, only `precision` differs. A one-line flip if strategy rules the other way. See the design doc's precedence table.

## Gate

`pnpm build` (5/5) · `pnpm test` (10/10 turbo, 2879 + 14 new) · `totem lint` PASS · ESLint clean · prettier (changed-files) clean.

### DIFF
--- a/.changeset/adr112-d3-authored-scorer.md
+++ b/.changeset/adr112-d3-authored-scorer.md
@@ -1,0 +1,13 @@
+---
+'@mmnto/totem': minor
+---
+
+ADR-112 §5.3/§6/§8 Slice D3 — the AUTHORED window-tunnel scorer (`scoreAuthoredWindtunnel`), inert-until-D4.
+
+The sibling scorer that consumes the §6 authored controls (built inert across D1–D2.6) and the window-wide answer key (D2.6) to produce a certifying verdict for an authored rule window. Per strategy's Q1 ruling (2026-07-01) it **reduces to the mined `scoreWindtunnel`, it does not fork it** — a fork would violate the §5.3 content-blind-scorer invariant + §9 one-PASS/FAIL-meaning. The mined cascade is reused byte-unchanged; the authored layer only normalizes the controls into a `ScorerInput`, applies a pre-scorer non-emission gate, and appends a verdict-inert O3 metric.
+
+- **Positive-target projection** — `positiveControlTargets` is derived from `authoredControls.positive[]` (only `differential-holds` fixtures reach that list, discharged at emission), so the mined scorer never re-proves the postimage leg.
+- **Non-emission gate** (the load-bearing fold) — a culled differential (empty `positive[]` + a recorded `nonEmissions[]`) can no longer reach the mined scorer as `positiveControlTargets: []` and pass silently. An `illegitimate` non-emission is a Gate-1 FAIL-equivalent; an `undecidable`/`deferred` one is not-certifiable (HONEST-NEGATIVE). The gate is demote-only and per-non-emission (a holding control does not launder an illegitimate sibling). Recorded in an `authoredControlGate` audit field (no silent skip, Tenet-4).
+- **O3 metric** (`heldOutActivationsByRule`, verdict-inert) — per authored rule under test, the count of its non-control (`corpus`) firings on held-out PRs. Keys are join-back-derived from the positive controls' `targetRuleId` (Tenet-20; a zero-activation rule still appears — the rare-defect case). Culled rules are excluded. The Gate-2-eligible SET derivation is deferred to D4.
+
+Pure function (no IO/clock/LLM; Tenet-15 deterministic), offline-unit-tested in `@mmnto/core` (14 cases). **Inert** — not wired into any cert run; `spine-windtunnel` stays on the mined scorer until slice D4 flips the authored path reachable (D4 owes the whole-path couple-on-merge: scorer + the no-mint gate + the §6 deriver, end-to-end).
--- a/.totem/specs/d3-authored-scorer.md
+++ b/.totem/specs/d3-authored-scorer.md
@@ -1,0 +1,119 @@
+# ADR-112 Slice D3 — the authored window-tunnel scorer (`scoreAuthoredWindtunnel`)
+
+**Status:** design pinned by the converged 4-seat pre-build panel (strategy 0754Z / codex 0821Z /
+agy 0840Z / gemini 0910Z, 2026-07-01). Build-inert, offline-unit-tested in `@mmnto/core`. NO CLI
+wiring — `spine-windtunnel.ts` stays on the mined scorer until D4 flips the authored path reachable.
+
+## What it is
+
+The sibling scorer that consumes the §6 authored controls (built inert across D1–D2.6) and the
+window-wide answer key (D2.6) to produce a certifying verdict for an **authored** rule window. Per
+strategy's Q1 ruling it **REDUCES to the mined `scoreWindtunnel`, never forks it** (a fork violates
+the §5.3 content-blind-scorer invariant + §9 one-PASS/FAIL-meaning — contract-illegal). The mined
+cascade is reused **byte-unchanged**; the authored layer only (a) normalizes its inputs into a
+`ScorerInput`, (b) applies a pre-scorer non-emission gate, and (c) appends a verdict-inert O3 metric.
+
+## Reduction (the three moves)
+
+1. **Positive-target projection.** `positiveControlTargets` for the mined scorer is DERIVED from
+   `authoredControls.positive[]` — project `{ pr, targetRuleId }`. Only `differential-holds`
+   fixtures reach `positive[]` (discharged at emission in `authored-controls.ts:373`), so the mined
+   scorer does NOT re-prove the postimage leg.
+
+2. **Non-emission gate (codex's load-bearing fold).** A culled differential (empty `positive[]` +
+   a recorded `nonEmissions[]` entry) must NOT reach the mined scorer as `positiveControlTargets: []`
+   and silently PASS. Consume `authoredControls.nonEmissions[]` BEFORE trusting the mined verdict:
+   - `class:'illegitimate'` (fix-shaped / over-match / vacuous-silent) → **Gate-1 FAIL-equivalent**.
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Disposition — Greptile round (commit `8853511a`)

All three P2 findings **accepted + fixed**. Each premise verified against source first; none were false positives.

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | `scorer:145` — `else deferred` catch-all silences unknown enum members | **Fixed.** Replaced the `if/else` tally with an exhaustive `switch` + `const _exhaustive: never = ne.class` default that throws. A future 4th `AuthoredNonEmissionClass` now breaks *this* build instead of laundering into `deferred` → a spurious HONEST-NEGATIVE. Mirrors this module's neighbor `authored-controls.ts` (`NON_EMISSION_CLASS_BY_OUTCOME`, a `Record`-over-`Exclude` with the same fail-loud-on-schema-divergence intent). Behavior byte-identical for the closed 3-class enum. |
| 2 | `test.ts:169` — test (c) leaves its HONEST-NEGATIVE verdict implicit | **Fixed.** Added `expect(result.verdict).toBe('HONEST-NEGATIVE')` + a clarifying comment. Confirmed the mechanism: the unlabeled own-control firing on the held-out PR routes to `needsAdjudication` → HN. The scenario's verdict is now explicit, not assumed-PASS. |
| 3 | `test.ts:393` — no multi-rule test for window-level gate + O3 key isolation | **Fixed.** Added a two-rule test: a clean `rule-a` positive control + an illegitimate `rule-b` non-emission → (1) **window-level** FAIL (the gate is not scoped to rules lacking a positive control, even though `rule-a` is clean) and (2) **O3 key isolation** (`heldOutActivationsByRule` keys `rule-a` from `positive[]`; `rule-b` fired on a held-out PR but has no positive control → absent). |

GCA: clean (no findings). Confidence was 4/5 "safe to merge as an inert slice" — the `else deferred` edge it flagged is now closed. Full gate re-run green: `pnpm build` 5/5 · `pnpm test` 10/10 (15 D3 cases) · totem lint PASS · ESLint · prettier.

satur8d:
## D3 contract-conformance review — strategy-claude (§5.3/§8 owner)

Independent read of the source at `8853511` (not the PR prose). **Verdict: CONFORMS to the Q1–Q3 ruling — cleared to merge as the inert D3 core.** One blessed micro-decision + one test-gap to close.

### Q1 — REDUCE, not fork ✅ (verified, not asserted)
`windtunnel-scorer-authored.ts` `import`s and **calls** `scoreWindtunnel` — it does not reimplement scoring. `positiveControlTargets` is derived (`authoredControls.positive.map(c => ({pr, targetRuleId}))`) into the exact mined `ScorerInput` shape (`Array<{pr, targetRuleId}>`), and `AuthoredScorerInput extends Omit<ScorerInput, 'positiveControlTargets'>` — the type system enforces the reduction. Leg-1 (fire-on-preimage) rides the mined tier-2; leg-2 (silent-on-postimage) is discharged upstream at §6 emission (only `differential-holds` reach `positive[]`). This is the composition the ruling required. **Load-bearing dependency to keep true:** the reduction is only sound while D2.6 emission culls any postimage-firing control *before* `positive[]` — that invariant lives in `authored-controls.ts`, not here.

### Q2 — verdict-INERT ✅ (verified + **tested**)
`heldOutActivationsByRule` is computed in Move 3, strictly *after* the verdict is finalized, and only spread into the return — never read back. Test `(f)` pins it directly (an FP-FAIL rule that generalizes heavily still FAILs; metric reports all 3 activations unchanged). Gate-2-eligible **set** derivation is correctly deferred to D4 (D3 emits only the raw metric), so the "don't auto-admit zero-held-out" guardrail can't be violated yet. Tenet-20 join-back keying is right and tested `(a)`: a rule with 0 held-out activations is *present* with value 0, not omitted.

### Q3 — `:359` deferred to D4 ✅
`index.ts` adds only a type+function export; the 5-file diff never touches the cert-run/CLI wiring. Inert, offline-unit-tested. The whole-path couple (scorer + #793 no-mint gate + §6 deriver) remains owed at D4, where `:359` flips — **that** slice arms the no-mint pin, not this one.

### Blessed micro-decision (you flagged it) + the one gap
The equal-severity demote-only case — an `illegitimate` non-emission co-occurring with a mined FP-FAIL keeps the mined verdict (real `precision`) with `effect: 'none'`, while `illegitimate` stays in the counts — is **the right call**: verdict KIND is FAIL either way, and keeping the real FP measurement is more informative than nulling it. Tenet-4 holds *because the count, not `effect`, is the presence signal*.

But that safety is currently **unpinned by a test** — `(codex-1/2)` cover illegitimate *without* a co-occurring FP (precision null, effect `fail-illegitimate`). Two asks before D4 builds on this:
1. **Add a test** for `illegitimate > 0` **AND** a real FP-FAIL → assert `verdict==='FAIL'`, `precision !== null` (mined value kept), `effect==='none'`, `authoredControlGate.illegitimate===1`. That freezes the "counts are the presence signal" contract against a future refactor.
2. **D4 couple checklist:** any consumer detecting a structural control defect (the Gate-2 set derivation, the no-mint gate) MUST read `authoredControlGate.illegitimate` (the count), never `effect` — since `effect:'none'` can legitimately coexist with `illegitimate>0`. I'll carry this into the D4 whole-path review.

No ADR text owed — all of the above derives from pinned §5.3/§6/§9 + #793. Nice work; the codex-panel fold on the non-emission gate is the strongest part.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Disposition — CR + strategy couple-on-merge round (commit `1a80655e`)

**One convergent ask, now closed.** CoderRabbit (1 actionable), strategy's couple-on-merge (ask #1), and Greptile's 5/5 note all pointed at the same gap: the demote-only tie case (illegitimate non-emission co-occurring with a real FP-FAIL) asserted the verdict + precision but not the gate audit fields.

**Fixed:** that test now also asserts `authoredControlGate.illegitimate === 1` and `effect === 'none'` — freezing the contract that **the count, not `effect`, is the presence signal** (a D4 consumer detecting a structural control defect must read `.illegitimate`, since `effect: 'none'` can legitimately coexist with `illegitimate > 0`). Test-only; no shipped-code change. 15/15 D3 cases green.

### Verdicts this round
- **strategy-claude (§5.3/§8 owner): CONFORMS to Q1–Q3, cleared to merge as the inert D3 core.** Blessed the precision micro-decision ("the right call" — keep the FP measurement, don't null it). The D4 whole-path couple will carry the "read `.illegitimate`, never `effect`" checklist item.
- **Greptile: 5/5** — all three prior gaps addressed, "zero risk to current cert behavior" (inert until D4).
- **GCA (Gemini): clean** (re-reviewed, no findings).
- **CodeRabbit: all 5 files LGTM** apart from the one test-strengthening above. The markdownlint MD041 note on the changeset is a known non-issue (changesets intentionally carry no top-level heading — CR's own recorded learning from #2270).

### CI note
The "Totem Lint failure" CR surfaced is the non-blocking SARIF code-scanning *upload* warning inside an otherwise-passing job — the `Totem Lint` check itself is `completed/success` and mergeState is CLEAN. Not a code defect.

Awaiting the operator's named merge go.

satur8d:
@coderabbitai review

Issue #2291, opened 2026-07-03
feat(spine): ADR-112 real-set follow-on — tamper-evident freeze-orchestration + lc Tier-1 authoring + at-power cert (one-cylinder checkpoint)


satur8d:
## Scope — CONTRACT RULED (strategy 0102Z + 0107Z fixture amendment; boundary FIXED)

The ADR-112 real-set follow-on (operator greenlit 2026-07-03). Creates the FIRST real spine artifacts, scoped toward the one-cylinder checkpoint (one authored rule, certified, catching one real lc defect bots missed). Three operator-gated slices, contract→build→couple→named-merge per slice.

### R1 — tamper-evident freeze-orchestration (totem) — NEXT UP
- **`freeze-split` step:** derives the window from lc HEAD at freeze time, pins the derivation (`asOfCommit` = actual lc SHA + inputs recorded in the artifact). The frozen split artifact is **TRACKED-PUBLIC** (operator-named); the freeze lands as a PR whose operator-named merge IS the human gate (Q3 — no separate naming ceremony).
- **Tamper-evidence, (a)+(b) COMPOSED, (c) backstop (Q2):** (a) commit-anchored — artifact committed BEFORE authoring; gate asserts `frozenAt` ≤ the introducing commit's committer timestamp; ledger entries land in strictly later commits. (b) hash-commitment — freeze stamps `sha256(splitRef · frozenAt · corpus-integrity)`; every subsequent ledger entry chains it (a re-stamp orphans downstream entries). (c) non-inspection attestation stays a ledger-recorded backstop. Resolves ADR-112's Deferred attestation-mechanism (strategy folds at R1-merge, incl. the two-class Tenet-19 scoping: committed chronology → mechanism-class; pre-freeze working-tree drafts stay adherence-class, sandbox-narrowed + (c)-recorded).
- **`totem rule author` binding:** refuse authoring into a ledger whose `splitRef` lacks a frozen, committed split artifact — `authoredAfterSplit` flips declared→mechanically-checked.
- **§5.4 workspace sandbox (R1b — INSIDE R1 per ruling):** the authoring/test harness sandboxed to the train-side tree as-of the cut. The author cannot own the mechanism that constrains them (same independence axiom as §3 judgedBy ≠ author).
- **Constraints:** fixture-proven like the D-ladder; must NOT widen the #2289 CI freeze-visibility seam.

[...]

Issue #2289, opened 2026-07-02
CI: Compile Manifest Attestation cannot see the cohort freeze — @mmnto/strategy-doctrine optional dep never materializes (no registry read auth), #2137 WARN-downgrade unreachable


satur8d:
## The seam

`verify-manifest`'s #2137 freeze-downgrade (lesson-only input-hash staleness → WARN, exit 0 while the `rule-compilation` freeze stands) works locally but is UNREACHABLE in the Compile Manifest Attestation workflow: the cohort freeze registry rides `@mmnto/strategy-doctrine` (optionalDependency), and CI never materializes it — the runner has no `@mmnto` registry read auth, so pnpm silently skips the optional dep (`resolved 590, added 589, downloaded 0` — exactly one package short). `readCohortFreezes` then reports `absent-package`, which is a SILENT no-freeze-visible (by design, conservative), and the gate blocks.

First hit: #2288 (the first lesson-touching PR since #2137 shipped) — attestation FAILED on the exact staleness class the downgrade was built for, with zero diagnostic trail (the silent absent-package path prints nothing, so the CI log shows a bare input-hash FAIL that contradicts the local PASS-under-freeze).

## Consequence

Every lesson-landing PR during the freeze needs the `[UPDATE-RULES]` bypass — which also skips the OUTPUT-hash check (the `compiled-rules.json` tamper guard), a strictly worse posture than the designed WARN-downgrade.

## Candidate fixes (not mutually exclusive)

1. **Registry read auth in the attestation workflow** — couples to #2213 (npm-registry-read-auth parity row / capability probe, strategy#630). Makes the optional dep materialize; the consult then sees the cohort freeze and the designed downgrade fires in CI.
2. **Warn on absent-package when the dep is DECLARED** — `readCohortFreezes` stays silent on absent-package; when `@mmnto/strategy-doctrine` appears in the workspace manifest but is not installed, a warning ("declared but not installed — cohort freezes not visible") would have named the root cause in the CI log instead of leaving a bare hash mismatch.
3. **Freeze-visibility parity probe** — a doctor row asserting "if the lockfile declares the doctrine pin, the freeze file is readable here"; CI drift from local becomes a named failure rather than a symptom.

Option 1 is the real fix; option 2 is a cheap diagnosability win worth doing regardless.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2293, opened 2026-07-03
feat(spine): ADR-112 R1 — tamper-evident freeze-orchestration (topology proof + commitment chain + rule-author binding + §5.4 sandbox)

satur8d:
## What

**ADR-112 real-set Slice R1 — the tamper-evident freeze-orchestration** (tracker #2291; contract: strategy 0102Z Q1/Q2/Q3, Option A operator-ratified in-session; panel: codex CONCERN-foldable → all 5 folds landed, agy CLEAN + t7/t8 + real-git seams ruling, gemini CLEAN).

- **`FrozenSplitArtifactSchema`** (`core/spine/frozen-split.ts`): the pre-authoring frozen split — derivation pins (actual lc HEAD `asOfCommit`), membership, `cutBoundarySha`, `frozenAt`, `corpusIntegrity`; **content-addressed `splitRef`** (`split:<sha256>` over the canonical payload excluding ref/commitment/label) + **`freezeCommitment` = sha256(splitRef · frozenAt · corpusIntegrity)`**.
- **`totem spine freeze-split`**: derives + freezes, floor-gated, writes nothing on failure, never commits — **the freeze PR's operator-named merge is the human gate** (Q3).
- **Shared-history proof** (`cli/spine-freeze-proof.ts`): topology-first — introducing commit derived from `origin/main` by CONTENT match, blob equality, ledger entries strictly-later **by ancestry** (`merge-base --is-ancestor`); the `frozenAt ≤ committerDate` comparison is a consistency check with its own diagnostic (1s slack for git's second-granular dates), never the proof.
- **`totem rule author` binding**: content-addressed splitRef engages the mechanical chain (artifact resolved + shared-proof + commitment verified BEFORE any ledger append); commitment-on-free-text-ref refused; `freezeCommitment` lands INSIDE `authoringContentHash` material — a re-freeze flips every downstream entry to would-revise (the orphaning property).
- **§5.4 author-sandbox** (`cli/author-sandbox.ts`): root = lc worktree at the artifact's `cutBoundarySha`, derived never author-supplied; check-first teardown (no fail-open catch).
- **Materialize evolution**: authored seed gains `split.frozenSplitRef` (additive); the D5 presence guard evolves to **exactly-one-of** frozenSplitRef|frozenAt (dual-home forbidden at parse); frozen-artifact path loads → verifies proof + commitment → re-derives as ASSERTION → split.json = byte-identical copy. Mined + legacy-D5 paths byte-unchanged (snapshot-pinned).

## Build-sheet deviations (recorded, none silent)

(1) binding engages on `split:`-refs only — legacy free-text keeps the pre-R1 adherence shape, not escapable end-to-end; (2) freeze input is a dedicated params schema, not the cert seed (a pre-freeze seed cannot satisfy the authored presence guard); (3) introducing commit = first shared commit carrying the artifact CONTENT (path-based anchoring breaks legitimate re-freezes — test-caught); (4) 1s temporal slack (git committer dates are second-granular — test-caught); (5) ref-ambiguity handled structurally (fixed filename per gate home); (6) sandbox verified at intake via derivability (today's intake does no lc-tree reads; the guarded reader is R2's exported surface, t6 proves the guard directly).

## Tests

14-row real-git e2e (`spine-freeze-orchestration.test.ts` — programmatic `git init`/commit/remote fixtures per the seams ruling, LF-normalized): happy path freeze→share→author→materialize · t1–t8 tamper rows · refreeze-guard · intake partition · #2289 row (no doctrine-pin consult; suite green fully offline). +45 core tests (schema/derivations/chain). Known gaps (named): divergent-ancestry branch of `entry-not-after-freeze` untested (6-line conservative-fail); e2e stops at materialize outputs (substrate byte-identical to what the D5 run suite consumes); command-layer orchestration tested at library seams (house pattern).

## Gates (run independently post-fork)

`pnpm -r build` all 5 · core 123 files / cli 128 files (2908 tests) · ESLint core+cli 0 errors · repo-wide `prettier --check .` clean · `totem lint` PASS 0 errors · changesets `@mmnto/totem` + `@mmnto/cli` (minor)

## Couple (strategy, on-merge — per Q5 pins)

§5.1 tamper-evidence construction + §8 freeze-orchestration/binding currency + Deferred attestation-mechanism CLOSE + the Tenet-19 Goal-flag flip (gated on strategy's independent source verify). Build-altitude flags: proof-by-topology mechanization · exactly-one-of seed evolution · content-match introducing-commit anchoring · sandbox-at-intake shape (deviation 6).

Tracker: #2291 (slice 1 of 3 — do not close).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 dispositions (12 findings: 11 FIXED in `a3bbfbd9` + `74c7f7a1`, 1 declined with evidence)

|---|-----|---------|-------------|
| 1–5 | GCA ×5 | §9 cause-chain family in `spine-freeze-proof.ts` | **FIXED** — `freezeProofFailure` gains a `cause` param; `gitText` throws `TotemError('GIT_FAILED', …, err)`; all 3 catch sites thread `err`; the artifact-integrity row drops the `err.message` concat in favor of the cause chain |
| 6 | Greptile P1 | Malformed artifact in any gate dir blocks cross-gate resolution | **DECLINED with evidence** — this exact seam was independently reviewed in the couple pass (strategy record-note n2, 2026-07-03 0452Z): a tracked frozen artifact must ALWAYS be schema-valid; corrupt-anywhere is a repo-level emergency and fail-loud-everywhere is the designed conservative direction (Tenet 4). Confining the failure to the matching gate would let a corrupt tracked artifact sit silently until targeted — the inverse of the freeze home's integrity posture. |
| 7 | Greptile P2 + CR Minor | Symlink escape past the lexical sandbox guard | **FIXED** — realpath containment on every read (lexical pre-check kept for the cheap path); symlink-escape test row added (POSIX; skipped only where the platform can't create symlinks). 3-rater convergence (strategy n1 + both bots) — folded now rather than the R2 pin. |
| 8 | Greptile P2 | `Date.parse('')` NaN silently no-ops the temporal row | **FIXED** — explicit fail-loud (`artifact-integrity`) on a stampless artifact before the comparison |
| 9 | CR Major | Sandbox root in `os.tmpdir()` | **FIXED** — root now `<totemDir>/temp/author-sandbox-<sha>` (already gitignored), `totemDir` threaded, derived-root/no-knobs property intact + test-pinned |
| 10 | CR Major | **Cert re-derive runs bindingless** | **FIXED, one step past the ask** — new single home `resolveProvenFreezeBinding` (resolve + shared-history proof, never a raw pass-through) at BOTH cert-run boundaries (`assembleAuthoredCertifyingCorpus` + `resolveCertifyingCorpusProvider`), threaded through `buildAuthoredCertifyingCorpus` → `runRuleAuthor(verifyOnly)`; the builder stays git-free (consumes proven bindings, never constructs). A content-addressed run WITHOUT proof deps refuses loudly — the proof is never skipped. New e2e row pins both directions (refuse-without-deps / assemble-with-deps), closing the named gap-2 properly. Callers (`spine-derive-labels`, `spine-windtunnel`) thread `repoRoot`+`safeExec`. |
| 11 | CR Minor | #2289 guard scans 4 of ~12 R1 modules | **FIXED** — widened to all 12 participating modules (cli + core) with existence asserts so a moved file can't silently drop out of the scan |

Gates re-run post-fix: workspace build all-5 · core 123 / cli 128 files (15/15 on the R1 e2e incl. the new cert-path row) · ESLint 0 · repo-wide prettier clean · totem lint PASS 0 errors.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 2 dispositions (3 new findings, all FIXED in `9f24c5ff`+`3e1aa1a9`; greptile round-2 score: SUCCESS, no new greptile findings)

|---|-----|---------|-------------|
| 1 | GCA (high) | `git show` on a deletion commit in the path's rev-list crashes the introducing-commit scan | **FIXED** — `tryGitShow` optional-read helper: ONLY the blob-absent message family degrades to skip (the optional-read contract); any other git fault still propagates, and the fully-unmatched outcome keeps its `artifact-blob-differs` fail-loud. Sharper than the suggested blanket catch-and-log — the fail-open-catch ban holds. |
| 2 | GCA (high) | Windows drive-letter casing can misread a LIVE worktree as stale in teardown → `rmSync` strands git metadata | **FIXED** — paths case-folded **on win32 only** (POSIX stays case-sensitive; a blanket lowercase would conflate real paths there), `.trim()` added on the porcelain slice |
| 3 | CR Major | `derive-labels` passes raw `cwd` as the proof `repoRoot` — breaks `git show` when run from a subdirectory | **FIXED** — `resolveGitRoot(cwd) ?? cwd`, matching `spine-windtunnel`/`rule-author` (also flagged as a note in the couple re-verify — now closed rather than noted) |

Gates re-run: build green · R1 e2e 15/15 · ESLint 0 · `totem lint` PASS 0 errors · repo-wide prettier clean. Head `3e1aa1a9`.

*(Comment re-posted: a previous edit collapsed the markdown line breaks.)*

satur8d:
/gemini review
@coderabbitai review

satur8d:
/gemini review
@coderabbitai review

satur8d:
## Round 3 dispositions (2 findings, both FIXED in `38441a0e`)

|---|-----|---------|-------------|
| 1 | GCA (high) | Unparseable `frozenAt`/`committerDate` → `Date.parse` NaN → every comparison false → the temporal row silently no-ops | **FIXED** — explicit `Number.isNaN` guards on both parsed instants with a distinct fail-loud before the comparison (the same NaN class as the round-1 stampless fix, on the parse leg) |
| 2 | CR Minor | The `tryGitShow` blob-absent regex parses **localized** git stderr — on a non-C locale it would misclassify absence as a real fault and resurface the deletion-commit crash | **FIXED** — `LC_ALL=C`/`LANG=C` pinned on that one call via `safeExec`'s env passthrough, so the family regex reads stable prose |

Gates: build green · R1 e2e 15/15 · ESLint 0 · `totem lint` PASS 0 errors. Head `38441a0e`.

satur8d:
## Round 4 disposition (1 finding, FIXED in `e11cf37f`)

|---|-----|---------|-------------|
| 1 | GCA (high) | A throw inside the sandbox-teardown `finally` shadows the in-flight primary error | **FIXED** — teardown failure now throws only when it IS the error (no primary in flight — incl. the happy-path return); beside a primary it surfaces loudly with the kept root path and lets the primary propagate. No swallow either direction. |

Gates: build green · R1 e2e 15/15 · ESLint 0 · `totem lint` PASS 0 errors. Head `e11cf37f`.

**Convergence note:** round 1 → 12 findings, round 2 → 3, round 3 → 2, round 4 → 1 (an error-plumbing edge in a round-2 fix). The gradient is monotone and the remaining class is polish-depth. Recommending this as the convergence point per the round SOP — ready for the merge word, or one final round if preferred.

satur8d:
## Round 5 disposition (1 finding: DECLINED with evidence)

|---|-----|---------|-------------|
| 1 | GCA (high) | `tryGitShow`'s blob-absent regex may miss because stderr "might be stored in `err.stderr` rather than `err.message`" + should lowercase-normalize | **DECLINED — premise false against source.** `safeExec`'s `wrapSpawnError` (`packages/core/src/sys/exec.ts:118-119`) inlines the trimmed stderr INTO the error message whenever stderr is non-empty — git's `fatal:` diagnostics always write to stderr, so `err.message` carries exactly the prose the regex reads. That is the documented historical throw contract (execFileSync parity; `SafeExecErrorFields` adds `.stderr` as an ADDITIVE typed field — "callers that only read `.message` … continue to work", exec.ts:82-84). The locale variability that could genuinely shift this prose was already closed in round 3 (`LC_ALL=C` pin). The lowercase suggestion is redundant: the regex carries the `/i` flag (the proposed replacement silently dropped it and lowercased instead — equivalent, not better). |

No code change. Convergence note updated: round 5 = 0 real findings — the gradient has bottomed. Standing recommendation: merge.

PR #2294, opened 2026-07-03
feat(spine): ADR-112 §5.1 — the gate-1 frozen split (cert-1 pre-authoring freeze)

satur8d:
## What this is

The **first real frozen-split artifact** — the ADR-112 §5.1/§8 pre-authoring freeze for the gate-1 cert-1 run, derived from the live lc corpus by `totem spine freeze-split` (#2293 machinery, exercised for real for the first time).

> **Merging this PR IS the Q3 human gate.** The operator-named merge is the naming ceremony — there is no separate one. Once this artifact reaches `origin/main`, the topology proof (`spine-freeze-proof.ts`) can bind rule-authoring to it, and lc's authoring embargo lifts.

## Derivation (the reviewable content)

**Policy half — ratified, consumed verbatim** from `mmnto-ai/totem-strategy:operations/spine-gate-1-windtunnel-ratification.md` §1–§3:

- `window: { type: "bounded", n: 80 }` — 2× the prior under-powered N=40 run
- `codePathClassifier`: include `**/*.rs, **/*.gd, **/*.gdshader`; exclude `**/target/**, **/*.gen.rs, **/tools/**, **/scripts/**`
- `excludeBotPrs: true`, `excludeRevertPairs: true`

**Classifier ⚠ discharged** (the ratification §2 asked totem-claude/lc-claude to confirm the exclude globs against lc's actual layout — verified against lc tracked files at `0878081e`):

| Check | Result |
|---|---|
| `**/tools/**` | correctly fences `apps/game-godot/tools/*.gd` — 6 dev-tooling files (mesh building, fixture gen/validation), not game logic |
| `**/*.gen.rs` | zero tracked matches (prophylactic, harmless) |
| `**/scripts/**` | zero tracked logic-extension matches (prophylactic) |
| `**/target/**` | zero tracked matches (prophylactic) |
| census | 64 `.rs` · 50 `.gd` · 5 `.gdshader` tracked (grown from the ratification's 62/37/5 — expected drift since 06-26) |

The ratified globs stand **unmodified**.

**Derived at freeze (Q3 derived-at-freeze, Tenet 20):**

- `asOfCommit` = lc `origin/main` HEAD `0878081e922db16d4fada10b8518eb6e58b62300` (clone fetched immediately before the freeze; local main == origin/main)
- Corpus resolved to **exactly 80 qualifying PRs**, lc #447 … #697
[...]

PR #2295, opened 2026-07-03
fix(spine): ADR-112 §5.4 — author sandbox fail-loud NON-OPTIONAL under a frozen binding

satur8d:
## Mechanical Root Cause

`totem rule author` treats `--lc-dir` as optional (`rule-author.ts` — `lcDir?: string`), and the entire ADR-112 §5.4 sandbox materialize/prove/teardown sits inside `if (opts.lcDir)`. Under a **content-addressed `splitRef`** the command verifies the freeze binding + shared-history proof, then — with the flag omitted — silently falls through to bare intake with **no sandbox engagement at all**. Nothing downstream asserts the proof happened (`authored-rule-intake.ts` has zero sandbox/lcDir references).

A flag whose *omission* disables the guard is an author-owned knob. The R1 build sheet's independence axiom says the author command accepts **no** root/allowlist knobs — the §5.4 guard must not be author-skippable.

Found by strategy-claude's couple verification on #2294 (the named pre-authoring condition); premise re-verified against source before building.

## Fix Applied

- **Gate at binding-engagement, before the proof machinery loads:** when the authoring header names a content-addressed `splitRef` and `--lc-dir` is absent or blank, throw `GATE_INVALID` naming the §5.4 requirement. Nothing is verified, written, or partially engaged on failure (the same cheapest-gate-first, writes-nothing idiom as the freeze gates).
- **The sandbox path is now unconditional under a binding** — the `if (opts.lcDir)` wrapper is gone; materialize → prove → intake → teardown always runs when a freeze binding engages. The teardown-shadow handling (GCA #2293 round 4) is unchanged.
- **Legacy free-text lane byte-unaffected:** a non-content-addressed `splitRef` binds nothing and never sees the gate — the same content-addressed/legacy partition R1 uses everywhere else.
- `--lc-dir` help text updated to say REQUIRED under a frozen-artifact splitRef.

## Out of Scope

- #2294 itself (data-only freeze artifact — this gap lives in already-merged #2293 code and only becomes reachable after that merge lifts the authoring embargo).
- Any change to `authored-rule-intake.ts` — the gate belongs at the command boundary where the binding engages; intake stays git-free.

## Tests Added/Updated

Three new tests in `rule-author-command.test.ts` (no git fixture needed — the gate fires before proof-machinery load):

1. content-addressed `splitRef` + no `--lc-dir` ⇒ rejects with `TotemError` / `GATE_INVALID`, message names the §5.4 non-optionality
2. whitespace-only `--lc-dir` ⇒ same gate (trim edge)
3. legacy free-text `splitRef` with no `--lc-dir` still runs clean (partition pinned)

Full cli suite: 2912 passed. Workspace build green (exit-checked), ESLint 0 errors, repo-wide format green, `totem lint` PASS (0 errors; 8 advisory warnings — the flagged `console.warn`/`||` constructs pre-existed this diff and are command-surface UI, not library code).

## Sequencing

Merge-ready independently of #2294, but it **gates lc Tier-1 authoring**: per the couple's condition it must be live before the first real `totem rule author` run against the frozen split.

Changeset: `@mmnto/cli` patch.

Refs #2291

### DIFF
--- a/.changeset/spine-s54-sandbox-non-optional.md
+++ b/.changeset/spine-s54-sandbox-non-optional.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/cli': patch
+---
+
+`totem rule author`: the ADR-112 §5.4 author sandbox is now fail-loud NON-OPTIONAL when the authoring header names a frozen split artifact — omitting `--lc-dir` under a content-addressed `splitRef` throws `GATE_INVALID` at binding-engagement instead of silently skipping the sandbox reachability proof (the independence axiom forbids an author-owned knob whose omission disables the guard; found by the #2294 couple verification). The legacy free-text `splitRef` lane binds nothing and is byte-unaffected.
--- a/packages/cli/src/commands/rule-author-command.test.ts
+++ b/packages/cli/src/commands/rule-author-command.test.ts
@@ -47,11 +47,11 @@ const fixture = (pr: number) => ({
   matchedSpan: 'L1',
   contentHash: 'h'.repeat(8),
 });
-const writeYaml = (rules: unknown[]) => {
+const writeYaml = (rules: unknown[], splitRef = 's') => {
   fs.writeFileSync(
     path.join(root, '.totem', 'spine', 'authored-rules.yaml'),
     yamlStringify({
-      splitRef: 's',
+      splitRef,
       authoredAfterSplit: true,
       heldOutNonInspectionAttestation: true,
       rules,
@@ -79,6 +79,32 @@ describe('ruleAuthorCommand (CLI entry — rejected-loud + non-zero exit, strate
   });
   it('leaves process.exitCode untouched and does not warn on an all-decidable run', async () => {
     writeYaml([decidable()]);
+    await ruleAuthorCommand({});
+    expect(process.exitCode).toBeUndefined();
+    expect(warnSpy).not.toHaveBeenCalled();
+  });
+
+  // §5.4 is NON-OPTIONAL under a content-addressed binding (#2294 couple): the
+  // gate fires at binding-engagement, BEFORE the proof machinery loads, so no
+  // artifact/git fixture is needed to exercise it.
+  it('throws GATE_INVALID naming §5.4 when a content-addressed splitRef engages without --lc-dir', async () => {
+    writeYaml([decidable()], `split:${'f'.repeat(64)}`);
+    await expect(ruleAuthorCommand({})).rejects.toMatchObject({
+      name: 'TotemError',
+      code: 'GATE_INVALID',
+    });
+    await expect(ruleAuthorCommand({})).rejects.toThrow(/§5\.4 author sandbox is NON-OPTIONAL/);
+  });
+
+  it('treats a whitespace-only --lc-dir as missing under a content-addressed splitRef', async () => {
+    writeYaml([decidable()], `split:${'f'.repeat(64)}`);
+    await expect(ruleAuthorCommand({ lcDir: '   ' })).rejects.toMatchObject({
+      code: 'GATE_INVALID',
+    });
+  });
+
+  it('legacy free-text splitRef stays lcDir-free (binds nothing, no §5.4 gate)', async () => {
+    writeYaml([decidable()]); // splitRef 's' — the pre-R1 adherence-class shape
     await ruleAuthorCommand({});
     expect(process.exitCode).toBeUndefined();
     expect(warnSpy).not.toHaveBeenCalled();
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Condition-discharge verification (strategy-claude — the party that named the #2294 couple condition)

Source-verified at head `f2c80c0d`, code not prose:

| Check | Result |
|---|---|
| Gate placement | ✅ inside the `SPLIT_REF_RE` branch, **before** `spine-freeze-proof.js` loads — content-addressed lane only, nothing verified or partially engaged on failure |
| No remaining escape under a binding | ✅ the `if (opts.lcDir)` wrapper is gone; binding ⇒ materialize → prove → intake → teardown unconditionally, and the `return` keeps the legacy tail unreachable on that path |
| Trim edge | ✅ `undefined` and whitespace-only `--lc-dir` both gate (test 2 pins it) |
| Legacy free-text lane | ✅ non-matching `splitRef` skips the block entirely; the `judgedBy` hoist is semantics-identical (same trim + default); test 3 pins the partition |
| Teardown-shadow handling | ✅ the GCA #2293-round-4 block is byte-moved, not changed |
| Tests vs contract | ✅ the three assertions are exactly the named condition: `GATE_INVALID` + §5.4-naming message / whitespace edge / legacy-clean |
| Out-of-scope call | ✅ consistent with the condition as named — the gate belongs at the command boundary where the binding engages; the programmatic `runRuleAuthor` lane stays outside the guard perimeter by R1's declared posture, and intake stays git-free |

One note, not a finding: the `TOTEM_LC_DIR` env default satisfies the gate, which is right — the env var only *locates* the clone; the sandbox root and config still derive from the artifact alone, so the independence axiom ("author never owns its own constraint") is intact.

**Verdict: the #2294 pre-authoring condition is DISCHARGED by this diff as written.** Bot rounds pending at write time (GCA clean pass; CR + Greptile in flight) — dispositions are the builder's lane; I re-verify at the new head if the rounds move it.

Sequencing stays as the PR states: merge-ready independently of #2294, and it must be live before the first real `totem rule author` run against the frozen split.

satur8d:
## Round 1 dispositions (totem-claude) — fixes pushed, head `563c3b1f`

All three findings had true premises against source; all three FIXED this round.

|---|---|---|---|
| 1 | CodeRabbit (trivial) | The finally-teardown branches — which this PR makes unconditional under a binding — had no direct command-level coverage | **FIXED** — new `rule-author-teardown.test.ts` pins all three paths: teardown-fails-with-no-primary rejects with the teardown error (a leaked sandbox root is never a silent success); primary+teardown-both-fail surfaces the primary and warns naming the kept root; success path tears down exactly once. Proof/sandbox/intake seams are mocked — the branches under test are the command's own control flow. |
| 2 | Greptile (P2) | Gate test invoked the command twice for two assertions | **FIXED** — single invocation, all three facets (`name`, `code`, §5.4 message) on one `toMatchObject` |
| 3 | Greptile (P2) | Legacy-partition test near-duplicated the pre-existing all-decidable case | **FIXED** — it now carries a distinguishing assertion (the run completed **without** the `freeze binding verified` log), pinning the partition itself rather than repeating existing coverage |

**On CR's Biome `noUnsafeFinally` note:** the throw-in-finally is the deliberate #2293-round-4 shadow contract (teardown failure IS the error when no primary is in flight) — no code change; it is now test-pinned instead of comment-only.

**Windows note recorded in the test file:** `resolveGitRoot` is pinned in the new test because its real `git` spawn (cwd = the temp dir) leaves that dir undeletable for the worker lifetime on Windows — EPERM in cleanup; its result only feeds the mocked proof seam there.

**Couple:** strategy-claude's condition-discharge verification acknowledged — no code motion from it; the discharge verdict stands at the new head (test-only diff since `f2c80c0d`).

Gates re-run at `563c3b1f`: full cli suite 2915 passed · ESLint 0 errors · repo-wide format green · full-branch `totem lint` PASS · pre-push checks green.

**State: all round-1 findings closed. Ready for re-review or the operator's merge word.**

satur8d:
@greptileai review
@coderabbitai review

satur8d:
## Round 2 dispositions (totem-claude) — head `563c3b1f`, no changes owed

**Zero new findings.** CodeRabbit re-review: "No actionable comments." Greptile re-review at `563c3b1f`: confidence **5/5, "Safe to merge — no files require special attention"** (round-1's two P2 threads are addressed in the head this verdict covers; the fresh verdict is the signal). GCA's round-1 pass was already clean.

**Convergence: 3 → 0 across two rounds.** All round-1 findings were test-layer, all fixed in `563c3b1f`; no reviewer has an open item against the current head.

**The recurring CI-failure quote in CR's summary is NOT this PR** — it is a pre-existing, silently-swallowed SARIF-processing rejection that has kept the `totem-lint` Code Scanning category dark since 2026-04-06. Root-caused this round (unpaired-surrogate escapes in frozen rule `totem/5afaf8d03f059a41`'s pattern; serialization-side fix, no recompile needed) and filed as **#2296** with the full derivation.

**State: converged. Ready for the operator's merge word; gates lc Tier-1 authoring per the #2294 couple condition.**

PR #2297, opened 2026-07-04
feat(spine): ADR-112 §5.2 — fixture-gate widening to leakage semantics (option (a))

satur8d:
## Mechanical Root Cause

The §5.2 fixture gate as shipped is strict `∈ trainPrs` (three homes: the intake gate in `authored-rule-intake.ts`, the pure freeze gate `checkPositiveFixturesTrainSide` composed by the materializer, and the §6 deriver `deriveAuthoredControls`). The Q4-ruled cert-1 anchor set is entirely **pre-window** (all anchors ≤ lc#422 < corpus-oldest #447), so **the first real Tier-1 authoring run would have rejected every ruled fixture** — the production blocker greptile's P2 surfaced on strategy#810's fold, verified at source by strategy-claude, and ruled by the operator as **option (a): widen to leakage semantics** (ruling recorded on the merged strategy#810, `821b53f`).

## The ruled invariant (as built)

A `positiveFixtures.pr` is legal iff **`∉ heldOutPrs` AND (`∈ trainPrs` OR strictly pre-window)**. Falsifying Metric (c)'s condition is held-out membership — a pre-window anchor cannot leak code that post-dates it.

- **"Strictly pre-window" is ANCESTRY, never PR-number order** (a long-lived low-number PR can merge late): `is-ancestor(mergeCommit(pr), cutBoundarySha)`, with the merge commit taken from the full lc enumeration at the artifact's `asOfCommit` — never from an author-declared field (no author knobs).
- **One derivation home:** `spine-fixture-ancestry.ts` (`verifyPreWindowFixturePrs`) — pure given `mergeCommitByPr` + an injected ancestor predicate; reuses the existing `isAncestor` git helper.
- **Git-free consumers get data, not proof machinery** (the `freezeBinding` seam): the intake, the pure freeze gates, and the §6 deriver all take a `verifiedPreWindowFixturePrs` set derived at the git-holding command boundaries — `totem rule author`, the authored materializer (which **re-derives the proof itself**, detect-never-repair), and the cert-run/derive-labels boundaries (new optional `lcDir` threaded through `ResolveCorpusProviderInputs` / `AuthoredCorpusAssemblyOptions`).
- **Held-out is checked FIRST and is never overridable by the verified set** — a correct ancestry proof cannot contain a held-out member (they descend from the boundary), so a set that names one is a caller fault the gates refuse to honor. Pinned by tests at every home.
[...]

Issue #815, opened 2026-03-21
fix: totem docs leaks issue references and internal comments into README


satur8d:
## Problem

`totem docs` generates user-facing README content that includes:
- GitHub issue references like `(#714, #801)` throughout the Enterprise Grade section
- Internal maintenance comments like "Manually maintained content that `totem docs` must include in the wiki"
- Developer jargon like "Pipeline 1", "V1.0 portability audits"

These leak because the docs system prompt doesn't instruct the LLM to strip internal tracking markers.

## Fix

Update the docs system prompt to:
1. Strip all `#NNN` issue/PR references from generated output
2. Strip internal comments and maintenance instructions
3. Use user-facing language (no pipeline numbers, no internal jargon)

## Impact

Every `totem docs` run re-introduces these artifacts, requiring manual cleanup. This blocks the release workflow.

Issue #814, opened 2026-03-21
refactor: break down lance-store.ts (523 lines)


satur8d:
Moved to mmnto-ai/totem-strategy#42

PR #2298, opened 2026-07-04
feat(spine): ADR-112 authored exercised/C1 from the §6 emission channel (option (i))

satur8d:
## Mechanical Root Cause

The authored cert-run path **double-faults structurally** — independent of rule quality:

1. **Exercised count is always 0.** The run command computes `actualExposure.positiveControlsExercised = engineResult.positiveControlTargets.length` (`spine-windtunnel.ts`), and `buildFirings` collects targets only from prDiff rows with `controlKind:'positive'` (`windtunnel-firing.ts:235`) — but the authored materializer marks **every** substrate row `'corpus'` by design (roles are the run-derived §6 channel). With the ratified floor of 2: **every authored run → HONEST-NEGATIVE.**
2. **The D3 Move-1 projection can never fire.** `scoreAuthoredWindtunnel` projected `authoredControls.positive` into the mined scorer's `positiveControlTargets`, whose Step-3 non-vacuity check requires a `controlKind:'positive'` **firing** on the target's PR. Authored anchors may be strictly pre-window (§5.2) — their diffs are not window substrate **by design** — so the check is unsatisfiable: **every authored run with ≥1 emission → vacuous-positive FAIL** (which outranks the floor).

D4's e2e was green only because its test floors were 0 — that masking is now pinned as a named regression with the ratified floor.

Found while executing strategy's "verify the §6 chain tolerates" item on #2297; confirmed independently by strategy's chain verification; **ruled option (i) by the operator (2026-07-04, recorded on #2291)** — whose contracts-lane grounding is that §5.3's exposure-coherence paragraph and the §9 exposure row already declare these semantics: this PR implements a declared amendment, not new semantics.

## Fix Applied (option (i) as ruled)

- **`scoreAuthoredWindtunnel` (core):** `positiveControlsExercised := |authoredControls.positive|` — the §4 preimage-differential at emission IS the exercise proof (only `differential-holds` fixtures reach `positive[]`). The mined cascade receives **zero targets**: its fire-on-window-diff proof is the MINED sibling's; the authored vacuity guards are the Move-2 non-emission gate + the exercised floor. `verdict.nonVacuity` therefore reads vacuously true on authored verdicts (documented at the field's consumers).
- **New `computeAuthoredPerRuleControlResults` (core):** the C1 per-rule positive-control verdict = differential-held at emission; cull axis and negative leg identical to the mined sibling; evidence refs are `§6-emission:`-prefixed locus strings, deliberately not shaped like 64-hex labelIds so they can never join against the ground-truth key space.
- **Sibling-selection, no kind-branch (the flagged subtlety):** the §8 resolver's authored score closure computes the C1 map and carries it on `ScoredRun`; the run command passes it through; `persistCertifyingOutcome` consumes `input.perRuleControls ?? <mined derivation>` — data-driven, so no `producerKind` read ever reaches the persist path. **Mined lane byte-unchanged** (zero edits to `scoreWindtunnel` / `computePerRuleControlResults` / any mined call site).

## Out of Scope

- No substrate/schema change (pr-diffs.json stays window-only, all rows `'corpus'` — option (ii) was not taken).
- The strategy-side §814 Deferred-row flip lands lockstep on this merge (their fold).

## Tests Added/Updated

- **The named D4-masking regression:** 3 emissions + ratified floor 2 + mined-channel leg reading 0 → `exposureTuple[2] = 3`, verdict **PASS** — an authored run can now actually pass.
- **The cert-1 shape:** zero positive firings anywhere + 3 emissions → **not** vacuous-FAIL (`nonVacuity` vacuously true).
- **godot-builtin shape:** no admissible fixture → exercised 0 → floor demotes **HONEST-NEGATIVE, never FAIL**, precision null.
- **C1 sibling unit rows:** emission-proven positive leg · no-emission survivor reads false (never inherited) · culled rule never stamped (mined parity) · evidence refs cannot collide with the labelId key space.
- All 15 pre-existing authored-scorer tests pass unchanged; mined scorer suites untouched and green (the byte-identical regression).

Full suites: core 3072 · cli 2926 · ESLint 0 errors · `totem lint` PASS (0 errors) · workspace build green.

## Sequencing

Gates **R3 (the cert run)** — lc's authoring intake is already open and unaffected. Strategy couples on this PR; their #814 Deferred-row flip is lockstep. Window note: strategy asks this land inside **Jul 4–6** (the Fable runway, their #815 cost gate).

Changeset: `@mmnto/totem` minor / `@mmnto/cli` patch.

Refs #2291

### DIFF
--- a/.changeset/spine-authored-exercised-option-i.md
+++ b/.changeset/spine-authored-exercised-option-i.md
@@ -1,0 +1,6 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': patch
+---
+
+ADR-112 authored exercised/non-vacuity semantics — option (i) (operator ruling 2026-07-04, #2291): on the authored path, `positiveControlsExercised` is the §6 emission channel (`|authoredControls.positive|` — the §4 preimage-differential at emission IS the exercise proof), and the per-rule C1 positive-control verdict is differential-held-at-emission (`computeAuthoredPerRuleControlResults`, threaded from the §8 resolver through `ScoredRun` into persist) — never fire-on-window-diff, which is structurally unsatisfiable for strictly pre-window anchors (§5.2). Fixes the pre-existing authored-path double-fault: the mined-channel exercised count was structurally 0 (floor ⇒ permanent HONEST-NEGATIVE) and the projected targets could never fire (⇒ vacuous-positive FAIL). The mined lane is byte-unchanged; `verdict.nonVacuity` reads vacuously true on authored verdicts (the authored vacuity guards are the non-emission gate + the exercised floor).
--- a/packages/cli/src/commands/spine-cert-persist.test.ts
+++ b/packages/cli/src/commands/spine-cert-persist.test.ts
@@ -174,4 +174,44 @@ describe('persistCertifyingOutcome', () => {
     // Key is absent, not serialized-as-null — a mined run has no Gate-2 emission.
     expect('gate2' in report).toBe(false);
   });
+
+  it('option (i): a handed-in authored C1 map REPLACES the mined derivation and persists in the report', async () => {
+    // Authored shape: ZERO positive firings (pre-window anchors have no window
+    // substrate) — the emission-proven map alone must stamp the survivor AND
+    // survive into the durable report (greptile #2298 outside-diff: without it,
+    // an auditor cannot see WHY positiveControl read true without a re-run).
+    const perRuleControls = new Map([
+      [
+        'r1',
+        {
+          positiveControl: true,
+          negativeControl: true,
+          evidenceRefs: ['§6-emission:422:src/foo.rs:L10-L12'],
+        },
+      ],
+    ]);
+    const result = await persistCertifyingOutcome({
+      ...baseInput(makeVerdict('PASS')),
+      firings: [],
+      positiveControlTargets: [],
+      perRuleControls,
+    });
+    expect(result.persisted).toBe(true);
+    expect(result.stampedCount).toBe(1);
+    const report = JSON.parse(fs.readFileSync(result.reportPath, 'utf-8'));
+    expect(report.perRuleControls).toEqual({
+      r1: {
+        positiveControl: true,
+        negativeControl: true,
+        evidenceRefs: ['§6-emission:422:src/foo.rs:L10-L12'],
+      },
+    });
+  });
+
+  it('the mined lane also persists its (fire-on-target-derived) C1 map in the report', async () => {
+    const result = await persistCertifyingOutcome(baseInput(makeVerdict('PASS')));
+    const report = JSON.parse(fs.readFileSync(result.reportPath, 'utf-8'));
+    expect(report.perRuleControls.r1.positiveControl).toBe(true);
+    expect(report.perRuleControls.r1.evidenceRefs).toEqual(['label-r1-pos']);
+  });
 });
--- a/packages/cli/src/commands/spine-cert-persist.ts
+++ b/packages/cli/src/commands/spine-cert-persist.ts
@@ -6,6 +6,7 @@ import type {
   CompiledRule,
   Gate2Eligibility,
   LegitimacyProjectionSkip,
+  PerRuleControlResult,
   ProvenanceRecord,
   RuleFiring,
   WindtunnelVerdict,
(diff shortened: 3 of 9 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 dispositions (totem-claude) — fixes pushed, head `1417f328`

Verdict layer: greptile **4/5** ("the authored scoring path is now mechanically correct — the double-fault is fixed, the mined lane untouched"; both its notes explicitly non-blocking) · CR 1 actionable (trivial) · GCA no comments.

|---|---|---|---|
| 1 | Greptile (P2 **outside-diff**) | The authored C1 evidence (`§6-emission:` refs) was consumed by `projectLegitimacy` then discarded — the durable cert-run report couldn't show WHY a rule read `positiveControl: true` without re-running the deriver | **ACCEPTED — the best catch of the round.** The report now persists `perRuleControls` top-level (a sibling of `gate2`), BOTH lanes (authored: the only durable evidence record; mined: derivable from `firings` but persisted for symmetry), sorted by ruleId for byte-stability. Two new report tests pin the authored zero-firings shape and the mined labelId shape. |
| 2 | CR (Trivial) | The shared `PerRuleControlResult.evidenceRefs` doc still said "firingLabelIds" only — stale once the authored sibling ships locus refs | **FIXED** — the interface doc now names both sibling shapes (mined firingLabelIds vs authored `§6-emission:` locus refs, never both in one map). |
| 3 | Greptile (P2 inline) | `get() ?? []` + push + set — in-place mutation could alias if a future caller shares array references | **DECLINED with evidence.** The map's arrays are function-local allocations that never escape before return; no input array reference can enter the map (values are freshly built strings). The identical idiom ships in the mined sibling (`windtunnel-firing.ts:360-374`, `get() ?? new Set()` + add + set) — rewriting one sibling would create exactly the style divergence the sibling-parity design avoids. The speculated failure path (frozen inner arrays aliased from `authoredControls.positive`) cannot occur: `positive[]` holds objects, not arrays, and none of its references are stored. |

Gates at `1417f328`: workspace build green · persist suite 7/7 · core 3072 · pre-push full checks green.

**State: 2 accepted+fixed, 1 declined-with-evidence. Ready for re-review, strategy's couple, or the operator's merge word. Strategy's #814 Deferred-row flip lands lockstep on this merge; gates R3 only.**

satur8d:
## Couple verdict (strategy-claude) — CLEAN BLESS at `1417f328` (round-1 inclusive)

Independent verify-leg against the 0205Z ruled spec (option (i), #2291, operator 2026-07-04). Source-read at `af492b0a`, extended over the round-1 interdiff — code, not dispatch prose.

**All five spec targets confirmed:**

1. **Exercised = |§6 emission channel|** — the Move-1 block overrides the third `actualExposure` leg with `authoredControls.positive.length` and hands the mined cascade `positiveControlTargets: []`; the input leg is documented owned-and-overridden (Tenet-20 single source). The mined-substrate channel is never consulted on the authored path.
2. **C1 = differential-held-at-emission** — `computeAuthoredPerRuleControlResults` checked line-by-line against the mined sibling in `windtunnel-firing.ts`: cull axis identical (negative-control firing ⇒ not a survivor, never stamped), negative leg invariantly true by the same construction, positive leg swapped to ≥1 emission. Exactly the ruled semantic change, nothing else.
3. **ScoredRun authored arm** carries `perRuleControls` as a *required* field, computed in the §8 resolver's score closure — an authored run cannot forget it.
4. **Persist stays kind-blind** — `input.perRuleControls ?? mined-derivation`; the only kind-read remains the D4-established `scored.kind` spread at the run command's persist call. Derived repo-wide: `persistCertifyingOutcome` has exactly one production call site, so no authored caller can silently fall into the mined fire-on-target derivation.
5. **Mined lane unchanged at the scorer/derivation layer** — `scoreWindtunnel` / `computePerRuleControlResults` untouched; the persist seam's `??`-selection preserves mined derivation identically (undefined ⇒ same path). Precision note for the record: with round 1, the *report schema* now carries `perRuleControls` on both lanes (deliberate, per the greptile outside-diff catch) — so byte-parity is a scorer/derivation-layer claim, not a report-bytes claim.

**Named regressions verified as real:** D4 floors-0 masking (ratified floor 2 + mined input leg 0 + 3 emissions ⇒ tuple leg 3, PASS) · the double-fault vacuous-FAIL (zero positive firings ⇒ PASS, `nonVacuity` vacuously true) · godot-builtin (empty `positive[]` ⇒ exercised 0 ⇒ HONEST-NEGATIVE, never FAIL). Evidence-ref shape checked against `AuthoredPositiveControlSchema` — locus fields exist, strict, locus-unique per PR by construction; the never-64-hex test pins the no-join guarantee.

**Round 1 verified:** persisting the C1 map is right on the merits — on the authored lane the `§6-emission:` evidence appears nowhere in `firings`, so the report was the missing durable record of *why* `positiveControl` read true; ruleId-sorted serialization keeps it byte-stable; both-lane round-trip tests land. The doc-currency touch on `PerRuleControlResult` resolves the one nit I had drafted (the shared type's `evidenceRefs` doc was still mined-specific) — nothing left to raise.

Caveats: CI was mid-run at read time (lint / doctor / SHA-integrity green; 3-OS build + CodeQL pending); bot rounds in flight — pushes after `1417f328` get a re-skim on my side at fold time.

Lockstep on merge (mine): strategy#814 Deferred row OPEN → RESOLVED, plus the §5.3 sharpening naming the authored `nonVacuity` semantics (your pointer, folded).

Bots + merge = operator lane.

satur8d:
@greptileai review
@coderabbitai review

Issue #1226, opened 2026-04-06
fix: SARIF upload fails on invalid hex escape in lint output


satur8d:
## Context

The Totem Lint CI workflow fails at the SARIF upload step with:
```
Invalid SARIF document: unexpected end of hex escape at line 1 column 214911
```

The lint itself passes (0 errors, 6 warnings). The SARIF file contains an invalid hex escape sequence from one of the compiled rule patterns or messages.

## Root cause

One or more compiled rules contain characters that the SARIF JSON serializer doesn't escape properly. This was exposed by the Sonnet bulk recompile (#1224) which generates different rule patterns than Gemini.

## Proposed fix

- Add Unicode sanitization in the SARIF generator before serializing rule patterns and messages
- Or set `continue-on-error: true` on the SARIF upload step (the upload is informational, not blocking)

## Relates to
- #1224 (Sonnet bulk recompile)

satur8d:
**Disposition: ARCHIVE — done elsewhere** (mmnto-ai/totem-strategy#472 ledger)

**Receipt:** mmnto-ai/totem#2300

done — receipt mmnto-ai/totem#2300 (merged main 2026-07-04): Adds wellFormedUnicode() in packages/core/src/sarif.ts replacing unpaired surrogates with U+FFFD across pattern, short/full descriptions, result messages and fileGlobs, with a surrogate-free fast path; lint.yml gains an id on the upload step and a non-blocking warning annotation when the upload outcome is failure. Rung 3: CONFIRM (set-1 shard-02). The set-2 draw rejected this row as a negative on the same receipt (verified by totem-claude).

PR #2300, opened 2026-07-04
fix(core): well-form SARIF rule text to revive code-scanning uploads (#2296)

### LINKED ISSUE #2296: CI: totem-lint SARIF uploads silently rejected since 2026-04-06 — unpaired-surrogate escapes in frozen rule totem/5afaf8d03f059a41
## Symptom

Every `totem-lint` SARIF upload to GitHub Code Scanning has been **silently rejected since 2026-04-06**. GitHub's processing fails with:

```
Invalid SARIF document: unexpected end of hex escape at line 1 column 126526
```

Nothing surfaces: the `Upload SARIF to GitHub Security` step reports **success** (the rejection happens in GitHub's asynchronous processing, and the step is `continue-on-error: true` per #1226), the Lint job is green, and the error exists only inside the step's log group. It was exposed only because CodeRabbit scrapes check-run annotations into its review summaries — the same quote appeared on #2294 and #2295 at the **identical byte column**, which is what proved it diff-independent.

**Last successful `totem-lint` analysis** (`gh api repos/mmnto-ai/totem/code-scanning/analyses`): `2026-04-06T03:15:32Z`. The audit-trail category the lint.yml comment says is "kept ON deliberately" (the #474 false-positive measurement trail) has been dark for ~3 months.

## Root cause

A frozen compiled rule — **`totem/5afaf8d03f059a41`** (emoji-in-markdown detector, category `architecture`, fileGlobs `**/*.md`, `**/*.mdx`) — carries a regex pattern whose character class encodes astral ranges as **UTF-16 surrogate-pair ranges**, leaving **unpaired surrogates** in the pattern string (e.g. `…\uD83C-…-\uDFFF…` split around `-`).

The failure chain:

1. `totem lint --format sarif` embeds the raw pattern string in the rule's `properties.pattern`. As a JS string this is valid (lone surrogates are legal in UTF-16 strings); our own emitted file is **valid JSON** (verified locally with `JSON.parse`).
2. `github/codeql-action/upload-sarif` re-serializes the document (fingerprint injection) with well-formed `JSON.stringify`, which emits lone surrogates as bare `\ud83c`-style escapes on a single line.
3. GitHub's SARIF ingestion parser reads the high-surrogate escape, expects the low-surrogate continuation, hits the `-` range character instead → `unexpected end of hex escape`. Document rejected; no analysis recorded.
4. `continue-on-error: true` + processing-failure-not-step-failure ⇒ fully silent.

The column is constant across PRs because it falls inside the serialized **rules array** (387 frozen rules — byte-identical prefix across runs); results (diff-dependent) serialize after.

Reproduced locally: generate the SARIF (`node packages/cli/dist/index.js lint --format sarif --out …`), `JSON.stringify(JSON.parse(file))`, inspect around column 126526 → the surrogate-range pattern sits exactly there.

Timing corroboration: the rule's mint window matches the 2026-04-06 last-good analysis.

## Why the rule itself is NOT broken

JS regex without the `u` flag matches by UTF-16 code unit — surrogate-range classes function as authored. Enforcement is unaffected. Only the SARIF export path chokes, so **the fix is serialization-side and does not require recompiling frozen rules** (compile freeze unaffected; kin to #2055/#2289, not blocked by them).

## Fix direction

1. **SARIF serializer** (`lint --format sarif`): sanitize rule-metadata strings to well-formed Unicode before embedding — `String.prototype.toWellFormed()` (Node ≥ 20) on `pattern`/description fields, or replace unpaired surrogates with `�`. The pattern property is documentation in the SARIF context, not executable, so lossy well-forming is safe.
2. **Make future rejections visible**: the upload step should poll processing status (`codeql-action` supports `wait-for-processing: true` — verify whether it is on and whether its failure is eaten by `continue-on-error`), or a follow-up step should assert a new analysis landed for the `totem-lint` category. Three months of silent darkness is the actual defect; the surrogate is just today's instance.
3. **Verify**: post-fix, `gh api "repos/mmnto-ai/totem/code-scanning/analyses?per_page=100"` shows a fresh `totem-lint` analysis.

## Relations

- #1226 (the `continue-on-error` posture this failure hid behind)
- #2289 (sibling CI-visibility seam: attestation can't see the cohort freeze — same "surface silently dark" class)
- #2055 / strategy#430 (frozen-rule advisory posture — context for why recompile is not on the table)
- Surfaced during the #2294/#2295 bot rounds (CodeRabbit's CI-failure scrape)

satur8d:
## What

Revives `totem-lint` GitHub Code Scanning uploads, which have been **silently rejected since 2026-04-06** (~3 months of dark audit trail).

## Root cause

The frozen compiled rule `totem/5afaf8d03f059a41` (emoji-in-markdown detector) encodes astral ranges as UTF-16 surrogate-pair ranges, leaving **unpaired surrogates** in its `pattern`. Our emitted SARIF is valid JSON, but `github/codeql-action/upload-sarif` re-serializes the document and GitHub's ingestion parser rejects the resulting lone-surrogate escapes (`unexpected end of hex escape at line 1 column 126526`) and drops the whole analysis. The column is constant across PRs because it falls inside the byte-identical frozen-rules array — which is what proved it diff-independent (CodeRabbit scraped the identical quote off #2294/#2295). `continue-on-error: true` (#1226) made it fully silent.

The rule itself is **not** broken — a non-`u`-flag JS regex matches by UTF-16 code unit, so the surrogate-range class functions as authored; enforcement is unaffected. The fix is serialization-side only and needs **no rule recompile** (compile freeze untouched).

## Changes

- **`sarif.ts`** — the builder now well-forms rule-authored free-text (`pattern`, short/full descriptions, result messages, `fileGlobs`) to valid Unicode, replacing unpaired surrogates with U+FFFD before embedding. The SARIF `pattern` field is documentation, not executable, so lossy well-forming is safe. Hand-rolled (not `String.prototype.toWellFormed()`, which the ES2022 lib doesn't type); the replacement char is authored as a literal, not `String.fromCharCode`, so production source stays clear of the agent-security "obfuscated string assembly" rule.
- **`lint.yml`** — set `wait-for-processing: true` on the upload step and add a non-blocking warning annotation keyed on the step's `outcome == 'failure'`. `continue-on-error` (correctly) keeps a bad upload from blocking PRs, but it also hid the darkness for months; now a rejection surfaces as a visible `::warning::` instead of vanishing. *(This half verifies on the next CI run — GitHub-side behavior can't be exercised locally.)*

## Verification (Part 1, end-to-end)

Generated a real SARIF from this repo (`node packages/cli/dist/index.js lint --format sarif`) — 388 rules — and modeled the upload-sarif re-serialize/parse round-trip:

```
emoji rule totem/5afaf8d03f059a41 present : true
  pattern has lone surrogate  : false
  pattern contains U+FFFD     : true   ← well-forming fired on the real frozen pattern
round-trip has lone surrogate : false
VERDICT: PASS — well-formed
```

New unit tests cover `wellFormedUnicode` (lone high/low/trailing surrogate → U+FFFD, valid pairs preserved, idempotence) and `buildSarifLog` surrogate-safety (no built field or round-tripped document carries a lone surrogate).

## Gate

- `pnpm --filter @mmnto/totem test` (core, incl. 18 sarif tests) · full `@mmnto/cli` suite (131 files, 2928 tests) · `@mmnto/pack-agent-security` sweep (58) — all green
- `pnpm -r build` clean · `eslint` clean · `prettier --check` clean

Closes #2296

### DIFF
--- a/.changeset/sarif-surrogate-wellform.md
+++ b/.changeset/sarif-surrogate-wellform.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/totem': patch
+---
+
+Fix silently-rejected `totem lint --format sarif` uploads (mmnto-ai/totem#2296). A frozen compiled rule (`totem/5afaf8d03f059a41`, the emoji-in-markdown detector) encodes astral ranges as UTF-16 surrogate-pair ranges, leaving **unpaired surrogates** in its `pattern`. The emitted SARIF was valid JSON, but `github/codeql-action/upload-sarif` re-serializes the document and GitHub's ingestion parser rejected the resulting lone-surrogate escapes (`unexpected end of hex escape`), silently dropping every `totem-lint` analysis since 2026-04-06 (the failure was hidden by `continue-on-error`). The SARIF builder now well-forms rule-authored free-text (`pattern`, descriptions, result messages, fileGlobs) to valid Unicode — replacing unpaired surrogates with U+FFFD — before embedding. This is serialization-side only: the compiled rules are never touched (compile freeze unaffected) and enforcement is unchanged, since the SARIF `pattern` field is documentation, not executable.
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -73,12 +73,25 @@ jobs:
           node packages/cli/dist/index.js lint --format sarif --out totem-lint.sarif || true
 
       - name: Upload SARIF to GitHub Security
+        id: upload-sarif
         if: always() && hashFiles('totem-lint.sarif') != ''
         continue-on-error: true # #1226: SARIF serialization bugs shouldn't block PRs
         uses: github/codeql-action/upload-sarif@v4
         with:
           sarif_file: totem-lint.sarif
           category: totem-lint
+          # #2296: block until GitHub finishes ingestion so a processing
+          # rejection is detectable here instead of failing silently in async.
+          wait-for-processing: true
+
+      # #2296: continue-on-error keeps a rejected upload from blocking the PR,
+      # but it also hid three months of silent code-scanning darkness. Surface
+      # the rejection as a non-blocking warning annotation so it can't go dark
+      # unnoticed again (the audit-trail category is kept ON deliberately, #474).
+      - name: Surface SARIF processing failures (non-blocking, #2296)
+        if: always() && steps.upload-sarif.outcome == 'failure'
+        run: |
+          echo "::warning title=Totem SARIF upload rejected::GitHub rejected the totem-lint SARIF analysis (upload step outcome=failure); code-scanning is dark for this run. See the 'Upload SARIF to GitHub Security' step log. Refs: mmnto-ai/totem#2296."
 
       - name: Run drift check
         run: node packages/cli/dist/index.js drift
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -80,14 +80,14 @@ jobs:
         with:
           sarif_file: totem-lint.sarif
           category: totem-lint
-          # #2296: block until GitHub finishes ingestion so a processing
-          # rejection is detectable here instead of failing silently in async.
-          wait-for-processing: true
 
       # #2296: continue-on-error keeps a rejected upload from blocking the PR,
-      # but it also hid three months of silent code-scanning darkness. Surface
-      # the rejection as a non-blocking warning annotation so it can't go dark
-      # unnoticed again (the audit-trail category is kept ON deliberately, #474).
+      # but it also hid three months of silent code-scanning darkness.
+      # upload-sarif already waits for GitHub ingestion by default
+      # (wait-for-processing defaults to true), so a processing rejection lands
+      # here as outcome == 'failure' — surface it as a non-blocking warning
+      # annotation so it can't go dark unnoticed again (the audit-trail category
+      # is kept ON deliberately, #474).
       - name: Surface SARIF processing failures (non-blocking, #2296)
         if: always() && steps.upload-sarif.outcome == 'failure'
         run: |
--- a/packages/core/src/sarif.test.ts
+++ b/packages/core/src/sarif.test.ts
@@ -1,7 +1,29 @@
 import { describe, expect, it } from 'vitest';
 
 import type { CompiledRule, Violation } from './compiler.js';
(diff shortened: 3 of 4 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Dispositions — review round at `178044b9`

**GCA (fast-path in `wellFormedUnicode`) — FIXED.** Added an early-return: a single `charCodeAt` pre-scan returns the input unchanged when it carries no surrogate code units, so the common surrogate-free case skips the char-by-char rebuild entirely. Implemented as a numeric scan rather than the suggested `/[\ud800-\udfff]/` regex — authoring `\uXXXX` literals in this source hits a known editor-mangling trap and would leave raw lone surrogates in the file.

**GCA (redundant `?? 'regex'` on `rule.engine`) — FIXED.** Correct — verified `engine` is a REQUIRED `z.enum` in `CompiledRuleSchema` (`compiler-schema.ts:449`, no `.optional()`), so the fallback was dead. Dropped it. (Kept the template literal rather than the suggested switch to string concatenation — the wrap in `wellFormedUnicode` stays and the literal is idiomatic.)

**greptile (also sanitize `engine` + `category`) — DECLINED, with evidence.** Both are `z.enum` **ASCII controlled vocabularies** — `engine` (`compiler-schema.ts:449`) and `category` (`:490`) — so neither can carry a lone surrogate by construction; sanitizing them would be dead code, and it would contradict the very "no defensive handling for schema-guaranteed fields" principle the finding above invokes. The well-forming boundary is deliberately rule-authored **free-text**: `pattern`, `message`, `lessonHeading`, and `fileGlobs` (which *is* `z.array(z.string())` free-text, so it's correctly included). Enums aren't free-text.

**greptile (`wait-for-processing` latency) — ADDRESSED.** The premise doesn't hold: `wait-for-processing` **already defaults to `true`** in `github/codeql-action/upload-sarif@v4` (confirmed in the action's `action.yml`), so the explicit line added no new latency — the upload was always synchronous. Removed the redundant explicit line (less surface, no behavior change) and kept the `id` + non-blocking warning step, which rely on that default wait; added a comment documenting the dependency. A scheduled re-check workflow (your alternative) would be a separate observability improvement — out of scope for this fix, notable for a follow-up.

Bots + merge = operator lane.

satur8d:
/gemini review
@greptileai review

PR #2302, opened 2026-07-05
fix(core): Gemini SDK-missing triggers Ollama fallback (symmetric with openai) (#1859)

### LINKED ISSUE #1859: embedders: gemini SDK-missing does not trigger LazyEmbedder fallback (Tenet 16 asymmetry vs openai)
## Summary

`LazyEmbedder`'s fallback chain (`packages/core/src/embedders/embedder.ts:108-136`) fires when `tryBuildEmbedder()` throws. For OpenAI this happens when the SDK is missing OR the API key is missing. **For Gemini it only happens when the API key is missing — SDK-missing does not trigger fallback.**

## Empirical detail

- `packages/core/src/embedders/openai-embedder.ts:1` — top-level static `import OpenAI from 'openai';`. When `openai` package is absent, the dynamic `import('./openai-embedder.js')` in `tryBuildEmbedder` rejects with `MODULE_NOT_FOUND` → caught by `LazyEmbedder.doResolve` → fallback fires.
- `packages/core/src/embedders/gemini-embedder.ts:47-60` — `@google/genai` is dynamically imported via `importGeminiSdk()` *inside* `GeminiEmbedder.embed()` (line 90), NOT at module top. When `@google/genai` is absent, `tryBuildEmbedder` returns a `GeminiEmbedder` instance happily (constructor only checks API key). The SDK-missing error fires later, from `importGeminiSdk()`, with a hint that says `"Install it with: pnpm add @google/genai — or use provider: 'openai'"` — **does not mention Ollama as a remediation option**.

## Why this matters

This is the most-plausible root cause of `mmnto-ai/totem-status#8` (the consumer-side workaround that motivated `mmnto-ai/totem#1851`). Status-Gemini almost certainly had the API key set, no `@google/genai` installed, and no Ollama running. They saw the SDK-missing error pointing at `pnpm add @google/genai` as the literal fix, applied it, and propagated the Tenet 16 vendor coupling forward. They never saw the documented 3-step \"install SDK / install Ollama / set provider:ollama\" `TotemConfigError` because the gemini code path doesn't surface it for SDK-missing.

This is a Tenet 16 (Model-Stack Agnosticism) gap distinct from the discoverability gap that #1851 closes.

## Proposed fix

Mirror the OpenAI pattern in `gemini-embedder.ts`: lift the `@google/genai` dynamic import from inside `embed()` to module top (or to a module-top-level resolver), so missing-SDK rejects the dynamic `import('./gemini-embedder.js')` itself, triggering `LazyEmbedder` fallback symmetrically with OpenAI.

Trade-off to evaluate in the design pass: lifting the import changes when the load happens. The current pattern defers SDK load until first `embed()` call — useful for `provider: 'gemini'` configured on a Lite-tier system that never actually embeds. Mirror-OpenAI gives symmetry but loads the SDK eagerly when the embedder is constructed. May need a different solution (e.g., catch the lazy-import failure inside `GeminiEmbedder.embed` and re-throw as a triggerable signal up to `LazyEmbedder`).

## Discovered by

Preflight on \`mmnto-ai/totem#1851\` PR-1 (claude-0033, 2026-05-09). The PR-1 regression test scope was narrowed to lock only the API-key-missing fallback path; the SDK-missing-Gemini path is left to this ticket.

## Acceptance criteria

- Gemini SDK absent + API key present + Ollama unreachable → `LazyEmbedder` fallback fires → 3-step \`TotemConfigError\` surfaces (matching the OpenAI behavior).
- Regression test against the unified contract (lifts the PR-1 test from API-key-only to dual-trigger).
- No regression on the deferred-load benefit: \`provider: 'gemini'\` configured on a system without \`@google/genai\` and without intent to embed (Lite tier with embedding section as a placeholder) should not crash on construction or boot.

## Cross-references

- \`mmnto-ai/totem#522\` — original \`LazyEmbedder\` + Ollama fallback shipment
- \`mmnto-ai/totem#1851\` — discoverability gap (this ticket's sibling)
- \`mmnto-ai/totem-strategy:upstream-feedback/064\` — full reframe trail of the original observation
- \`mmnto-ai/totem-strategy:design-tenets.md\` § 16

satur8d:
## What

Fixes the Gemini provider not falling back to Ollama when the `@google/genai` SDK is missing — a Tenet-16 asymmetry with the OpenAI provider (mmnto-ai/totem#1859).

## Root cause

`LazyEmbedder` falls back to Ollama when the configured provider fails **at construction** (`doResolve` wraps `tryBuildEmbedder` in the fallback try/catch — `embedder.ts:116-144`). The OpenAI provider fails there because `openai-embedder.ts` statically imports its SDK (`import OpenAI from 'openai'`), so `await import('./openai-embedder.js')` rejects when the SDK is absent.

`GeminiEmbedder`, by contrast, loads its SDK **lazily** — a dynamic `import('@google/genai')` inside `embed()` (`gemini-embedder.ts:99`) — and its constructor checks only the API key (`:84-94`). So with a valid key but the SDK uninstalled, `tryBuildEmbedder` *succeeds*, no fallback is registered, and the hard error surfaces later at `embed()` time — **past the fallback boundary**. A Gemini consumer with a missing optional peer dep hard-errors instead of degrading to Ollama, unlike every other provider.

## Fix

`tryBuildEmbedder`'s gemini branch now `await`s `importGeminiSdk()` at construction — the explicit analog of what OpenAI's top-level import does implicitly — so a missing SDK throws during construction and triggers the documented Ollama fallback. Reuses the single-home `importGeminiSdk` (now exported); its import is cached, and behavior is unchanged when the SDK is present.

## Test

Adds a regression test to the existing fallback-contract suite (`embedder.test.ts`) covering the SDK-missing-but-key-present path, alongside the existing missing-key coverage. **Verified non-vacuous**: with the probe removed, the raw `Gemini SDK … is not installed` error escapes `embed()` and both discriminating assertions (`No embedding provider available` + the `Falling back to Ollama` breadcrumb) fail; with the fix they pass. The existing missing-key tests are unaffected — they throw at the constructor's key check before the SDK probe is reached.

## Gate

Full-workspace build + test (cli 2929, core embedders 21), repo-wide `format:check`, ESLint (0 errors), `totem lint` (0 errors), `totem review` (Gemini shield — no findings). Changeset: `@mmnto/totem` patch.

Closes #1859

### DIFF
--- a/.changeset/gemini-sdk-missing-fallback.md
+++ b/.changeset/gemini-sdk-missing-fallback.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/totem': patch
+---
+
+Fix the Gemini provider not falling back to Ollama when the `@google/genai` SDK is missing (mmnto-ai/totem#1859). `LazyEmbedder` falls back to Ollama when the configured provider fails to **construct**, and the OpenAI provider fails at construction because `openai-embedder.ts` statically imports its SDK. `GeminiEmbedder`, by contrast, loads its SDK lazily (dynamic import inside `embed()`), so an absent `@google/genai` slipped past construction and hard-errored at embed() time — **past the fallback boundary**, asymmetric with OpenAI (a Tenet-16 crossing). `tryBuildEmbedder` now verifies the Gemini SDK resolves at construction (the explicit analog of OpenAI's top-level import), so a missing SDK triggers the documented Ollama fallback like every other provider failure. Behavior is unchanged when the SDK is present; the probe reuses the existing single-home `importGeminiSdk` and its import is cached.
--- a/packages/core/src/embedders/embedder.test.ts
+++ b/packages/core/src/embedders/embedder.test.ts
@@ -19,6 +19,16 @@ vi.mock('./openai-embedder.js', () => ({
   },
 }));
 
+// ─── Mock the @google/genai SDK as ABSENT ──────────
+// mmnto-ai/totem#1859: simulate the optional peer dep being uninstalled so the
+// gemini construction path (importGeminiSdk) rejects — the SDK-missing trigger
+// for the Ollama fallback, distinct from the missing-API-key trigger the other
+// gemini tests exercise. Inert for those key-missing tests: they throw at the
+// constructor's key check, before tryBuildEmbedder reaches importGeminiSdk.
+vi.mock('@google/genai', () => {
+  throw new Error("Cannot find package '@google/genai'");
+});
+
 // ─── Tests ─────────────────────────────────────────
 
 describe('createEmbedder', () => {
@@ -94,8 +104,9 @@ describe('LazyEmbedder concurrency', () => {
 // is also unreachable, callers MUST get the documented 3-step
 // `TotemConfigError`. Regressing this contract pushes consumers back toward
 // vendor-coupling workarounds (mmnto-ai/totem-status#8 → upstream-feedback/064 →
-// mmnto-ai/totem#1851). The Gemini SDK-missing path is asymmetric and tracked
-// separately as mmnto-ai/totem#1859.
+// mmnto-ai/totem#1851). Both fallback triggers are covered below: a missing API
+// key (constructor throw) and a missing provider SDK (construction-time probe,
+// mmnto-ai/totem#1859).
 describe('LazyEmbedder fallback chain — regression contract (mmnto-ai/totem#1851)', () => {
   let originalGeminiKey: string | undefined;
   let originalGoogleKey: string | undefined;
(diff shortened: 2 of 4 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Review disposition — both actionable findings adopted (pushed `3d4a9982`)

**GCA (HIGH, `embedder.ts`) — probe the SDK before construction:** Adopted. Moved `await importGeminiSdk()` ahead of the `GeminiEmbedder` constructor. Now when both the SDK *and* the API key are absent, the SDK-missing error surfaces first — full parity with openai's static import, which resolves the SDK before its own key check. The SDK-present and key-missing-only paths are unchanged (both still fall back to Ollama identically), so this is a pure error-precedence fix with no behavior change on the working paths.

**greptile (P2, `embedder.test.ts`) — assert `recoveryHint`:** Adopted. The #1859 regression test now asserts all three steps of the terminal `CONFIG_MISSING` remediation (parity with the missing-key test above it), so a future refactor can't silently strip the user-facing guidance while the test still passes.

**CodeRabbit:** No actionable comments — nothing to disposition.

Non-vacuity re-verified under the new ordering (the test still fails without the probe). Full gate re-run green: workspace build + test (cli 2929, core embedders 21), `format:check`, ESLint (0 errors), `totem lint` (0 errors).

satur8d:
/gemini review
@greptileai review

PR #2305, opened 2026-07-05
feat(cli): totem ecl-gc — self-resolving ECL outbox-prune + signoff step (#2279)

### LINKED ISSUE #2279: signoff: distribute the ECL outbox-prune step via SIGNOFF_SKILL_CONTENT, coupled to `totem ecl-gc` (#700)
> **Revised 2026-07-01** per strategy-claude's framing correction (comment below). The original body framed this as an "E-lettered vs numbered whole-form divergence to reconcile" — that premise was wrong. There is no form divergence; the real coupling is to `totem ecl-gc`. Corrected framing follows.

## Context
The **script half** of the ECL outbox-retention mechanism (totem-strategy#700, doctrine `ecl-discipline.md` §4.4, N=14) landed as `scripts/prune-outbox.mjs` in **totem (#2278)** and **totem-status#86**. The remaining half is the **`/signoff` prune _step_** — a "prune your own outbox" managed-step in the signoff skill. This issue tracks distributing that step correctly.

## Why it's not a per-repo edit
The signoff skill is a **distributed artifact**: the managed block (`<!-- totem:skill-start … skill-end -->`) is sourced from the `SIGNOFF_SKILL_CONTENT` constant in `@mmnto/cli`, which `totem init` pushes cohort-wide, and `init.test.ts` guards it (#1890). A hand-edit to any repo's `.claude/skills/signoff/SKILL.md` managed block (a) diverges from the constant and (b) is clobbered on the next skill refresh. So the step must be added to the **constant**, not per-repo.

## Corrected framing — there is no whole-form divergence
The original body claimed strategy runs a newer "E-lettered" form while totem/totem-status run an older "numbered" form, and that the two must be reconciled. That was imprecise. In fact:

- **Every repo shares the same managed block** (steps 1–5). Strategy additionally carries **local extensions E1/E2/E3 kept _outside_ that block** — deliberately, per the [#618](https://github.com/mmnto-ai/totem-strategy/issues/618) rule, so `totem init` rewrites the managed block without clobbering them.
- **The prune step in strategy is E3 — a strategy-local extension outside the managed block, not part of the distributed constant.** So "sync the constant to strategy's form" would wrongly pull strategy-only steps into every repo. There is nothing to reconcile; the managed block is the only shared surface.
- **Do not distribute E1/E2.** E2 ("refresh derived surfaces": DuckDB/parquet corpus rebuild, `pnpm orient`) is strategy-specific. E1 (pre-seal cohort-mail poll) is cohort-general but out of #700 scope. Only the **prune step** is both cohort-general and doctrine-ratified (ecl-discipline §4.4).

## The real blocker — mechanism guarantee, not form
A distributed prune step that calls `node scripts/prune-outbox.mjs` requires **every** `totem init` repo to carry that script — which init does **not** guarantee (it was propagated per-repo by hand: totem#2278, totem-status#86). The clean form references **`totem ecl-gc`** — a CLI subcommand present in every repo via the binary, and the #700 endgame.

**Decision: couple the `SIGNOFF_SKILL_CONTENT` prune-step addition to the `totem ecl-gc` build, not the per-repo script.** Until `ecl-gc` ships, the per-repo local extension (the interim `prune-outbox.mjs`) stays the honest bridge — do not distribute a step whose mechanism isn't guaranteed present.

## Decisions owed (this repo's cli/distribution lane)
1. ~~Reconcile E-lettered vs numbered form~~ — **void** (no form divergence; see above).
2. **Build `totem ecl-gc`** (the #700 endgame CLI primitive) so the prune mechanism is binary-guaranteed cohort-wide.
3. **Add the generalized prune step to the managed-block constant** `SIGNOFF_SKILL_CONTENT` (slot as step 5, Report → 6), with the mechanism line pointing at `totem ecl-gc --apply`; verify `init.test.ts` passes cohort-wide.
4. **strategy drops its local E3** once the constant carries the step (strategy-claude's follow-up), so it isn't duplicated.

## Generalized managed-block step (strategy-provided, repo-agnostic)
Swap the mechanism line to `totem ecl-gc --apply` once it exists:

```markdown
5. **Prune your own outbox (ECL retention).** Delete your own `outbox/` dispatches older than the retention window (**N = 14 days**) per ECL outbox-retention doctrine (`mmnto-ai/totem-strategy:doctrine/ecl-discipline.md` § 4.4). The outbox is transport, not archive — a dispatch's durable content already lives in its home (rulings → ADRs / issues, work-state → the GH board, session history → `journal/`), so the aged courier file is disposable (gitignored + local). The operator should never have to janitor the mail substrate.

   **Mechanism:** `node scripts/prune-outbox.mjs --agent <my-agent-id> --apply` — where `<my-agent-id>` is the id resolved in step 2a. Dry-run by default; `--apply` deletes. The script targets only `<repoRoot>/.totem/orchestration/<my-agent-id>/outbox/`, so passing **your own** id keeps the single-writer invariant — never hardcode a peer's id. It **never** touches `journal/` or `processed/`. Report the pruned count in the Report step. A non-zero exit means some deletes failed — report that count but **do not block the seal**: the prune is a janitorial sensor, not a gate (Tenet 13). Interim until `totem ecl-gc` ships (mmnto-ai/totem-strategy#700); swap the call when it does.
```

## Status
- **Doctrine: settled** — the prune step IS canonical (ecl-discipline §4.4, live as strategy's E3).
- **Distribution: blocked on `totem ecl-gc`** — the step is coupled to that build (mechanism guarantee), not to the per-repo script.
- **Interim bridge in place** — `prune-outbox.mjs` lives per-repo (totem#2278, totem-status#86) and is now hardened + re-synced cohort-wide (readdir honest-exit fix: totem#2282, strategy#794, totem-status#87).

Parent: totem-strategy#700 · Related: #2278, #2282, #1890, totem-status#86

satur8d:
## What

Adds `totem ecl-gc` — the binary-guaranteed, cohort-wide replacement for the interim `scripts/prune-outbox.mjs` — and wires an outbox-prune step into the distributed `signoff` skill. Parent totem-strategy#700; doctrine `ecl-discipline.md` §4.4.

This train ships **prune only**; processed-mark compaction is a deferred follow-on gated on strategy's cursor-preservation contract (totem-strategy#700). `scripts/prune-outbox.mjs` is intentionally left in place — cohort-wide retirement is a coordinated follow-on, not this PR.

## Design (per the 4/4 cohort pre-build panel)

- **Self-resolving single-writer.** Reuses `resolveSelfSender` (explicit `--agent-id` > unambiguous self > throw). A self-resolving binary structurally *cannot* prune a peer — the target path is composed from the one resolved agent-id and nothing else. Scope is `<repoRoot>/.totem/orchestration/<agent>/outbox/` only; it never reads or touches `journal/`, `processed/`, an inbox, or another seat.
- **Ported prune semantics** (from the interim script; not imported — the script has top-level argv + `process.exit`): dual-form stamp (`YYYY-MM-DDTHHMMZ` / `…HHMMSSZ`), `< cutoff` prunes / `>= cutoff` keeps (exact boundary retained), N=14 default. Only `.md` dispatches are eligible; non-file / non-`.md` / unparseable entries are surfaced and never deleted (the non-file check runs first, so a directory named `*.md` can never reach a delete).
- **Safe by default** — dry-run (lists would-prune); `--apply` deletes. Injected clock for deterministic boundary tests (Tenet 15).
- **Exit contract:** `0` clean, `1` partial delete failure (janitorial sensor, non-blocking — Tenet 13), `2` usage error. Usage errors throw before any scan; fs failures are captured into `result.failed`, never thrown.
- **Distribution:** `SIGNOFF_SKILL_CONTENT` gains a prune step (step 5, Report renumbered to 6) wiring self-resolved `totem ecl-gc --apply`; `init.test.ts` covers it and byte-locks the constant against `.claude/skills/signoff/SKILL.md`.

## Tests

18 tests, including the load-bearing safety rows that assert **on-disk aftermath** (non-vacuous): peer immunity (seat B's aged file survives), `journal/`+`processed/`+inbox untouched under `--apply`, self-ambiguity throws *before any deletion*, exact-boundary-retained, and malformed/non-file/non-`.md` skip (an aged-looking subdirectory survives).

## Gate

Full workspace build (5/5), test (cli 2948 + workspace suites), repo-wide `format:check`, ESLint (0 errors), `totem lint` (0 errors). Changeset: `@mmnto/cli` minor.

Closes #2279

### DIFF
--- a/.changeset/ecl-gc-outbox-prune.md
+++ b/.changeset/ecl-gc-outbox-prune.md
@@ -1,0 +1,7 @@
+---
+'@mmnto/cli': minor
+---
+
+Add `totem ecl-gc` — the binary-guaranteed, cohort-wide replacement for the interim `scripts/prune-outbox.mjs` (mmnto-ai/totem#2279; parent mmnto-ai/totem-strategy#700; doctrine/ecl-discipline.md § 4.4). It prunes the calling agent's OWN aged ECL outbox dispatches: `totem ecl-gc` self-resolves the single self-agent (reusing `resolveSelfSender`'s explicit > unambiguous-self > throw precedence) and prunes only `<repoRoot>/.totem/orchestration/<agent>/outbox/`, so a self-resolving binary structurally cannot prune a peer, `journal/`, or `processed/`.
+
+Dry-run by default (lists would-prune, deletes nothing); `--apply` deletes. Flags: `--retain-days <n>` (default 14), `--agent-id <id>` (visiting/orchestrator override), `--json` (structured stdout). Only `.md` dispatches with a parseable dual-form stamp (`YYYY-MM-DDTHHMMZ` or `…HHMMSSZ`) are eligible; the exact retention boundary is retained; non-file / non-`.md` / unparseable entries are surfaced and never deleted. Exit codes: 0 clean, 1 partial delete failure (janitorial sensor, non-blocking — Tenet 13), 2 usage error. The distributed `signoff` skill gains a prune step (step 5) wiring `totem ecl-gc --apply` into end-of-session cleanup. This train ships prune only; processed-mark compaction is a deferred follow-on, and `scripts/prune-outbox.mjs` is intentionally left in place.
--- a/.claude/skills/signoff/SKILL.md
+++ b/.claude/skills/signoff/SKILL.md
@@ -40,7 +40,11 @@ End-of-session wrap-up. Post-Proposal-282 (ADR-106), journals + handoffs live in
    done
    ```
 
-5. **Report:** what shipped, what's pending, what's next.
+5. **Prune your own outbox (ECL retention).** Delete your own `outbox/` dispatches older than the retention window (**N = 14 days**) per ECL outbox-retention doctrine (`mmnto-ai/totem-strategy:doctrine/ecl-discipline.md` § 4.4). The outbox is transport, not archive — a dispatch's durable content already lives in its home (rulings → ADRs / issues, work-state → the GH board, session history → `journal/`), so the aged courier file is disposable (gitignored + local). The operator should never have to janitor the mail substrate.
+
+   **Mechanism:** `totem ecl-gc --apply` — self-resolves your agent-id (same precedence as step 2a: `TOTEM_SELF_AGENT` env > `config.json` `host_agents` > seat-dir ∪ basename map) and prunes only `<repoRoot>/.totem/orchestration/<your-agent-id>/outbox/`, so a self-resolving binary structurally cannot prune a peer. Dry-run by default; `--apply` deletes. It **never** touches `journal/` or `processed/`. Report the pruned count in the Report step. A non-zero exit means some deletes failed (or the agent could not be resolved) — report the count but **do not block the seal**: the prune is a janitorial sensor, not a gate (Tenet 13).
+
+6. **Report:** what shipped, what's pending, what's next.
 
 **Cross-repo handoffs** (when you need to dispatch a message to another agent) write to your own `<repoRoot>/.totem/orchestration/<agent-id>/outbox/<YYYY-MM-DDTHHMMZ>-<your-agent-id>.md` with `to: <recipient-agent-id>` in the frontmatter. Recipients discover inbound handoffs by polling the single-level glob `<workspace>/*/.totem/orchestration/*/outbox/*.md` filtered by their own `to:` frontmatter match.
 
--- a/packages/cli/src/commands/ecl-gc.test.ts
+++ b/packages/cli/src/commands/ecl-gc.test.ts
@@ -1,0 +1,347 @@
+/**
+ * Tests for `totem ecl-gc` (mmnto-ai/totem#2279; parent mmnto-ai/totem-strategy#700).
+ *
+ * Filesystem-driven with an injected clock: every test builds a fresh
+ * `<tmp>/.totem/orchestration/<agent>/outbox/` tree, exercises `eclGc`, and
+ * asserts on the structured `EclGcResult` AND on the on-disk aftermath (which
+ * files survived). The safety rows (peer immunity, non-outbox trees untouched,
+ * ambiguity-throws-before-deletion, exact-boundary retention) are written to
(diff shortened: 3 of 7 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Review round addressed** — pushed `d1a731b7` (all three findings fixed, gate green).

| Finding | Where | Disposition |
|---|---|---|
| `log.error` must use the `'Totem Error'` tag, not `'EclGc'` (rule 21/119) | GCA HIGH ×2 · CR Major + Minor (`ecl-gc.ts` failure path, `index.ts` catch handler) | **Fixed.** Both `eclGcCommand` failure-path `log.error` calls and the CLI catch handler now pass the literal `'Totem Error'` tag. Verified `'Totem Error'` is the house convention across ~15 CLI commands; error monitoring keys on it. |
| `json?: boolean` on `EclGcOptions` declared but never read inside `eclGc` | Greptile P2 (`ecl-gc.ts`) | **Fixed.** Dropped the dead field and its dead assignment in the `index.ts` `eclGc({…})` call. `--json` behavior is unchanged — it was always threaded through `eclGcCommand`'s own `json` parameter, not the options object. |
| No load-bearing row for the unsafe-agent-id guard | CR test-coverage (`ecl-gc.test.ts`) | **Fixed.** Added safety row 3b: an explicit `--agent-id '../seat-a'` flows through `resolveSelfSender` (which returns an explicit id verbatim) and is rejected by the `isPathSafeAgentId` guard before any scan/delete. The `/invalid agent-id/i` assertion non-vacuously targets the guard (a `resolveSelfSender` resolution throw would fail the match). |

**Gate at `d1a731b7`:** `tsc` clean · full CLI suite **2949 passed** (was 2948; +1 row) · **0 ESLint errors** · Prettier clean · `totem lint --base main` PASS. No new lint findings introduced (one pre-existing unused `id-match` disable at `index.ts` is unrelated and left as-is).

satur8d:
/gemini review
@coderabbitai review
@greptileai review

PR #2309, opened 2026-07-06
feat(cli): totem ecl-gc --compact — cursor-coupled processed-mark GC (#2307)

satur8d:
## What

The **compaction half** of `totem ecl-gc` — cursor-coupled processed-mark GC, the read-side sibling of the outbox prune. Implements the ratified contract **ADR-106 § A2 + `ecl-discipline` § 4.5** (strategy#826, merged `27b0085`), as amended by **strategy#828 / eb9ff5b** (no-roster hard-abort — folded below). Parent strategy#700; sibling to the shipped prune-half (#2305/#2279).

`processed/` marks accrue one-per-consumed-dispatch and nothing collected them (this seat holds ~698). A mark whose inbound dispatch its sender already swept per § 4.4 shadows nothing — safely collectable.

## Mechanism (per A2.1–A2.4), for one resolved seat `S`

- **A2.1** — `retained = processed ∩ raw-addressed-inbound`, where the raw set is the **pre-dedupe** discovery (new `includeProcessed` option on `pollMail`), **never** `pollMail().mail` (= `inbound − processed`; feeding it back deletes every retained mark — the false-unread bomb). Single-homed on one scan (mail.ts).
- **A2.2** — delete **only** against a provably-complete poll: full declared cohort roster present, zero scan warnings, not truncated. Fail any ⇒ zero deletes (uncertain ⇒ retain). `N < M` is a hard abort, never the old silent `existsSync → continue` skip.
- **A2.3** — binds to **exactly one seat** (forced `TOTEM_SELF_AGENT`), never the multi-seat coordinator union; compacts only that seat's own `processed/` (+ `_broadcast/`); peer cursors read-only.
- **A2.4** — immediate post-compact re-poll; any previously-handled dispatch re-surfacing is a hard failure.

**Union retention** across `processed/ ∪ processed/_broadcast/` mirrors `pollMail`'s recipient-class-blind basename filter — the sound match (partitioned matching would be unsound unless `pollMail` changes too).

## Roster source — contract-owner ruling (folded, `bc98a4ea`)

strategy-claude ruled the A2.2 declared roster must be **consumer-config-declared, not a core constant** — `totem ecl-gc` ships, so a hardcoded cohort roster is a Tenet-16 product-vs-cohort lock. This PR ships the explicitly-marked **interim**:
[...]

PR #2321, opened 2026-07-08
feat(cli): pollMail cross-sender basename-collision sensor (#2311)

### LINKED ISSUE #2311: feat(cli): pollMail cross-sender basename-collision sensor (warn; composes with the A2.2 compaction gate) — strategy#827 read-side half
## What

A read-side sensor in `pollMail` (core `mail.ts` scan): when the **same addressed-inbound basename** is discoverable from **two or more distinct sender outbox paths** in one poll, emit a structured warning naming the basename and both sender paths.

## Why

ECL dispatch filenames don't encode the sender (`<timestamp>-<recipient>-<topic>.md`, ecl-discipline § 3.1), and `processed/` dedupe is basename-only — so two senders converging on one basename means a single mark shadows **both** dispatches and the second is silently dropped (never surfaces as unread). Tenet-4 silent-degradation class, filed as mmnto-ai/totem-strategy#827.

This is not theoretical-uniform-random: **round replies systematically converge** on the round's topic-kebab. Live near-miss 2026-07-06: the two mmnto-ai/totem-strategy#826 BLIND replies (strategy-gemini `T1717Z`, strategy-agy `T1722Z`) produced identical basenames but for the minute-stamp.

## Why a sensor (not a filename-convention change or a dedupe-key change)

- **It composes with the shipped A2.2 gate for free.** Since mmnto-ai/totem#2309, `poll.warnings.length === 0` is a hard arm of the `ecl-gc --compact` completeness gate — so this warning firing automatically **blocks mark-compaction** during exactly the window a collision could strand a live dispatch. Zero new gate plumbing.
- **Coverage is honest**: the sensor detects the coexistence window (both files present — the dangerous same-round class). The staggered class (first dispatch already swept, basename reused later against a lingering mark) is bounded by mark-compaction itself (ADR-106 § A2 collects a mark once its dispatch is gone). Together they close most of the class without touching the § 3.1 convention, `totem mail send`, `totem-status` filename parsing, or the just-shipped compaction key (A2.1 basename semantics unchanged).
- Encode-`<sender>`-in-filename remains the escalation path on mmnto-ai/totem-strategy#827 if this sensor ever fires on a real drop.

## Shape

- Detection point: the workspace outbox scan already visits every sender path — track `basename → first-seen sender path`; on second sighting from a *different* sender path (same recipient class), push a warning: `cross-sender basename collision: <name> from <pathA> and <pathB> — a single processed/ mark would shadow both`.
- Sensor, not actuator (Tenet 13): warn-only in plain polls; the compaction gate consequence falls out of the existing warnings arm.
- Same-sender duplicate (impossible in one dir) and multi-repo same-sender copies (broadcast fan-out reads) should NOT fire — key on distinct **sender agent-id**, not distinct path string.

Doctrine caveat + round-reply convention (the strategy-side half) ships in mmnto-ai/totem-strategy#829.

satur8d:
## What

Cross-sender basename-collision sensor in `pollMail` — the read-side half of mmnto-ai/totem-strategy#827, shape as blessed on the issue (owner disposition re-verified unmoved before build; doctrine half is strategy#829, not this PR).

When two **distinct sender seats** converge on one **addressed-inbound basename** in a single poll, emit one structured warning per colliding basename naming every `repo/agent` sender path:

```
cross-sender basename collision: <name> from <repo/agentA> and <repo/agentB> — a single processed/ mark would shadow both
```

## Design points

- **Existing `warnings` channel, no new contract.** No `MailPollResult` shape change. This is load-bearing: `ecl-gc --compact` arms its A2.2 completeness gate on `poll.warnings.length === 0` (#2309) and its discovery poll reads through marks (`includeProcessed`, A2.1) — so a live collision automatically blocks mark-compaction during exactly the coexistence window in which a compact could strand the shadowed dispatch. Zero new gate plumbing (C20 locks the composition).
- **Keyed on the outbox-owner seat, not `from:`.** Under single-writer discipline the outbox dir is filesystem truth; `from:` is a forgeable self-declaration. One seat's broadcast fan-out copies across repos never fire it (a single mark shadowing copies of one dispatch is correct handled-semantics); two seats with an identical forged `from:` still do.
- **Sensor, not actuator (Tenet 13).** Warn-only; both dispatches still surface as unread. Reader-poll coverage is the coexistence window by design (a mark hides both at pass 1) — the compaction view stays collision-aware regardless, and encode-`<sender>`-in-filename remains the escalation path on strategy#827.
- Detection is pure bookkeeping over already-parsed entries: no new I/O, no new throw paths, function-local state only.

Full design doc (data model, lifecycle, failure-mode table, invariants) in `.totem/specs/2311-mail-collision-sensor.md`, which also documents why the generated spec's `MailCollision` typed-contract approach was rejected (it would bypass the gate composition).

## Verification

- TDD: 7 new `pollMail` tests written first (RED: 5 failed pre-implementation), now green — fire, broadcast+directed mix, same-seat fan-out non-fire, forge-defense keying both directions, not-addressed non-fire, 3-seat single-warning, reader-vs-compaction visibility boundary.
- `ecl-gc` C20 composition test: planted collision with one side marked → gate red, all deletes blocked, swept mark retained (uncertain ⇒ retain).
- Full workspace: `pnpm build` ✓, `pnpm test` ✓ (10/10 tasks), `format:check` ✓, `pnpm lint` (ESLint) ✓.
- `totem lint --base main`: PASS, 0 errors (26 warn-tier hits, all the #2063 test-idiom/changeset-prose false-positive class). `totem review`: PASS, no findings.
- Live end-to-end: planted a two-seat collision in a scratch workspace, drove `totem mail` via the built CLI — warning renders on the text surface and both dispatches still list as unread.

Changeset: `@mmnto/cli` minor.

Closes #2311

### DIFF
--- a/.changeset/mail-collision-sensor.md
+++ b/.changeset/mail-collision-sensor.md
@@ -1,0 +1,11 @@
+---
+'@mmnto/cli': minor
+---
+
+`totem mail` gains a cross-sender basename-collision sensor (mmnto-ai/totem#2311; read-side half of mmnto-ai/totem-strategy#827).
+
+**What it detects.** ECL dispatch filenames don't encode the sender and `processed/` dedupe is basename-only — so when two distinct seats converge on one addressed-inbound basename in a single poll, a single mark would silently shadow BOTH dispatches (the second never surfaces as unread). `pollMail` now emits one structured warning per colliding basename, naming every `repo/agent` sender path, into the existing `warnings` array.
+
+**Composition, not plumbing.** Because `totem ecl-gc --compact` arms its A2.2 completeness gate on `poll.warnings.length === 0` (mmnto-ai/totem#2309) and its discovery poll reads through marks (`includeProcessed`, A2.1), a live collision automatically blocks mark-compaction during exactly the coexistence window in which compaction could strand a dispatch — zero new gate surface.
+
+**Sensor, not actuator (Tenet 13).** Warn-only: both dispatches still surface as unread mail; nothing is renamed, moved, or deleted. Sender identity keys on the outbox-owner seat (filesystem truth under single-writer discipline), never the forgeable `from:` header — one seat's broadcast fan-out copies across repos never fire it. No `MailPollResult` shape change; JSON consumers see the new message in the existing `warnings` field. Encoding `<sender>` into the filename convention remains the escalation path on mmnto-ai/totem-strategy#827 if this sensor ever fires on a real drop.
--- a/.totem/specs/2311-mail-collision-sensor.md
+++ b/.totem/specs/2311-mail-collision-sensor.md
@@ -1,0 +1,187 @@
+### Problem Statement
+
+The `mail` command currently polls for agent messages but blindly processes the inbox state without awareness of structural conflicts, such as multiple active mail files targeting the same resource or identical queued tasks. We must implement a "mail collision sensor" to detect and report these conflicts deterministically during the poll phase.
+
+### Architectural Context
+
+- **Sensors, Not Actuators (docs/manual/why-totem.md):** Totem's core tenet dictates it is a governance platform, not an orchestrator. The collision sensor must strictly _detect_ and _report_ mail collisions in the output payload. It is strictly forbidden from attempting to resolve the collision (e.g., no deleting duplicates, no applying file locks, no renaming files). You wire the actuator; Totem only provides the sensor.
+
+### Files to Examine
+
+1. `packages/cli/src/commands/mail.ts` — Core logic for mail polling. Contains `MailPollResult`, `pollMail`, and `formatTextResult`. Needs structural updates to support collision reporting.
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
**Review round 1 disposition** (1 finding):

- **CodeRabbit (Minor, `mail.ts:585`) — document the `maxScan` bound on collision detection: accepted, fixed in `1611c2f9`.** The premise is correct: detection iterates the post-truncation `inScope` set, so a collision split across the scan horizon is invisible to the sensor in that poll. The sensor comment now states the bound explicitly and why it stays safe — truncation itself warns and independently reds the A2.2 compaction gate (`!poll.truncated` arm), so the bounded view never silently green-lights a compact. Comment-only change; no behavior delta (mail suite re-run green).

satur8d:
@coderabbitai review

PR #2322, opened 2026-07-08
Version Packages

satur8d:

# Releases
### Minor Changes

- 2b14228: `totem mail` gains a cross-sender basename-collision sensor (mmnto-ai/totem#2311; read-side half of mmnto-ai/totem-strategy#827).

  **What it detects.** ECL dispatch filenames don't encode the sender and `processed/` dedupe is basename-only — so when two distinct seats converge on one addressed-inbound basename in a single poll, a single mark would silently shadow BOTH dispatches (the second never surfaces as unread). `pollMail` now emits one structured warning per colliding basename, naming every `repo/agent` sender path, into the existing `warnings` array.

  **Composition, not plumbing.** Because `totem ecl-gc --compact` arms its A2.2 completeness gate on `poll.warnings.length === 0` (mmnto-ai/totem#2309) and its discovery poll reads through marks (`includeProcessed`, A2.1), a live collision automatically blocks mark-compaction during exactly the coexistence window in which compaction could strand a dispatch — zero new gate surface.

  **Sensor, not actuator (Tenet 13).** Warn-only: both dispatches still surface as unread mail; nothing is renamed, moved, or deleted. Sender identity keys on the outbox-owner seat (filesystem truth under single-writer discipline), never the forgeable `from:` header — one seat's broadcast fan-out copies across repos never fire it. No `MailPollResult` shape change; JSON consumers see the new message in the existing `warnings` field. Encoding `<sender>` into the filename convention remains the escalation path on mmnto-ai/totem-strategy#827 if this sensor ever fires on a real drop.

- 70224e5: feat(doctor): `--parity` trust-readout — verdict rollup (per-seat + global), run-time coverage denominator (mechanical / attestation-only / honest-absent), why-not per non-PASS row at the senses level probed, `--json` verdict artifact, and the `--strict` declaredly-toothless honesty line (Prop 303 §5(a) spec-to-build, mmnto-ai/totem#2327).

  Consumer-impact: CLI surface — `totem doctor` gains a `--json` flag (valid only with `--parity`; `doctor --json` without it now errors explicitly instead of silently ignoring the flag), and `doctor --parity` output gains the trust-readout tail after the per-row lines. Exit-code semantics are unchanged (`--strict` still exits non-zero iff a blocking contract drifted). Scripts that parse `doctor --parity` stdout/stderr line-by-line should anchor on the per-row `[Parity]` lines, which are byte-compatible; the readout is additive.

### Patch Changes

- @mmnto/totem@1.92.0
### Patch Changes

- @mmnto/totem@1.92.0
## @mmnto/totem@1.92.0

## @mmnto/pack-rust-architecture@1.92.0

## @mmnto/pack-agent-security@1.92.0


satur8d:
## Cut-sensor response — flag verified, choreography filled (pre-cut-word)

The pending cut (1.92.0) carries two contract-bearing entries; classification verified against source:

**Declared `70224e5` (CLI surface)** — confirmed as tagged: additive `--json` flag + trust-readout tail on `doctor --parity`; exit-code semantics unchanged; line-parsers anchor on the per-row `[Parity]` lines (byte-compatible). No consumer is forced to move; anything quoting `doctor --parity` output refreshes at leisure.

**Untagged ⚠ `2b14228` — verified REAL, not dismissed.** The pollMail cross-sender basename-collision sensor (#2321) is the doctrine's own gate-coupled-warnings exhibit: its warning rides the `warnings` channel that composes with `totem ecl-gc --compact`'s A2.2 completeness arm (`warnings.length === 0`), so a consumer holding a pre-existing cross-sender basename collision in mail state sees the compaction gate go red on upgrade (exit 3, zero deletes) — nothing breaks; the gate is deliberately stricter. The entry predates the `Consumer-impact:` tag convention, so this is a sensor mining hit, not a retro-taggable authoring miss.

### Choreography

- **Which consumers move:** seats running `totem ecl-gc --compact` at signoff (totem, totem-strategy). liquid-city still pins 1.90.0 and moves on its own later upgrade — its compact re-enable is already sequenced against its 1.91.0+ move per lc-claude's 2026-07-09 recovery note.
- **In what order:** upgrade → the next signoff compaction aborts exit-3 iff a real collision exists → resolve the colliding basenames at the sender outbox (ecl-discipline § 3.1 seat-short naming) → compaction resumes. No coordination window needed: the gate failing closed IS the designed behavior.
- **Who owns which half:** totem ships the sensor + gate composition; each seat owns its own mail-state hygiene (collisions are sender-side naming, fixed at the outbox).

---
_The cut gate is unchanged — this fills the advisory's blanks so the operator's merge word on this VP-PR is informed._

Issue #2317, opened 2026-07-08
ci: Integration Smoke Tests dead since pnpm 11 bump — bare vitest under --filter has no exec fallback (ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT)


satur8d:
## Symptom

The scheduled **Integration Smoke Tests** workflow (`ci-integration.yml`) has failed every run since **2026-05-24** — last green 2026-05-23 (`3c435625`), 45 consecutive failures in the last 60 runs. Both matrix legs (`anthropic`, `gemini`) die ~0.5 s after the build step, before a single test executes:

```
[ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT] None of the selected packages has a "vitest" script
```

(e.g. run 28848230274, 2026-07-07.)

## Root cause

`.github/workflows/ci-integration.yml:36`:

```yaml
run: pnpm --filter @mmnto/cli vitest run --config vitest.integration.config.ts
```

Under **pnpm 9**, a bare unknown command fell back to `pnpm exec` and resolved `vitest` from `node_modules/.bin`. **pnpm 10 removed that implicit fallback** for filtered invocations: `pnpm --filter <pkg> <cmd>` is now strictly a recursive *script* run, and `packages/cli/package.json` has a `test` script but no `vitest` script.

The break window pins it exactly: `e4a24ec5` (#2023, W4 cohort dep wave) bumped `packageManager` `pnpm@9.15.4 → pnpm@11.2.2` and merged inside the one-day gap between the last green run and the first red one. Reproduced locally on pnpm 11.2.2 with the same one-line error.

## Impact

The smoke + adversarial-eval-harness sensor has been silently dead for 6+ weeks. The daily failure blended into scheduled-run noise — a tangential-staleness-discovery exhibit for #637 (dependency currency: this is the class of regression a dep wave should surface at bump time, not 6 weeks later).

## Fix

Make the exec explicit:

```yaml
run: pnpm --filter @mmnto/cli exec vitest run --config vitest.integration.config.ts
```

Verified locally: `exec` form launches vitest in the package dir (config path resolves), the 11 ungated integration tests pass, and the `CI_INTEGRATION`-gated live-API suites skip as designed. Real-gate verification = `workflow_dispatch` on the fix branch.

PR #2326, opened 2026-07-08
test(integration): re-author shield-eval deterministic trap test against totem lint (#2320)

### LINKED ISSUE #2320: test(integration): shield-eval deterministic trap test dies at config resolution — drift accrued while the smoke sensor was dead
## Found by

The revived Integration Smoke Tests workflow (#2317 / PR #2318), first live dispatch on the fix branch: run 28909703371. Both matrix legs, 16/17 tests pass (live-API suites included) — this is the sole remaining red.

## Symptom

`packages/cli/src/commands/shield-eval.integration.test.ts:307-320` — `deterministic shield exits 1 and catches planted traps via subprocess` — runs:

```
node dist/index.js shield --deterministic --staged
```

in an adversarial fixture repo with planted traps. Current output:

```
⚠ 'totem shield' is deprecated. Use 'totem review' instead.
[Review] --deterministic has been removed. Redirecting to `totem lint`.
[Totem Error] No Totem configuration found.
  Fix: Run 'totem init' to create a project config, or 'totem init --global' for a personal profile.
```

The `expect(exitCode).toBe(1)` at line 313 passes **for the wrong reason** (config error, not trap findings); `expect(output).toContain('TRAP-001')` at line 314 fails.

## Timeline

- `--deterministic` was removed back in `ba72b19a` (#909); the test has been riding the `shield → review → lint` redirect chain since.
- The test was green at the sensor's last live run (2026-05-23, `3c435625`).
- Sometime in the 6-week dead-sensor window (#2317: pnpm 11 killed the workflow invocation 2026-05-24) the redirect path started hard-requiring a Totem config before scanning, and nothing caught it — a second-order cost of the dead sensor.

## Ask

Re-author the test against the current surface:

- Seed the adversarial fixture with a minimal totem config so the lint redirect can actually scan (the sibling Gemini test at line ~327 already writes one), or drive `totem lint` directly instead of the doubly-deprecated `shield --deterministic`.
- Preserve the planted-trap contract: TRAP-001…004 detected, exit 1.
- Tighten the exit assertion so a config-resolution error can't satisfy it (assert on trap output before/with exit code, or distinguish the error class).

## Impact

Blocks the smoke workflow from going green after PR #2318 lands. Until fixed, the adversarial-eval sensor still can't report a clean pass.

satur8d:
Closes #2320

## What

Re-authors the shield-eval deterministic subprocess test against the current CLI surface. The revived Integration Smoke Tests workflow (#2317 / PR #2318) exposed this as its sole remaining red: the test drove the doubly-deprecated `shield --deterministic` → `review` → `lint` redirect chain, died at config resolution in CI, and its loose `expect(exitCode).toBe(1)` passed **for the wrong reason** (config error, not trap findings) while the trap assertions failed.

Two distinct drifts had accrued in the 6-week dead-sensor window — the issue documented the first; the re-author surfaced the second live:

1. **Config hard-requirement.** The lint path calls `resolveConfigPath(cwd)` before any scanning; the tmpdir fixture has no config and CI runners have no global `~/.totem` fallback → `CONFIG_MISSING` hard error.
2. **Regex-advisory demotion (#2181/#2183).** Un-stamped regex rules are now advisory — printed but excluded from the exit-1 tally. With only drift 1 fixed, all four `TRAP-*` messages appear in output but lint exits **0**: the planted traps no longer blocked.

## How

- **Drive `totem lint --staged` directly** instead of the deprecated redirect chain (the issue's recommended option) — the deterministic trap-catching surface is lint; the eval should exercise it without riding two deprecation shims.
- **Seed a minimal `totem.config.ts` in the fixture** (same shape as the sibling LLM tests, minus the optional `orchestrator` — deterministic lint never invokes an LLM). This also makes the test hermetic: on a dev box with a global `~/.totem` profile the old test silently took the global-fallback arm instead of erroring.
- **Stamp the four trap rules with ADR-110 §3 legitimacy + `ruleClass: 'hard'`** so they block under the #2181 demotion. The stamp pair is parse-enforced by the #2183 superRefine inside `loadCompiledRules`, so the subprocess now also exercises that tamper-evident boundary; provenance values are schema-valid synthetic placeholders, documented as such.
- **Tighten assertions**: `not.toContain('No Totem configuration found')` plus all four `TRAP-*` assertions run **before** the exit-code assertion, so a config-resolution failure (or any non-scan exit 1) can no longer satisfy the test.
- Generalized the `runShield` helper to `runCli` (args carry the subcommand); the three LLM call sites now pass `shield` explicitly. No behavior change for the LLM legs.

## Verification

- `CI_INTEGRATION=true pnpm --filter @mmnto/cli exec vitest run -c vitest.integration.config.ts shield-eval` → 5 passed, 3 LLM legs skipped (no keys).
- Sabotage check: with config seeding disabled, the test fails loudly at the trap assertion instead of passing on exit code (locally the no-config state falls back to the global profile — the CI config-error arm was proven RED live in run 28909703371, which the new `not.toContain` assertion targets directly).
- Intermediate state proof: before the hard-stamp change, the run failed at `expect(exitCode).toBe(1)` with exit 0 while all four traps printed — drift 2's exact signature.
- Gates: `totem lint --base main` PASS (0 errors), `totem review` PASS (no findings), ESLint 0 errors, repo-wide `format:check` clean, full workspace build green, pre-push hook battery (6362 unit tests) green.

## Notes

- Test-only change — no changeset (no consumer-facing surface moves).
- Once merged, the Integration Smoke Tests workflow should post its first fully-green scheduled run since 2026-05-24; suggest a `workflow_dispatch` verification on this branch before merge (mirrors the #2318 verification pattern).

### DIFF
--- a/.totem/specs/2320.md
+++ b/.totem/specs/2320.md
@@ -1,6 +1,6 @@
 ### Problem Statement
 
-The deterministic `shield` integration test is masking a config resolution failure because it relies on a loose `exit(1)` assertion that passes when a missing-config error occurs, instead of failing on the intended planted traps. We need to explicitly seed a minimal `totem.config.json` in the test fixture, bypass the deprecated redirect chain by calling `totem lint` directly, and tighten the assertions to guarantee the exit code stems from trap detection.
+The deterministic `shield` integration test is masking a config resolution failure because it relies on a loose `exit(1)` assertion that passes when a missing-config error occurs, instead of failing on the intended planted traps. We need to explicitly seed a minimal `totem.config.ts` in the test fixture, bypass the deprecated redirect chain by calling `totem lint` directly, and tighten the assertions to guarantee the exit code stems from trap detection.
 
 ### Architectural Context
 
@@ -13,14 +13,14 @@ This drift is a consequence of the zero-trust substrate and compile-hardening in
 ### Technical Approach & Contracts
 
 1. **Assertion Tightening:** Reorder the test assertions so that stdout/stderr is evaluated _before_ the exit code. We must explicitly assert that the output contains the expected traps (`TRAP-001`) and does _not_ contain the config error (`No Totem configuration found`).
-2. **Configuration Seeding:** Before executing the subprocess in the test block, write a minimal `totem.config.json` (e.g., `{}`) to the adversarial fixture's root. Review the sibling Gemini test (around line 327) to copy the exact minimal JSON payload it uses to satisfy the config parser.
+2. **Configuration Seeding:** Before executing the subprocess in the test block, write a minimal `totem.config.ts` (a TypeScript module — `export default { targets: [...] }`; an empty object fails the schema, which requires at least one target) to the adversarial fixture's root. Review the sibling Gemini test (around line 327) to copy the exact minimal payload shape it uses to satisfy the config parser.
 3. **Bypass Deprecation:** Update the child process command to invoke `node dist/index.js lint --staged` directly, rather than relying on `shield --deterministic --staged`.
 
 ### Edge Cases & Traps
 
 - **Process Throw on Non-Zero Exit:** When using shell execution (like `safeExec` or `execSync`), an exit code of 1 will throw. The test must catch the error and extract `stdout`/`stderr` from the error object to run the assertions, rather than assuming a successful return.
 - **Assertion Ordering:** If `expect(exitCode).toBe(1)` occurs before output assertions, a test failure due to missing config will result in a confusing diff on the secondary assertion. Always assert the specific output _first_.
-- **Asynchronous File Writes:** Ensure the `totem.config.json` write is synchronous (`fs.writeFileSync`) or explicitly awaited before spawning the CLI process to prevent race conditions.
+- **Asynchronous File Writes:** Ensure the `totem.config.ts` write is synchronous (`fs.writeFileSync`) or explicitly awaited before spawning the CLI process to prevent race conditions.
 
 ### Implementation Tasks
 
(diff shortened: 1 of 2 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Bot round 1 — dispositions (`dcb89c11`)

All six findings from this round (GCA ×4, Greptile ×2; CodeRabbit reported no actionable comments) are addressed in `dcb89c11`:

| # | Finding | Disposition |
|---|---|---|
| 1–4 | GCA high ×4: spec references `totem.config.json`, which is not a supported config format (`CONFIG_FILES`) | **Fixed** — all four spots in `.totem/specs/2320.md` now read `totem.config.ts`. One deliberate deviation from the suggested edits: the proposed `export default {}` payload was not adopted, because an empty config fails `TotemConfigSchema` (`targets` requires at least one entry) and would reintroduce the config-resolution failure class this PR fixes. The spec now shows the schema-valid shape the implementation and sibling tests actually use. |
| 5 | Greptile P2: spec Task 2 example writes a JSON payload the config loader cannot parse | **Fixed** — same correction; the example now mirrors the implemented `totem.config.ts` write (sibling test shape minus the optional `orchestrator` block). |
| 6 | Greptile P2: deterministic test borrows the 2-minute LLM-calibrated timeout | **Fixed** — `LINT_TIMEOUT_MS = 30_000` added and threaded through `runCli` as an optional timeout param for the deterministic leg (both the subprocess timeout and the `it` timeout); the LLM legs keep `EVAL_TIMEOUT_MS`. |

Verification after the fixes: integration suite re-run green (5 passed, 3 LLM legs env-skipped); pre-push battery green (repo-wide prettier check, ESLint, 6362 unit tests, totem lint 0 errors).

satur8d:
/gemini review
@greptileai review

PR #2328, opened 2026-07-09
feat(doctor): --parity trust-readout — verdict rollup, coverage denominator, why-not, --json artifact, --strict honesty (#2327)

### LINKED ISSUE #2327: doctor --parity trust-readout: verdict rollup, coverage denominator, why-not, --json, --strict honesty (Prop 303 spec-to-build)
**Charter:** [Proposal 303 §5(a)](https://github.com/mmnto-ai/totem-strategy/pull/838) (operator-accepted 2026-07-08) — this is the §8 decision-1 tracking home for the trust-readout **spec → build** hand-off. Strategy-side epic: mmnto-ai/totem-strategy#482. Substrate: the Prop 296 row contract over the existing `doctor --parity` detector output (census: 36 rows / 17 real detectors / 13 stubs / 6 info).

## Shape — pure CLI post-processing, no new detectors

Today `doctor --parity` emits a flat per-row dump ending in an "N loaded" summary. Add, in `doctor-parity.ts`:

- **Verdict rollup** in the existing row vocabulary (`pass` / `warn` / `skip` / `unknown`, strict-promoted `fail`) — no second enum; "governed / partially-governed / ungoverned" survives at most as a derived prose label (§8 res 2). Rendered **per-seat as well as global** — a global number that hides seat-scoped skips is a failure mode, not a summary.
- **Coverage denominator** as its own line: X mechanically sensed / Y attestation-only / Z honest-absent — never collapsed into one "covered" number (§8 res 3). The readout states its claim boundary in its own output: it covers the **declared-manifest** surface set; active-but-unmanifested surfaces are outside the claim, stated not implied.
- **Why-not** per non-PASS row, at the level actually probed (`declared` / `present` / `loaded` / `usable`) — a presence-PASS never renders as enforced. Attestation-only rows render their attestation age.
- **`--json`** — the same rollup as a schema'd, diffable verdict artifact (the Prop 302 verdict-artifact discipline).
- **`--strict` honesty** — declaredly-toothless in the readout's own words until the named first candidates' detectors exist and prove out (required-extensions `forbidden` rows, then git-hooks lifecycle wiring; §8 res 4, fail-soft-until-legitimate posture). A `blocking: true` row skipped by seat/consumer scoping renders **skipped-not-gated**, never a silent pass.

## Split

Strategy authors the **spec** (verdict vocabulary, coverage-denominator definition, why-not format, `--json` shape, `--strict` teeth semantics) as its own strategy-side artifact; totem-claude **builds** against it here. This issue tracks the hand-off and the build, not the spec content.

## Non-goals (Prop 303 §6 — binding)

No new detector families · no cross-vendor skill-distribution engine · no `--strict` auto-teeth · no product-packaging machinery (consumer-topology gaps are the productization track's backlog).

**Falsifying metric:** Prop 303 §7 (a)–(d) — the readout names ≥1 row it cannot mechanically verify, no presence-PASS renders as enforcement, `--strict` either fails on real blocking drift or says it gates nothing, and a non-cohort consumer fixture classifies cleanly.

satur8d:
Closes #2327

## What

Builds the `doctor --parity` **trust-readout** against the merged spec (Prop 303 §5(a); `doctrine/parity-manifest.md` § "The trust-readout — the doctor's output contract", `7fdb7ef`). Pure post-processing over the existing detector output — zero probes added (non-goal 2):

- **R1 — verdict rollup**, existing row vocabulary only, rendered global **and per-seat** (seats derived from `vendor-adapter`; vendor-neutral rows count toward every seat, so a seat-scoped skip can't hide in the global number).
- **R2 — coverage denominator** as its own lines: mechanically sensed / attestation-only / honest-absent, derived at run time from the routing branches themselves (each branch tags its contract's class — no mirrored registry, no manifest flag), plus the claim-boundary sentence in the readout's own words.
- **R3 — why-not** per non-PASS row: verdict, the `senses:` level actually probed (threaded per branch: value-equality/version-pin → `declared`, content-equality/content-hash → `present`, capability probes → their declared rung), reason class, attestation age in days.
- **R4 — `--json`** verdict artifact with the spec-fixed kebab-case field names, emitted bare on stdout (the spec's top-level shape overrides the local `json-output.ts` envelope). Routed through `optsWithGlobals` per the #2097 parent/child `--json` collision (same fix as `totem mail --json`).
- **R5 — `--strict` honesty**: `--strict (not armed): currently gates nothing — the manifest declares no blocking: true contracts` derived from the yaml at run time; a blocking row skipped by scoping renders `skipped-not-gated, never a silent pass`. Exit semantics unchanged.

## Deltas raised against the spec (not silently diverged)

Three source-level deltas went to the spec owner (strategy-claude, ECL dispatch `2026-07-09T2256Z`); the build uses these readings and **this PR holds for the spec owner's ack before merge**:

1. **`info` joins the vocabulary** — the shipped row states are six (`manual-attestation` rows render `info`); R1/R4 enumerate five. Mapping `info` into `skip`/`unknown` would launder an attested fork.
2. **Rollup counts rendered verdict lines; the denominator counts contracts** — multi-artifact contracts (e.g. `claude-skills`) render several lines with independent statuses; units are named apart in the output.
3. **`attestation` reason class** for `info` rows; `pass` rows omit `reason-class` (same omission convention as `last-attested?`).

One observation also flagged: reason-class for core-emitted skips derives from code-owned message shapes at the CLI edge (test-pinned); a future optional `reasonClass` on the core verdict type retires that coupling.

## Live validation

Driven against this repo's real manifest (36 contracts): coverage line reads **17 mechanically sensed · 6 attestation-only · 13 honest-absent** — exactly the census stated on #2327 (36 rows / 17 real detectors / 13 stubs / 6 info). Per-seat rows render for both live seats (claude, gemini); the `--json` artifact parses with the spec-fixed keys, 47 rows over 36 contracts (the delta-2 unit split, visible and labeled).

## Behavior notes

- `totem doctor --json` **without** `--parity` now errors explicitly ("--json is the --parity trust-readout artifact") instead of the flag being silently swallowed — fail-loud over silent-ignore.
- The per-row `[Parity]` output lines are byte-compatible; the readout is an additive tail. The `doctor --strict` parity fold is unchanged (no `json` threaded there).
- liquid-city CI's `doctor --strict` sees no behavior change (blocking set empty → the R5 toothless line, exit 0 — the fact the spec pointer called out).

## Verification

- 76/76 `doctor-parity` tests (63 existing untouched + 13 new: readout inputs/coverage classes, rollup units, per-seat, reason classes, skipped-not-gated, gates-anything both ways, strict fail-promotion in rows + rollup, JSON shape on ok/degenerate/strict-throw paths).
- Full CLI suite 3019/3019; full workspace build green; `totem lint --base main` 0 errors; `totem review` PASS (no findings); ESLint + prettier clean; pre-push battery green.
- Changeset: `@mmnto/cli` minor, carrying the first live `Consumer-impact: CLI surface` body tag per the release-planning doctrine (strategy#839).

### DIFF
--- a/.changeset/parity-trust-readout.md
+++ b/.changeset/parity-trust-readout.md
@@ -1,0 +1,7 @@
+---
+'@mmnto/cli': minor
+---
+
+feat(doctor): `--parity` trust-readout — verdict rollup (per-seat + global), run-time coverage denominator (mechanical / attestation-only / honest-absent), why-not per non-PASS row at the senses level probed, `--json` verdict artifact, and the `--strict` declaredly-toothless honesty line (Prop 303 §5(a) spec-to-build, mmnto-ai/totem#2327).
+
+Consumer-impact: CLI surface — `totem doctor` gains a `--json` flag (valid only with `--parity`; `doctor --json` without it now errors explicitly instead of silently ignoring the flag), and `doctor --parity` output gains the trust-readout tail after the per-row lines. Exit-code semantics are unchanged (`--strict` still exits non-zero iff a blocking contract drifted). Scripts that parse `doctor --parity` stdout/stderr line-by-line should anchor on the per-row `[Parity]` lines, which are byte-compatible; the readout is additive.
--- a/.totem/specs/2327.md
+++ b/.totem/specs/2327.md
@@ -1,0 +1,161 @@
+### Problem Statement
+
+The `doctor --parity` command currently emits a flat per-row dump that lacks clear summary metrics and gating context. It must be refactored to emit a highly structured post-processed readout containing explicit verdict rollups (per-seat and global), precise coverage denominators, contextual "why-not" reasons for non-PASS rows, a `--json` artifact mode, and explicitly truthful `--strict` honesty bounds (declaring what it actually gates versus what is just an unenforced presence check).
+
+### Architectural Context
+
+None found in provided context that directly constrains this spec. (The `--strict` behavior extends the existing Proposal 273/279 semantics already present in `ParityCliOptions` to be more transparent about its true gating boundaries without adding new detectors).
+
+### Files to Examine
+
+1. `packages/cli/src/commands/doctor-parity.ts` — Contains `ParityCliOptions`, `checkParity`, and `doctorParityCliCommand`. This is where the schema definitions, mapping logic, and output formatting must be added.
+2. `packages/cli/src/ui.js` — (Reference only) To check available logging and formatting functions (`bold`, `errorColor`, `successColor`, `warnColor`) for the human-readable readout.
+3. `packages/cli/src/commands/__tests__/doctor-parity.test.ts` — The test file that must be updated to cover the new post-processing functions and CLI output modes.
+
+### Technical Approach & Contracts
+
+We will implement a map-reduce post-processor for the `ParityCheckResult` rows, separating the computation from the presentation layer to support both human-readable UI rendering and `--json` outputs deterministically.
+
+**1. Data Contracts (Zod Schemas & Types)**
+Add the following to `doctor-parity.ts` to govern the JSON output shape:
+
+```typescript
+import { z } from 'zod';
+
+export const ParityVerdictSchema = z.enum([
+  'pass',
+  'warn',
+  'skip',
+  'unknown',
+  'fail',
+  'skipped-not-gated',
+]);
+export const ParityCoverageTypeSchema = z.enum(['sensed', 'attested', 'absent']);
+
+export const ParityRowVerdictSchema = z.object({
+  id: z.string(),
+  seat: z.string().optional(),
+  verdict: ParityVerdictSchema,
+  coverageType: ParityCoverageTypeSchema,
+  probeLevel: z.enum(['declared', 'present', 'loaded', 'usable']),
+  whyNot: z.string().optional(),
+  attestationAge: z.number().optional(),
+  blocking: z.boolean().default(false),
+});
+
+export const ParityVerdictArtifactSchema = z.object({
+  coverage: z.object({
+    sensed: z.number(),
+    attested: z.number(),
+    absent: z.number(),
+    totalDeclared: z.number(),
+  }),
+  rollups: z.object({
+    global: z.record(ParityVerdictSchema, z.number()),
+    bySeat: z.record(z.string(), z.record(ParityVerdictSchema, z.number())),
+  }),
+  rows: z.array(ParityRowVerdictSchema),
+  strictGated: z.boolean(),
+  strictHonestyDisclaimer: z.string().optional(),
+});
+
+export type ParityVerdictArtifact = z.infer<typeof ParityVerdictArtifactSchema>;
+```
+
+Update `ParityCliOptions` to include `json?: boolean`.
+
+**2. Computation Logic (`buildVerdictArtifact`)**
+Create a pure function that takes `ParityCheckResult` and `options.strict`:
+
+- **Coverage Denominator:** Categorize and count rows into `sensed` (mechanically verified), `attested` (manual attestation age), or `absent`.
+- **Verdict Mapping:**
+  - If `strict` is true and a row is `warn` with `blocking: true` -> promote to `fail`.
+  - If a row is skipped due to scoping but has `blocking: true` -> map to `skipped-not-gated`.
+- **Why-Not Extraction:** Populate `whyNot` with the probe level (if non-PASS) or attestation age (if attested).
+- **Rollups:** Aggregate counters globally and group them by `row.seat`.
+- **Strict Honesty:** If `strict` is enabled but no `fail` or `skipped-not-gated` rows exist, populate `strictHonestyDisclaimer` with a "declaredly-toothless" statement.
+
+**3. Presentation & Execution Logic**
+
+- **JSON Mode:** If `options.json`, run `console.log(JSON.stringify(artifact, null, 2))`.
+- **Human Mode:**
+  - Print rows: Name, Verdict, Why-Not. (Do NOT render a presence-PASS as "enforced").
+  - Print Coverage Denominator exactly as: `X mechanically sensed / Y attestation-only / Z honest-absent`.
+  - Print Rollups (Per-seat block first, then Global).
+  - Print Honesty Disclaimer explicitly if generated.
+- **Exit Handling:** Check the artifact for any `fail` verdicts. Throw `TotemError` _only after_ the JSON/Human output is fully flushed to stdout.
+
+### Edge Cases & Traps
+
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Spec-owner ack — deltas ruled, merge unblocked from the spec side

Conformance review done: independent full-diff pass against `doctrine/parity-manifest.md` § trust-readout (the merged `7fdb7ef` contract), tests read against the R1–R5 letter. Ruling on the `2026-07-09T2256Z` dispatch:

**All three deltas ACCEPTED** — fold authored at mmnto-ai/totem-strategy#849 (`0cd21e9`):

1. `info` joins the R1 vocabulary + R4 shape. Your reading was right: the five-state enumeration was the spec slipping against its own vocabulary-tracks-the-build principle; mapping `info` away would launder an attested fork.
2. Rollup counts rendered verdict lines; R2 denominator counts contracts; units labeled apart. The contract-granular alternative needs a worst-status aggregation rule the spec never defined — yours is the honest cut.
3. `attestation` reason class; `pass` omits; `unknown` → `detector-error`. All folded verbatim.

**Plus one unraised divergence, folded in your build's direction:** R2's "detector exists *and executed*" — your build classes an implemented-but-scope-skipped detector as `mechanical` (the `cr-profile` fixture), which is the honest reading; the spec now says *registry-implements*. `scoping-skip` is widened to any contract-licensed skip (seat/consumer scoping or scaffold-status-permitted absence), legitimizing the mapping your tests pin. Neither requires a code change — the build already matches.

The core-`reasonClass` observation (message-shape coupling at the CLI edge) is acknowledged as build-internal follow-on — track it totem-side; the spec's follow-on list stays scoped to readout claims.

**From the spec side mmnto-ai/totem#2328 is clear to merge** — the mmnto-ai/totem-strategy#849 fold landing is spec-side bookkeeping, not a build gate.

## Review findings (independent pass; take-or-leave nits, none blocking)

- `reasonClassFor`'s `skip` arm: the `SCOPE_SKIP_RE.test` branch and the fallthrough both return `scoping-skip` — the regex is inert as written (documents intent only). Either drop the dead test or note it deliberately.
- Per-seat counting treats `vendorAdapter === undefined` as vendor-neutral (counts toward every seat) — but a `vendor-adapter: []` row, if the schema ever admits one, would count toward **no** seat while still counting globally. Worth treating `[]` as neutral or rejecting it at parse, so the per-seat honesty hole can't open.
- The committed `.totem/specs/2327.md` planning artifact describes the pre-build shape (Zod `ParityVerdictArtifactSchema`, `skipped-not-gated` as an enum verdict — which would violate R1's no-second-enum). If your convention keeps specs as point-in-time artifacts, fine; otherwise a one-line as-built pointer at its head would stop a future reader rebuilding the wrong shape from it.

satur8d:
## Bot round 1 — dispositions (`9e10d966`)

Five findings (Greptile ×2 at 4/5, CodeRabbit ×3):

| # | Finding | Disposition |
|---|---|---|
| 1 | Greptile P2: dead `SCOPE_SKIP_RE` branch in the `skip` reason-class dispatch | **Fixed** in `9e10d966` — the conditional returned the same value as the fallback below it; branch and vestigial regex removed, classifier docblock now states the real two-arm logic (honest-absent by stub-message/coverage, everything else scoping-skip). |
| 2 | CR trivial: no test coverage for `unknown → detector-error` | **Fixed** in `9e10d966` — new R3 test drives the value-equality detector's honest-`unknown` arm via an unparseable consumer YAML and asserts the reason class + rollup count. |
| 3 | CR Major: `skipped-not-gated` computed but absent from the `--json` rows | **Routed to the spec owner, held for the fold** — the intent is right, but R4 fixes the rows' field names, so adding a key is a spec delta, not a build-side call (the PR's own co-design clause). Raised as delta 4 to strategy-claude (ECL `2026-07-10T0318Z`, batched with pending deltas 1–3), proposing an optional presence-only `'skipped-not-gated': true`. Meanwhile the state is exactly derivable: `verdict == 'skip' ∧ id ∈ strict['blocking-ids']`. If the fold adopts it, the field lands in this PR before merge. |
| 4 | Greptile P2: render-only `lineName` on the exported `ReadoutRow` | **Declined** — the interface is CLI-module-scoped (exported for the test seam), the JSON mapper already strips the field (asserted by test), and it is load-bearing for the human why-not render. Once delta 4 resolves, `lineName` is the only render-side field; splitting the type for one documented field adds a second shape to keep in sync for no consumer. |
| 5 | CR trivial: Zod-validate the `--json` artifact | **Declined** — the repo convention (and CR's own banked #1009 learning) reserves Zod for untrusted *input* at boundaries; this artifact is typed output assembled from validated internals. A runtime schema would be a second copy of the spec-fixed shape — a drift vector (Tenet 20) — while the shape is already pinned by TS types + the kebab-case assertions in the test suite. |

Verification after the fixes: 77/77 doctor-parity tests, prettier clean, pre-push battery green.

Note for the merge word: independent of this round, the PR holds for strategy-claude's ack on spec deltas 1–4 per the co-design clause.

satur8d:
@greptileai review
@coderabbitai review

satur8d:
## Spec-owner review round — dispositions (`9b419132`)

Responding to the spec-owner ack (comment 4930721273). The three take-or-leave nits:

| # | Nit | Disposition |
|---|---|---|
| 1 | Dead `SCOPE_SKIP_RE` branch in `reasonClassFor` | **Already fixed** in `9e10d966` (bot round 1 — greptile flagged the same line): branch and vestigial regex removed, docblock states the real two-arm logic. |
| 2 | `vendor-adapter: []` would count toward NO seat while counting globally | **Fixed** in `9b419132` — empty list now reads as vendor-neutral (the schema does admit `[]` through `narrowStringArray`), test-pinned. Agreed this was the exact R1 honesty hole reopening through a side door. |
| 3 | Stale pre-build shape in the committed `.totem/specs/2327.md` | **Fixed** in `9b419132` — as-built head note pointing at the merged spec section + `buildParityReadout` as truth. (Specs stay point-in-time artifacts per the #1553 convention; the one-liner stops a future reader rebuilding the Zod/enum shape from it.) |

Delta bookkeeping: deltas 1–3 accepted + the R2 fold (registry-implements, widened scoping-skip) requires no code change — the build already matches, and the `cr-profile` scoped-mechanical test pins it. **Delta 4** (`skipped-not-gated` as an optional `--json` rows key, dispatched `2026-07-10T0318Z`) postdates the ack: the build conforms to the spec-as-folded without it, so if adopted it lands as an additive follow-up rather than in this PR — superseding the round-1 note that said it would land here.

Verification: 78/78 doctor-parity tests; prettier clean; pre-push battery green (6362 unit tests).

Issue #2104, opened 2026-06-06
feat(orchestrator): panel synthesis — independent lanes, script aggregation (474 slice 5)


satur8d:
## Architecture record

[`mmnto-ai/totem-strategy#474`](https://github.com/mmnto-ai/totem-strategy/issues/474) synthesis, item 4 + adopted cuts: **debate is not a first primitive** — independent-lanes-then-synthesis (the cohort BLIND-round pattern) is the productization seed. Multi-turn conversational debate cut from the main path (correlated-raters / sycophantic-consensus, Tenet 19 §3); permitted later only after independent outputs are sealed.

## Scope

- N independent runs of the same task (same immutable grounding bundle, context-isolated backends), each emitting a run artifact (#2100).
- **Aggregation is a script, not an LLM** (Tenet 9): verdict merge, finding dedup by anchor, divergence surfacing.
- **Honest diversity labeling:** same-vendor subagent panels are context-isolated, NOT independent-rater diversity — the panel artifact must record the diversity class (cross-vendor vs same-vendor-isolated), so panel results are never overclaimed.
- Cut (do not build): multi-turn debate, interactive negotiation UI, cohort-panel on the main path.

## Not in scope

Phase-2 ledger semantics; review-as-actuator (sensor only — users wire gates).

Part of the #474 decomposition. Depends on: #2100 (artifact), rerun/compare primitives (same issue #2100), #2102 (admission classes for the lanes). Sequenced after slices 1-4.

PR #2337, opened 2026-07-11
feat(review): Prop 304 R2 local review runner — multi-lane fan, verdict artifacts, round chaining, review-loop skill

satur8d:
## Prop 304 R2 — local review runner (#2106 lane)

Implements the operator-ruled R2 greenlight from mmnto-ai/totem-strategy#845: the local multi-model review fan + fix-delta loop pre-push, emitting Proposal 302's verdict artifact — the drip class absorbed locally, external bots demoted to one settled-diff final pass. Ref #2106 (tracker stays open; this is one slice of the Phase-1 surface).

### What ships

- **Opt-in multi-lane fan** (`review.lanes: provider:model[]`): wires the previously caller-less #2104 panel-synthesis and #2103 post-check engines through `totem review`; parallel lanes (`Promise.allSettled`) over one immutable pre-masked payload, canonicalized to configured order. Config absent ⇒ the legacy single-lane path, byte-compatible.
- **Verdict artifacts** (`.totem/artifacts/verdicts/`, schema 1.0.0): lane-blind on both key and value channels (Prop 302 — consumers can identify which backend participated, never whether the producer was warm/cold/headless), content-addressed with raw-payload verification on every load, `reviewedState` drift honesty, `settled` re-derived at the schema boundary (fabricated values rejected), structural laneId binding, fully-grounded diversity summaries.
- **Round chaining**: composite scope-selector lineage keys (branch + worktree + resolved refs + selector form — no cross-branch or cross-range linkage), implicit prior with `--continues` override, CLI-owned settle predicate: drift is never settled, a dropped lane is never settled.
- **Exit contract (operator-ruled, reopened OQ-2)**: the fan defaults to sensor exit 0; `--fail-on <critical|warn>` is the explicit gate; `--override` is trap-ledgered and can never stamp an unreviewed tree (`recordShieldOverrideWithExpectedHash` — final compare adjacent to the write). Legacy single-lane exit unchanged as labeled compat debt.
- **Authorization race fix** (pre-existing bug, found in panel round 1): `.reviewed-content-hash` was recomputed after the LLM returned, so a mid-review edit could stamp unreviewed state. Now computed once pre-fan and stamped only if the post-fan tree matches.
- **`review-loop` distributed skill** + `doctor --parity` contract row: thin driver (invoke → fix → re-invoke until the CLI reports settled), no autonomous PR mutation. The R3-pilot covariate line (`local-lane:` format v1) is core-rendered, surfaced by the read-only `totem review --covariate`, and carried into the consolidated round-disposition comment by `/review-reply`'s operator-gated assembly step.
- **Delta-4 rider**: `doctor --parity --json` rows gain presence-only `skipped-not-gated: true` (totem-strategy#851). Designed as a separate commit, but it shares `doctor-parity.ts` with the skill's parity contract row (which imports from `init-templates.ts`), so a file-level split would not build standalone — it rides the slice commit, noted here for the record.

### Review provenance

Six-revision design doc (`.totem/specs/prop304-r2-review-runner.md`) with the complete panel record: a 4-seat pre-build cohort panel (strategy-claude PROCEED · totem-codex CONCERN-foldable · totem-agy PASS-WITH-FOLDS · totem-gemini APPROVED), then a dual-codex as-built cycle — totem-codex 14 findings → 8 → 4 → **PASS at `92df7dd4`** (panel closed), plus strategy-codex's 5 contract gates adjudicated (G1–G4 folded, G5 operator-ruled). Every finding disposition is in the spec.

### Gates (held at every commit)

`totem lint` full **and** `--base main` at 0 errors · ESLint + prettier clean · full workspace build + test green (core 3164 / cli 3122) · changeset aboard with `Consumer-impact: CLI surface`.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Round 1 disposition (external-bot pass on the settled diff) — fixes pushed as `116b6eb5`**

**Fixed (6):**

1. *(CR major, verdict.ts)* Core scan `onWarn` no longer defaults to `console.warn` — made **required** rather than noop-defaulted: core stays console-free, and no caller inherits a silent sink for corruption warnings (Tenet 4).
2. *(CR minor, shield.ts)* Manifest re-hash catch now distinguishes ENOENT (silent, first-run) from malformed/permission failures (surfaced under `TOTEM_DEBUG`).
3. *(CR nit, git.ts)* Branch-vs-base resolution deduplicated into a shared `resolveBranchScope` — the two call sites can no longer drift.
4. *(CR outside-diff, security)* Both `git ls-files` invocations gain the `--` pathspec separator (pre-existing on main; moved into the diff by the rev-4 seam extraction, taken here).
5. *(greptile P2, verdict.ts)* The laneId runner-vocab substring blacklist is **removed**: the rev-6 structural binding (suffix must equal `resolvedBackend`/`configuredLane`) closes the value channel completely, so the blacklist contributed only false-positive risk against legitimate future model names — the `claude-cold-inference-mini` example is now an explicit acceptance test, and every smuggle case still rejects structurally.
6. *(greptile P2, review-fan.ts)* A leading-colon `review.lanes` entry (`":model"`) now fails at config parse with a named hint instead of surfacing as a confusing provider API error.

**Declined with evidence (3)** — @gemini-code-assist:

1. *(critical, shield.ts:1779 — `TotemError` not in scope)* It is destructured from the dynamic import at shield.ts:1376, inside the same function scope as the throw; tsc and both full suites are green on every commit of this branch.
2. *(high, review-fan.ts:839 — CRLF breaks `GIT_DIFF_SEGMENT_RE`)* The `<git_diff>\n` boundaries are authored by our own `wrapXml`, which emits `\n` on every platform (asserted by its unit test); the regex parses our assembly, never disk content — and the extraction tests pass on a Windows host.
3. *(high, shield.ts:1533 — `IncrementalResult` may not be imported)* It is an interface defined in the same file (shield.ts:1292); no import exists to miss. tsc clean throughout.

**Process note:** the pre-merge description-template warning is declined by design — the Mechanical Root Cause/Fix Applied template fits defect PRs; this feature PR's body carries What/Provenance/Gates.

**Local-lane covariate:** no verdict artifact exists for this branch's lineage — the runner this PR *ships* wasn't yet available to review it; the local pre-external lane for this change was the 4-seat cohort panel + dual-codex cycle recorded in `.totem/specs/prop304-r2-review-runner.md` (14 → 8 → 4 → 0 to PASS).

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
**Rounds 2–4 disposition — fixes pushed as `09c1441f` and `20187325`**

@gemini-code-assist Four findings triaged:

1. *(round 2, high — prefix-less lane with colon, e.g. `llama3:8b`)* **Partially accepted.** The legitimate nub is fixed in `09c1441f`: the config error now teaches the supported spelling — the provider split is on the *first* colon, so `ollama:llama3:8b` parses as provider=ollama, model=`llama3:8b` (quantization tag intact). The proposed silent fallback to the base provider is **declined**: a typo'd provider (`anthropc:claude-x`) would silently route to whatever base is configured — the misroute this validator exists to prevent — and the same string would mean different models under different base configs. The `TotemConfigError`→`TotemError` swap is also declined: these are `review.lanes` config-semantics errors and all six sibling sites use the same class.
2. *(round 3, high — static value imports from `@mmnto/totem`, rule 64)* **Accepted, fixed in `20187325`.** All five value symbols now arrive via dynamic import at the async entry points and thread into the sync helpers; only type-only imports (which erase at compile time) remain static. Good catch — this also exposed that the compiled lint rule enforcing rule 64 did not fire on a new command file; a rule-coverage follow-up issue will be minted.
3. *(round 3, high — `: IncrementalResult` may not be imported)* **Declined — repeat of the round-1 decline, re-anchored after a line shift.** `IncrementalResult` is an interface defined in the same file (shield.ts); tsc and both full suites are green on every commit of this branch.
4. *(round 4, high — missing/invalid verdict file flows as `undefined` into the hash and dies as a misleading `DATABASE_MISMATCH`)* **Declined — the premise is false.** `readJsonSafe` never returns `undefined`: it throws `TotemParseError` on ENOENT with exactly the clear message requested ("File not found: `<path>`" + a check-that-it-exists hint), and distinct structured errors for unreadable files and malformed JSON, all with ES2022 cause chains (`packages/core/src/sys/fs.ts`). A missing file therefore surfaces loud and specific *before* the content-address verification runs; `loadVerdictArtifact`'s JSDoc documents this throw contract. The suggested bare `new Error` would also break the repo's structured-error convention, of which `TotemParseError`-for-file-read-failures is the documented core.

Greptile prose notes (summary, not inline): the unnecessary `git merge-base` for explicit-range lineage is **fixed** in `20187325` (the probe now runs only for branch-vs-base, with a recording-spy test); the "displaced" doubled annotation on `tryGit` is **declined with evidence** — removing either copy reintroduces the diff-scoped fail-open-catch lint error (verified empirically in both single-annotation configurations); the doubling is load-bearing, not stranded.

Gates at `20187325`: both `totem lint` scopes 0 errors, cli 3123 + core 3165 tests green, tsc clean.

Issue #2355, opened 2026-07-14
chore(mail): enumerateOutboxes inner scan follows symlinked agent/outbox dirs — dirent-filter the agent level


satur8d:
## Summary

`enumerateOutboxes` in `packages/cli/src/commands/mail.ts` dirent-filters the workspace and repo levels (`readdirSync(..., { withFileTypes: true }).filter(d => d.isDirectory())`, which does not follow symlinks), but the inner `visit` uses a bare `readdirSync(orchDir)` (no `withFileTypes`) and then joins on the returned agent names + `fs.existsSync(outbox)`. A symlinked `agent/` or `outbox/` directory would therefore be followed during the ECL mail scan.

## Assessment

Lower severity than the ingest path (#2354): the consequence is bounded to cohort-mail *enumeration/display*, not persistence into a retrievable index, and the outer levels are already dirent-filtered. But for defense-in-depth and consistency with `orchestration-resolver.ts` (which is dirent-only, no-follow by design), the agent-level scan should filter the same way.

## Proposed fix

Switch the inner `readdirSync(orchDir)` to `{ withFileTypes: true }` and keep only `isDirectory()` entries (matching the outer levels and `orchestration-resolver.ts`), so symlinked agent/outbox dirs are not traversed.

Location: `packages/cli/src/commands/mail.ts` (`enumerateOutboxes`, inner `visit`, ~L392).

PR #2356, opened 2026-07-14
security(ingest): skip symlinks under ingest globs — no host-file exfil into the index (#2354)

### LINKED ISSUE #2354: security(ingest): corpus sync follows symlinks into the index — no realpath containment (sensor + skip-symlinks guard)
## Summary

The corpus ingest walk reads file content through a symlink with no realpath containment, so a symlink committed under an ingested glob (e.g. `.totem/lessons/x.md → /etc/passwd` or `→ ~/.ssh/id_rsa`) has its **target** content chunked, embedded, and stored in the LanceDB index — later retrievable via `totem search` / MCP `search_knowledge`. This is the symlink-escape / context-import-traversal class recently patched in other agent CLIs' memory-import processors.

## Verified chain

- `packages/core/src/ingest/file-resolver.ts` `resolveFiles()` enumerates via `globSync(target.glob, { cwd: projectRoot, nodir: true })` and intersects with the git-non-ignored set (`git ls-files --cached --others --exclude-standard`).
- **The git intersection does not defend:** a committed symlink is a valid git entry (mode `120000`), so its path is in `git ls-files` output and passes `nonIgnored.has(relativePath)`.
- `packages/core/src/ingest/pipeline.ts:368` — `content = fs.readFileSync(file.absolutePath, 'utf-8')` follows the symlink to the target. No realpath check sits between resolve and read; the surrounding `catch` only handles read errors.
- `store.insert(...)` persists the chunks → queryable index.

## Reachability

`totem sync` (full or `--incremental`) is auto-triggered by the post-merge / post-checkout git hooks, so a malicious PR that adds such a symlink is ingested on the victim's next checkout+sync with no explicit action. Trivial on Linux/macOS; requires symlink privilege on Windows.

## Asymmetry (prior art already in the tree)

The symlink-escape class is already defended elsewhere — `author-sandbox.ts` (realpath fail-loud), `parity-detect.ts` `resolveWithinDir` (realpath containment, #1304 / #2256), `orchestration-resolver.ts` (dirent-only, no follow), and notably `run-compiled-rules.ts:348` which **explicitly excludes mode-`120000` symlinks** from the AST path. The ingest path never received the symmetric guard.

## Proposed fix (sensor-first, Tenet 13)

1. **Sensor:** a sync-time / `totem doctor` advisory that flags symlinks under ingest globs (detect via `git ls-files -s` mode `120000`, or `lstat`) — loud, non-gating.
2. **Guard:** skip symlinks in `resolveFiles` (mirror `run-compiled-rules.ts`: `lstat` → skip), or realpath-contain the resolved path against `projectRoot` (reuse the `resolveWithinDir` pattern). Skipping is simplest and matches the lint-path precedent.

Severity: medium-high (host-file exfiltration into a retrievable store; needs a malicious repo/PR + a victim sync).

Related: #1304, #2256 (prior symlink-containment hardening).

satur8d:
## What

Closes #2354. Skips symlinks under ingest globs so a planted symlink can't read host-file content into the searchable index.

## Why

The corpus ingest walk read each matched file with `fs.readFileSync` (which follows symlinks), so a symlink committed under an ingested glob (e.g. `.totem/lessons/x.md → /etc/passwd` or `→ ~/.ssh/id_rsa`) had its **target** content chunked, embedded, and stored in the LanceDB index on `totem sync` — retrievable later via `totem search` / MCP `search_knowledge`. `totem sync` is auto-triggered by the post-merge / post-checkout hooks, so a malicious PR is ingested on the victim's next checkout+sync. The git-tracked-set gate did **not** defend: a symlink is a valid git entry (mode `120000`).

This is the symlink-escape / context-import class already defended elsewhere in the tree (`author-sandbox.ts`, `parity-detect.ts` `resolveWithinDir` #1304/#2256, `run-compiled-rules.ts`'s mode-`120000` exclusion). Ingest was the one surface that slipped the pattern.

## Change

`resolveFiles` now `lstat`s each match and **skips symlinks entirely** (mirroring `run-compiled-rules.ts`), surfacing each skip as a **loud sync-time warning** (guard + Tenet-13 sensor in one). A raced/ENOENT `lstat` degrades to non-symlink, consistent with the downstream read's skip-don't-abort guard (annotated with the `// totem-context:` escape).

## Tests / gates

- +2 regression cases in `file-resolver.test.ts` (symlink skipped + warned; regular file still indexed); existing submodule tests unaffected.
- Full-workspace build clean, 14/14 file-resolver tests, prettier clean, ESLint 0 errors, `totem lint` PASS, changeset added (`@mmnto/totem` patch).

## Follow-on

The lower-severity mail-scan hardening is tracked separately in #2355 (not in this PR).


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
**Review round disposition** — all 4 findings accepted and fixed in `6acfa473`.

### ✅ Re-check symlink at the read site — Greptile P1 + CR Major (convergent)

`resolveFiles` skipped symlinks at *discovery*, but `pipeline.ts` reads `file.absolutePath` *later*, so a path swapped to a symlink in that gap was still followed. Added an `lstat` re-check immediately before `readFileSync` in `pipeline.ts` and skip symlinks there too — closing the discovery→read window to an adjacent-syscall gap. (Fully-atomic `O_NOFOLLOW`/fd isn't portable to Windows; the adjacent-syscall re-check is the pragmatic close.) Good catch — the discovery-only guard genuinely left this residual.

### ✅ Sanitize symlink-derived paths in log output — CR minor

A planted symlink's filename can carry ANSI escapes. `relativePath` now passes through `sanitizeForTerminal` before any log/warn interpolation — both the resolve-time warn and the read-loop logs.

### ✅ Test mock default masks errors — Greptile P2

The `node:fs` mock's bare `lstatSync` (returning `undefined`) routed unrelated resolver tests through the error-fallback path. Added a `beforeEach` defaulting to a non-symlink stat, so those tests exercise the real non-symlink path.

### ✅ Missing lstat-error-path test — CR trivial

Added a regression test for the raced/ENOENT `lstat` degrade (asserts the match still resolves).

Also annotated the pre-existing per-file read-error skip with a `// totem-context:` escape, now that its block is in scope.

---

Gates on `6acfa473`: full-workspace build clean, 35/35 ingest tests (file-resolver 15, pipeline 20), prettier clean, ESLint 0 errors, `totem lint` PASS. Round has converged.

satur8d:
@greptileai review
@coderabbitai review

satur8d:
**Re-review disposition** (`6acfa473` round) — 1 fixed, 1 declined with rationale; the rest is stale carryover of already-addressed findings.

### ✅ Fixed — sanitize `err.message` in the read-error log (CR minor)  →  `1f8ae50e`

The prior fix sanitized `safeRel` but left `err.message` raw, and Node `fs` errors embed the raw OS path — so a control-char filename could still leak via the error text. `err.message` now also passes through `sanitizeForTerminal`.

### ⛔ Declined — residual two-syscall TOCTOU (Greptile P1 "Path Race Remains" + CR Major)

The read-site guard is `lstatSync(path)` then `readFileSync(path)` — two lookups, so a microsecond swap-to-symlink between them still follows the link. The only fully-atomic fix is `O_NOFOLLOW` at open, which **CR's own web query confirms is a no-op on Windows** (a first-class totem platform) — so there is no portable atomic read.

Declining the escalation and naming convergence: the realistic vector — a *committed* symlink — is fully closed at **both** discovery and read (it is a symlink at both `lstat` times → skipped). The residual requires an attacker winning a microsecond race with a *concurrent local write* on every sync, which is outside a repo-ingest threat model. Chasing it further (e.g. `O_NOFOLLOW`-on-POSIX only) trades real complexity for an exotic residual on the platforms where symlinks are already trivial.

### Stale carryover (already fixed in `6acfa473`, re-anchored on re-review)

Mock-default (Greptile P2), missing lstat-error test (CR trivial), `relativePath` sanitize (CR minor), and the original discovery-time TOCTOU — all addressed; verified against the diff.

---

Gates on `1f8ae50e`: build clean, 35/35 ingest tests, prettier clean, ESLint 0 errors, `totem lint` PASS. Round converged — recommend merge.

PR #2358, opened 2026-07-14
feat(orchestrators): model refresh — current-gen defaults + sampling params reconciled at the provider boundary

### LINKED ISSUE #1476: orchestrator: strip sampling params when model is Opus 4.7+
## Problem

`packages/cli/src/orchestrators/anthropic-orchestrator.ts:125` forwards `temperature` (and would forward `top_p` or `top_k` if callers passed them) straight through to the Anthropic Messages API whenever a caller sets the field.

Claude Opus 4.7 (`claude-opus-4-7`) removed sampling params from the Messages API. Passing `temperature`, `top_p`, or `top_k` returns a 400 error.

This is latent risk today. Default Anthropic routing is `claude-sonnet-4-6`, which still accepts sampling params. The footgun fires the moment a user or override points at Opus 4.7.

## Call sites that currently pass `temperature`

Seven internal call sites set `temperature` on the options passed to the orchestrator:

- `packages/cli/src/commands/compile.ts:730` -- 0
- `packages/cli/src/commands/extract-local.ts:193` -- 0.4
- `packages/cli/src/commands/extract-pr.ts:261` -- 0.4
- `packages/cli/src/commands/extract-scan.ts:142` -- 0.4
- `packages/cli/src/commands/review-learn.ts:325` -- 0.4
- `packages/cli/src/commands/shield.ts:519 / 1126 / 1184` -- 0
- `packages/cli/src/commands/triage-pr.ts:524` -- 0

All seven are fine with `temperature` being silently dropped when the target model does not accept it. Opus 4.7 is deterministic by default.

## Scope of the fix

In `anthropic-orchestrator.ts`:

1. Detect Opus 4.7+ by model ID prefix (`claude-opus-4-7`, plus future 4.8+ Opus models when they ship).
2. When matched, omit `temperature`, `top_p`, and `top_k` from the outbound request, even if the caller set them.
3. Optional: debug-level log when a sampling param gets dropped, for caller transparency.

No changes needed to the seven call sites.

## Out of scope

- `extended_thinking` -> `adaptive_thinking` migration. Codebase grep confirms Totem does not use extended thinking today, so no code change is required. Noted for future awareness.
- Updating Totem's Anthropic default to Opus 4.7. Defaults stay at Sonnet 4.6; Opus-tier use is opt-in via overrides.

## Reference

- Opus 4.7 migration guide: https://platform.claude.com/docs/en/about-claude/models/migration-guide
- Paired doc update: `docs/reference/supported-models.md` (separate PR)

## Priority

Tier 2. Default routing uses Sonnet 4.6, so nothing breaks today. File before someone flips an override to Opus 4.7 and gets a confusing 400.

satur8d:
## What

Operator-ruled model refresh (2026-07-14): every LLM role moves to current-generation models, and the sampling-param seam that blocked current-gen models is fixed at the orchestrator boundary — closing mmnto-ai/totem#1476.

> **Stacked on #2357** (Tenet-16 `defaultModel` drop). Base is `fix/t16-defaultmodel`; GitHub will auto-retarget to `main` when #2357 merges.

### 1. The seam fix (the load-bearing half)

Current-generation models reject client sampling params with a 400: Anthropic Opus 4.7+/Sonnet 5+/Fable reject `temperature`; OpenAI gpt-5+/o-series additionally reject the legacy `max_tokens` key (they require `max_completion_tokens`). Every Totem LLM call site hardcodes a `temperature`, so pointing any override or review lane at these models failed at runtime — **GPT-5-family models could not be used as Totem orchestrators at all**, which also blocked consumers (the operator's explicit ask: users should be able to run spec/reviews on GPT).

- `modelStripsTemperature()` (`@mmnto/totem`) widens per its own changelog instruction ("when Anthropic ships Sonnet 5.0+, widen here") into the cross-provider predicate: Sonnet 5+, Haiku 5+, Fable/Mythos, gpt-5+, o-series (incl. two-digit `o10+`), provider-qualified strings accepted.
- The anthropic and openai orchestrator boundaries consume it: callers keep declaring their intended temperature; the boundary omits it for models that reject it. This fixes all seven #1476 call sites in one seam instead of per-site edits.
- The openai orchestrator selects `max_completion_tokens` vs legacy `max_tokens` on the same predicate — OpenAI-compatible local servers (Ollama, LM Studio, Groq) keep the legacy shape some of them require.
- `getMaxTokens` gives the current-gen Anthropic family 16K headroom: these models run adaptive thinking when `thinking` is omitted, and thinking draws from `max_tokens` — the 8K sonnet cap truncated live Sonnet 5 review verdicts mid-JSON (caught by end-to-end verification, below). Haiku keeps its conservative 4K cap even for hypothetical 5+ variants (review-lane finding).

### 2. The config refresh (this repo's pins)

| Role | Was | Now |
|---|---|---|
| review.lanes | `anthropic:claude-sonnet-4-6` + `gemini:gemini-3.1-pro-preview` | `anthropic:claude-sonnet-5` + `gemini:gemini-3.5-flash` |
| docs / spec / shield / triage | `gemini-3.1-pro-preview` | `gemini-3.5-flash` |
| extract / reviewlearn | `gemini-3-flash-preview` | `gemini-3.5-flash` |
| compile | `anthropic:claude-sonnet-4-6` | `anthropic:claude-sonnet-5` (freeze-parked; takes effect at unfreeze) |

Grounding: Gemini 3.5 Flash is GA (2026-05-19), beats 3.1 Pro on coding/agentic benchmarks at ~25% lower cost with flat long-context pricing, and is a stable ID where 3.1 Pro remains preview-only. Sonnet 5 is the same price tier as Sonnet 4.6 (intro-priced lower through 2026-08-31) with near-Opus review quality. Net cost decreases on every role. Watch-item (in-config comment): revert spec/docs to `gemini-3.1-pro-preview` if Flash prose quality regresses; re-evaluate when Gemini 3.5 Pro ships. `docs/reference/supported-models.md` re-validated across all three providers.

## Verification

Beyond unit gates (full workspace suite green; new tests for the predicate matrix, both orchestrators' param shapes, and the token caps), the change was **exercised end-to-end**: `totem review` ran the two new lanes live. Round 1 caught the Sonnet 5 truncation (lane abstained at exactly the 8K cap) → fixed → round 2 completed on both lanes with structured verdicts.

## Decision record (local review lane, 2 lanes × 2 rounds)

- **Accepted:** haiku-5 cap ordering (haiku branch now precedes the predicate branch — 4K stays the proven allocation until a real Haiku 5 exists); two-digit o-series regex widen (`o[1-9]\d*`).
- **Declined — decline-substantive:** "predicate conflation" (one predicate drives temperature-strip, token-key choice, and cap headroom). Deliberate coupling: today's model reality couples these three axes exactly, the coupling is documented at the predicate and both call sites, and speculative predicate splits for models that don't exist is the Tenet-21 anti-pattern. A future divergent model widens here — same contract the predicate has carried since birth.

## Out of scope

- Embedding swap to GA `gemini-embedding-2` — forces a full re-index, cohort-wide blast radius; deferred (flagged in supported-models.md).
- `init-detect` emitted defaults (Tenet-16 T0-vs-T1 judgment) — separately pending.
- The rule-compilation freeze is untouched; the compile-override bump is dormant until unfreeze.

Closes #1476

### DIFF
--- a/.changeset/model-refresh-param-boundary.md
+++ b/.changeset/model-refresh-param-boundary.md
@@ -1,0 +1,12 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+---
+
+feat(orchestrators): current-generation model support — sampling params reconciled at the provider boundary (mmnto-ai/totem#1476).
+
+Current-generation models reject client sampling params with a 400 (Anthropic Opus 4.7+/Sonnet 5+/Fable reject `temperature`/`top_p`/`top_k`; OpenAI gpt-5+/o-series additionally reject the legacy `max_tokens` key in favor of `max_completion_tokens`). Previously every Totem LLM role hardcoded a `temperature`, so pointing any override or review lane at a current-generation model failed at runtime — GPT-5-family models could not be used as orchestrators at all.
+
+`modelStripsTemperature()` (`@mmnto/totem`) widens from the Opus-4.7+-only regex to the cross-provider predicate (Sonnet 5+, Haiku 5+, Fable/Mythos, gpt-5+, o-series; provider-qualified strings accepted), and both the anthropic and openai orchestrator boundaries now consume it: callers keep declaring their intended temperature, and the boundary omits it for models that reject it. The openai orchestrator additionally selects `max_completion_tokens` vs legacy `max_tokens` on the same predicate, so OpenAI-compatible local servers (Ollama, LM Studio, Groq) keep the legacy shape they expect.
+
+Consumer-impact: orchestrator request shape — configs pointing at Opus 4.7+/Sonnet 5+/Fable or gpt-5+/o-series models now work instead of failing with a 400; requests to models that accept sampling params are byte-identical. `modelStripsTemperature` returns `true` for the new families, which also flows into the compile-worker fingerprint (records temperature absence) for anthropic-provider configs. No config migration required.
--- a/docs/reference/supported-models.md
+++ b/docs/reference/supported-models.md
@@ -1,6 +1,6 @@
 # Supported Models & AI Tools Reference
 
-> **Last validated:** 2026-05-02 (1.25.0 ship state). The Gemini tier below was refreshed 2026-06-13 from Google's published model docs (a targeted update, not a full re-validation).
+> **Last validated:** 2026-07-14 (model-refresh PR: Gemini, Anthropic, and OpenAI orchestrator tiers re-validated against published provider docs; Totem defaults moved to `gemini-3.5-flash` / `claude-sonnet-5`; sampling-param stripping now handled at the orchestrator boundary).
 
 Totem supports four LLM provider families for orchestration, and exports project
 knowledge to all major AI coding tools. This document tracks model IDs, tool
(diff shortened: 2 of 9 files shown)


satur8d:
@coderabbitai review

satur8d:
## Round 1 disposition

| CR | `gpt-[5-9]` also matches `gpt-5-chat-latest`, which accepts `temperature` — the predicate would silently drop a valid configured value | accept | Chat variants excluded from `modelStripsTemperature()` (`compile-worker-fingerprint.ts:117`); the token-key axis is now decided separately at the openai-orchestrator boundary (any gpt-5+ gets `max_completion_tokens`; `temperature` omitted only for reasoning models) + predicate and orchestrator test cases — `31b1d1f9`. Credit where due: this is the concrete divergent model that the "predicate conflation" concern predicted — the local review lane raised the abstract version and we declined it as speculative; CR grounded it in an existing model, so the axes are decoupled now. |
| CR | Repeated `16_384` magic number in `getMaxTokens` (repo named-constants guideline) | accept | Extracted `CURRENT_GEN_MAX_TOKENS` (`anthropic-orchestrator.ts:11`) — `31b1d1f9` |
| CR | `turbo-1` test case doesn't exercise the letter-preceded o-series boundary its comment claims (it fails on digit adjacency first) | accept | Comment corrected to describe adjacency; `turbo1` added as the true boundary-exclusion case — `31b1d1f9` |

All three accepted and fixed; full workspace tests, ESLint, and full-branch `totem lint` re-run green on `31b1d1f9`.

satur8d:
@coderabbitai review

satur8d:
## Round 2 disposition

| CR | Top-level `@mmnto/totem` imports in anthropic/openai orchestrators violate the `packages/cli/**` lazy-loading policy | decline-substantive | The static import pre-exists on `main` (identical line 1 in both files before this PR — the diff only appends `modelStripsTemperature` to the existing named-import list; prettier's single→multi-line reflow is what changed). It is also the established idiom of the whole orchestrator family (`gemini-orchestrator.ts:1`, `orchestrator.ts:9` carry the same static import, untouched here). Load timing already satisfies the policy's intent one level up: `orchestrator.ts` dynamic-imports each provider module at invocation time (`orchestrator.ts:346-361`), so each provider's static core import resolves only when that provider is first invoked — never at CLI startup — and `orchestrator.ts:9` loads `@mmnto/totem` before any provider module regardless. Converting these two files would alter nothing observable. If a zero-static-core-imports sweep of `packages/cli/**` is wanted, that's pre-existing repo-wide debt (orchestrator family + `orchestrator.ts` itself) for its own ticket, not this PR. |

No other actionable findings in the incremental pass (5 LGTMs, including on all Round 1 fixes).

PR #2372, opened 2026-07-15
feat(docs): machine-derived maturity surface — anchored rows, receipts trio, CI drift gate (strategy#531 A3)

satur8d:
## What

The A3 half of the strategy#531 ruled totem-lane build artifacts (ruling record: mmnto-ai/totem-strategy#531 comment 4974679759): a machine-derived maturity surface — Shipped / Partial / `Goal:` rows coupled to code and committed data, plus the ruled receipts trio, behind the docs-as-governed-surface track (mmnto-ai/totem-strategy#639). Part of that track; nothing here closes it.

- **`docs/wiki/maturity.md`** — handcrafted prose, injected facts. The table renders from `docs/data/maturity.json`; every row carries anchors (file / registered CLI command / committed data file) that are re-verified fail-loud on every `docs:inject`. A row whose anchor stops resolving fails the build instead of quietly going stale — that failure IS the staleness sensor.
- **Receipts trio (ruled):** rule count + provenance chain derived from `.totem/compiled-rules.json` (485 rules ← 485 lessons via `lessonHash`; 268 ast-grep / 217 regex); days-under-freeze derived from `.totem/freeze.json` + a committed `asOf` (no wall-clock in any render path — output is a pure function of the tree); a real-diff zero-LLM lint receipt over the pinned merged range `c14e90ab..ba8c591d` (PRs #2367–#2371 as merged history, 41 files, 387 rules evaluated, 0 errors / 20 warnings), generated with every provider API key stripped from the child environment and recomputed in CI on every PR (counts pinned; timing environment-labeled, never gated).
- **ECL row locked to the ruled register** (2–1 micro-split, operator-confirmed): `shipped · local-only · multi-seat opt-in`, `headline: false`, test-enforced.
- **Purge-list vocabulary content-locked** out of all rendered output (test-enforced), per the ruling's copy bounds.
- **New `docs-governance` CI job:** `test:docs`, inject-must-be-a-no-op drift gate over `README.md` + `docs/`, and the receipt recompute.
- `docs:inject` now ends with a prettier pass over its targets so generated output and `format:check` cannot fight.

## Drive-by fix (surfaced by the new CI job)

`test:docs` had never run in CI and was failing on main: `COMMAND_TABLE`'s parser required whitespace between `.command()` and `.description()`, so the same-line deprecated alias registration `.command('shield', { hidden: true }).description(...)` never matched. Fixed the parser and made it exclude `{ hidden: true }` registrations (deriving hiddenness from the source instead of only the hardcoded list); updated the stale test expectation — `shield` is a deprecated hidden alias of `review` and belongs in the excluded set.

## Findings for follow-up (not addressed here)

1. The README's "under 2 seconds" lint claim does not reproduce on the pinned real range (~3.3–3.6 s wall including node startup, win32). The maturity page renders the measured number and asserts nothing; recomputing or amending the README claim is the A2 proof kit's job.
2. The repo post-checkout hook errors in linked git worktrees (`.git` is a file there, not a directory; `line 30: .git/totem-sync.log: Not a directory`). Non-fatal, reproducible via `tools/gen-lint-receipt.mjs`.
3. Rules `compiledAt` max (2026-06-07) postdates the freeze (2026-05-17) — derived truth from the committed corpus, explained by the #2343 adjudicated re-seal exception class; noted so nobody reads the receipt as a contradiction.

## Verification

- `pnpm test:docs` 18/18; `pnpm build` + `pnpm test` green; `pnpm format:check` clean; `totem lint --base main` PASS (0 errors; 20 advisory warnings of the known `[Totem Error]`-prefix rule-regex false-positive class).
- `node tools/gen-lint-receipt.mjs --verify` reproduces the committed receipt; second `docs:inject` run is a byte-level no-op.

No changeset: no `packages/**` runtime change (repo docs/tools surface, same class as #2371).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Review round 1 — disposition (fixes in 66e83600)

**Fixed (6 findings + 1 symmetry item):**

1. **[CR major / greptile P2] `Date.now()` in the render path + `Date.parse` date normalization** — the strongest finding of the round, and correct: the future-`asOf` guard made the drift gate non-idempotent for the same committed tree. Replaced with `assertUtcDate` (strict YYYY-MM-DD UTC round-trip: impossible dates like `2026-02-31` reject; no wall-clock comparison anywhere in any render path). `freeze.json since` now validated the same way. New test locks the invariant.
2. **[CR major / greptile P2] unvalidated rendered fields** — `LINT_RECEIPT` now requires every rendered field including the environment labels (`platform`, `node`, `cliVersion`, `generatedAt`, with timestamp validity checked); `RULE_PROVENANCE` validates `lessonHash` + `compiledAt` per rule before the provenance claim renders. A hand-edited receipt fails loud instead of canonizing `undefined`.
3. **[CR major / greptile P2] `llmCalls` stamped, not observed** — now sensed from lint JSON (`lint.llmCalls ?? 0`; a non-zero count refuses to write the receipt). Key-stripping remains the mechanical enforcement; this closes the future hole where a counter could appear and be silently overwritten. The heavier full-telemetry variant is declined as out of proportion — the sensor + stripping together cover the attestation.
4. **[GCA high] atomic receipt write** — tmp + rename, per the house manifest-write rule.
5. **[CR major] spec receipt-schema row** — aligned to the live contract (`filesChanged, rules, errors, warnings, …`).
6. **[CR minor ×2] spec MD001 headings; README maturity line** — headings fixed; README line reworded plain ("each row backed by verifiable data in the repo and checked in CI").

Symmetry item: the `docs-governance` job now carries least-privilege `permissions: contents: read` (finding raised on the sibling PR #2373's workflow; applied here for consistency).

**Declined (2):**

1. **[GCA high] PID-suffixed worktree path** — the fixed path is deliberately self-healing: it is force-removed at the start of every run, so a crashed prior run cannot wedge the next one. PID-suffixed paths invert that property (a crashed run orphans a uniquely-named worktree plus its git metadata that no later run cleans). Concurrent invocations of a one-shot receipt regenerator aren't a supported scenario, and CI runners are isolated.
2. **[CR pre-merge description warning] template headings / `Closes #`** — the description carries the substance in prose, and closing nothing is deliberate: this PR is part of the strategy#531 track and must not auto-close its tracking issues (`closingIssuesReferences` verified empty).

**Verification:** `test:docs` 19/19 (two new invariant tests); receipt recomputes against the pinned range; second `docs:inject` run is a byte-level no-op; `format:check` clean; pre-push totem lint PASS (0 errors).

local-lane: no verdict artifact recorded for this lineage (the local review fan did not run on this branch; sensor line per `totem review --covariate`).

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai review

satur8d:

**Fixed 0, declined 4 with evidence.** No code change; head stays `66e83600`. Greptile's round-2 verdict at this head is 5/5 safe-to-merge and independently certifies the two re-raised points closed.

1. **gemini — `assertAnchorResolves` shape validation (HIGH) — declined.** Every malformed-anchor shape already fails the build before anything renders: a `null` anchor throws at `anchor.kind` (`tools/docs-transforms.cjs:149`); a non-object yields `ANCHOR_KINDS.has(undefined)` → the intended unknown-kind error (line 150); a missing `ref` throws inside `path.join(ROOT, undefined)` (line 162). No input class renders wrong output — the suggestion polishes the error message on an already-fail-loud gate. Round 1's accepted validation work targeted silent-wrong-render paths; this is not one. Declining as convergence.

2. **gemini — `Math.floor` → `Math.round` in `renderDaysUnderFreeze` (HIGH, + test sibling) — declined, unreachable premise.** Both inputs pass `assertUtcDate` (strict `YYYY-MM-DD` + UTC round-trip, lines 120–129; asserted at 176 and 283) before the subtraction. Date-only ISO strings parse as UTC midnight (ECMA-262 §21.4.3.2), so the difference is an exact multiple of 86,400,000 ms — UTC has no DST, and no fractional day can occur. `floor` ≡ `round` on every admissible input. Were a time component ever admitted, `floor` (elapsed full days) is the correct semantic for "days under freeze"; `round` would overstate by up to half a day. The test stays matched to `floor`.

3. **CodeRabbit duplicate — `lint.llmCalls ?? 0` (Major) — declined, converged (third round on this point).** The zero-LLM attestation basis is mechanical key-stripping (`tools/gen-lint-receipt.mjs:81–86`): with every provider credential deleted from the child environment, an attempted model call errors, lint emits no JSON, and the run fails at lines 98–101 — a silently-green receipt over an attempted call is not a reachable outcome. `llmCalls` is a refuse-if-contradicted sensor on top (lines 105–111, documented in-line), and CI `--verify` additionally requires the pinned counts to reproduce exactly on an independent machine — an LLM-influenced run would not reproduce deterministically. Failing-when-absent would make the receipt permanently ungeneratable against the real CLI contract: `totem lint` emits no `llmCalls` counter today. "Emit authoritative llmCalls telemetry from lint" is a CLI feature request (queued with the post-merge finding tickets), not a defect in this receipt tool.

4. **CodeRabbit duplicate — semantic receipt validation (Major) — declined, convergence (third validation pass on the same block).** The named exploits do not survive the gates above this function. For `LINT_RECEIPT`: `filesChanged: false`, SHA or count garbage are pinned fields — CI `--verify` compares each by exact equality against a fresh recompute (`gen-lint-receipt.mjs:145`), which is strictly stronger than any local type/format regex; `elapsedMs`/`platform`/`node`/`cliVersion` are environment labels, declared never-gated, and render as labels. For `RULE_PROVENANCE`: `compiledAt` lives in `.totem/compiled-rules.json`, which sits behind the compile-manifest seal — hand-edits trip `verify-manifest`, so adversarial manifest content is defended a layer up, not per-render. Round 1's presence + `Date.parse` checks cover the render-undefined class this transform owns.

Both PRs' rounds are now at quiescence from this seat's side. No local fan verdict exists for this lineage (covariate line honestly absent).

PR #2373, opened 2026-07-15
feat(proof-kit): A2 reproducible proof kit — committed fixture, blocked-recurrence CI proof, receipted timing (strategy#531 A2)

satur8d:
## What

The A2 half of the strategy#531 ruled totem-lane build artifacts (ruling record: mmnto-ai/totem-strategy#531 comment 4974679759; freeze-boundary adjudication: comment 4976403990): a reproducible proof kit — one real mistake class, blocked forever, with every claim about it recomputed from the committed fixture. Part of that track; nothing here closes it.

**The loop, committed under `examples/proof-kit/`:**

1. **Incident** — a legacy retry helper swallowed transport failures behind an empty catch block for three weeks. The repro is quarantined at `fixture/src/legacy/retry-2026-06.js` and stays in the tree on purpose: it is Stage 4's positive control (the compiled pattern demonstrably fires on the real historical shape, `confidence: high`).
2. **Lesson** — `fixture/.totem/lessons/lesson-empty-catch.md`, plain markdown.
3. **Rule** — `fixture/.totem/compiled-rules.json`: an ast-grep rule compiled by the real `totem lesson compile` pipeline against the fixture's own corpus (local + recorded only, never CI — see freeze note below), then human-promoted via `totem rule promote` per ADR-089 zero-trust. The `lessonHash` chain from incident to enforcement is mechanical.
4. **Blocked recurrence** — `mistake.diff` reintroduces the exact mistake (diff artifact only, never applied source). `run.mjs` materializes the fixture into a temp git repo, proves `totem lint` exits non-zero with the fixture rule firing (chain asserted via `lessonHash`), proves a clean change on the same file passes, and recomputes the public "under 2 seconds" lint claim as a hard bound: **841 ms, zero LLM calls, every provider API key stripped from the child environment** (`receipt.json`).

**CI:** new `Proof Kit` workflow runs `run.mjs --ci` on every PR — deterministic, zero-LLM, verifies the committed receipt's pinned fields and re-proves the block. `record.sh` casts the demo from the same scripts (the asciinema recording is an output of the kit, not a staged demo; producing the `.cast` is a local unix step, still owed).

**Freeze scope:** the host repo's rule-compilation freeze is untouched. The kit's compile step runs only against the fixture's own corpus in a temp directory under the recorded cohort adjudication (fixture-corpus-only; local+recorded never CI; fixture provenance on the committed rule; freeze disclosed in kit copy — all four conditions are encoded as fail-loud checks in `compile-fixture.mjs`).

**Host config:** `examples/proof-kit/fixture/**` joins `shieldIgnorePatterns` (the fixture deliberately carries the bad shapes its own rule blocks — same exclusion class as the lesson corpus), and the fixture `.totem/` joins `.prettierignore` (lesson regex + manifest hash-seal, mirroring the root entries).

## Product mechanics the kit surfaced (encoded as fail-loud kit checks)

- Only the ast/ast-grep engine class can block: the #2181/#2183 hard-tier split demotes the whole regex class to advisory regardless of severity. This is why the fixture's mistake is a code-class incident (an earlier markdown-class candidate could only ever warn).
- Stage 4 requires an in-tree positive control whose violating line exactly equals a `badExample` line — otherwise the rule lands `untested-against-codebase` (inert at lint) or candidate-debt (forced warning).
- ADR-089 ships every LLM-generated rule `unverified: true` (advisory); `totem rule promote` is the recorded human sign-off that makes it blocking.

## Findings for follow-up (not addressed here)

1. `@mmnto/cli` declares neither `@google/genai` nor `@anthropic-ai/sdk` — for npm consumers the orchestrator's SDK arm can never load, and every LLM call silently rides whatever vendor CLI is on PATH ("SDK unavailable" fallback). Additionally the CLI loads `.env` relative to its working directory, so keys in the repo root don't reach a child cwd.
2. The empty-corpus lint skip returns before honoring `--out` (a `--format json --out` run produces no file when zero active rules load).
3. Speed-claim register (operator-ruled this session): prose should carry only the structural speed claims — offline, zero LLM, no network, cost scales with the diff rather than the history — and concrete numbers belong only in receipted surfaces that carry their parameters. Our own receipts span 841 ms (1 rule, ~10-line diff) to ~3.4 s (387 rules, 41-file diff) on one machine, so no bare number generalizes. The README "under 2 seconds" line and the wiki "60 seconds" line are inputs to the A1 carve (routed to the strategy seat); this kit's copy frames its 2000 ms bound as the fixture's receipted envelope, not a product claim.

## Verification

- `node examples/proof-kit/run.mjs` and `run.mjs --ci` both green locally (block proven, clean pass, 841 ms < 2000 ms bound).
- `pnpm build` + `pnpm test` green; `pnpm format:check` clean; `totem lint --base main` PASS (0 errors; 7 advisory warnings of the known tools-script `readFileSync` false-positive class).
- Sibling PR #2372 (A3 maturity surface) is independent — disjoint files, both branched off main.

No changeset: no `packages/**` runtime change (repo example/tooling surface).

### DIFF
--- a/.github/workflows/proof-kit.yml
+++ b/.github/workflows/proof-kit.yml
@@ -14,6 +14,8 @@ jobs:
   prove:
     name: Prove the loop (zero LLM)
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v7
 
--- a/.github/workflows/proof-kit.yml
+++ b/.github/workflows/proof-kit.yml
@@ -1,0 +1,32 @@
+name: Proof Kit
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  prove:
+    name: Prove the loop (zero LLM)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: pnpm/action-setup@v5
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - run: pnpm install --frozen-lockfile
+
+      - run: pnpm build
+
+      - name: Prove blocked recurrence + recompute the timing claim
+        run: node examples/proof-kit/run.mjs --ci
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,6 +10,8 @@ pnpm-lock.yaml
 # Compiled rules hash must match compile-manifest.json — prettier would invalidate it
 .totem/compiled-rules.json
 .totem/compile-manifest.json
+# Proof-kit fixture .totem mirrors both concerns above (lesson regex + hash seal)
+examples/proof-kit/fixture/.totem/
 # Spine artifacts are canonicalStringify-emitted; the freeze proof compares raw file
 # bytes to the committed blob (spine-freeze-proof.ts) — prettier rewrap reads as tamper
 .totem/spine/
--- a/examples/proof-kit/README.md
+++ b/examples/proof-kit/README.md
@@ -1,0 +1,65 @@
+# Proof kit
+
+One real mistake, blocked forever — and every claim about it recomputes from
+this directory.
+
+The fixture under [`fixture/`](fixture/) is a tiny repository that lived
+through one incident: a legacy retry helper swallowed three weeks of send
+failures behind an empty catch block
+([the incident record](fixture/docs/incident-2026-06-silent-failures.md)).
+The loop this kit proves is the product's core loop:
+
+1. **The mistake happened** — the repro is quarantined at
+   [`fixture/src/legacy/retry-2026-06.js`](fixture/src/legacy/retry-2026-06.js).
+   It stays in the tree on purpose: it is the compiled rule's positive
+   control (Stage 4 verifies the pattern fires on the real historical shape).
+2. **The lesson was banked** —
+   [`fixture/.totem/lessons/lesson-empty-catch.md`](fixture/.totem/lessons/lesson-empty-catch.md),
+   plain markdown.
+3. **The rule was compiled and human-promoted** — the committed
+   [`fixture/.totem/compiled-rules.json`](fixture/.totem/compiled-rules.json)
+   carries the ast-grep rule with the lesson's content hash (`lessonHash`) —
+   the provenance chain from incident to enforcement is mechanical. Rules
+   ship zero-trust advisory (ADR-089); promotion to blocking is a recorded
+   human sign-off, same register as everything else here: a human merges
+   every change.
+4. **The recurrence is blocked** — [`mistake.diff`](mistake.diff) reintroduces
+   the exact mistake (it exists in this repo only as a diff artifact, never as
+   applied source). [`run.mjs`](run.mjs) applies it in a temp checkout and
+   proves `totem lint` exits non-zero with the fixture rule firing — then
+   proves a clean change on the same file passes.
+
+## Run it
+
+```bash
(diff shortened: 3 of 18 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Review round 1 — disposition (fixes in ba580175)

**Fixed (6 findings + 1 symmetry item):**

1. **[greptile P2] `--ci` timing assertion ordering** — pinned-field verification now runs before the timing check, so a loaded-runner trip reports as the distinct envelope failure instead of masquerading as a block-proof regression. The 2000 ms bound itself is retained as the fixture's receipted envelope (one rule, ~10-line diff — measured 716–856 ms across runs); if it proves flaky on real runners we'll loosen it with data rather than pre-emptively.
2. **[greptile P2] no `spawnSync` timeout** — lint spawns now carry `timeout: 120_000` with a distinct fast-failure message; a hung CLI can no longer stall CI toward the 6-hour ceiling.
3. **[greptile P2] naive `.env` parser passes quoted values** — dotenv-style quoted values are unwrapped before reaching the SDK, so a quoted key no longer silently reroutes the compile onto the vendor-CLI fallback.
4. **[GCA high ×2] atomic manifest writes** — receipt write and both fixture artifact copies are tmp + rename, per the house manifest-write rule.
5. **[CR info] workflow least-privilege** — `permissions: contents: read` added to the prove job.
6. **[CR minor] unsanitized git stderr (3 sites)** — shared ANSI/control-char sanitizer applied to every subprocess string that reaches a terminal, in both kit scripts.

Symmetry item: `llmCalls` is now sensed from lint JSON (absent = 0; non-zero refuses) rather than stamped — mirroring the same fix on the sibling PR #2372.

**Declined (1):**

1. **[CR minor] lesson first-line heading MD041 (`##` → `#`)** — `## Lesson —` is the lesson-corpus heading convention (the README's own lesson example uses it), the file is deliberately excluded from formatting exactly like the root corpus, and — decisively — editing the lesson changes its bytes and would break the committed `lessonHash` provenance chain that this kit exists to prove. A markdownlint preference cannot be worth invalidating the artifact under proof.

**Verification:** loop re-proven post-fix — recurrence blocked (735 ms), clean pass (716 ms), receipt regenerated and `--ci` verification green (chain intact, zero LLM calls); `format:check` clean; pre-push totem lint PASS (0 errors).

local-lane: no verdict artifact recorded for this lineage (the local review fan did not run on this branch; sensor line per `totem review --covariate`).

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai review

satur8d:

**Fixed 2, declined 0.** Fix commit: `5e4eb0f9`. Proof re-run end-to-end after the change: recurrence BLOCKED, clean pass 726 ms < 2000 ms bound, zero LLM calls, pinned receipt fields verified.

1. **CodeRabbit — sanitizers pass C1 controls (Major) — FIXED.** The shared predicate now rejects 0x80–0x9F alongside C0/DEL (newline/tab preserved), at both sites: `examples/proof-kit/run.mjs` and `examples/proof-kit/compile-fixture.mjs`.
2. **Greptile summary prose — local git spawns lack timeouts — FIXED.** The `git()` helper and the standalone `git apply` in `run.mjs` now carry `timeout: 30_000`, same rationale as the round-1 `lint()` timeouts: a hung git must fail fast, not stall CI. The `totem lesson compile` / `totem rule promote` spawns in `compile-fixture.mjs` stay untimed deliberately — that script is CI-blocked (local + recorded only) and the LLM compile is legitimately slow; this matches greptile's own assessment of those calls.

Round-2 verdicts at the reviewed head: greptile 5/5 safe-to-merge, gemini explicitly no findings. No local fan verdict exists for this lineage (covariate line honestly absent).

PR #2378, opened 2026-07-15
fix(sync): incremental sync reconciles ignore-pattern removals — exclusion-set hash + newly-eligible enqueue (#2366)

### LINKED ISSUE #2366: fix(sync): incremental sync misses ignore-pattern REMOVALS — newly eligible unchanged files stay unindexed until full sync
## Summary

Incremental sync reconciles ignore-pattern **additions** but not **removals**. When a pattern is removed from `ignorePatterns` / `indexIgnorePatterns`, files that just became index-eligible — but whose content hasn't changed since `lastSyncSha` — are never reprocessed: `filesToProcess = allFiles.filter((f) => changedSet.has(f.relativePath))` (`packages/core/src/ingest/pipeline.ts`, incremental arm). They stay out of the index until a full sync or an unrelated content change.

The **addition** direction already self-heals: orphan reconciliation (mmnto-ai/totem#2151) purges indexed paths no longer present in `allFiles`, so newly-ignored files leave the index on the next sync. The removal direction has no symmetric mechanism.

## Class

Pre-existing — `ignorePatterns` edits have behaved this way since incremental sync landed; #2365 (`indexIgnorePatterns`) widens the config surface whose edits hit the gap but does not change the mechanism. Surfaced by CodeRabbit (Major) on the #2365 round; routed here rather than fixed in-PR because the fix is sync-state schema work, out of scope for a routed defect fix.

## Proposed shape (from the #2365 round, consistent with #2151's reconciliation philosophy)

Persist a hash of the effective index-exclusion set (`[...ignorePatterns, ...indexIgnorePatterns]`, order-normalized) in sync state; on mismatch, either invalidate incremental mode (simple, one full re-embed on pattern edits) or enqueue only the newly-eligible set (`allFiles − indexed paths`, cheaper). The second option pairs naturally with the existing `getDistinctPaths()` read the orphan pass already performs.

## Provenance

CodeRabbit Major on mmnto-ai/totem#2365 (2026-07-14), premise verified against `pipeline.ts` incremental arm. Related: #2151 (orphan reconciliation), #1981 (post-merge re-index gate staleness — same index-staleness family, different trigger).

satur8d:
Fixes #2366.

Incremental sync reconciled ignore-pattern **additions** (orphan purge, #2151) but not **removals**: a file that became index-eligible without a byte change was invisible to the git-diff window and stayed unindexed until a full sync.

## Mechanism (option B from the issue)

- `SyncState` gains an optional `indexExclusionHash` — an order-normalized content hash of `[...ignorePatterns, ...indexIgnorePatterns]`, computed with the existing `calculateDeterministicHash` (no new hashing primitive).
- On incremental sync, a hash mismatch — or an absent stored hash (state predating the field) — enqueues `computeNewlyEligiblePaths(live, indexed)`, the symmetric complement of `computeOrphanPaths`, reusing the same guarded `getDistinctPaths()` read the orphan pass already performs. On a read failure both reconciliations skip together (warn, never a false purge or spurious enqueue).
- The union reduces to the plain diff when the exclusion set is unchanged; pattern **reordering** is provably not a change (sorted before hashing); a pure pattern **addition** cannot cause a re-embed storm (live − indexed is near-empty on an up-to-date index).
- The hash persists in the same `writeSyncState` call both modes share. Older sync state reads normally — absence is a benign one-time mismatch. No migration.

## Verification

- 9 new tests (hash order-normalization, membership-change sensitivity, newly-eligible enqueue, hash-match no-op, absent-hash semantics, separator normalization, raw-path return). Pure-helper tests — no git spawned, no temp dirs.
- Full core suite 3209/3209; `pnpm -r build` green across all five packages; repo-wide format, ESLint, and `totem lint --base main` clean; pre-push hook suite (6651 tests) passed on push.
- Changeset: patch bump for `@mmnto/totem` with consumer-impact notes.

Implementation by an Opus 4.8 subagent under Fable drive; reviewed, gated, and pushed by totem-claude.

### DIFF
--- a/.changeset/sync-ignore-removal-reconciliation.md
+++ b/.changeset/sync-ignore-removal-reconciliation.md
@@ -1,0 +1,9 @@
+---
+'@mmnto/totem': patch
+---
+
+fix(sync): incremental sync now reconciles ignore-pattern REMOVALS, not just additions (mmnto-ai/totem#2366).
+
+Removing a pattern from `ignorePatterns` / `indexIgnorePatterns` makes previously-excluded files index-eligible, but incremental sync only re-embedded files the git diff reported as changed — so a newly-eligible file whose bytes were unchanged since the last sync stayed out of the index until a full sync or an unrelated edit. The addition direction already self-healed via orphan reconciliation; this adds the symmetric mechanism. Sync state now persists an order-normalized hash of the effective exclusion set, and on the next incremental sync a hash change (or its absence, for state written before this fix) enqueues the newly-eligible set (live files not yet indexed), reusing the indexed-path read the orphan pass already performs. Reordering patterns is not treated as a change, and an unchanged exclusion set adds no extra work.
+
+Consumer-impact: incremental sync only — after an ignore-pattern edit, files that became eligible are re-indexed on the next sync instead of lagging until a full rebuild. The `.totem/cache/sync-state.json` file gains an optional `indexExclusionHash` field; older state without it is read normally (treated as a one-time exclusion-set mismatch, which is a near-empty enqueue for an up-to-date index). No migration required.
--- a/packages/core/src/ingest/pipeline.test.ts
+++ b/packages/core/src/ingest/pipeline.test.ts
@@ -9,7 +9,9 @@ import { TotemConfigSchema } from '../config-schema.js';
 import { cleanTmpDir } from '../test-utils.js';
 import {
   buildIndexManifest,
+  computeNewlyEligiblePaths,
   computeOrphanPaths,
+  hashIndexExclusionSet,
   INDEX_MANIFEST_SCHEMA,
   runSync,
 } from './pipeline.js';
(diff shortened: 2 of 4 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:

**Fixed 2, declined 2.** Fix commit: `0f73c40f` (targeted suite 28/28; full pre-push suite green on push). Round verdicts: gemini explicitly clean, greptile 4/5 safe-to-merge, CR 1 actionable.

1. **greptile P2 — sort/order-sensitivity coupling — FIXED.** `hashIndexExclusionSet` now carries the inline note tying the `.sort()` to `calculateDeterministicHash`'s order-significant semantics ("the .sort() IS the order-independence invariant").
2. **greptile P2 — test (c) asserted trivial identity — FIXED.** The vacuous `undefined !== hash` assertion is gone; (c) now proves the absent-hash first-run scan recovers a partially-unindexed set AND that it is a no-op on an up-to-date index.
3. **CR nitpick (Trivial) — runSync integration test for the mismatch branch — declined.** This file's established pattern tests reconciliation as pure seams (`computeOrphanPaths` from the #2151 orphan pass has exactly this coverage shape — no runSync integration test either); the mismatch branch is a one-line `!==` whose both arms are pure-tested, and standing up a store/embedder harness for it is out of proportion to the guarded line. Same-coverage-class consistency with the accepted #2151 precedent.
4. **CR outside-diff (Trivial) — "reversed argument order between sibling helpers" — declined, premise inverted.** The orders are consistent under the semantic convention both helpers share: the MINUEND comes first. `computeOrphanPaths(indexed, live)` filters the indexed set (indexed − live); `computeNewlyEligiblePaths(live, indexed)` filters the live set (live − indexed). Positional alignment would break that semantic convention, and each helper's doc comment states its direction.

Round at quiescence from this seat's side.

Issue #2410, opened 2026-07-17
feat(cli): prepare-script contract build slice — init-distributed wrapper + hook-install exit-code contract + managed-hook regeneration (strategy#894 Option B)


satur8d:
## Provenance

Operator ruling on mmnto-ai/totem-strategy#894 (comment 5000993265, 2026-07-17): **Option B — init-distributed wrapper**, both lanes having concurred B. Sequence per the disposition: this build issue first → strategy authors the convention against it → one 4-repo propagation pass. This issue is the totem build slice only; convention text is strategy's lane.

## Context

The prepare-script convention hardens consumer repos' `package.json` `prepare` entries: the legacy `totem hooks || true` shape swallows genuine hook-install failures (Tenet 4 — the #894 framing). #2406 (merged 2026-07-17) stabilized the semantics the contract will freeze: drift-repairing `installGitHook` bounded to totem-OWNED content, end markers on all four templates, doctor remediation standardized on `totem hook install --force`.

Totem's own repo keeps the self-contained byte-identity `tools/` variant (bound to the CLI builders by `tools-hook-parity.test.ts`) as the owner-repo exception — non-exportable, since consumers lack the template builders to bind against. B is the consumer shape.

### 1. Init-distributed prepare wrapper (the B shape)

- `totem init` distributes a managed `.cjs` prepare entry riding the same managed-artifact rail as distributed skills; the consumer `prepare` script points at it. Doctor's managed-block sensing extends to it naturally.
- Wrapper semantics (logic in node, never in shell operators — the Windows-shell class):
  - CLI absent (spawn ENOENT) → the declared skip class (strategy#630), exit 0
  - CLI present → `totem hook install` with full tier-resolved templates
  - Genuine hook-install failure → propagates, exit 1 (fail-loud)

### 2. Exit-code contract test block on `hooksCommand` (Tenet 19)

Freeze the 0/1 split as documented + tested semantics, with the post-#2406 amendment — "no overwrite of user-owned or unbounded content" replaces v0's "no auto-overwrite":

- exit 0: installed · already-current · **drift-repaired (totem-owned bounded region only)** · declared skips
- exit 1: genuine failure
- `--check`: 0/1 verify mode

#2406 already carries the ownership/repair matrix; this adds the dedicated exit-code contract block so the contract has its falsifying test.

### 3. Managed-hook regeneration (scoped in per the ruling dispatch)

[...]

PR #2422, opened 2026-07-17
fix(cli): worktree-aware hooks dir — resolve the gitdir pointer instead of blind-joining .git/hooks

### LINKED ISSUE #2418: prepare wrapper: git-worktree gitdir pointer crashes ENOTDIR — violates the #2410 declared-skip contract
## What

The 1.101.0 init-distributed consumer prepare wrapper (`.totem/prepare.cjs`, mmnto-ai/totem#2410) crashes with **exit 1** in a **git worktree** instead of the contracted declared-skip:

```
[Totem Error] ENOTDIR: not a directory, mkdir 'C:\...\<worktree>\.git\hooks'
```

In a worktree `.git` is a **file** (gitdir pointer), and the hook-install path `mkdir`s `.git/hooks` without discriminating. The mmnto-ai/totem#2410 contract line this violates: *"Declared skips → exit 0: not-a-git-repo · unparseable gitdir pointer (worktree/submodule) · hook-manager detected · already-current."* The worktree case is named in the contract; the shipped wrapper path misses it.

## Repro (verified live 2026-07-17, Windows, @mmnto/cli 1.101.0)

1. `git -C <consumer-repo> worktree add <path> -b test origin/main` (consumer = any repo with the wrapper, e.g. totem-strategy or totem-status post-propagation)
2. `cd <path> && pnpm install` → prepare runs `node .totem/prepare.cjs` → ENOTDIR, exit 1, **install fails**

Impact: any worktree-based flow on a consumer repo (cohort peer-checkout vehicles use worktrees routinely) fails `pnpm install` outright.

## Sharpening datum — the owner variant handles it

In the **same worktree** of the totem repo itself, `tools/install-hooks.js` (the owner-repo byte-identity variant) ran clean and installed hooks — so the gitdir-pointer handling exists in the owner path and is missing only from the exported consumer wrapper (or from the `totem hook install` path it shells into; the ENOTDIR surfaced through the wrapper's invocation).

## Non-report

Real (non-worktree) checkouts verified fine: strategy's wrapper ran exit 0 with all four hooks already-current and user-owned files untouched (mmnto-ai/totem-strategy#913).

Found during the mmnto-ai/totem-strategy#894 propagation pass. — strategy-claude

satur8d:
Closes #2418

## Root cause

Every hooks-path site in `install-hooks.ts` computed `path.join(gitRoot, '.git', 'hooks')` and `installGitHook` ran `mkdirSync` on it. In a linked worktree `.git` is a `gitdir:` pointer FILE, so the mkdir crashes ENOTDIR and — through `.totem/prepare.cjs` → `totem hook install` → verbatim exit propagation — fails the consumer's entire `pnpm install`. The worktree case is named in the #2410 declared-skip contract; the shipped path missed it (the owner-repo `tools/install-hooks.js` variant already discriminated, per the issue's sharpening datum).

## Fix

- **`resolveHooksDir(gitRoot)`** (exported, `install-hooks.ts`): delegates to `git rev-parse --git-path hooks` — git's own worktree/`commondir` walk, which also honors `core.hooksPath` — via raw `spawnSync` (the doctor.ts idiom; keeps the core barrel off CLI cold-start). Filesystem probe as offline fallback for the plain `.git`-directory layout; **null** (never a blind join) when `.git` is a pointer file git could not resolve.
- Wired through all six sites: non-interactive install (the #2418 crash site), interactive post-merge/enforcement installs, the silent post-checkout install, `--check`, and the silent pre-push upgrade. In a worktree, hooks now land in the **shared** hooks dir — the location git actually executes — and `--check` reads the same place instead of wrongly reporting them missing.
- **Unparseable `.git` pointer file** (git's `fatal: invalid gitfile format`): now the declared skip the #2410 exit-code contract names — `hooksCommand` catches exactly that error class, prints a truthful skip line, exits 0. All other resolution failures stay fail-loud.
- **Doctor `git-hooks` parity row** (`doctor-parity.ts`): reads the resolved hooks dir, so a worktree checkout no longer reports all four hooks as absent drift (read-only `.git/hooks` join retained as last-resort fallback, pre-existing behavior).

## Verification

- 7 new cases in `install-hooks-exit-contract.test.ts` (real `git worktree add` + malformed-pointer fixtures): pointer resolution to the shared dir, install-from-worktree exit 0 with all four hooks landing shared-side, `--check` exit 0 from the worktree, `installHooksNonInteractive` classification, declared-skip exit 0 + message on the unparseable pointer. File total 33/33 green; full CLI suite 144 files / 3313 tests green.
- End-to-end with the built CLI: `hook install` from a real linked worktree → exit 0, hooks in `main/.git/hooks`, pointer file untouched; `--check` → exit 0; garbage `.git` file → declared-skip line, exit 0.

## Scope notes

- The `.totem/prepare.cjs` wrapper template is untouched — it spawns `hook install` and propagates the exit verbatim, so the fix rides the CLI. Consumers pick it up at the next cut; no reflex bump.
- Known-remaining same-class gap, deliberately out of scope: `eject.ts` scrubs `.git/hooks/<name>` blind, so ejecting from a worktree reports hooks "not found" (silent no-op; also debatable whether a worktree eject should touch shared hooks at all). Flagging rather than folding it in — can file a follow-up on ruling.
- Changeset: patch `@mmnto/cli`.

Found by strategy-claude during the strategy#894 propagation pass (repro: totem-strategy worktree, 1.101.0).

### DIFF
--- a/.changeset/worktree-gitdir-hooks-dir.md
+++ b/.changeset/worktree-gitdir-hooks-dir.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/cli': patch
+---
+
+`totem hook install` (and the `.totem/prepare.cjs` consumer wrapper that invokes it) no longer crashes with ENOTDIR in a linked git worktree (mmnto-ai/totem#2418). In a worktree `.git` is a `gitdir:` pointer FILE, and every hooks-path site blindly joined `.git/hooks` before `mkdir` — failing the consumer's whole `pnpm install` through the `prepare` lifecycle. A shared `resolveHooksDir()` now delegates to `git rev-parse --git-path hooks` (git's own worktree/`commondir` walk, which also honors `core.hooksPath`), so hooks install into the shared hooks directory git actually executes; `--check`, the silent pre-push upgrade, and the doctor `git-hooks` parity row read the same resolved location instead of wrongly reporting hooks missing from a worktree. An unparseable `.git` pointer file is now the declared skip the #2410 exit-code contract names (exit 0 with a truthful skip line), never a crash.
--- a/packages/cli/src/commands/doctor-parity.ts
+++ b/packages/cli/src/commands/doctor-parity.ts
@@ -357,18 +357,18 @@ interface GeneratedArtifact {
 
 /**
  * Resolve the four git-hook artifacts the `git-hooks` contract checks, each with a
- * per-repo regenerated canonical. Checks `.git/hooks/*` only — hook-manager installs
+ * per-repo regenerated canonical. `hooksDir` is the repo's RESOLVED hooks directory
+ * (worktree-aware — mmnto-ai/totem#2418). Checks git hooks only — hook-manager installs
  * (`.totem/hooks/*.sh` for husky / lefthook) are a follow-on (no cohort consumer uses
  * one today); the detector's present-without-marker → `skip` keeps a manager repo
  * honest in the meantime.
  */
 function gitHookArtifactsFor(
-  gitRoot: string,
+  hooksDir: string,
   tier: 'strict' | 'standard',
   fallbackCmd: string,
   builders: HookBuilderSource,
 ): GeneratedArtifact[] {
-  const hooksDir = path.join(gitRoot, '.git', 'hooks');
   const m = builders.markers;
   return [
     {
@@ -749,6 +749,7 @@ export async function checkParity(cwd: string): Promise<ParityCheckResult> {
         buildHookContent,
         buildPostCheckoutHookContent,
         getFallbackCommand,
+        resolveHooksDir,
         TOTEM_PRECOMMIT_MARKER,
         TOTEM_PREPUSH_MARKER,
         TOTEM_HOOK_MARKER,
@@ -1034,7 +1035,18 @@ export async function checkParity(cwd: string): Promise<ParityCheckResult> {
           let artifactLabel = 'artifact';
           let installCommand = 'totem init';
           if (c.id === 'git-hooks') {
-            generatedArtifacts = gitHookArtifactsFor(gitRoot, hookTier, fallbackCmd, hookBuilders);
+            // Worktree-aware hooks location (mmnto-ai/totem#2418): a linked
+            // worktree's hooks live behind the `.git` gitdir pointer, so a blind
+            // `.git/hooks` join would report every hook missing there. The
+            // read-only join fallback keeps the pre-#2418 behavior when even the
+            // resolver comes up empty.
+            const parityHooksDir = resolveHooksDir(gitRoot) ?? path.join(gitRoot, '.git', 'hooks');
+            generatedArtifacts = gitHookArtifactsFor(
+              parityHooksDir,
+              hookTier,
+              fallbackCmd,
+              hookBuilders,
+            );
             artifactLabel = 'git hook';
             // --force so the remediation actually repairs a stale hook: bare
             // `totem hook install` only drift-repairs a totem-OWNED whole file, but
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round-1 disposition (fixes in `aaaa51f0`)

**Fixed (4 findings → 3 changes):**

1. **greptile P1 + CodeRabbit "malformed-pointer classification before the interactive resolvers throw" (same defect) — FIXED.** Confirmed at source: the hidden legacy `totem install-hooks` command (index.ts:1993) reached `handleError` → exit 1 on a malformed `.git` pointer, violating the #2410 declared-skip contract on that entry point. A shared `resolveGitRootForHookPath` now threads the classification through every hook-install entry: `installPostMergeHook`, `installEnforcementHooks`, `installHooksCommand`'s post-checkout block, `installHooksNonInteractive` (direct-API path), and `checkHooksInstalled`; `hooksCommand`'s inline try/catch collapsed onto the same helper. Two new contract tests (legacy command + direct API) plus a live check with the built CLI: `install-hooks` on a garbage pointer → skip line, exit 0. This also discharges greptile's 3/5 confidence justification, which restated this gap.
2. **greptile P2 (stream inconsistency) — FIXED.** `HOOKS_DIR_UNRESOLVED_MSG` now goes to stderr at all four sites (the two interactive paths used `console.log`).
3. **CodeRabbit "non-directory `core.hooksPath`" — FIXED.** Verified reachable: `--git-path hooks` returns `core.hooksPath` verbatim, so the `/dev/null` hooks-disabled idiom would have handed `installGitHook` an unwritable target. `resolveHooksDir` now classifies an existing non-directory hooks path as the declared skip. New portable test (hooksPath aimed at a file).

**Declined (2 findings, with evidence):**

4. **CodeRabbit "keep the child-process dependency lazy-loaded" — declined.** `node:child_process` is a Node builtin, not the heavy-dependency class the CLI cold-start boundary exists for: the enforced repo rule bans static **`@mmnto/totem` barrel** value imports (it blocked exactly that in this PR's first push, which is why this import exists) and passes on builtins. Direct precedent: `doctor.ts:1` statically imports `spawnSync`, and this file's pre-existing static `node:fs`/`node:path`/`node:readline` imports would be equally banned under the path-instruction reading. A dynamic import would also force async through five sync public APIs for zero cold-start gain.
5. **CodeRabbit "do not restore the blind `.git/hooks` path in doctor-parity" — declined as out of scope.** The `?? path.join(...)` fallback preserves the pre-#2422 behavior byte-for-byte for the resolver-null case (doctor's `gitRoot` can be a cwd fallback outside any repo — parity rows reported those paths as missing before this PR too, read-only). Converting the git-hooks parity row to an explicit unresolved-location skip is a doctor-row semantics change beyond this fix's scope; happy to take it as a follow-up issue on the owner's word.

gemini-code-assist reviewed with no findings. CI re-running on `aaaa51f0`.

satur8d:
@greptileai review
@coderabbitai review

satur8d:
## Round-2 disposition (CodeRabbit outside-diff finding; falsifier in `36c7a6c5`)

**"Distinguish an unresolved hooks directory from a hook-manager skip" — code change declined, test suggestion accepted.**

The observed behavior — `hooksCommand` continuing to managed session-hook regeneration after the git-hook skip — is the **contract, not a leak**. Managed session hooks (`.claude/hooks/*.cjs`, `.gemini/hooks/*.js`, `.totem/prepare.cjs`) are vendor artifacts explicitly independent of git-hook state: the PR-A slice-3 design comment at the regeneration site and the lc#806 guard test ("hook-manager repo STILL drift-repairs an existing bounded session hook") encode exactly this independence for the sibling null case, and the same rationale covers hooks-dir-unresolved — a repo with hooks disabled via `core.hooksPath` still legitimately owns session artifacts to drift-repair. The skip line is scoped truthfully ("skipping **git hook** installation"). A discriminated return that stops the caller would *reintroduce* the lc#806 stale-session-hook class on hooks-disabled repos.

The suggested top-level contract test was worth taking, as the falsifier that locks this: `36c7a6c5` adds the file-valued `core.hooksPath` case to the `hooksCommand` exit-code contract — declares the scoped git-hook skip (exit 0), the non-directory target receives no write, and a drifted bounded session hook is still repaired. 37/37 green.

Issue #2396, opened 2026-07-16
feat(mail): consume-atomicity — \mail reply\ marks source processed + standalone \mail mark\ (ADR-106 A1.4)


satur8d:
## Premise (verified at @mmnto/cli 1.99.0, main `c2c33528`)

Mail consumption is not atomic with reply. `totem mail reply <source>` (`packages/cli/src/commands/mail.ts:1251` `mailReply`) reads the source dispatch, infers recipient/subject, and writes the reply — but never marks the source as consumed (no copy to `processed/`). There is also no standalone mark command: the `mail` subcommand surface is `send` / `reply` only. The consuming seat must remember to hand-copy the source into `.totem/orchestration/<seat>/processed/` as a separate manual step, and the two live lc repros show exactly that step being dropped — the dispatch then re-surfaces as phantom-unread on every subsequent `totem mail` poll.

ADR-106 A1.4 names consumption-marking as the consumer's obligation; the mechanism to make it atomic with the reply action is unbuilt.

## Ask (lc-codex's recommendation, concurred)

1. `totem mail reply` marks the source dispatch processed **atomically** with writing the reply (copy source → `processed/` in the same command; `--no-mark` opt-out for stage-only workflows).
2. A standalone `totem mail mark <source>` for consume-without-reply (read-and-acted-elsewhere), so the manual `processed/` copy has a binary-guaranteed path (same self-resolving discipline as `ecl-gc`: a seat can only mark into its OWN `processed/`).

## Provenance

lc-codex DX batch (leverage item 1 of 4), routed via strategy-claude dispatch 2026-07-16 0728Z; two live repros in liquid-city. Premise re-verified against source at triage time.

Related: #2204 (seat-selector + documented consume path), mmnto-ai/totem-strategy#700 (ecl-gc / processed-cursor), ADR-106 §A1.4. Broadcast marks land in `processed/_broadcast/` (`packages/cli/src/commands/mail.ts:326-328`) — `mail mark` must handle both shapes.

Issue #2323, opened 2026-07-08
feat(ecl-gc): mass-collection sensor — loud advisory when one compact run would collect most of the cursor (wipe-shaped-but-correct class)


satur8d:
## Incident anchor

liquid-city's first-ever `ecl-gc --apply --compact` (1.90.0, 2026-07-06 seal; regression report dispatched 2026-07-08, lc#707 context) collected **100% of the seat's marks** in one run. Triage verdict: the collection was CORRECT by the A2 contract (every mark was inert — dispatches sender-pruned past the §4.4 horizon; clean-room 1.90.0 repro + healthy strategy-claude control 299/299 + the A2.4 falsifier all exonerate the compactor). But the run was *wipe-shaped*, and it took a HIGH-priority cross-agent regression report plus a multi-hour forensic pass to establish that after the fact.

## Gap

A compact that collects a large fraction of the cursor in one run is either (a) correct first-run/backlog janitorial work, or (b) the exact signature of the failure class the gate exists to prevent (blind poll → everything reads inert). The tool currently cannot tell the operator which — it prints counts and exits 0. A human-attention signal at the moment of mass collection would have converted this incident from "forensics after a counterpart's regression report" to "read the run output."

## Ask

A **mass-collection sensor** on `eclCompact`: when `collectable.size / marks.size` exceeds a threshold (e.g. ≥ 90%, or any first run against a cursor with no prior compact record), emit a loud advisory naming the fraction and the § 4.5 inert-mark rationale.

Design constraints (from the shipped gate mechanics):

- **Must NOT ride the `warnings` array** — `gateComplete` arms on `warnings.length === 0`, so an advisory there would self-block every legitimate first compact. It needs its own result field + a prominent text/JSON surface (same class as the A2.4 `resurfaced` field: sensor output, not gate input).
- Sensor, not actuator (Tenet 13): advisory only in the default shape. Whether a `--force-mass`-style opt-through should gate the ≥ threshold case under `--apply` is an OPEN question — it trades Tenet-13 purity against the wipe-shaped blast radius; needs an owner ruling.
- Dry-run should carry the same advisory so the operator sees it before `--apply`.

## Relates

- #700 (ecl-gc remainder), #2309 (A2 gates), #2311 (collision sensor — sibling warn-tier sensor on the same machinery)
- Signoff-skill phrasing refinement filed separately (the "don't block the seal on 1/3" line).

satur8d:
[...]

Issue #2204, opened 2026-06-19
mail CLI: add a seat-specific --self/--to selector + document the skill-owned consume path (multi-seat repo ergonomics)


satur8d:
## Problem

The `totem mail` CLI under-exposes two things that exist in the mechanism but not the front-door, making multi-seat repos awkward to operate by hand. Surfaced by **lc-codex** (2026-06-19) during an lc blind-round reply, sanity-checked by strategy-claude (the doctrine is correct; this is CLI exposure/docs, not a doctrine change).

### 1. No seat-specific self selector

`totem mail --json` resolves `selfAgents.source = "dirs"` → **every** agent dir in the repo is treated as "self." From `liquid-city` that returned `["lc-agy", "lc-claude", "lc-codex", "lc-gemini"]` and listed mail addressed to `lc-claude`, not just the querying seat. Fine for a repo-level dashboard, surprising when the operator asks **one seat** to "check for a dispatch."

The seat-specific mechanism **already exists** — `TOTEM_SELF_AGENT` env (the SessionStart hooks call `pollMail({ env: { TOTEM_SELF_AGENT: <seat> } })`, and `mmnto-ai/totem-strategy:scripts/poll-cohort-mail.mjs` takes `--to <seat>`) — but `totem mail` surfaces no `--self <seat>` / `--to <seat>` flag.

### 2. No consume command, and no docs pointer saying so

`totem mail --help` exposes show/send/reply but no consume/mark-processed op. This is **intended** — processed markers are written by the consuming tooling (`/signoff`, the handoff CLI) at consumption time, holding the single-writer invariant ([ADR-106 Amendment 1 §A1.3](https://github.com/mmnto-ai/totem-strategy/blob/main/adr/adr-106-ephemeral-coordination-layer.md)). But the CLI doesn't *say* it's skill-owned, so a seat has to guess (lc-codex correctly avoided hand-writing a marker, but only after uncertainty).

## Asks

- [ ] Add a seat selector to `totem mail` (`--self <seat>` / `--to <seat>`), surfacing the existing `TOTEM_SELF_AGENT` resolution as a flag.
- [ ] `totem mail --help` note: processed markers are written by `/signoff` (read/send/reply only here).
- [ ] (Optional, same area) fold the known **ECL-no-derived-handled-state** gap — an inbound "unread" poll doesn't subtract sent replies, so handled-state isn't derivable (`unread = inbound − handled`).

## Not in scope

[...]

PR #2424, opened 2026-07-18
fix(mail): broadcast processed-marks are per-seat — stop the flat-union first-consumer-wins subtraction

### LINKED ISSUE #2412: fix(mail): multi-seat processed-mark union hides broadcasts — one seat's mark subtracts it for every seat in the poll's self set (first-consumer-wins; confirmed)
## Observation

strategy-claude, 2026-07-17 signon: totem-claude's 2026-07-16T2213Z seat-attribution **broadcast** (`to: broadcast`) was surfaced by strategy's SessionStart orientation poll but **not** by a live `totem mail` run in the same session. The CLI poll matches `to:` against the seat's self-agents only — the known frontmatter-routing hole class (mmnto-ai/totem-strategy#811). A seat relying on the CLI poll alone would never consume a broadcast.

## Why it matters

Broadcasts are the cohort's fan-out primitive (consumers-move announcements, contract changes). If the canonical poll surface can't see them, delivery depends on which orientation path a seat happens to run — a silent-miss class, not a degraded one (Tenet 4).

## Expected shape (sketch — design belongs to the build)

- `totem mail` treats `to: broadcast` as addressed to every seat, unioned into each seat's unread set.
- Consumption stays per-seat: each seat's `processed/` mark clears it for that seat only.
- Interaction with `TOTEM_SELF_AGENT` scoping (#2204) and consume-atomicity (#2396) should be specified together — this is a sibling of the lc-codex mail-semantics batch (#2396–#2399).

## Provenance

Bus-routed fold from strategy-claude's 2026-07-17T0902Z dispatch (their observation, attributed not re-verified here).

satur8d:
Closes #2412

## Root cause (confirmed forensically on the issue)

`buildProcessedSet` drained `processed/` ∪ `processed/_broadcast/` for **every resolved self agent** into one flat basename set, and the unread filter subtracted by bare basename globally. Under multi-seat self-resolution (unscoped polls, `config.json` `host_agents`), one seat's broadcast consumption put its mark in the union and hid the broadcast from every other seat — **first-consumer-wins**. The live specimen: strategy-codex's 2026-07-16T22:48:45Z mark made the 2213Z broadcast invisible to strategy-claude's 4-seat poll while their seat-scoped SessionStart poll correctly listed it.

## Fix (the shape confirmed on-issue)

- `buildProcessedMarkCounts` replaces the flat union with a per-basename **count of distinct seats** holding a mark (either `processed/` root or `_broadcast/` — both count; the live specimen sat in the root).
- **Broadcast-named files** subtract only when the count reaches **every resolved seat** — broadcasts are per-seat consumable by design (ecl-gc keeps per-seat `_broadcast` stores), and a multi-seat poll's unread line then truthfully asserts "unread for at least one of my seats", the only claim it can make.
- **Directed mail** keeps the any-seat union (per-recipient nuances stay in the #2311 basename-collision lane, per the issue).
- **Single-seat polls** (`TOTEM_SELF_AGENT`-scoped, the recommended shape per #2204) are bit-identical: `required = max(seats, 1)` collapses to 1.
- Discriminator is the **positional `broadcast` filename token** — the same zero-I/O signal the self-priority bucket already trusts (#2144). Subtraction runs pre-parse so the scan cap isn't spent opening consumed files. Disclosed residual: a `to: broadcast` dispatch **without** the filename token stays on the union rule — no worse than before, and the ecl filename discipline makes the token standard on new dispatches (locked by a test).

## Verification

- 6 new tests in `mail.test.ts` (file now 119/119): the live-repro shape (one seat's root-mark, multi-seat poll → still unread), all-seats-marked subtraction across mixed mark locations, directed-union unchanged, single-seat bit-identity, the tokenless-legacy residual, and `includeProcessed` still returning the RAW set for compaction (ADR-106 § A2.1).
- **Live specimen flipped end-to-end:** the 2213Z broadcast is still preserved in `totem-claude`'s outbox with 2 of strategy's 4 seats holding marks. The pinned 1.101.0 unscoped strategy poll shows 4 unread (broadcast invisible); the fixed build's identical poll shows 5 unread including the broadcast.
- Full workspace suite green (10/10 tasks).

Changeset: patch `@mmnto/cli`. Durable-fix note: the mass-collection sensor (#2323) and consume-atomicity (#2396) lanes are untouched by this change.

### DIFF
--- a/.changeset/mail-broadcast-per-seat-marks.md
+++ b/.changeset/mail-broadcast-per-seat-marks.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/cli': patch
+---
+
+`totem mail` no longer hides a live broadcast from a multi-seat poll after a single seat consumes it (mmnto-ai/totem#2412, first-consumer-wins). The unread filter previously drained every resolved seat's `processed/` marks into one flat basename union — so under multi-seat self-resolution (unscoped polls, `config.json` host_agents), one seat's broadcast mark subtracted the broadcast for every other seat, silently. Broadcasts are per-seat consumable by design (ecl-gc keeps per-seat `_broadcast` stores): a broadcast-named dispatch now subtracts only when EVERY resolved seat holds the mark, so a multi-seat poll truthfully asserts "unread for at least one of my seats". Directed mail keeps the any-seat union, and single-seat (`TOTEM_SELF_AGENT`-scoped) polls are bit-identical to before. A `to: broadcast` dispatch without the positional `broadcast` filename token stays on the union rule (subtraction runs pre-parse; the ecl filename discipline makes the token standard).
--- a/packages/cli/src/commands/mail.test.ts
+++ b/packages/cli/src/commands/mail.test.ts
@@ -462,6 +462,17 @@ describe('pollMail — broadcast marks are per-seat, not union (mmnto-ai/totem#2
     expect(poll({ env: { TOTEM_SELF_AGENT: 'totem-claude' } }).mail).toEqual([]);
   });
 
+  it('duplicate seat ids are normalized — the requirement counts DISTINCT seats (GCA #2424)', () => {
+    writeOutbox('totem-strategy', 'strategy-claude', [
+      { name: BROADCAST_FILE, to: 'broadcast', subject: 'consumers move' },
+    ]);
+    writeProcessed('totem', 'totem-claude', [BROADCAST_FILE]);
+    writeProcessed('totem', 'totem-gemini', [BROADCAST_FILE]);
+    // Both distinct seats hold marks; the duplicated id must not raise the bar to 3.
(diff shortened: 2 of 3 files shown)


satur8d:
**GCA round disposition (fix in `d1316907`).**

**Duplicate self-agent ids — accepted as invariant hygiene, with one precision note.** The premise is real: neither `parseEnvAgentList` nor the config `host_agents` path dedupes, so duplicates reach the poll. The stated impact is half right — a duplicated seat inflates `selfSeatCount` AND double-drains that same seat's marks, so the broadcast subtraction stays behavior-equivalent by symmetry (a dupe raises the bar and the count in lockstep; no reachable wrong verdict was found). What IS real: redundant per-dupe directory reads, and the "seats counted = seats that can hold marks" equality being assumed rather than enforced while the per-seat rule load-bears on it. Normalized once at the poll boundary (`[...new Set(agents)]`), redundant reads gone, invariant locked by a duplicate-ids test.

Issue #1630, opened 2026-04-23
core: TotemErrorCode taxonomy audit -- coherence sweep across 23 codes and 6 subclasses


satur8d:
## Signal

Surfaced by GCA on PR mmnto-ai/totem#1629: GCA flagged a `TotemConfigError(CONFIG_INVALID)` use for a CLI-layer flag-conflict error and cited a (hallucinated) rule that `TotemConfigError` is "reserved for schema validation in core." The specific rule doesn't exist in `.gemini/styleguide.md`, and the project already uses `TotemConfigError` for CLI-layer flag conflicts in 10+ CLI files (7× in `compile.ts` alone).

The finding was declined on-PR because the specific citation was hallucinated. But GCA's substantive instinct has some truth to it: the Totem error-class taxonomy currently conflates config-schema violations with CLI-layer argument-validation errors under the same class.

## Current state

`packages/core/src/errors.ts` defines:

- `TotemError` — base class with ~23 distinct `TotemErrorCode` values
- `TotemConfigError` — subclass restricted to `CONFIG_MISSING | CONFIG_INVALID`
- `TotemDatabaseError`, `TotemCompileError`, `TotemParseError`, `TotemOrchestratorError`, `TotemGitError` — other subclasses

`TotemConfigError(CONFIG_INVALID)` is used for:

- CLI flag conflicts (`--upgrade` + `upgradeBatch`, `--force` + `upgradeBatch`, `--refresh-manifest` + `--force`)
- Schema validation failures in config loaders
- Missing orchestrator / targets / etc.

These are arguably three different failure classes sharing one error type.

## Candidate refinements

1. **New `TotemFlagError` subclass** for CLI-layer option / argument conflicts. Distinct from config-schema errors. Semantic: "your invocation is incompatible, change flags and retry."
2. **Promote `UPGRADE_*` codes into a class** (`TotemHashResolutionError`?) for hash-lookup failures (`UPGRADE_HASH_NOT_FOUND`, `UPGRADE_HASH_AMBIGUOUS`, duplicate-hash corruption). Distinct from schema / config errors. Adjacent to the new duplicate-hash detection added in PR mmnto-ai/totem#1629.
3. **Audit each `TotemErrorCode` for its natural class** — ~23 codes spread across 6 subclasses; the boundaries may not be coherent after 1.14-1.15's rapid growth.

## Tier / priority

[...]

PR #2431, opened 2026-07-18
feat(mail): consume-atomicity — reply marks source processed + standalone mail mark (ADR-106 A1.4)

### LINKED ISSUE #2396: feat(mail): consume-atomicity — \mail reply\ marks source processed + standalone \mail mark\ (ADR-106 A1.4)

satur8d:
Closes #2396

## What (ADR-106 § A1.4 — consume-atomicity)

The copy-into-`processed/` marking obligation § A1.3 puts on the recipient previously lived on **no command surface** — a consuming seat had to hand-copy the source into its cursor, and a dropped copy re-surfaced the dispatch as phantom-unread on every later poll. Two surfaces close it:

- **`totem mail reply` marks the source atomically** (on by default; `--no-mark` opts out for stage-only workflows). The reply lands first — a mark that cannot land never suppresses a reply that did — and a mark failure is loud, never a silent drop. Marking is a side-effect of the read tool, § A1.3's "never a separate ritual".
- **`totem mail mark <source>`** — the standalone consume-without-reply path, with `--agent-id` (same single-seat self-resolving precedence as `ecl-gc`: explicit flag > unambiguous resolved self > hard usage error).

A mark is a **same-basename full copy** of the source under the consuming seat's own `processed/` (directed) or `processed/_broadcast/` (broadcast) — the exact stores `buildProcessedMarkCounts` (#2424) drains and `ecl-gc` compacts. Full copy per § A1.4's falsifying test (a): the recipient retains forensic evidence after the sender sweeps its outbox. Idempotent (same-basename mark present → no-op); atomic temp+rename (§ A1.2 — no torn mark visible to the `.md`-only drain); single-writer preserved (§ A2.3 — a seat marks only its OWN subtree, never guesses the consuming seat).

## Conservative readings taken (flagged, not invented contract)

1. **Mark content + placement** aren't spelled out by A1.4; full-copy derives from falsifying test (a) + the issue's "copy source → processed/", and the directed-vs-`_broadcast/` split from the existing reader/`ecl-gc` stores.
2. **Broadcast discriminator is the parsed `to:` field** (authoritative), not the reader's zero-I/O filename-token proxy — the reader drains both stores per seat, so subtraction is unaffected either way; the split keeps `ecl-gc`'s per-store compaction honest.
3. **Reply marks into the replying seat** (`result.header.from`). No recipient-vs-`to:` validation added — judgment beyond the contract.
4. **Error code reuses `MAIL_SEND_FAILED`** (mail-actuator family, mark-specific message text) rather than adding a core `TotemErrorCode` — keeps the change cli-scoped; the taxonomy audit lane is #1630.

## Verification

- 11 new tests in `mail.test.ts` (120 → 131): directed subtract-after-mark, broadcast `_broadcast/` placement, idempotency, single-writer own-cursor, missing/unparseable/ambiguous hard errors, reply default-mark + full-copy, replied-to source no longer surfaces on poll, `--no-mark` opt-out, broadcast-source mark under a directed reply.
- Full workspace gates: `pnpm -r build` PASS; `pnpm test` PASS (3346 tests, 10 tasks); repo-wide `format:check` PASS; `totem lint --base main` PASS (freeze-era stale-manifest warning only); ESLint 0 errors.

Changeset: patch `@mmnto/cli`.

Provenance: implemented by an Opus 4.8 builder under totem-claude orchestration (2026-07-18 quota-stretch directive); diff, contract conformance, and gates reviewed by totem-claude before push.


satur8d:
**Amendment `1a709700` — NTFS colon/ADS sanitization for the mark path (sensor-trust class, #2437).**

Mid-review, the cohort live-verified a Windows defect that lands on this PR's new writer: a colon in an ECL filename is an NTFS Alternate Data Stream separator — a mark written under a colon-bearing legacy source basename "succeeds" into a 0-byte base file with the bytes in an invisible stream `readdirSync` never lists, so the mark exists yet never subtracts and the dispatch reports UNREAD forever, while `mail mark` reports success. Rather than ship a fresh instance of the trap, the fix rides here:

- **`sanitizeEclBasename`** (colon-strip only, all platforms, idempotent — deliberately narrow so the reader's positional-token bucketing, which keys on the raw filename, is untouched): the single normalization seam.
- **`markSource` stores under the sanitized basename** (idempotency check included).
- **Two-sided read normalization:** `buildProcessedMarkCounts` keys by sanitized mark basename **with per-seat dedup** (a legacy colon mark plus its sanitized heal in one seat's stores counts as ONE seat — load-bearing for the per-seat broadcast requirement), and `pollMail`'s subtraction looks up the sanitized inbound name. Net: sanitized marks subtract colon-bearing inbound names, and pre-existing corrupted marks heal on the next poll.
- **`ecl-gc` compaction comparison normalizes both sides** — without this, the compactor would read a healed mark as inert against its raw-named live inbound and collect the very mark it must retain. No-op in the colon-free common case (59 existing ecl-gc tests unchanged).
- **Emitter audit: already colon-free** — `compactStamp` drops the `HH:MM` colon (mail.ts:1162), `fileToken`/`slugify` allowlists exclude `:`. Now regression-locked with a seconds-bearing-ISO test.

Known residual (accepted): two distinct on-disk names that sanitize identically (`…T05:10Z-x.md` vs `…T0510Z-x.md`) share one mark — the basename-collision class already tracked at #2311.

Gates on the final commit: build PASS; full workspace tests 3351 (+5) PASS; `format:check` PASS; `totem lint --base main` 0 errors; ESLint 0 errors.

Tracking: #2437 (item 3 there — a doctor/mail advisory on colon-bearing orchestration files — remains its residual). Provenance: amendment built by the original Opus builder under totem-claude orchestration; diff and contract reviewed by totem-claude before push.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round disposition (fixes in `9ccc3f52`)

**Accepted — 8 of 10, every premise verified against source before fixing:**

1. **Mark-failure distinguishability + anti-retry (greptile P1 + its 4/5 prose, convergent — the round's headline) — FIXED.** A post-send mark failure now rethrows with the landed reply's path in the message and an explicit "do NOT retry mail reply — run `totem mail mark <source>`" hint (code stays `MAIL_SEND_FAILED`; message text is the discriminator). Kills the duplicate-reply-on-retry hazard for automated consumers. Test locks the message, the hint, and the reply's presence on disk.
2. **Broadcast classification on the sanitized basename (CR @623) — FIXED, verified load-bearing.** The `T\d{4}Z` stamp regex rejects colon-bearing stamps, so an un-normalized colon-bearing BROADCAST fell to the any-seat rule — the #2412 first-consumer-wins class resurrected for that name shape. `isBroadcastNamed` now classifies on `sanitizeEclBasename(file)`. The second `stampPrefix.exec` site (self-priority bucketing) had the same shape — a colon-bearing self/broadcast name would mis-bucket into the low-priority tail and could fall past the MAX_SCAN horizon on a colon-legal FS — normalized identically (ordering-only; delivery truth stays the parsed `to:`). Every-seat test added.
3. **A2.4 resurfacing comparison (CR sibling site) — VERIFIED RAW, FIXED.** The `resurfaced` computation compared raw verify-poll basenames against raw on-disk marks; now both sides normalize, matching the A2.1 retention comparison. ecl-gc's 59 tests unchanged-passing (no-op for colon-free).
4. **`markSource` repoRoot parity (CR @1480) — FIXED.** Now `resolveTotemRepoRootSync`, same as `pollMail`/`eclCompact`; a subdirectory invocation lands the mark in the real root's store instead of a phantom `<subdir>/.totem`. Subdir test added.
5. **Directed-mark ownership (CR @1532) — FIXED.** Schema verified first: `to:` is single-valued everywhere (`parseHeader` captures one scalar; no list form exists in the codebase), so equality — not membership — is correct. A directed dispatch whose recipient ≠ the resolving seat is now a hard refusal with the `--agent-id <recipient>` hint; broadcast exempt. Reject + accept tests.
6. **`noMark` spread leak (GCA) — FIXED** (destructured off before the `mailSend` spread).
7. **Test/prose batch (CR ×3) — FIXED:** pre-reply unread assertion (CR's own shape); changeset precision — "atomically" → same-command sequential fail-loud, and the healing claim corrected to "superseded on re-mark; polling rewrites nothing"; Commander wiring tests (5, new file). **Disclosed limitation on the wiring tests:** `index.ts` builds its program at module scope and auto-runs `parseAsync` on import, so the live program cannot be imported without executing the CLI — the tests lock the exact negation-translation and dispatch semantics via a wiring-shape replica with handlers stubbed. Making the program importable is an index.ts refactor out of this PR's scope.
8. **Race-safe publish (CR heavy-lift + greptile P2) — PARTIAL:** unique per-invocation temp (`.{pid}.tmp`) removes the shared-temp spurious-failure case, and a rename failure with the destination now present returns `alreadyMarked: true` (both writers carry identical bytes — the copy of the same source — so overwrite/no-op are equally benign).

**Declined — the remainder of the concurrency asks, with rationale:** no locks or no-clobber machinery. Single-writer per seat is the ADR-106 § A2.3 contract, greptile's own P2 concedes unreachability under it ("cannot occur in practice today"), and the identical-content property means the only observable artifact of an out-of-contract double-launch is a cosmetic `alreadyMarked:false` double-report — accepted residual, stated here so it isn't rediscovered as a surprise.

**Gates on `9ccc3f52`:** full workspace build PASS; `pnpm test` PASS (10/10 tasks; cli 3351 → 3361, files 144 → 145); `format:check` PASS; workspace ESLint EXIT 0 (zero findings in changed files); `totem lint --base main` PASS 0 errors (freeze-era warnings only).

Round is at quiescence from my side — no re-invokes per the quota-wall discipline; merge on the operator's word.

Issue #2308, opened 2026-07-06
feat(core): generic fallback chunker — explicit-opt-in strategy for file types without a dedicated chunker (unblocks Rust/GDScript retrieval pre-AST)


satur8d:
## Summary

Add a **generic, language-agnostic fallback chunker** as a built-in strategy, registered via `chunker-registry.ts` `register()`, selectable **only by explicit opt-in** in `totem.config.ts`. It gives consumers *some* retrieval on file types that have no dedicated chunker yet (Rust, GDScript, …) instead of the current all-or-nothing: either omit the target (no retrieval) or point it at a mismatched strategy.

**Strategy → totem upstream-feedback**, authored by `strategy-claude` on behalf of `liquid-city`. Two-seat demand (`lc-claude` + `lc-agy` independently converged) plus a standing operator directive (the Opus-era retrieval push). LC has an interim stopgap merged (a code-map generator with a CI freshness gate, `mmnto-ai/liquid-city#712`) that this upstream feature lets them retire.

## Motivation

`liquid-city` needs Rust + GDScript files indexed for retrieval (see its `totem.config.ts` targets, lines 67-74). The dedicated chunkers for those languages are not yet shipped:
- Rust AST chunker — `mmnto-ai/totem#1537`
- Godot/GDScript pack — `mmnto-ai/totem#1782`
- `type: 'reference'` content type — `mmnto-ai/totem#1596`

Until those land, a consumer wanting retrieval over an unsupported file type has no good option. `createChunker()` is fail-loud by contract (Tenet 4) — an unregistered strategy throws:

```ts
// packages/core/src/chunkers/chunker.ts
export function createChunker(strategy: string): Chunker {
  const Ctor = lookup(strategy);
  if (!Ctor) {
    throw new Error(
      `Unknown chunk strategy: '${strategy}'. The strategy is not registered — either install the pack that provides it or correct the strategy name in totem.config.ts.`,
    );
  }
  return new Ctor();
}
```

So today the consumer's only choices are: (a) don't declare the target → zero retrieval on those files, or (b) point the target at a semantically-wrong built-in → bad chunks. A generic fallback closes that gap with *good-enough* chunks now, superseded per-language by the dedicated chunkers later.

## Proposed change

[...]

Issue #1645, opened 2026-04-23
cli: appendTelemetry uses cwd instead of resolvedTotemDir — same class as #1644 fix


satur8d:
## Summary

`appendTelemetry` in \`packages/cli/src/utils.ts\` writes LLM-call telemetry to \`path.join(cwd, totemDir, 'temp')\`. When the caller runs a command from a sub-directory and passes an explicit \`configRoot\`, telemetry lands in the wrong location (relative to \`cwd\`, not to the config root).

CR flagged this class during mmnto-ai/totem#1644 review for the regex-evaluation telemetry writer in \`run-compiled-rules.ts\`. That one shipped fixed in PR #1644. The \`appendTelemetry\` helper in \`utils.ts\` has the same pattern for the LLM-telemetry path and was scoped out of that PR to avoid broadening surface on an unrelated code path.

## Scope

- \`packages/cli/src/utils.ts\` \`appendTelemetry\` signature / callers.
- Every call site that passes \`cwd\` + \`totemDir\` should instead pass (or resolve) \`configRoot + totemDir\` so telemetry lands next to \`compile-manifest.json\`.

## Severity

Tier-3 — affects operators who run \`totem\` from a sub-directory (uncommon but legitimate). Pre-existing behavior, no regression; the regex-telemetry write in PR #1644 now uses the correct path and is the forward pattern to match.

## Acceptance

- [ ] \`appendTelemetry\` accepts \`configRoot\` (or takes a single resolved path) instead of \`cwd + totemDir\`.
- [ ] All call sites updated.
- [ ] Regression test confirms telemetry lands under \`configRoot/totemDir/temp\` when \`configRoot !== cwd\`.

## Source

Flagged by CodeRabbit on PR mmnto-ai/totem#1644 at \`packages/cli/src/commands/run-compiled-rules.ts:244\` (sibling-instance call-out).

Issue #2428, opened 2026-07-18
chore(core): consolidate the twin hand-rolled glob matchers (rule-engine + spine) — picomatch declined with rationale (kimi cold-read item 3)


satur8d:
## Origin

The 2026-07-18 browser-kimi cold read (first out-of-distribution rater; verbatim + dispositions at mmnto-ai/totem-strategy#916) flagged the hand-rolled glob matcher's **ownership cost** vs adopting picomatch, while crediting the tests and edge-case comments. Routed to this repo's lane for an accept/decline-with-rationale on the record — this issue is that record.

## Disposition: decline the dependency, accept the consolidation

**Declining picomatch (for now):**

1. The rule-engine matcher (`packages/core/src/rule-engine.ts` `matchesGlob`) evaluates `fileGlobs` from compiled rules — repo-controlled but pipeline-generated input. A deliberately **bounded dialect** with its own tests is a smaller reasoning surface than a full glob engine, the same posture that has `safe-regex2` gating rule regexes. Kimi itself credited the tests + edge-case comments — the cost side of the ledger is real but currently paid.
2. Behavior migration risk is asymmetric: a full engine's edge semantics (`extglob`, brace expansion, dot-file defaults) differ from the bounded dialect; a silent widening of what a rule's `fileGlobs` matches changes enforcement scope without any rule changing on disk.

**Accepting the consolidation half — the fact that strengthens kimi's point:** there are TWO hand-rolled matchers, not one:

- `packages/core/src/rule-engine.ts:25` `matchesGlob`
- `packages/core/src/spine/selection-rule.ts:89/133` `globToRegExp` + `matchGlob`

Twin implementations of the same dialect can drift (the divergent-twin class #2413 hit with `isBoundedOwnedFile`). Core also already ships `glob@^11` (with minimatch inside) for FS walking, so "zero-dep" is not the argument — bounded-dialect determinism is.

## Ask

Consolidate the two matchers into one shared, dialect-documented helper (likely `core/src/sys/`), with the union of both test suites. Re-evaluate picomatch adoption only if the consolidated dialect's feature needs grow toward a full engine's.

PR #2434, opened 2026-07-18
fix(rules): span-aware #[cfg(test)] exemption — Rust unwrap-baseline false-positives in inline test modules

### LINKED ISSUE #2397: fix(rules): Rust unwrap baseline false-positives inside inline #[cfg(test)] modules — span-aware test exemption

The Rust unwrap baseline rule `b2c3d4e5f6a70001` (`packages/cli/src/assets/compiled-baseline.ts:406-419`, "Unwrap in production code causes unrecoverable panics") exempts test code by **file glob only**: `['**/*.rs', '!**/tests/**', '!**/*test*.rs']`. The dominant Rust unit-test idiom — an inline `#[cfg(test)] mod tests { ... }` block inside a production file — is invisible to file globs, so every `.unwrap()` inside an inline test module false-positives. There is zero `#[cfg(test)]` awareness anywhere in the packages tree (grep-verified), and the rule's regex engine is line-based, so it cannot become span-aware in place.

## Ask

Exempt `#[cfg(test)]` spans from production-only Rust rules. Cleanest shape: re-author the unwrap rule as ast-grep with a `not: inside: #[cfg(test)]`-annotated-module constraint (the tree-sitter-rust grammar already ships in `@mmnto/pack-rust-architecture`), keeping the glob exclusions for integration-test trees. Same treatment applies to any future production-only Rust rule — worth a shared helper/constraint snippet rather than a one-off.

## Provenance

lc-codex DX batch (leverage item 2 of 4), routed via strategy-claude dispatch 2026-07-16 0728Z; hit in liquid-city on `totem lint` over files with inline test modules. Premise re-verified against source at triage time.

Related: #1755 (re-author-as-ast-grep class), #2063 (warning-tier rules false-positive on standard test idioms), #1614 (analogous exemption shaping for the catch-ban rule).

satur8d:
Closes #2397

## What

Span-aware `#[cfg(test)]` exemption for production-only Rust rules, across all three engine paths (regex, bounded-regex, AST/ast-grep). `unwrap()`-class violations inside inline test modules — test code living in production `.rs` files, which is exactly why filename-glob exemption can't work — no longer false-positive; matches outside test spans still flag.

Mechanism: `isProductionRustRule` (a rule targeting `.rs` that excludes or doesn't target test files, with the `fileGlobs`-entries-may-be-objects guard) gates a dependency-free scanner, `getRustTestSpans`, that locates `#[cfg(test)] mod` blocks by brace matching with comment/string/raw-string/char-literal skipping. Span reads fail toward **flagging** (a read failure yields no spans, so the rule still fires — never toward suppression).

## Two commits, two provenances

- `26220ca0` (totem-gemini): the feature — detection heuristic, scanner, three engine-path exemptions, regex + bounded e2e tests, changeset.
- `b287f9ac` (totem-claude review fixes):
  1. **Lifetime-aware scanner** — the char-literal skip treated a lifetime tick (`'a`, `'static`) as a string opener; two lifetimes could swallow the braces between them, corrupting depth tracking into an over-long span. Over-exemption suppresses REAL violations after the test module — the unsafe direction for a lint gate. Now a `'` is a char literal only when it closes as one; otherwise only the tick is consumed. Regression test added (lifetime-laden test mod + production `unwrap` after it, asserted outside every span).
  2. **Removed the `lessonHash === 'unwrap-ast-rule'` production hardcode** and the AST e2e test that only passed through it (the fixture put Rust source in a `.ts` file under `.ts` globs — a path no real rule can reach). Disposition note left in-file.
  3. **Per-file span hoisting** in both AST branches — was per-match, with a file re-read per match in the tree-sitter branch.
  4. **`pub` / `pub(crate)` visibility support** before `mod` (a `#[cfg(test)] pub mod` is still a test module; previously unspanned → residual FPs).
  5. **Wind-tunnel call threads `cwd`** into `applyRulesToAdditions` instead of silently defaulting to `process.cwd()` (the #1645 class).

## Disclosed residuals (not blockers, stated so they aren't discovered as surprises)

- **AST-path exemption has no e2e test:** an honest one needs the Rust grammar registered, which core's test env doesn't have (the #2308/#2387 language-pack lane). The AST paths share the two unit-tested helpers and the same exemption predicate as the e2e-tested regex/bounded paths.
- **Wind-tunnel era-mismatch:** the sync span read uses the current tree, while replayed diffs carry historical post-images. A missing/changed file degrades to no-exemption (fails toward flagging). If replay fidelity ever matters here, #2189 is the lane.
- **`isProductionRustRule` is name-heuristic** (glob substrings `test`/`tests`); a production rule that deliberately targets test files by another naming convention would still be exempted inside `#[cfg(test)]` spans — which is the semantically correct direction for the rule classes this serves.

## Verification

Full workspace gates on the final commit: `pnpm -r build` PASS; `pnpm test` PASS (10/10 tasks, 144 CLI files); repo-wide `format:check` PASS; `totem lint --base main` PASS (0 errors; warnings are the freeze-era + heuristic-lesson-on-scanner-code class, same set the branch's first push cleared).

Changeset: patch `@mmnto/totem` (core).

Provenance: built by totem-gemini, review-fixed and gated by totem-claude (2026-07-18 quota-stretch cohort distribution).

### DIFF
--- a/.changeset/rust-cfg-test-span-aware-exemption.md
+++ b/.changeset/rust-cfg-test-span-aware-exemption.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/totem': patch
+---
+
+Add span-aware `#[cfg(test)]` module exemption for production-only Rust rules to eliminate false-positives inside inline unit test modules.
--- a/packages/core/src/regex-safety/apply-rules-bounded.ts
+++ b/packages/core/src/regex-safety/apply-rules-bounded.ts
@@ -92,7 +92,7 @@ export async function applyRulesToAdditionsBounded(
       const spans = getRustTestSpans(content);
       rustTestSpansCache.set(file, spans);
       return spans;
-      // totem-context: intentional cleanup
+      // totem-context: span-read failure yields no spans → no exemption → the rule still FIRES; the exemption fails toward flagging, never toward suppression.
     } catch {
       rustTestSpansCache.set(file, []);
       return [];
--- a/packages/core/src/regex-safety/apply-rules-bounded.ts
+++ b/packages/core/src/regex-safety/apply-rules-bounded.ts
@@ -13,6 +13,9 @@
  * compound-rule pipeline under separate bounds.
  */
 
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
 import type {
   CompiledRule,
   DiffAddition,
@@ -22,6 +25,8 @@ import type {
 import { TotemParseError } from '../errors.js';
 import {
   extractJustification,
+  getRustTestSpans,
+  isProductionRustRule,
   isSuppressed,
   matchesGlob,
   type RuleEngineContext,
@@ -78,6 +83,22 @@ export async function applyRulesToAdditionsBounded(
 
   const regexRules = rules.filter((r) => r.engine === 'regex' || !r.engine);
 
+  const rustTestSpansCache = new Map<string, { startLine: number; endLine: number }[]>();
+  const getRustSpans = async (file: string) => {
+    if (rustTestSpansCache.has(file)) return rustTestSpansCache.get(file)!;
+    try {
+      const fullPath = path.resolve(options.repoRoot, file);
+      const content = await fs.promises.readFile(fullPath, 'utf-8');
+      const spans = getRustTestSpans(content);
+      rustTestSpansCache.set(file, spans);
+      return spans;
+      // totem-context: intentional cleanup
+    } catch {
+      rustTestSpansCache.set(file, []);
+      return [];
+    }
+  };
+
   for (const rule of regexRules) {
     // Partition additions by file so the evaluator can batch one rule per
     // file at a time. File granularity matches the fileGlobs scoping and
@@ -131,6 +152,15 @@ export async function applyRulesToAdditionsBounded(
       for (const matchedIndex of result.matchedIndices) {
         const addition = fileAdditions[matchedIndex];
         if (!addition) continue;
+
+        // Exempt matches inside inline Rust test modules for production-only Rust rules
+        if (isProductionRustRule(rule)) {
+          const spans = await getRustSpans(file);
+          const isExempt = spans.some(
+            (s) => addition.lineNumber >= s.startLine && addition.lineNumber <= s.endLine,
+          );
+          if (isExempt) continue;
+        }
 
         if (isSuppressed(ctx, addition.line, addition.precedingLine)) {
           onRuleEvent?.('suppress', rule.lessonHash, {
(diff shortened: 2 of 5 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round disposition (fixes in `a1b330ff`)

**Accepted — 5 (2 substantive, both premise-verified before fixing):**

1. **Escaped-backslash string termination (greptile P1) — FIXED.** Confirmed: `content[i-1] !== '\\'` mis-reads a string ending in an escaped backslash (`"...\\"`) as unterminated — the scan over-reads, swallows braces, and over-extends the test-module span into production code (over-exemption, the same unsafe direction as the lifetime bug fixed in `b287f9ac`). Now counts the consecutive-backslash run before a candidate quote — odd = escaped, even = terminator. Regression test: `"\\"` inside a test mod, production `unwrap()` after it asserted outside every span.
2. **Nested block comments (the concrete half of CR's Major) — FIXED.** Rust block comments nest; both `indexOf('*/')` skips stopped at the first terminator, leaving an unconsumed tail whose `}` corrupts depth. Both loops now use a depth-counting comment consumer. Regression test with a `}` inside a nested comment.
3. **Span-exemption telemetry (greptile P2) — FIXED** at all four exemption sites (sync regex, bounded, tree-sitter, ast-grep): each emits `onRuleEvent('suppress', …, justification: 'exempt: inline #[cfg(test)] module span (#2397)')` mirroring the `totem-ignore` suppress shape, so metrics distinguish "matched but exempt" from "never matched". Event-emission test added.
4. **CodeQL dead stores (×2) — FIXED**; the write-only `hasSemicolon` went with them (its only reader was a removed dead branch). Control flow unchanged — the `;`/`{` inner-loop breaks and the unconditional outer `break` are preserved.
5. **Await async `dispose()` (CR Minor) — FIXED.** Verified `async dispose(): Promise<void>` at `regex-safety/evaluator.ts:183`; the bounded test's `finally` now awaits it.

**Declined — 2, with rationale:**

- **"Use a lexically aware Rust scanner" (CR Major, the rewrite half).** The bounded hand-rolled scanner with a growing edge-case regression suite is this repo's deliberate posture for exactly this class of surface — #2428 declined picomatch on the same bounded-dialect-determinism reasoning, on the record. CR's own Python probes surfaced two concrete failure classes; both are fixed above (items 1–2) and regression-locked. A tree-sitter-grade lexer is also not available to core (no Rust grammar in core's env — the same #2308/#2387 gap that shaped this PR's test strategy, disclosed in the body).
- **"Resolve spans from the caller's post-image content" (CR outside-diff Major).** This is the wind-tunnel era-mismatch residual **already disclosed in this PR's body** ("Disclosed residuals" § 2): the sync span read uses the current tree while replay evaluates historical post-images, and the degradation direction is no-exemption — it fails toward flagging, never toward suppression. Threading a readStrategy through `applyRulesToAdditions`' span resolution belongs to the diff-replay wire-up; pointer posted on #2189 (its lane) citing this finding, so it is carried, not dropped.

**Gates on `a1b330ff`:** full workspace build PASS; `pnpm test` PASS (10/10 tasks; core 3229 → 3232); `format:check` PASS; workspace ESLint PASS 0 errors; `totem lint --base main` PASS 0 errors (freeze-era warnings only).

Round is at quiescence from my side — no re-invokes per the quota-wall discipline; merge on the operator's word.

Issue #2329, opened 2026-07-09
fix(review): secret-redaction rewrites env-var reads to REDACTED before the LLM sees the diff — mints false CRITICAL ReferenceError findings


satur8d:
## Exhibit

\	otem review --base main\ on the #2325 branch returned:

> CRITICAL [1] tools/vp-pr-cut-sensor.mjs:231 — The variable \REDACTED\ is not defined, which will cause a ReferenceError at runtime. The token should be read from the environment (e.g., \process.env.GITHUB_TOKEN\) as indicated by the usage comment.

Actual source at the flagged lines (verified, and the script executed live twice in the same session):

~~~js
const token = process.env.GITHUB_TOKEN;
const repo = process.env.GITHUB_REPOSITORY;
~~~

The reviewer's own remediation text names \process.env.GITHUB_TOKEN\ — i.e. the model was shown a diff where the identifier had already been replaced with \REDACTED\ by the pre-send sanitizer, then correctly reasoned the placeholder was undefined. The finding class is a sanitizer seam artifact, not a code defect, but it renders as CRITICAL and fails the review exit code — a false-positive gate red on any tools script that legitimately reads a token env var.

## Ask

- Redact secret VALUES, never identifier/env-var NAMES in code diffs (\process.env.GITHUB_TOKEN\ is not a secret; its runtime value is).
- Or: mark redacted spans in the prompt so the model knows \REDACTED\ is sanitizer output, not source text.

Sibling class: #1637 (CRITICALs for symbols outside diff hunks) — both are reviewer-input-fidelity defects where the model is judged on text that is not the source.

umudhasanli:
This is an important distinction for source diffs: identifier names are evidence, secret values are the sensitive part. A source-code policy should not reuse a generic log policy.

Flare Redact can be configured for this boundary by disabling assignment-name heuristics while keeping format/content detection:

```ts
import { compilePolicy } from "flare-redact";

const sourcePolicy = compilePolicy({
  redactKeys: false,
  disable: ["generic_assignment"],
  enable: ["high_entropy"],
  refineConfidence: true,
  minConfidence: 0.6,
});

const safeDiff = sourcePolicy.redact(diff);
```

That preserves `process.env.GITHUB_TOKEN` and similar identifiers, while still masking distinctive provider tokens, PEM blocks, JWTs, credentials embedded in URLs, and high-confidence unknown literals. The confidence refinement specifically guards UUID/SHA/digest-like code noise.

[...]

PR #2443, opened 2026-07-18
feat(review): exclude generated-artifact bytes from LLM synthesis input — inject summary + semantic hash, not silent drop

### LINKED ISSUE #2398: feat(review): exclude generated-artifact bytes from LLM synthesis input — inject summary + semantic hash, not silent drop

`totem review` sends generated-artifact bytes to the LLM as part of the review diff. The only exclusion mechanism is config `ignorePatterns` (applied in `getDiffForReview`, see `packages/cli/src/git.test.ts:310-328`): opt-in, empty by default, and a **silent drop** — an excluded file simply vanishes from the payload, losing the signal that it changed at all. Nothing classifies generated artifacts (lockfiles, `compiled-rules.json`, `dist/**`, `*.wasm`, regenerated dashboards) by default, so a PR that regenerates artifacts burns review-context tokens on bytes no reviewer should read — and in liquid-city's case dominates the synthesis input.

## Ask (lc-codex's recommendation, concurred)

1. A default generated-artifact classification for the review payload (seeded globs + honor `linguist-generated` / `.gitattributes` where present).
2. Excluded-but-changed files are **summarized, not silently dropped**: inject a per-file line into the synthesis input — path, change shape (regenerated/added/deleted), size delta, and a semantic hash — plus the gate summary when a deterministic gate (e.g. regeneration check) covers the artifact. The reviewer keeps the "this regenerated" signal without the bytes.

This aligns with the Prop 297 taxonomy (GENERATED spans take regeneration-diff treatment, never prose review) — the review payload should route classes the same way (mmnto-ai/totem-strategy#639 operating spec).

## Provenance

lc-codex DX batch (leverage item 3 of 4), routed via strategy-claude dispatch 2026-07-16 0728Z. Premise re-verified against source at triage time.

Related: #2329 (payload rewriting before the LLM — the masking seam this would compose with), #1637 (findings outside diff hunks).

satur8d:
Closes #2398

## What

`totem review` no longer feeds generated-artifact BYTES to the LLM — a new Stage 0.5 (between diff resolution and file classification) classifies generated artifacts, strips their diff sections, and injects a per-file **summary** — path, change shape, size delta, semantic hash — in a distinctly-tagged block, so the "this regenerated" signal survives without the bytes and nothing is silently dropped (every exclusion is also named loudly on stderr).

**Classification:** a changed file matching any seeded default glob (lockfiles, `.totem/compiled-rules.json` + manifest, `dist`/`build`/`out`/`.next`/`coverage`, `*.wasm`/`*.min.js`/`*.min.css`/`*.map`/`*.tsbuildinfo`) OR a repo `.gitattributes` `linguist-generated` pattern, minus `linguist-generated=false`/`-linguist-generated` un-marks. The Git-native attribute is both the extension point and the false-positive escape hatch — deliberately no new config knob.

**Mislabel guard (#2329's lesson):** the injected block is wrapped in its own tag with a "TOTEM-GENERATED SUMMARY, not diff content" preamble so the model cannot mistake summarization artifacts for real diff content.

**All-generated fast-path:** when every changed file is a generated artifact, the review completes deterministically (same shape as the existing all-non-code fast-path, including the reviewed-content hash write) — artifact correctness is a gate concern (`totem lint`, lockfile-sync), not the LLM's.

## Conservative readings taken (disclosed, not invented)

1. **"Semantic hash"** = sha256 (12 hex) over the file's own diff segment — the always-available, scope-consistent artifact (works for deletes/binaries); labeled as exactly that, not over-claimed as a full-content hash.
2. **Per-artifact gate verdicts:** no data source exists at payload-assembly time, so none were fabricated; the preamble names the deterministic-gate treatment generally. Concrete per-artifact verdict wiring is deferred (needs a data source outside this path).
3. **False-positive risk** on `build`/`out`/`coverage` dirs used as source by some repos: never silent (stderr + in-prompt naming with hash), and the `.gitattributes` un-mark is the remedy.

## Verification

- 27 new tests (24 in `shield-generated.test.ts`: classification, gitattributes parsing incl. un-marks, section splitting/hashing, summary building; +3 prompt-injection integration tests in `shield.test.ts`).
- Full workspace gates: build PASS; `pnpm test` PASS (cli 3376 → 3403, 10/10 tasks); `format:check` PASS; workspace ESLint PASS 0 errors; `totem lint --base main` PASS 0 errors.

Changeset: patch `@mmnto/cli`.

Provenance: implemented by an Opus 4.8 builder under totem-claude orchestration (2026-07-18 quota-stretch distribution); classification definition, integration points, fast-path semantics, and gates reviewed by totem-claude before push.


satur8d:
/gemini review
@greptileai review

satur8d:
/gemini review
@greptileai review

satur8d:
## Round disposition (fixes in e08af623)

**Both distinct findings accepted (GCA posted the first twice); both verified at source before fixing. Greptile's 5/5 is on record as the clean counterpart.**

1. **Newline-less section truncation (GCA high ×2) — FIXED.** Confirmed: \section.slice(0, section.indexOf('\n'))\ on a section with no newline (a truncated diff tail) feeds -1 into slice, chops the final character, and breaks the \$-anchored filename match — the section would fall out of classification. Whole-section fallback added; regression test uses exactly that shape (\diff --git a/dist/out.js b/dist/out.js\ with no trailing newline → filename parses intact).
2. **Bare \+++\/\---\ prefix check swallows real hunk content (GCA high) — FIXED.** Confirmed: an added line like C's \++i;\ renders \+++i;\ in a hunk and was skipped as a header, undercounting the summary's size-delta stats (symmetric for \---x;\). Headers always carry a space after the triple marker (\+++ b/...\, \+++ /dev/null\) — now the discriminator. Regression test locks both directions (content counted, headers still skipped; the /dev/null header shapes in the existing fixtures keep passing).

Gates on e08af623: full workspace build PASS; pnpm test PASS (10/10 tasks); format:check PASS; workspace ESLint exit 0 (pre-existing warnings only, none in changed files); totem lint --base main PASS 0 errors.

Round at quiescence — no re-invokes per the quota-wall discipline; merge on the operator's word.

Issue #2045, opened 2026-05-30
research: totem review redesign — grounding > override-memory > model-diversity (lc#458 empirical anchor; "eventually", not committed)


satur8d:
## Provenance & status

satur8d floated redesigning `totem review` **"eventually — might just use the cohorts, probably codex or something."** This is a **musing, not a committed decision.** Filed as a tracked product item (vs. an ECL note that would archive) to preserve the empirical anchor below. Routed to the **shield/gate lane** (totem-claude). Surfaced via lc-claude dispatch `2026-05-30T0205Z` (LC `mmnto-ai/liquid-city#458` session). **Tier: exploratory / low-priority.** Nobody is committed to building this yet.

## The empirical anchor (the reason to track)

On `mmnto-ai/liquid-city#458`, the local shield re-emitted the **same false-positive 5× across 5 pushes**: it claimed `_agent_buffer` was a `PackedByteArray` (→ "offsets +3/+7 wrong, CRITICAL 0.92") when it's a `PackedFloat32Array` — **declared at line 436, four lines above the function but *outside* the diff hunks.** Each push re-emitted the identical finding despite the prior override + trap-ledger entry.

**The tell:** the *external* PR bots, reviewing the full PR, **never made that error.** Greptile validated the offsets + `.duplicate()`; GCA caught **two real bugs** lc's own 5-skeptic verification workflow missed. Same model class (shield orchestrator = `claude-sonnet-4-6` per LC `totem.config.ts`), **opposite outcome. The differentiator is full-file vs diff-scoped context, not the model.**

## Three levers (lc-claude, in priority order)

1. **Grounding (biggest win, model-agnostic).** Before emitting a finding that depends on a symbol's type/shape, resolve its *declaration* (read it, or feed a lightweight symbol/type index of touched identifiers) — not just the hunk. Kills the recurring offset/type-class of false-positives. Cheapest fix; addresses the actual failure.
2. **Override-memory.** Hash the finding's anchored code-span; if span unchanged + a prior override exists → suppress/demote. 5 identical overrides on one PR is the symptom, and it trains reflexive override (erodes gate signal).
[...]

PR #2451, opened 2026-07-19
refactor(core): consolidate glob matcher ownership

### LINKED ISSUE #2428: chore(core): consolidate the twin hand-rolled glob matchers (rule-engine + spine) — picomatch declined with rationale (kimi cold-read item 3)

satur8d:
## Mechanical Root Cause

Totem maintained separate hand-rolled glob implementations for rule enforcement and spine path classification. Their overlapping syntax had no single owner, while their historical dialect differences were implicit in caller-local code. That duplication made fixes drift-prone and left the bounded evaluator with a third copy of rule matching behavior.

## Fix Applied

- Centralized both compatibility contracts in `packages/core/src/sys/glob.ts` as explicit rule-engine and path-classifier profiles over one tokenizer, compiler, and matching loop.
- Preserved the historical dialects rather than widening either one, including negative-glob precedence and bounded rule syntax.
- Routed the rule engine, bounded evaluator, spine selection, and git diff filtering through the shared implementation.
- Added a bounded 512-entry LRU cache for the hot rule-profile path while retaining the fixed classifier-pattern cache.
- Added a patch changeset and the issue design record.

## Out of Scope

- Unifying the two historical glob dialects into a new mode-less language.
- Repairing or reinterpreting the existing frozen glob corpus.
- Replacing the deterministic implementation with a third-party matcher.
- Changing selection policy, glob configuration schema, or unrelated rule semantics.

## Tests Added/Updated

- [x] Added/updated automated tests covering the root-cause mechanism
- [ ] N/A (explain why verification logic is not applicable)

- Pinned all 255 unique signed frozen globs over 815 deterministic rendered paths, including all 39 zero-match rows.
- Covered the eight ruled divergence axes, negative precedence, zero-segment and middle globstars, slash-bearing legacy suffixes, line terminators, and bounded-cache eviction.
- Replayed the current frozen `selectionRule` cover at N=80 and the historical ADR-110 lock at N=40 under both revert modes; all pre/post memberships and applicable frozen-cover comparisons were identical.
- Independent legacy comparison found zero mismatches across the frozen corpus, crafted edge cases, and 200,000 deterministic structured comparisons.
- `pnpm test` — 10/10 workspace tasks passed.
- `pnpm -r run lint` — 0 errors (7 pre-existing CLI warnings).
- `pnpm run format`, `pnpm run format:check`, core TypeScript compilation, `git diff --check`, and `pnpm exec totem lint --base main` passed; Totem lint reported 387 rules and 0 errors.

Local `totem review` remains provider-unsettled with no override or bypass. The initial full run produced two CLI fallback `invoke-error` failures; the staged review-cleanup rerun produced one unextractable Anthropic abstention and one Gemini `invoke-error`, leaving 0/2 usable lanes. The review-lane defect is tracked separately in #2452.

## Related Tickets

Closes #2428

Related: #2452, #2045


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## `selectionRule` oracle replay

The outstanding frozen-membership condition is now recorded in `.totem/specs/2428.md` at commit `6f4db81f`.

### Current frozen cover

- corpus: offline `mmnto-ai/liquid-city`; no GitHub API
- frozen `asOfCommit`: `0878081e922db16d4fada10b8518eb6e58b62300`
- pre-swap source: `b06f9fa460bd7d22b33e3353739512d4af0f36ed`
- post-swap source: `6219a5c3dbdf0a5e343305d649ad16822027b329`
- enumeration: 408 byte-identical ordered `PrMeta` records, SHA-256 `af17a9f9dcdc6192f0079ef0bfd964700a7b79f372895cbec14d1b9293935a79`
- `excludeRevertPairs=false`: 80/80, membership SHA-256 `3433189c04fca03295263ff633af22ece795a33b5e45d3a36cc3e15c2102f2c8`, pre/post diff empty, frozen-cover diff empty
- `excludeRevertPairs=true`: 80/80, same digest, pre/post diff empty, frozen-cover diff empty

The two flag values intentionally agree because this freeze contains no excluded revert pair. Each result also equals the corresponding ADR-111 cover from the tracked split.

### Historical replay

The prior ADR-110 lock at `b5bea41e50bfc5247e3a41e436daccb908ed9fe7` (`asOfCommit=09b21bc3bf5c1933d7f2f313659d5d687a01c599`, `N=40`) produced 355 byte-identical `PrMeta` records and 40/40 byte-identical members under both revert modes. Membership SHA-256: `741549ed334ec7fd1d36a3a49e2ef81db08843f63bac15d92f7cb27368822f4b`; symmetric differences were empty. No historical ADR-111 split exists for that lock, so frozen-cover comparison is explicitly not applicable.

### Checks

- focused `selection-rule.test.ts`: 36/36 passed
- Prettier write/check: passed
- `git diff --check`: passed
- `totem lint --base main`: 387 rules, 0 errors

The separate review-lane reliability follow-up requested by strategy is filed as #2452, linked to #2045.

Ready: invoke CR / Greptile / GCA, or merge as-is. Recommendation: all three if the operator wants an external pass because this is a substantive core-code diff; no bot has been invoked by this comment.

satur8d:
## Bot round disposition

Commit `4f6f56e9` · CI green across all lanes · selectionRule oracle-replay satisfied (spec §Direct selectionRule Oracle Replay; comment 5014084089). 4 findings, 0 blockers, all in `sys/glob.ts`.

**Fixed (`4f6f56e9`):**
- **[greptile] `tokenizeGlob` neverMatch early-return** — now returns `[{ kind: 'never' }]` before token allocation/traversal. Behavior-preserving; the frozen 255×815 corpus SHA-256 pin is unchanged.
- **[CodeRabbit] `collectRuleEngineStarIndexes` annotations** — added as comments only, mapping each branch to its legacy shape (`*.`, `**/`, `/**/`, `/**`, `/*.`); no logic or control-flow change.
- **[CodeRabbit] PR description** — rewritten under the repo template (Mechanical Root Cause / Fix Applied / Out of Scope / Tests Added-Updated / Related Tickets).

**Declined, with rationale:**
- **[CodeRabbit] `fileMatchesGlobs` `typeof` string guard** — no non-string entry is reachable through a supported production flow: callers pass `CompiledRule.fileGlobs` (`CompiledRuleBaseSchema` → `z.array(z.string())`); raw `compiled-rules.json` parses via `CompiledRulesFileSchema`; stage-4 config arrays and all TS/YAML/TOML config reach `TotemConfigSchema.parse`; `.totemignore` entries are regex-captured strings. The one object case is a synthetic `as unknown as string` cast in `rule-engine.test.ts`. A silent filter would be fail-soft — an all-invalid list would default-allow — so loud validation at a public boundary is the right home if this is ever wanted, not a silent drop here.
- **[greptile] `BoundedRegexCache.set` no-op `delete`** — the `delete`+`set` is the canonical LRU move-to-end idiom: a harmless no-op on the current miss-only call path that preserves correct MRU promotion if `set` is ever reused as an upsert. Removing it would couple `set()`'s correctness to the current single call site.

**Local review lane:** unsettled this round — 0/2 usable lanes (Anthropic abstained on non-extractable output; Gemini `invoke-error`). #2452 tracks the review-lane defect; both reruns are disclosed in the PR description. No `local-lane:` covariate line to carry.

Greptile confidence 4/5 "safe to merge"; both its findings are dispositioned above. Round converged.

Issue #2466, opened 2026-07-20
bug(review): three fast-paths mint the reviewed-content stamp without reviewing — a non-review issues push authorization, not just exit 0


satur8d:
## Symptom

`totem review` skips every non-code file and **exits clean**. In a markdown-dominant repo the result is a review that examined nothing and reports it as success.

Observed 2026-07-20 in `mmnto-ai/totem-strategy`, on two separate branches:

```
[Review] Deterministic fast-path: all changed files are non-code
[Review] Skipped: doctrine/model-tiering.md
```

```
[Review] Deterministic fast-path: all changed files are non-code
[Review] Skipped: operations/world-claims-register.md
```

No verdict artifact, no findings, no `local-lane:` covariate line, exit 0. `totem review --help` shows no flag that forces the diff through — the fast-path is not overridable.

## Why this is sensor-trust class, not a feature gap

The failure is **silent-positive**. An agent runs the local lane, sees no findings, and reports "local review settled, nothing found." That statement is false in a specific and dangerous way: nothing was examined. Distinguishing the two required reading the skip lines rather than the exit code.

This is the same shape as a coordination sensor reporting zero unread when it never polled — a clean exit standing in for a clean result ([Tenet 4](https://github.com/mmnto-ai/totem-strategy/blob/main/design-tenets.md#4-fail-loud-never-drift): never silently degrade; [Tenet 13](https://github.com/mmnto-ai/totem-strategy/blob/main/design-tenets.md#13-sensors-not-actuators): a sensor that cannot measure must say so, not return a default).

It also inverts a live operating posture. The bot-quota response (`mmnto-ai/totem-strategy#845`) demoted external bots to final-pass and **pulled the local review lane forward**. In `totem-strategy` — a doctrine/strategy repo whose content is essentially all markdown — the lane that posture leans on covers approximately nothing, precisely where judgment density is highest.

## Second, compounding gap

`review.lanes` is not configured in `totem-strategy` (no match anywhere in the repo's JSON). Per the `review-loop` contract, an unconfigured fan takes the legacy single-lane path, which emits **no verdict artifact and no `local-lane:` line** — so `/review-loop`'s settle condition is unsatisfiable in that repo regardless of the fast-path.

Two independent reasons the loop cannot settle there, and neither surfaces as an error.

## The fix is honesty first, coverage second

Ranked, because the cheap half closes the dangerous part:

[...]

Issue #2465, opened 2026-07-20
feat(spec): totem spec must disclose its evidence set — #2452 grounding case for the selection-envelope contract


satur8d:
## Problem

`totem spec` synthesizes confident, load-bearing premises **without disclosing the evidence set they rest on**. The contract gap is not that a spec was wrong once — it is that a consumer cannot tell a well-evidenced premise from an under-evidenced one, so wrongness stays invisible until someone independently reads the source material.

This is the grounding case for the F1 selection-envelope contract in `mmnto-ai/totem-strategy#939` §F8.

## Grounding case (#2452)

`totem spec 2452` (gemini-3.5-flash, run artifact `a4d12d22ccb5…`) produced a spec asserting that Sonnet-5 emitted **chatty wrapped verdicts** which the extraction cascade needed broadening to accept. It read as authoritative and drove a concrete task list ("introduce robust parsing regexes to strip chatty markdown wrapper text").

That premise was **wrong, and load-bearing**. An implementer following it would have widened a parser to manufacture verdicts out of refusals — degrading the very trust surface the issue existed to fix.

The real cause, found by totem-codex reading the two actual run artifacts (`925e0d2ef092…`, `be1a57f1d35e…`): `CLI_FALLBACK_COMMANDS.anthropic` passed the temporary **file path** as the `claude -p` prompt, so Claude treated the path as the prompt and refused. The outputs were refusals, not chatty verdicts. Slice B fixed the transport and kept those outputs as **negative** extraction fixtures.

**The decisive evidence existed on disk and was outside the synthesizer's selected context** — and nothing in the spec's output indicated that a selection boundary had been drawn at all.

## Why disclosure, not accuracy, is the fix

Requiring the synthesizer to "be right" is unfalsifiable and unenforceable. Requiring it to **declare what it looked at** is mechanical and checkable, and it lets a reader do exactly what codex did — notice that the run artifacts were not consulted and go read them.

## Proposed direction

- Spec output declares its **evidence set**: what was retrieved/considered and selected (the index query results, files, artifacts).
- Ideally also what was **not** looked at within reach — the honest boundary, not just the inclusion list.
- A premise not supported by anything in the selected evidence is marked **unsupported** rather than asserted in the same register as a grounded one.
- Interacts with the `search_knowledge` confidence-signal issue from this round: a spec built on noise-floor retrieval (all hits at `0.016`) should say so.

## Acceptance criteria

[...]

Issue #2452, opened 2026-07-19
bug(review): zero-completed-lane fan can exit 0 after CLI fallback abstention


satur8d:
## Problem

A `totem review` fan can finish with zero completed verdict lanes yet exit 0 when one lane is recorded as `ABSTAINED`. The command is honest in its artifact (`settled=false`, `lanes=0/2`), but the process exit reads like a pass when `--fail-on` is omitted.

The same reproductions show that CLI fallback failures do not retain enough evidence to distinguish extraction failure, authentication/quota/model errors, or child-process timeout.

This is one review-lane reliability defect family, adjacent to the broader review redesign in #2045.

## Reproductions

Observed while preparing #2451:

### Abstain plus invoke failure exits 0

An 8-file uncommitted review produced:

- Anthropic: CLI fallback returned output that the structured-verdict cascade could not extract, recorded as `ABSTAINED`
- Gemini: `invoke-error`
- usable lanes: `0/2`
- `settled=false`
- process exit: 0 because `--fail-on` was omitted

Artifact prefix: `925e0d2ef092`; lineage `42d32d18`, round 1.

A narrowed code-only staged run showed the same zero-completed-lane shape and failed only because `--fail-on critical` additionally enforced lane coverage / structured verdict requirements.

Artifact prefix: `be1a57f1d35e`; lineage `734d508c`, round 0.

### Both lanes fail invocation and correctly exit 1

```
pnpm exec totem review --staged --fresh --fail-on critical
```

For the final exact #2451 staged diff:

- 8 changed files; 56,065-character diff
- two non-code files filtered; 58KB total prompt
- Anthropic SDK unavailable, Claude CLI fallback, then `invoke-error`
- Gemini SDK unavailable, Gemini CLI fallback, then `invoke-error`
- all-lanes-failed unsettled artifact written
- hard exit 1 after 184.3 seconds

This all-failed G3 behavior is correct. The contrast isolates the exit-legibility gap to abstention being counted as output even though it is not a completed verdict.

## Expected behavior

A round with zero completed verdict lanes must not be confusable with a passing review:

- abstain-only or abstain-plus-failed fans hard-error like the existing all-lanes-failed gate, or expose an equally unmistakable non-pass exit contract
- `ABSTAINED` remains distinguishable from invocation failure in the artifact
- provider unavailability remains an honest sensor-down state, not a code verdict

## Acceptance criteria

[...]

PR #2471, opened 2026-07-20
fix(review): deterministic skip paths no longer stamp the push gate (#2466)

satur8d:
Part of #2466 — **deliberately does not close it.** See "What this does not fix" below.

Three deterministic skip paths in `shieldCommand` drop the entire diff without examining it, then call `writeReviewedContentHash(...)` before returning — so a run that reviewed nothing minted `.totem/cache/.reviewed-content-hash`.

> **Premise correction (2026-07-20).** This PR was written believing that stamp is consumed as push authorization by a PreToolUse gate. **It is not, in this repo.** #2180 (merged 2026-06-15) decommissioned `review-gate.sh`; `tools/pre-push` now runs `totem review || exit 1` live, and nothing here reads `.reviewed-content-hash`. The hole this PR closes is therefore **dormant in totem and live for downstream consumers on old hooks**. The locally-live half of #2466 is the exit code, which this PR does not change. Detail in the disposition comment; #2472 tracks the stale doc that caused the error.

## The three are not equally dangerous

This is the part that changed the fix, and it is not visible from the field report. `.claude/hooks/content-hash.sh` — shared by the stamp writer and the gate verifier — hashes **only tracked files matching the review extensions** (default `.ts`/`.tsx`/`.js`/`.jsx`).

| Path | Fires when | Effect of its stamp |
|---|---|---|
| `1725` all-non-code | no code file in diff | hash unchanged → **no-op** |
| `1743` filtered-empty | no code survives filtering | hash unchanged → **no-op** |
| `1710` all-generated | diff is all generated | **can change the hash** |

For the first two the stamp authorized nothing that was not already authorized; the defect is the dishonest clean-pass surface (Tenets 4/13). The third is different: `.gitattributes linguist-generated` can mark a tracked `.ts` as generated (#2443), so that path can drop a hashed source file, stamp the new hash, and authorize code no reviewer saw.

**That inverts the obvious fix order.** The path carrying a stated "deterministic PASS" rationale (`shield.ts:1707–1709`) is the dangerous one. Not sending generated artifacts to the LLM stays correct — their correctness is a gate concern — but stamping the push gate on their behalf does not follow from it.

Reachability, stated plainly: not reachable in this repo today (no `generated` markers in `.gitattributes`, `dist/` untracked, default globs are lockfiles and `.totem/*.json`, none in the 658-file hash set). Reachable by design in any consumer, since marking generated files is what the feature is for.

## Change

None of the three stamp. Each emits a `NON-REVIEW:` warning naming what was skipped and stating that nothing was examined and no push is authorized. The notice is single-sourced as `NO_STAMP_NOTICE` so the three sites cannot drift into describing one guarantee three different ways, and its voice matches the existing worktree-drift notices — the other surface that declines to stamp.

Precedent this follows rather than invents: `shield-estimate.ts:50` already declines to stamp on empty-diff runs, for exactly this reason ("an estimate is a forecast, not a passing review"), and there is a standing compiled rule against stamping the cache on non-reviews.

**Prose-only pushes are unaffected.** The gate compares a hash over code files a prose change does not touch, so it stays satisfied by the last genuine review. No deadlock.

## What this does not fix

**The locally-live half of #2466.** Because `tools/pre-push` runs `totem review || exit 1`, a prose-only diff hits a fast path, returns, exits 0, and the push proceeds — a non-review passing the gate through its *exit code*. That is #2466's original framing, and for this repo it was the correct one.

This PR does not change any exit code. #2466 therefore stays **open** after this merges, for the exit-contract follow-up, which should use totem-codex's two-phase model: **admission/applicability** (`not-applicable` with a deterministic reason — `no-diff`, `all-non-code`, `filtered-empty`, `all-generated`) versus **fan execution** (`completed | abstained | failed`). A deterministic skip is not a fourth lane outcome and not an `InvokeFailureKind`; mapping it into #2452 slice B's taxonomy would erase the phase boundary and imply provider work that never occurred.

## Deliberately not in scope

- **The no-changes stamp (`1659`)** has a different premise — no diff at all, rather than a diff dropped after inspection. Arguably adjacent, but it is not what #2466 reports and widening a fix PR into it invites a review round about scope instead of correctness. Flagging rather than folding in.
- **`recordShieldOverride` (`696`)** is an explicit, ledgered operator override. Correct to stamp.
- **Exit state.** Whether a total skip should exit non-zero rather than 0 changes the error surface for every docs-only run, so it wants its own ruling. The honesty half — no stamp, loud notice — is what lands here, and it is the half that closes the authorization hole.

## Verification

- **Red/green proven.** Both stamp tests fail against the previous `shield.ts` and pass after; verified by stashing only the source change and re-running. The third test (mixed code+prose must not take a skip path) passes on both — it is a regression guard against over-firing, not a bug-encoder.
- Full workspace suite unfiltered: **147 files, 3455 tests, green**.
- `pnpm -r build` · `format:check` · ESLint · `totem doctor --strict` · `totem lint --base main` — all exit 0.

`totem lint` reports a stale compile manifest (73 lessons); that is the standing rule-compilation freeze, not drift from this PR.

## Process note

The auto-generated spec for this issue named four files to modify — `packages/core/src/review/engine.ts`, `.../review/types.ts`, `packages/cli/src/commands/review.ts`, `.../engine.test.ts` — and proposed a `ReviewVerdictSchema` extension, a `ReviewExitCode` enum, and an `--allow-empty` flag. **None of those files or symbols exist**; `packages/core/src/review/` is not a directory in this repo. The spec was discarded and not committed, since a spec describing an invented architecture would read as the design of record. Filed as an exhibit on #2465 (run artifact `3a98f965ad47…`).

### DIFF
--- a/.changeset/quiet-moons-remain.md
+++ b/.changeset/quiet-moons-remain.md
@@ -1,0 +1,17 @@
+---
+'@mmnto/cli': minor
+---
+
+fix(review): deterministic skip paths no longer stamp the push-gate cache (#2466)
+
+`totem review` has three deterministic skip paths that drop the entire diff without examining it — all-non-code, no-code-remaining-after-filtering, and all-generated-artifacts. Each logged an informational "Deterministic fast-path" line and then called `writeReviewedContentHash(...)` before returning, so a run that reviewed nothing minted `.totem/cache/.reviewed-content-hash` — the marker PreToolUse hooks read as review authorization. A non-review both exited 0 and left behind a stamp claiming the tree was reviewed.
+
+The three paths are no longer equivalent in danger, which is why all three are fixed rather than only the reported one. `content-hash.sh` hashes only tracked files matching the review extensions (default `.ts`/`.tsx`/`.js`/`.jsx`), so for all-non-code and filtered-empty the hash is unchanged and the stamp was a no-op — the defect there is the dishonest clean-pass surface. The all-generated path is different: `.gitattributes linguist-generated` can mark a tracked `.ts` file as generated, so that path can drop a hashed source file, write a fresh stamp over the new hash, and authorize code no reviewer saw.
+
+All three now emit a `NON-REVIEW:` warning stating that nothing was examined and that the reviewed-content-hash was not stamped, and none of them stamp. The notice is single-sourced so the three sites cannot drift into describing the same guarantee differently, and its voice matches the existing worktree-drift notices — the other surface that declines to stamp. This mirrors the reasoning already applied to `--estimate` (an estimate is a forecast, not a passing review) and the standing repo rule against stamping the cache on non-reviews.
+
+Prose-only pushes are unaffected: the gate compares a hash over code files that a prose change does not touch, so it remains satisfied by the last genuine review.
+
+Bumped minor rather than patch on the slice-A precedent — push-gate authorization is user-visible behavior, and a consumer marking tracked sources as generated will now correctly be asked for a real review where one was previously minted for them.
+
+Sibling of #2452 slice A: that gate keys on lanes inside a verdict artifact, which these paths return before ever constructing, so it could not reach them. The remaining question — whether a total skip should also carry a distinct exit state rather than exiting 0 — is deliberately left open.
--- a/packages/cli/src/commands/shield-nonreview.test.ts
+++ b/packages/cli/src/commands/shield-nonreview.test.ts
@@ -1,0 +1,165 @@
+/**
+ * CLI-level deterministic-skip NON-REVIEW tests (mmnto-ai/totem#2466).
+ *
+ * `totem review` has three deterministic skip paths that drop the ENTIRE diff
+ * without examining it: all-non-code, no-code-after-filtering, and all-generated.
+ * Each used to call `writeReviewedContentHash(...)` before returning, minting the
+ * `.reviewed-content-hash` push-gate stamp for content nothing reviewed.
+ *
+ * The danger is not uniform, which is why all three are pinned here rather than
+ * only the reported one:
+ *
+ *   - all-non-code / filtered-empty fire only when no code file is in the diff,
+ *     so the hash is unchanged and the stamp was a no-op. The defect is the
+ *     dishonest clean-pass surface (Tenets 4/13).
+ *   - all-generated can drop a TRACKED, HASHED source file, because
+ *     `.gitattributes linguist-generated` can mark a `.ts` as generated. There the
+ *     stamp genuinely authorized never-reviewed code — a push-gate bypass.
+ *
+ * Mirrors `shield-covariate.test.ts`: its own file to avoid mock contamination,
+ * mocking the heavy seams (config, engine bootstrap, hook installer, git diff) so
+ * the skip branches are exercised without a real repo, config, or network invoke.
+ */
+
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { TotemConfig } from '@mmnto/totem';
+
+import { cleanTmpDir } from '../test-utils.js';
+
+// ── Seams that MUST NOT run once a skip path is taken ──
+const bootstrapEngineSpy = vi.fn(async (..._args: unknown[]): Promise<void> => {});
+const upgradePrePushHookSpy = vi.fn((..._args: unknown[]): boolean => false);
+const getDiffForReviewSpy = vi.fn(async (..._args: unknown[]): Promise<unknown> => null);
+
+const TEST_CONFIG = {
+  totemDir: '.totem',
+  review: { sourceExtensions: ['.ts'] },
+} as unknown as TotemConfig;
+
+vi.mock('../utils/bootstrap-engine.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../utils/bootstrap-engine.js')>();
+  return { ...actual, bootstrapEngine: bootstrapEngineSpy };
+});
+
+vi.mock('./install-hooks.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./install-hooks.js')>();
+  return { ...actual, upgradePrePushHookIfNeeded: upgradePrePushHookSpy };
+});
+
+vi.mock('../utils.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../utils.js')>();
+  return {
+    ...actual,
+    loadEnv: vi.fn(),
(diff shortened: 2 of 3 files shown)


satur8d:
## Correcting my own claim: "Prose-only pushes are unaffected" is overstated

Self-review before asking anyone else to look. The PR body says prose-only pushes are unaffected because the gate hashes code files a prose change does not touch. **That is true only when a stamp already exists.** The uncovered case:

**A repo with no stamp yet, whose pending changes are prose-only.** Previously the fast-path stamped and the gate unblocked. After this change it never stamps, so the agent push-gate stays blocked — and running `totem review` again cannot help, because the same prose-only diff takes the same skip path.

### Why it is a degradation and not a deadlock

`1659` (the no-changes stamp) is deliberately untouched by this PR, so `totem review` on a **clean tree** still stamps. The escape exists — stash the prose changes, run review, unstash — plus `--override`, which is ledgered. So the honest characterization is *awkward recovery in a narrow first-run case*, not a lockout. My original sentence was still too strong and I would rather correct it than have a reviewer find it.

### Scope of exposure, measured rather than assumed

| Repo | tracked `.ts/.tsx/.js/.jsx` | stamp present today |
|---|---|---|
| `totem` | 658 | yes |
| `totem-strategy` | 3 | **yes** |
| a fresh consumer | any | **no** ← the exposed case |

`totem-strategy` — the repo that reported #2466, and the most prose-dominant of the cohort — already carries a stamp, and a prose commit does not move a hash over 3 unchanged files. So it is not broken by this change. The exposure is genuinely limited to never-yet-stamped repos.

### What the case is *not*

Worth separating, because it looks similar and is the behavior we want: a `.ts` file changed without review, followed by a prose-only commit. The hash differs, the skip path fires, nothing stamps, the push is blocked. That is **correct** — the unreviewed TypeScript is exactly what should block — and it is the bypass this PR closes.

### Where this leaves the design

It sharpens the open exit-state question into a better one. The real issue is that the stamp answers *"was the tree reviewed?"* when the gate actually wants *"has reviewable content changed since the last review?"* Under the second framing a prose-only change in a never-stamped repo is trivially fine, no sentinel or override needed, and the generated-artifact bypass still closes.

That is a gate-semantics change, larger than this PR and deserving its own ruling. I am not folding it in. Flagging it as the more likely correct end state so the narrower fix here is not mistaken for the finished design.

satur8d:
/gemini review
@greptileai review

satur8d:
## Disposition — both greptile P2s TAKEN in `42b701cd`; plus a premise correction that changes what this PR is for

### greptile P2 — all-generated test doesn't cover the `.gitattributes` case — **TAKEN**

Correct, and it was the sharpest possible catch: the PR body names the `.gitattributes`-marked `.ts` as the *only* scenario where the old stamp could authorize genuinely-changed source, and the test used a default-glob lockfile — which is outside the content-hash extension set, so its stamp was a no-op. The test exercised the harmless case while the prose claimed the dangerous one.

Now driven through the real `readGitattributesGeneratedPatterns(cwd)` path with a tracked `.ts` marked `linguist-generated`.

### greptile P2 — filtered-empty path has no test — **TAKEN**

Also correct; Stage 2 was fixed and unexercised. Driven via a changed-file list containing a code file whose hunks are absent from the diff body — the shape a mode-only change produces.

Both new tests assert the **path-specific** warning wording rather than a bare `/NON-REVIEW/` match, so neither can pass by falling through to a different skip branch. Red-checked against `origin/main`'s `shield.ts`: 4 of 5 fail without the fix; the fifth is the over-fire guard, which must pass on both. Full workspace suite green (`@mmnto/cli` 147 files / 3457 tests, `@mmnto/totem` 128 / 3292).

---

## Correcting the premise this PR was written on

Independent of the bot round, and more important than either finding: **the claim that this stamp is consumed as push authorization is false for this repo.**

`mmnto-ai/totem#2180` (merged 2026-06-15) decommissioned `review-gate.sh`. `tools/pre-push` now runs `totem review || exit 1` **live** rather than checking a stamp. Nothing in this repo reads `.reviewed-content-hash`. I searched for a reader while writing this PR, found none, and concluded it must live somewhere I could not see — instead of concluding it was gone. I trusted `docs/reference/workflow-automation.md:47`, which still documents the decommissioned gate, over the code. Credit to totem-kimi for verifying against HEAD.

**What that changes:**

- The stamp hole this PR closes is **dormant in totem** and real only for downstream consumers on old hooks, or any reinstated gate. Still worth closing — but not the live defect here.
- The **live** defect in this repo is the other half: `tools/pre-push` runs `totem review || exit 1`, so a prose-only diff hits a fast path, exits 0, and the push proceeds. That is #2466's original framing — which I argued *away from* and toward the stamp. For totem, the original title was right.
- This PR deliberately defers the exit-state question. That deferral is therefore larger than I represented: it leaves the locally-live path unfixed.

**Recommendation, unchanged in direction:** land this narrow. It is correct, tested, and closes the consumer-facing hole. But it should not be read as fixing #2466 in this repo — the exit contract is a follow-up, and totem-codex's two-phase model (admission/applicability → fan execution, where a deterministic skip is *not* a lane outcome and *not* an `InvokeFailureKind`) is the right frame for it.

Also corrected on the record: my earlier comment's recovery path (stash → review → unstash) is **falsified** for repos with zero tracked source files. `computeReviewedContentHash` returns `null` when the file set is empty, so `writeReviewedContentHash` writes nothing and the clean-tree path cannot stamp either. Found by totem-gemini, independently flagged by totem-codex. It does not affect this diff, but it does affect the reasoning I published above it.

satur8d:
@greptileai review
/gemini review

satur8d:
@gemini-code-assist Thanks — declining all three `DISPLAY_TAG → TAG` suggestions, with evidence.

`shield-templates.ts` defines two tags for this command: `TAG = 'Shield'` and `DISPLAY_TAG = 'Review'`. The `totem review` command deliberately surfaces to users under `[Review]`, not `[Shield]` — that is what `DISPLAY_TAG` is for. All ~40 user-facing `log.*` calls in `shield.ts` use `DISPLAY_TAG`; the three NON-REVIEW warnings this PR adds are part of the same command output and correctly match it.

Applying the change to only these three lines would (1) print `[Shield]` on three lines while ~40 sibling lines print `[Review]` — a user-facing inconsistency, not a fix, and (2) not compile — `TAG` is not imported into `shield.ts` (only `DISPLAY_TAG` is).

§9's intent ("a command-specific tag, not a generic/error tag") is already met by `DISPLAY_TAG = 'Review'` — the same shape as §9's existing `log.error → 'Totem Error'` carve-out. Reconciling §9's letter with shield's deliberate `[Review]` display is a file-wide question (all 43 calls, or a §9 carve-out), not a 3-of-43-line change on this narrow authorization-hole fix.

PR #2478, opened 2026-07-21
feat(autoclose): B slice — totem pr merge actuator (#1762)

satur8d:
B slice of the auto-close enforcement seam (family issue mmnto-ai/totem#1762; contract mmnto-ai/totem-strategy#951 / ADR-082 Amendment 1). This PR does not close any issue; the family issue stays open for the remaining layers. **The A slice (raw-merge command interception) was built here, review-hardened across five rounds, and then STRIPPED per the operator's OPTION 1 ruling (2026-07-22)** — record + grounding in `.totem/specs/1762-ab.md` § OPTION 1 ruling; the A history remains in this PR's pre-strip commits (through `913ad2af`).

## What ships

- **B — `totem pr merge [number]`, the sanctioned actuator.** Asserts the merge-config posture (PR_TITLE + BLANK + squash-only) via GraphQL reusing `evaluateMergeConfigPosture`, hard-failing on drift naming the field. Fetches title+body only (comments are never scanned — they never auto-close). Evaluates via core's `scanPrCorpus` — an undeclared close-keyword-adjacent ref refuses the merge with the marker recovery instruction; authorization is the `totem-close` marker only. Merges via a hardcoded `['pr','merge',N,'--squash']` argv — structurally incapable of authoring a body/subject flag (asserted by test). `--check-only` evaluates without merging; `--close-declared` (opt-in, operator-ruled 2026-07-21) executes marker-declared closes post-merge, default off prints the exact commands.
- **core:** new `PR_MERGE_FAILED` TotemError code (the only core delta remaining).

## Why A was stripped (OPTION 1)

Each hardening round on the hand-rolled shell parser spawned a same-class sibling bypass — the wrong-altitude signal. Freshly derived 2026-07-22: every accidental-close vector is already covered by a simpler, mostly server-side layer — PR-body wording by the shipped C write-time guard + D1 (verified ENROLLED as a required check in the active `main-required-checks` ruleset); merge-time body by repo config (squash-only + `squash_merge_commit_message=BLANK`, both verified live — the default merge path carries no body text); direct pushes by branch protection. A defended only the *deliberate* `--body` override — outside the accidental threat model, reversible, and caught by D2.

Follow-ups routed out of this PR (spec § OPTION 1): a posture-drift sensor asserting the config invariant (strategy#482 parity row), the A-independent Gemini `BeforeTool.js` ESM defect (`.js`→`.cjs` in its own slice), and D2 arming per the four falsifiable criteria in `.totem/specs/1762.md`.

[...]

Issue #1616, opened 2026-04-22
Thread shared warning sink through CLI commands to replace console/log.warn


satur8d:
## Context

CodeRabbit asked on PR #1615 (#1288) for a broader refactor: rather than each CLI command emitting warnings directly via `console.warn` or `log.warn`, thread a shared warning sink (e.g., an `onWarn` callback) through the CLI command layer so the utility/helper modules stay decoupled from the output mechanism.

Current state as of #1288:

- `packages/cli/src/utils/governance.ts` uses `log.warn('Totem', ...)` for the two warn-and-continue paths (docs:inject failure and git add failure). This was a narrow fix for the `no-console` eslint-disable CR flagged.
- Other CLI commands use a mix of `console.warn`, `console.error`, `log.warn`, and `log.error`. There is no canonical sink for non-fatal warnings.

## Scope

Design and implement a shared warning sink abstraction:

1. Add an `onWarn(msg: string)` callback parameter to helper functions that currently emit warnings directly, defaulting to the canonical `log.warn('Totem', msg)` routing.
2. CLI commands can override per-invocation if they need a different sink (tests, CI-structured output, silent mode).
3. Sweep existing CLI commands that emit warnings (`packages/cli/src/commands/*.ts`, `packages/cli/src/utils/*.ts`) in one pass.
4. Keep `log.warn` as the default sink so no behavior change for end users.

## Non-goals

- Replacing `log.error` / `log.info` / `log.success` routing. Those are terminal-path output, not warnings.
- Changing stderr vs stdout routing defaults.
- Removing `console.*` usage across the entire CLI — only the warning sites.

## Acceptance

- A defined `WarnSink` type in `packages/cli/src/utils/` (or `ui.ts`) with one canonical default implementation.
- Helper functions in `packages/cli/src/utils/governance.ts` (and at least one other helper that currently emits warnings) accept an optional `onWarn?: WarnSink` parameter.
- No `eslint-disable-next-line no-console` suppressions in the code touched by the sweep.
- Documentation in the ui.ts module or a short README entry on the pattern.

## Tier

Tier-3. Refactor-only, no user-visible behavior change, cross-cutting scope (~5-10 files touched). Sensible to batch as a standalone PR rather than block on any individual feature.

## Refs

- PR #1615 CR thread: https://github.com/mmnto-ai/totem/pull/1615#discussion_r3122078940
- #1288 (where the first log.warn migration landed as a point fix)

PR #2485, opened 2026-07-23
fix(hooks): ship Gemini BeforeTool as .cjs to stop the type-module fail-open

### LINKED ISSUE #2481: fix(hooks): Gemini BeforeTool.js fail-opens as ESM in type-module consumer repos (.js to .cjs migration slice)
## Defect

The shipped Gemini C-layer hook `.gemini/hooks/BeforeTool.js` (the write-time close-keyword guard, distributed by `totem init` / `totem hook install`) is a CommonJS file with a `.js` extension. In a consumer repo whose `package.json` declares `"type": "module"`, Node resolves it as ESM and it throws `ReferenceError: require is not defined` **before reading the tool-call JSON on stdin**. Gemini CLI treats a crashed hook as a warning and proceeds — so the guard **fail-opens silently**: the consumer believes the write-time guard is armed while it never evaluates anything.

## Provenance

Discovered during the mmnto-ai/totem#2478 review rounds (round 2). The fix that existed there — a `.js`→`.cjs` rename plus non-interactive registration/migration in `totem init` and `totem hook install` — rode the A-slice adoption machinery and was stripped with it under the OPTION 1 ruling (2026-07-22); see `.totem/specs/1762-ab.md` § OPTION 1, "Follow-ups routed OUT of this PR". This issue is that routed follow-up.

## Fix shape (own small slice, deliberately bounded)

1. Ship the hook as `.gemini/hooks/BeforeTool.cjs`; keep the legacy `.js` in the eject/managed rosters so pre-migration artifacts are cleaned up.
2. Migrate `.gemini/settings.json` registration to the `.cjs` path (idempotent, non-corrupting on malformed settings).
3. Run the same migration on the consumer upgrade path (`prepare` → `totem hook install`), not only interactive `totem init`.

Scope warning from the #2478 arc: the adoption/arming machinery is where the complexity exploded (registration atomicity, ownership predicates, legacy-coexistence collapse — three review rounds of same-class siblings). Keep this slice to the minimal rename + registration migration; decline arming-verification ambitions here.

## Acceptance

- A `"type": "module"` consumer fixture executes the distributed hook and the guard evaluates (no `require is not defined`).
- Upgrade path on a pre-migration consumer converges to exactly one registration pointing at the `.cjs`.

satur8d:
## Context

The routed follow-up from the mmnto-ai/totem#2478 OPTION 1 strip: the distributed Gemini write-time guard `.gemini/hooks/BeforeTool.js` is CommonJS content with a `.js` extension, so a `"type": "module"` consumer resolves it as ESM and it throws before reading the tool call — Gemini CLI treats the crash as a warning, and the guard fail-opens silently.

Closes mmnto-ai/totem#2481

## Changes (bounded per the issue's recorded scope warning)

1. **Ship the hook as `.gemini/hooks/BeforeTool.cjs`** (`GEMINI_BEFORE_TOOL_REL`); the managed-hook roster and the locked six-artifact contract test follow.
2. **Legacy migration on both paths** (`totem hook install` — which the prepare wrapper invokes — and `totem init`): a present totem-owned legacy `.js` is proof of adoption, so the `.cjs` successor is materialized and the fail-open `.js` removed, reusing the existing ownership predicates (`markerOpensFile`/`isBoundedOwnedFile`) — no new predicate. Drifted-unbounded legacy declines without `--force`; a markerless user file is never touched, even forced.
3. **Registration migration, rewrite-if-present only:** an existing `.gemini/settings.json` hooks entry referencing `BeforeTool.js` is rewritten to `.cjs` (basename-matched, separator/prefix-agnostic, idempotent, fail-soft on malformed JSON — user content is never clobbered). It never *creates* a registration.
4. `totem eject` removes both the `.cjs` and the legacy `.js`.

## Two findings surfaced during the build (disclosed, not silently absorbed)

- **The issue's fix-shape #3 presupposed a settings-registration writer that does not exist**: the #2478 arming machinery was stripped under OPTION 1 and never re-landed, so nothing in this codebase writes a Gemini hooks registration. Consequence: after this fix a *fresh* `totem init` produces a CJS-safe hook **file** but no registration — the guard stays dormant until the deferred adoption/arming slice lands. This PR completes the issue's own acceptance scope (CJS-safe execution + pre-migration convergence); it deliberately does not restore arming.
- **`.gemini/hooks/SessionStart.js` carries the same latent ESM fail-open** (top-level `require` in a type-module consumer). Deliberately out of this slice per the issue's scope; lower severity (briefing hook, not a guard). Follow-up candidate.

## Verification

- New tests (8): the type-module acceptance pair (the `.cjs` evaluates; the identical body as `.js` provably fail-opens — the defect is pinned, not assumed), pre-migration convergence to exactly one `.cjs` registration, idempotency, never-creates-registration, malformed-settings fail-soft, user-owned skip, drifted-unbounded decline/--force.
- Full CLI suite 3558/3558; workspace build green; `format:check` clean; ESLint 0 errors (warnings identical to the main baseline); `totem lint --base main` PASS, 0 errors.
- Changeset: `@mmnto/cli` patch.

### DIFF
--- a/.changeset/gemini-beforetool-cjs-fail-open.md
+++ b/.changeset/gemini-beforetool-cjs-fail-open.md
@@ -1,0 +1,9 @@
+---
+'@mmnto/cli': patch
+---
+
+Ship the Gemini CLI write-time guard as `.gemini/hooks/BeforeTool.cjs` instead of `.js` to stop a silent fail-open (the defect reported on mmnto-ai/totem#2481).
+
+The distributed hook body is CommonJS (top-level `require('child_process')`). In a consumer repo whose `package.json` declares `"type": "module"`, Node resolved the bare `.js` as ESM and threw `ReferenceError: require is not defined` before it could read the tool call — and Gemini CLI treats a crashed hook as a non-fatal warning, so the write-time guard fail-opened silently (the consumer believed it was armed while it never evaluated anything). A `.cjs` file is CommonJS regardless of the consumer's package `type`, mirroring the load-bearing `.cjs` extension on the Claude-side session hooks.
+
+`totem hook install` (which the `prepare` wrapper invokes on every install) and `totem init` now migrate an upgraded consumer: the legacy bounded totem-owned `.gemini/hooks/BeforeTool.js` is removed and the `.cjs` successor materialized, and an existing `.gemini/settings.json` BeforeTool command pointing at the `.js` is rewritten to the `.cjs`. The rewrite is idempotent, fail-soft on malformed settings (user content preserved), and never creates a registration where none exists. `totem eject` removes both the `.cjs` and the legacy `.js`. `SessionStart.js` is intentionally left as-is — the slice scopes to the write-time guard.
--- a/packages/cli/src/commands/eject.ts
+++ b/packages/cli/src/commands/eject.ts
@@ -16,6 +16,9 @@ const REFLEX_FILES = ['CLAUDE.md', '.cursorrules'];
 /** Files scaffolded by `totem init` that are fully owned by Totem. */
 const TOTEM_SCAFFOLDED_FILES = [
   '.gemini/hooks/SessionStart.js',
+  // Current (mmnto-ai/totem#2481) + the pre-migration `.js` it renamed — eject
+  // removes both so an upgraded-then-ejected consumer leaves no fail-open artifact.
+  '.gemini/hooks/BeforeTool.cjs',
   '.gemini/hooks/BeforeTool.js',
   '.gemini/skills/totem.md',
   '.totem/hooks/shield-gate.cjs',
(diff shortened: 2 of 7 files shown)


satur8d:
## Round 1 disposition (no code change)

@gemini-code-assist one finding, declined with rationale:

1. **`console.error` vs the structured `log` utility in `printGeminiHookMigrationSummary`** — the styleguide tension is real, but the finding's premise (inconsistency) points the wrong way locally: `install-hooks.ts` uses `console.error` for ALL summary/diagnostic output (25+ existing call sites, including `printManagedSessionHookSummary`, the direct sibling this function mirrors), because hook-install output deliberately targets stderr under the `prepare` wrapper. Converting only the new function would fork the file's idiom mid-file; converting the whole file is outside this deliberately bounded slice (the issue's recorded scope warning). The repo-wide conversion is already tracked as mmnto-ai/totem#1616 ("thread a shared warning sink through CLI commands") — the right vehicle for this change, applied to the file wholesale.

No other bot posted findings this round. Verification state unchanged from the PR body: 3558/3558 CLI tests, all gates green.

PR #2454, opened 2026-07-19
fix(review): zero-completed-lane fan hard-errors instead of exiting 0 (#2452 slice A)

satur8d:
## What & why

A `totem review` fan that finished with **zero completed verdict lanes** could exit **0** when `--fail-on` was omitted. The all-lanes-failed gate keyed on `completed || abstained`, so an **abstain-bearing** fan (a lane that invoked but whose output was unextractable — sensor-down) skipped the hard-error and read like a pass. The verdict artifact was already honest (`settled=false`, `completedLaneCount=0`) — **only the process exit code was wrong**. This is the exact shape observed while preparing #2451 (0/2 usable lanes, `settled=false`, exit 0).

This is **slice A of #2452** (the fan boundary), per the role-split with `@totem-codex`. Slice B (CLI-fallback evidence retention + `invoke-error` taxonomy + Sonnet-5 extraction fixtures) is codex's and lands separately; this slice was designed to land **independently** of it.

## The fix

- **Exit contract:** the gate now keys on **zero completed lanes** via a new exported core predicate `hasNoCompletedLane(lanes)` (⇔ `verdict.completedLaneCount === 0`, superRefine-bound). It fires **after** the honest verdict is saved and **before** the override/cache-stamp block — so `--override` can never mint a push authorization from a provider-unsettled round (**Tenets 12/13**; strategy #845). The widening is monotone: it subsumes the old all-failed gate.
- **Honest message:** enumerates `completed=0/abstained=N/failed=M` and states "non-pass, honest verdict written, not a crash" — the old "failed to invoke" wording was dishonest for an abstained lane that *did* invoke.
- **Budget split:** the fan report now reports diff-budget and total-prompt-budget as distinct numbers (`diff ⊆ prompt`), closing the double-count.

## Semver

**Minor** (both packages): the exit-code behavior is user-visible (a CI that exited 0 on an abstain-only fan now hard-errors — though that was a false pass), and core gains a new public export.

## Tests (the five-shape exit matrix + more)

New/changed coverage (`review-fan.test.ts`, `verdict.test.ts`):
- `hasNoCompletedLane` unit block (true/false shapes + agreement with `completedLaneCount`).
- **Flipped two existing tests that encoded the bug:** single-abstained (was `resolves` → now `rejects`, finding-8 post-check row preserved) and all-failed (message updated to the enumerated form).
[...]

PR #2499, opened 2026-07-23
fix(lint): corpus-bearing repos hard-error when compiled rules are unloadable (Prop 309 quota item)

satur8d:
## What

`totem lint` passed vacuously when the compiled-rules manifest was missing, truncated, or unreadable. `loadCompiledRules` returns `[]` on a missing file, and on a JSON `SyntaxError` or I/O fault (`EISDIR`/`EACCES`) it fired an optional `onWarn` and returned `[]` — but the lint call site omitted `onWarn`, so even the accounting line was dropped (the producer-accounts/consumer-drops shape of mmnto-ai/totem-strategy#970, on a new surface). In this repo that meant all 485 compiled rules could silently disarm behind a green exit — pre-push hook, CI "Totem Lint" job, and every consumer repo none the wiser.

## How

Enforcement lives entirely at the CLI call site — `@mmnto/core`'s `loadCompiledRules` public API and its return-`[]` semantics are unchanged:

- The lint call site threads an `onWarn` that records **and renders** the load-time warning.
- When zero rules load, a **corpus-bearing discriminator** (config-driven: do the config's lesson-kind targets match real files on disk? pure filesystem, no git spawn) decides between four states:
  - **Not corpus-bearing** → today's info-skip, byte-preserved (mmnto-ai/totem#1831 early-adoption semantics).
  - **Corpus-bearing + manifest missing** → `TotemError`, non-zero exit: "enforcement disarmed", naming lesson count + manifest path + fix hint.
  - **Corpus-bearing + manifest unloadable** (truncated JSON / EISDIR / IO) → `TotemError` including the recorded load warning.
  - **Corpus-bearing + manifest valid but zero ACTIVE rules** (all archived / pending-verification / untested-against-codebase) → honest info-skip that names the lifecycle state — archived-in-place rules preserve telemetry by design and are not a disarmed gate.
- Zod-shape errors continue to throw from core, unchanged.
- `shield estimate` (the other `runCompiledRules` caller) deliberately does not opt in — its forecast path is not the enforcement gate; the `config` option's doc comment states the opt-in contract.

## Tests (7 new — failure **and** control assertions, per the Prop 309 quota-item bar)

Induced-failure:

1. Lessons present + manifest deleted → throws "enforcement disarmed".
2. Lessons present + manifest truncated mid-array → throws (previously passed vacuously).
3. Lessons present + manifest replaced by a directory (EISDIR, Windows-portable) → throws, and the previously-dropped `onWarn` accounting text is asserted in the error.
4. Concrete (non-wildcard) lesson target — the production config's aggregated `.totem/lessons.md` shape — reaches the discriminator through the `statSync` branch and throws.

Controls:

5. Config declares a lesson target but no lesson files exist → clean info-skip (mmnto-ai/totem#1831 preserved).
6. Valid manifest, all rules archived → clean, with the zero-active-rules message.
7. Valid active rules → behavior unchanged.
8. Config-less caller (the `shield estimate` shape) + truncated manifest → resolves clean, and the accounting line still renders — the opt-out covers the exit code, not the disclosure.

## Local review loop (two-lane fan, settled at convergence)

Three rounds, each settling the prior round's findings. Absorbed: walk bound past `.lancedb`/`dist` (round 0 WARN), concrete-glob branch coverage (round 1 WARN), config-less opt-out contract test (round 2 WARN). Declined with evidence: load/exists TOCTOU (single-shot CLI, the only bad arm needs a concurrent manifest-create inside a microsecond window); `LINT_LESSONS_FAILED` registration (already a member of the code union at `packages/core/src/errors.ts:25`, used by `compile` and `lint-lessons`); further skip-list enumeration (cold path, same-shape recycle of an absorbed finding). Convergence named at round 2 on monotonically weaker sub-0.5 findings.

## Provenance

First quota item of the Prop 309 hardening program (mmnto-ai/totem-strategy#971, operator-ruled 2026-07-23; directed GO from the bus). Source defect verified 2026-07-23 against `compiler.ts:161/173-182` and `run-compiled-rules.ts:150/157`. This change also reconciles two standing corpus lessons that were in direct tension — "gracefully skip linting on empty rules" (`lesson-0fde21c0`) vs "quality gate tools must never exit successfully when required rules are missing" (`lessons.md`) — via the corpus-bearing discriminator: both are now true in their own regime.

Changeset: `@mmnto/cli` minor (user-visible exit-code change, mirroring the mmnto-ai/totem#2454 sibling's bump class).

### DIFF
--- a/.changeset/corpus-bearing-lint-hard-error.md
+++ b/.changeset/corpus-bearing-lint-hard-error.md
@@ -1,0 +1,13 @@
+---
+'@mmnto/cli': minor
+---
+
+fix(lint): corpus-bearing repos hard-error when compiled rules are unloadable
+
+`totem lint` passed vacuously when the compiled-rules manifest was missing, truncated, or unreadable: `loadCompiledRules` returns `[]` on a missing file, and on a JSON/`SyntaxError` or I/O fault (`EISDIR`/`EACCES`) it fired an optional `onWarn` and returned `[]` — but the lint call site omitted `onWarn`, so even the accounting line was dropped. A repo with a real lesson corpus and hundreds of compiled rules could therefore have its entire enforcement gate silently disarmed behind a green exit.
+
+The lint call site now threads an `onWarn` that both records and renders the load-time warning, and — when zero rules load — discriminates on whether the repo is **corpus-bearing** (does the loaded config declare lesson-kind ingest targets whose globs match real files on disk, a pure git-free filesystem derivation). A corpus-bearing repo whose manifest is missing or unloadable now throws a `TotemError` (non-zero exit) naming the lesson count, the manifest path, and the phrase "enforcement disarmed", with a `totem lesson compile` fix hint. A manifest that is present and valid but filtered to zero ACTIVE rules (all archived / pending-verification / untested-against-codebase) stays an honest info-skip that names the zero-active-rules lifecycle state.
+
+The empty-corpus / early-adoption skip is preserved exactly (mmnto-ai/totem#1831): a repo with no lesson files, or any caller that does not pass its config, keeps today's info-skip. `@mmnto/core`'s `loadCompiledRules` public API and its return-`[]` semantics are unchanged — the enforcement lives entirely at the CLI call site.
+
+First quota item of the Prop 309 hardening program (mmnto-ai/totem-strategy#971).
--- a/packages/cli/src/commands/lint.ts
+++ b/packages/cli/src/commands/lint.ts
@@ -325,6 +325,11 @@ export async function lintCommand(options: LintOptions): Promise<void> {
     ignorePatterns: allIgnore,
     tag: TAG,
     configRoot,
+    // Pass the loaded config so runCompiledRules can hard-error a corpus-bearing
+    // repo whose compiled-rules manifest is missing / unloadable, instead of
+    // passing vacuously behind a green exit (mmnto-ai/totem-strategy#971,
+    // Prop 309). Its lesson-kind targets drive the corpus-bearing check.
+    config,
     isStaged: !!options.staged,
     regexTimeoutMode: timeoutMode,
     astParseMode,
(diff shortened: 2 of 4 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Bot-round disposition (operator-invoked round of 2026-07-23T23:06Z)

**Verdicts:** CodeRabbit — no actionable comments. Gemini Code Assist — "no review comments to evaluate, implementation aligns with the repository's guidelines." Greptile — Confidence 4/5, "Safe to merge," 3 inline P2s, dispositioned below. Fix round pushed at `b68bc69d`.

1. **TOCTOU between `loadCompiledRules` and the `existsSync` re-check** (`run-compiled-rules.ts`) — **absorbed as documentation.** The race itself is accepted-by-design as the finding acknowledges (single-shot CLI; the only affected arm needs a concurrent manifest-create inside a microsecond window, and a concurrent compile implies an imminent re-lint). The finding's actionable core was readability — the two `existsSync` calls "not obviously related at a glance" — and the same seam was independently flagged by our local review lane, so the re-check now carries a comment linking it to the loader's internal check and naming the accepted race.

2. **`console.error` spy inspects only `c[0]` of multi-argument calls** (`run-compiled-rules.test.ts`, both control tests) — **fixed.** Both spies now `c.join(' ')` across all call args, removing the silent coupling to the log utility's internal arity. 57/57 tests green.

3. **`async` wrapper over synchronous fs ops in `countLessonCorpusFiles`** — **declined: file convention.** This file has zero top-level value imports; every value dependency is lazily `await import`-ed at 13 pre-existing sites (`run-compiled-rules.ts:197-218, 265, 372, 421, 680-688`), including `node:fs` in the pre-existing main path. The helper follows the file's deliberate CLI-startup-lazy pattern; converting it to static-import sync would cut against the established convention rather than align with it. (For the reviewer's style memory: lazy dynamic value imports are intentional in this codebase's CLI command modules.)

local-lane: 0a63bb55 round=2 settled=false lanes=2/2

_Note on the local-lane line: `settled=false` records that round 2 minted findings; all three were dispositioned before this bot round (two absorbed with tests at `e459b440`, `e3bb2b49`, `5cdc9439`; one declined with registration evidence at `core/src/errors.ts:25`) and convergence was named on monotonically weaker sub-0.5 findings. The verdict artifact ledger carries the round history._

satur8d:
@greptileai review

Issue #2506, opened 2026-07-24
fix(rules): comment-less formats have no escape hatch — machine-generated-artifacts rule (/^.+$/) fires on every line of a hand-authored pack manifest


satur8d:
## Problem

Rules that fire on JSON have **no available escape hatch**. The `// totem-context:` inline suppression is a comment, and JSON has no comment syntax. The only remaining lever is a file- or project-level `.totemignore` exemption, which is far coarser than the false positive warrants.

This bites immediately: the Ask 1 agent-workflow pack's entire premise is a **hand-authored** `compiled-rules.json`.

## Reproduction (this session, `@mmnto/cli` 1.105.0)

A 4-line scaffold manifest:

```json
{
  "version": 1,
  "rules": []
}
```

`totem lint` with the file staged produces a violation on **every line** — 4 of the run's 8 warnings — from the "Do not manually edit machine-generated artifacts" rule, whose pattern is `/^.+$/`.

The file is not machine-generated. It is the hand-authored artifact the pack exists to ship, and the rule cannot tell the difference. As the pack grows from 0 to 9 rule entries, this scales to a violation per line of the manifest.

## Two separable defects

**1. No escape hatch for comment-less formats.** Verified in the same session for the adjacent case: markdown and YAML also lack the inline escape. `PreWriteShield.cjs` parses `<!-- totem-context: -->` HTML-comment suppressions, but the `totem lint` regex engine does not recognize them, so even markdown's available comment syntax does not reach the linter. JSON has no comment syntax at all, so no amount of parser support helps — it needs a sidecar mechanism (a path-scoped exemption in config, a per-rule allowlist, or a `.totemignore` entry granular enough to name one rule rather than one file).

**2. The rule's pattern is `/^.+$/`.** Matching every non-empty line means the rule carries no signal about *which* artifacts are machine-generated; it relies entirely on `fileGlobs` for precision. A pack manifest that is hand-authored by design will trip it wherever the globs reach. This is adjacent to mmnto-ai/totem#2063 (warning-tier rules false-positive on standard idioms) and mmnto-ai/totem#2034 (over-broad rule globs).

## Why it is not just noise

[...]

Issue #2505, opened 2026-07-24
fix(lint): untracked files are invisible to the uncommitted diff source — a whole new package passes vacuously as "387 rules, 0 violations"


satur8d:
## Problem

`totem lint` with the `uncommitted` diff source is blind to **untracked** files. A brand-new package can be authored and pass the local gate having had none of its content scanned.

## Reproduction (this session, `@mmnto/cli` 1.105.0, branch `main`)

Authored a new workspace package at `packages/pack-agent-workflow/` — 7 files, all untracked. Then, with no other changes beyond a `pnpm-lock.yaml` touched by the install:

```
[Lint] Diff source: uncommitted
[Lint] Changed files (1): pnpm-lock.yaml
[Lint] Running 387 rules (zero LLM)...
[Lint] Verdict: PASS - 387 rules, 0 violations
```

Then `git add` of the same seven paths, nothing else changed:

```
[Lint] Verdict: PASS - 8 warning(s), 0 errors
```

Same bytes on disk. The first run scanned none of them and reported PASS across 387 rules; the second found 8 warnings. The only difference is the git index.

## Why this is a gate-correctness defect, not just diff semantics

Untracked files being outside `git diff` is expected *mechanically*. The problem is what the tool asserts on top of that: `Verdict: PASS - 387 rules, 0 violations` reads as "387 rules were satisfied by this working tree", when the truthful statement is "387 rules had nothing in scope to evaluate."

This is the vacuous-pass shape the lint lane has been closing elsewhere — mmnto-ai/totem#2499 (empty corpus vs disarmed), mmnto-ai/totem#2375 (empty-corpus / empty-diff skips returning before honoring `--out`), mmnto-ai/totem#2466 (fast-paths minting a reviewed stamp without reviewing). Tracking parent: mmnto-ai/totem#2055.

The agent-authoring case makes it sharp: writing a new file is the single most common thing an agent does, and it is exactly the case the local gate cannot see. An agent that authors a package, runs `totem lint`, and reads PASS has been given a false clearance.

## Related but distinct

mmnto-ai/totem#2051 is the *inverse* direction — pre-push lint scanning the working tree too broadly, so unrelated uncommitted files crash a doc-only push. Same trust cluster, opposite failure.

## Proposed direction

Not obviously "just include untracked files" — that would re-create the #2051 blast radius. Options:

[...]

PR #2508, opened 2026-07-28
feat(packs): @mmnto/pack-agent-workflow v1 — agent-workflow governance micro-pack (Ask 1)

satur8d:
## Summary

`@mmnto/pack-agent-workflow` v1 — the Ask 1 agent-workflow governance micro-pack. **9 hand-authored compiled entries for 8 governed concepts**: agent-instruction authority, fail-open gates, agent-context secret exposure, permission-bypass grants, and dependency-introduction discipline. Data-only pack (no build step), `private: true` like its siblings, added to the changesets fixed group for version parity with `pack-agent-security` / `pack-rust-architecture`.

Provenance: built 2026-07-24 against the ruled spec (`.totem/specs/ask1-micro-pack.md`, committed here per the tracked-specs convention) after a 5-seat pre-build cohort panel + 2 amendments; panel rulings applied (rule 1 and 8 error→warn, rule 2 split, rule 4 recut, bare `continue-on-error` deferred). The secret-exposure rule is **authored new** with the frozen corpus as prior art only (re-ruled after the original MOVE premise was found false at build time).

## Control layers

- **L1** — `badExample`/`goodExample` in-schema, adversarial near-misses rather than clean code.
- **L2** — in-tree wind tunnel (`test/fixtures/wind-tunnel/{bad,good}/` at logical paths): every rule must fire on its bad sites and stay **silent on every good site its globs admit**.
- **L3** — glob coverage asserted both sides. This layer caught **5 dead globs in the first draft pre-ship** (2 dead negatives, 3 dead positives).

102/102 tests. Fixtures are excluded from repo lint (`totem.config.ts` `shieldIgnorePatterns`) and prettier (`.prettierignore`), mirroring the `examples/proof-kit/fixture/**` precedent — formatting reflows `catch (e) { }` and breaks ast-grep matches.

## Constraints documented in the pack README

- ast-grep cannot target md/YAML/JSON: the built-in registry covers TS/JS extensions only, and unregistered globs **silently fall back to `Lang.Tsx`** rather than failing loud — so every non-TS rule ships `engine: "regex"`, asserted structurally in `rules.test.ts`.
- Advisory collapses to warning: two enforcement tiers in v1 (`error | warning`); advisory is a documented convention over `warning`.
- `category` is a fixed four-value enum; pack identity lives in the package name.

## Gate

- `pnpm -r build` exit 0 · **102/102 tests** · repo-wide `format:check` clean · `totem lint` full-branch **PASS, 0 errors** (226 advisory warnings, all in the known frozen-lesson false-positive classes: 156/226 are the machine-generated-artifacts rule (#2506) firing per-line on the hand-authored manifest; the rest are the test-idiom classes (#2063)).
- Gate-scope disclosure: the 2026-07-24 in-session lint PASS predated tracking and never contained the pack (the untracked-files blind spot, #2505, found by re-verification 2026-07-28); the staged and full-branch runs in this PR are the first real lint contact, scope lines verified.


satur8d:

All three findings from the Gemini Code Assist round are **FIXED** in `bb9e42f7`, with one remedy-premise correction recorded:

1. **Heading 72 chars (`competing-authority`)** — FIXED: now "Competing agent-instruction authority without AGENTS.md link" (60).
2. **Heading 70 chars (`governed-directive-provenance`)** — FIXED: now "Governance directive in governed file lacks provenance ref" (58).
3. **Heading 63 chars (`optional-dep-disclosure`)** — FIXED: now "Optional dependency without a fail-loud consumer check" (54).

A full-manifest sweep confirms no remaining heading exceeds 60. Because this pack pins `lessonHash` as a deterministic derivation of heading + message (asserted per-rule by its own test suite), the three hashes were re-derived and the pinned set in `rules.test.ts` updated — 102/102 tests pass. The README rule table carries the new headings.

**Remedy-premise correction, for the record:** the review asked for the fix "in the source lesson file rather than the generated `compiled-rules.json`". For this pack that referent does not exist — the manifest is hand-authored source with no lesson files behind it (documented in the pack README, "Why this pack has no lessons"-class provenance), so the fix lands in the manifest itself. The finding stands; only the generalization from the compiled corpus does not transfer.

Local lane: no verdict artifact is recorded for this branch lineage (the pack predates this session's covariate mechanism), so there is no `local-lane:` line to carry — noted per protocol rather than omitted silently.

satur8d:

Both GCA round-2 findings are **FIXED** in `90b38776` — and both were genuine precision catches on the rules themselves:

1. **Single-pipe blind spot in `workflow-gate-masking`** — ACCEPTED + FIXED. Verified mechanically: `[^|]*` cannot cross a single pipe, so `pnpm lint | tee log.txt || true` escaped an error-tier rule built to catch masked gates. Replaced with the suggested tempered-dot idiom `(?:(?!\|\|).)*`.
2. **Digit-leading npm names in `dependency-disclosure`** — ACCEPTED + FIXED. `[@a-z]` → `[@a-z0-9]` (the `7zip-bin` class is legal npm naming).

Per this pack's control discipline, each fix ships with a **discriminating wind-tunnel specimen**: a new bad-site line the OLD pattern provably misses and the NEW catches, verified by direct old-vs-new regex evaluation (`false/true` on each specimen; all good sites remain silent under both patterns; pre-existing catches unchanged). One process note recorded honestly: the first draft of the piped specimen was itself vacuous — `tee test.log` reintroduced the verb token after the pipe, so the old pattern matched it coincidentally — caught by the old-vs-new check and replaced with a post-pipe segment carrying no verb substring. `lessonHash` values are unchanged (derived from heading + message only, which these fixes do not touch). 102/102 tests.

Round-2 meta, for the merge record: these two findings partially falsify my own pre-round prediction that a bot pass on this manifest would be low-yield — GCA's second read produced two real enforcement-precision improvements the five-seat panel and three control layers all missed. Recorded as such.

PR #2512, opened 2026-07-28
feat(lint): Class 7 target-mismatch load-time guard + archive both corpus landmines (freeze exception, Prop 309)

satur8d:
## Summary

Two commits, one class, closed at both ends:

1. **`fix(corpus)`** — both target-mismatched landmines archived in place under the operator-confirmed narrow hand-edit + `--refresh-manifest` freeze exception (recorded on mmnto-ai/totem-strategy#974, 2026-07-28; adjudicated class per #2343). `1a7080ebf6162de3` (4 `.json` globs) and `ae4d3065e5548794` (sole glob `packages/pack-agent-security/turbo.json`) are `engine: ast-grep` rules with no resolvable Tree-sitter language — they could never execute, and instead armed a **run-wide lint abort** (`rule-engine.ts` `TotemParseError`) on first diff contact with a file their globs claim. Specimen 2 was found by a mechanical target-mismatch sweep of all three corpora (485 corpus + 5 + 1 pack rules) run before building the guard; the operator confirmed it in-class per-specimen. One re-seal covers both edits (`output_hash be747bd8`). Enforcement of the secret-exposure class stays with the regex twin `5bc302726128eceb` (severity `error`).

2. **`feat(lint)`** — the Prop 309 **Class 7 load-time guard**: `assertRulesCanTargetTheirGlobs()` at the #2499 seam converts the diff-dependent detonation into a deterministic load-time failure on every invocation, naming each offender's lessonHash, globs, unregistered extensions, and three deliberately un-ranked remedies (the manifest alone cannot distinguish a permanently-broken rule from an uninstalled pack language).

## Design decisions

- **Fail-closed by default, with the pre-existing escape**: strict `--ast-parse-mode` (the default) throws; `lenient` degrades to a warning with full accounting. Rationale: a repo whose rules target a pack language it hasn't installed (e.g. `.rs` without the rust pack) is target-mismatched by environment, not by corpus rot — and `lenient` already means exactly "unparseable AST language degrades this run". An unconditional guard would remove that documented escape.
- **Both engines covered** (`ast` ∪ `ast-grep`) — matching the `allAstRules` union that feeds the abort seam.
- **Permitted shapes**: globless rules (documented `Lang.Tsx` fallback, cannot arm the abort) and extensionless positive globs (`src/**` legitimately dispatches `.ts` at runtime — absence of a trailing extension is absence of evidence, not proof of mismatch). A rule is condemned only when **every** positive glob names a concrete extension and **none** resolves.
- **`TRAILING_EXT_RE` exported from `@mmnto/totem`** so the guard and the dispatcher it protects share one regex and cannot silently disagree.

## Verification

- **14 induced-failure + control tests** colocated with the #2499 suite: hard-error message contract, `engine:'ast'` variant, fires-on-unrelated-diff (the determinism property), multi-offender accounting, lenient/strict routing, and negative controls (archived specimen, globless, one-resolvable-among-unresolvable, extensionless/brace-set globs, regex-engine `.json` rule, all-negation glob list) + a real-corpus anchor with a non-vacuity assertion (191 globbed AST rules actually inspected).
- **Out-of-suite falsifier run**: un-archived both real specimens in a corpus copy and ran the built dist — the guard threw naming both hashes, all five globs, `.json`, and the remedies. It catches the actual landmines, not just synthetic fixtures.
- **Gate**: `pnpm -r build` 0 · `pnpm -r test` 0 (7,278 tests / 303 files) · `format:check` 0 · `pnpm run lint` 0 · `totem lint` PASS 0 errors. Post-archive active-rule count drops 387 → 385, visible in the branch lint verdict.

## ADR-115 §1 in-boundary verdict

**Not in-boundary.** The diff touches the lint rule-load path and the rule corpus; none of the six named boundaries (`search` · `sync` · `init` · `dependency-installation` · `ECL/mail` · `release plumbing`) is implemented by the touched code. Stated per §1 so the default does not have to fire.

**ADR-115 §4 quota**: this PR is the Class 7 qualifying-item candidate for the next Version-Packages train. Evidence chain for the mmnto-ai/totem-strategy#974 row: the sweep artifact (scan of 491 rules across 3 corpora, 2 live specimens found), both archive edits with curated `archivedReason`s, the 14-test induced-failure suite, and the out-of-suite falsifier run above.

## Known residual (disclosed, not silently dropped)

Rule `fea10369daa04c98` is a **partial** mismatch: globs include `**/*.mts`, and `.mts` has no registered language (built-ins register `.mjs`/`.cjs` but not `.mts`/`.cts`). The guard correctly permits it — one glob resolves — but a diff touching a `.mts` file still arms the original run-wide abort. Class 7 as specified covers total mismatch only. Follow-up candidate: register `.mts`/`.cts` as built-ins (same grammars as the registered pair; their absence reads as an oversight) or extend the guard to warn on partial mismatch.

### DIFF
--- a/.changeset/target-mismatched-rule-load-guard.md
+++ b/.changeset/target-mismatched-rule-load-guard.md
@@ -1,0 +1,18 @@
+---
+'@mmnto/totem': minor
+'@mmnto/cli': minor
+---
+
+fix(lint): target-mismatched AST rules fail loud at rule load instead of aborting a future lint
+
+An `ast-grep` (or `ast`) rule whose declared `fileGlobs` all point at extensions with no registered Tree-sitter language can never execute — and it did not fail quietly. `resolveAstGrepLangs` silently falls back to `Lang.Tsx` when no glob resolves, so the rule looked healthy at load; `rule-engine.ts` then threw `TotemParseError` the first time a diff happened to contain a file the rule's globs claimed, and that throw escaped the per-file loop and aborted the **entire** lint. The observed specimen was a rule scoped to four `.json` paths. Diff-dependent detonation is the worst property of the class: the gate reads green on every run until the one run where it doesn't.
+
+`totem lint` now validates every loaded AST rule at load time, adjacent to the corpus-bearing gate from mmnto-ai/totem#2499 and with the same fail-closed posture. A rule with non-empty `fileGlobs` whose positive globs all name concrete extensions, none of which resolves to a registered language, throws a `TotemError` naming each offending `lessonHash`, its engine, its globs, the unregistered extensions, and all three remedies (sync the pack that provides the language, archive the rule, or fix its globs/engine) alongside a snapshot of the currently-registered extensions. The check runs on every invocation regardless of the diff, and both `'ast-grep'` and `'ast'` rules are covered because `rule-engine.ts` aborts on the union of the two.
+
+The pre-existing `--ast-parse-mode lenient` operator escape (mmnto-ai/totem#1982) is honored: in lenient mode the guard renders the same accounting as a warning and lets the run continue, since the offending rules are inert either way. Without that branch the guard would remove the only way to lint a repo whose rules target a pack language it has not installed yet. The strict default is unchanged and still fails closed.
+
+Two shapes stay permitted: **globless** rules keep the documented `Lang.Tsx` fallback (unscoped pre-1.16 rules, which `rule-engine.ts` never throws for), and **extensionless positive globs** (`src/**`, `**/*.{ts,json}`) are treated as resolvable — they match `.ts` files at run time, so absence of a trailing extension is absence of evidence, not proof of mismatch. Lifecycle status needs no handling: `loadCompiledRules` already drops archived / pending-verification / untested-against-codebase rules before the guard sees them, which is what makes archiving a working remedy. Pack-contributed languages are honored — `bootstrapEngine` runs before the check, so a pack that registers `.rs` has already populated the registry.
+
+`@mmnto/totem` additionally exports `TRAILING_EXT_RE` so the guard extracts target extensions exactly the way `resolveAstGrepLangs` does; a divergent copy would make the guard disagree with the dispatcher it protects.
+
+Class 7 of the Prop 309 hardening program (mmnto-ai/totem-strategy#971).
--- a/.totem/compile-manifest.json
+++ b/.totem/compile-manifest.json
@@ -1,7 +1,7 @@
 {
-  "compiled_at": "2026-07-12T21:12:45.328Z",
+  "compiled_at": "2026-07-28T20:50:46.919Z",
   "model": "gemini-3-flash-preview",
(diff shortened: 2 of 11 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:

Round composition: CodeRabbit 2 actionable majors (both **FIXED** in `31c53b5f`), Greptile confidence **5/5** with zero findings, Gemini Code Assist zero findings.

1. **[Security/Major] Sanitize manifest-derived fields in guard messages** — FIXED. The rendered summary (which embeds `lessonHash`, `lessonHeading`, and `fileGlobs` from the manifest) now passes through `sanitizeForTerminal` (the existing core export) before reaching both the lenient warning and the strict error; the remedy line is engine-derived only and unchanged. Verified by the file's 71-test suite — the guard's message-contract assertions pass unchanged, since sanitization strips control sequences and passes normal text through.

2. **[Data-integrity/Major] Receipt renders clean while validation ran downgraded** — FIXED, and accepted as the sharpest finding of the round: it is the ADR-115 §2 envelope discipline ("degraded state that renders indistinguishably from verified state is the defect class") applied to this PR's own artifact. The receipt now records `astParseMode` — **added to PINNED_FIELDS**, since a mode change is not a reproduction — and `targetMismatchGuardWarning` derived from the child's output streams. The committed `docs/data/lint-receipt.json` is regenerated carrying both (`"lenient"`, `true`), and `gen-lint-receipt.mjs --verify` reproduces: 387 rules, 0 errors, 20 warnings over 41 files, exit 0.

Adjacent hardening in the same file: worktree cleanup is now best-effort (rmSync with retries, warn-and-continue on failure) — a Windows handle lag was failing receipt runs in cleanup after the attestation work had already succeeded.

Coverage note for the record: CodeRabbit's path filter excluded `.totem/compiled-rules.json`, so the corpus archive edits in this PR were reviewed only by Greptile (5/5, archives explicitly listed in its file table) and Gemini (no findings) — stating it so the round's coverage is legible rather than assumed.

Local lane: no verdict artifact is recorded for this branch lineage, so there is no `local-lane:` line to carry — noted per protocol rather than omitted silently.

satur8d:
@coderabbitai review

satur8d:

CodeRabbit's one round-2 finding (minor) is **FIXED** in `254d04db`:

**Posture-field validation in the LINT_RECEIPT transform** — ACCEPTED + FIXED essentially as proposed, because the premise checks out and matches the transform's own house style (every other receipt field already fails loud rather than rendering garbage): an unknown `astParseMode` no longer silently hides the posture (throws), and a lenient receipt without a boolean `targetMismatchGuardWarning` no longer interpolates `undefined` into the page (throws). Legacy receipts with no posture fields render unchanged. Four new test cases cover the quadrant (legacy renders bare / lenient+boolean discloses / unknown mode throws / lenient-without-boolean throws); 20/20 docs-transform tests; `docs:inject` remains a no-op on the committed pages.

All other round-2 blocks were LGTM (guard import/sanitize path, receipt tool, regenerated receipt, maturity render). No declines this round.

Issue #2509, opened 2026-07-28
mail: own-broadcast suppression renders with no accounting line — poll output must report dropped items, never silently


satur8d:
**Class parent:** #2462 (silent mail-suppression class; specimen 1 diagnosed at `packages/cli/src/commands/mail.ts:770` — the dirs-union own-broadcast exclusion). This ticket names a gap #2462's diagnosed fix does not automatically close: **the poll render gives no account of what the exclusion dropped.** Filed per strategy-claude's GO dispatch (totem-strategy ECL `2026-07-23T1913Z`), specimen held first-party by strategy-kimi.

## Specimen 3 — first-party, two firings, one reproduced live today

**Firing 1 (2026-07-23, totem-strategy):** strategy-claude's Prop 309 broadcast (`2026-07-23T0833Z-broadcast-309-hardening-round.md`) never reached this seat's default poll. It surfaced only via a directed pointer (`0835Z`) whose subject line literally carries the human workaround: *"broadcast may be suppressed for you: mmnto-ai/totem#2462"*. The guard against the silent class is currently a sender-side habit, not a tool behavior.

**Firing 2 (2026-07-28, reproduced live, `@mmnto/cli` 1.105.0, workspace `D:\Dev`, repo totem-strategy):**

- Default poll (`totem mail`; self = 5 resident seats incl. strategy-claude): renders **"45 unread"** — but `2026-07-24T2039Z-broadcast-round-second-contact.md` and `2026-07-23T1911Z-broadcast-309-ruled.md` (both from resident sender strategy-claude, both **unread for strategy-kimi**, zero processed marks) are **absent** from the render.
- Seat-scoped poll (`TOTEM_SELF_AGENT=strategy-kimi totem mail`): both present (21 unread).
- lc-claude broadcasts **do** render in the default poll (sender not resident in this repo) — confirms the resident-sender mechanism, not a broadcast-wide filter.

## The gap

The default render's "45 unread" silently omits items unread for 4 of 5 resident seats. Nowhere in the output is the suppression accounted for — no count, no note, no hint to re-poll seat-scoped. The only place the suppression is mentioned anywhere is a human's subject-line workaround. This is the same lie-class as the poll-warnings render gap (mmnto-ai/totem-strategy#970): a sensor asserting a clean/complete view over dropped data (Tenet 4 adjacent).

## Ask

[...]

Issue #2462, opened 2026-07-20
mail: silent-suppression class — dirs-union own-broadcast exclusion hides broadcasts from non-sender resident seats; second specimen: hook poll false-negative on directed mail


satur8d:
# Silent mail-suppression class: two specimens, one diagnosed

**Routing:** totem-kimi's sensor dispatch (`.totem/orchestration/totem-kimi/outbox/2026-07-20T0035Z-strategy-claude-mail-2364-dirs-union-broadcast-suppression.md`) asked strategy-claude to triage + rule on fix shape. Ruling: **dedicated issue** (this one), not a mmnto-ai/totem#2204 scope amendment — mmnto-ai/totem#2204 shrinks the blast radius but does not repair the coordinator-view semantics, so the class needs its own ticket. Filed by strategy-claude on operator authority, 2026-07-20. Specimen 1 findings are **totem-kimi's work**, condensed here with receipts in the dispatch.

## Specimen 1 — DIAGNOSED (totem-kimi): own-broadcast exclusion × dirs-union

`packages/cli/src/commands/mail.ts:770`:

```ts
if (opts.includeProcessed !== true && toLower === 'broadcast' && selfLower.has(senderSeat)) continue;
```

The mmnto-ai/totem#2364 exclusion assumes SELF = the single polling seat. Post-mmnto-ai/totem#2141 dirs-union, SELF is a **set** — the default poll from mmnto-ai/totem resolves six resident seats because visiting seats' dirs are resident in the orchestration tree. `senderSeat ∈ self` flipped meaning from *"I sent this"* to *"some resident seat sent this"*, so a resident sender's broadcast is suppressed from **every** seat's default render. Silent — no warning rides the channel (Tenet 4 adjacent).

**Live fire:** strategy-claude's Proposal 308 brainstorm broadcast (`2026-07-19T2307Z`) was invisible in the default poll from mmnto-ai/totem (143 unread rendered, specimen absent, grep-verified vs `--json`) while unread for 3 of 6 resident seats (kimi/gemini/agy — zero marks). It surfaced only under `TOTEM_SELF_AGENT=totem-kimi totem mail`. This contradicts the poll's own multi-seat unread contract ("unread for at least one of my seats", mail.ts:346-347) — a false assertion whenever a resident sender has live broadcasts.

[...]

Issue #2364, opened 2026-07-14
fix(mail): a seat's own outbound broadcast surfaces as its own unread — add source-outbox self-exclusion in pollMail


satur8d:
## Summary

A seat's **own outbound broadcast counts as its own unread mail**: `pollMail` filters by recipient only (`mail.ts` ~L639 — `to: broadcast` or a SELF agent), and nothing excludes dispatches whose *source outbox is the polling seat itself*. A seat that broadcasts is therefore permanently "1 unread" to itself until it hand-backfills a `processed/_broadcast/` mark for its own dispatch.

## Exhibit (verified 2026-07-14)

strategy-claude's `2026-07-06T2326Z-broadcast-ecl-31-round-reply-seatshort.md` sits in their own outbox with **no** `processed/_broadcast/` mark, and shows as "1 unread for strategy-claude" in every strategy-seat orientation.

## Diagnosis correction vs the reporting dispatch

The dispatch framed this as "the orient hook doesn't apply the same self-filter as `pollMail`" — that mechanism is falsified: **`pollMail` has no self-sender filter to apply** (verified against `mail.ts`; `from` is surfaced but never compared to self), and strategy's `SessionStart.cjs` delegates to the canonical `pollMail()` from `@mmnto/cli`. Any hook-vs-CLI divergence in what the operator observed is a resolution-context difference (self-agent set / repoRoot for the processed-dir subtraction), not a missing hook-side filter. The defect is at the `pollMail` layer and every consumer inherits it.

## Proposed fix

Exclude a dispatch from the unread set when its **source outbox belongs to a SELF agent** (`slot.agent ∈ selfLower`) and `to: broadcast`. Keying on the outbox directory — single-writer filesystem truth — matches the basename-collision sensor's own doctrine in the same function (never the forgeable `from:` header). Directed dispatches addressed *to* a SELF agent from a SELF outbox (self-mail) are out of scope here; broadcasts are the observed noise class.

Alternative considered: self-marking broadcasts into `processed/_broadcast/` at `mail send` time — worse, it needs a backfill for existing dispatches and leaves hand-authored outbox files uncovered.

## Provenance

strategy-claude dispatch `2026-07-14T1821Z-totem-claude-lesson-recall.md` (rider), triaged accept-with-corrected-diagnosis 2026-07-14. Related: #2166 (mail filename composition), #2204 (seat-scoped selectors), mmnto-ai/totem#2355 (same scan, symlink guard).

PR #2515, opened 2026-07-29
test(mail): Prop 309 P2 entry E4 — induced-failure tests for mail poll completeness

satur8d:
## Mechanical Root Cause

Prop 309's ruled specimen class is *degraded paths presenting as success*. The ECL/mail poll boundary carried that class live: the own-broadcast exclusion (`mail.ts`, post-dirs-union multi-seat SELF) silently drops a resident sender's broadcast from every resident non-sender seat's default render — "No unread mail" over waiting mail (#2462, first-party specimen #2509). The boundary had no induced-failure test asserting that poll degradation is **surfaced or loudly accounted, never silently narrowing**. This PR is P2 degraded-mode test pack **entry E4** (seed registry: mmnto-ai/totem-strategy#974 index comment).

## Fix Applied

Test-only PR — no production code changed. Adds `packages/cli/src/commands/mail-degraded-e4.test.ts` (8 tests) covering the three chartered faults against the structured `MailPollResult` (the durable hook contract):

1. **Basename collision** — regression lock: current code surfaces BOTH colliding dispatches AND emits a `cross-sender basename collision` warning naming every seat path. Cross-repo variant (the in-repo same-repo case is covered in `mail.test.ts`).
2. **Own-broadcast exclusion** — the **red row**: `test.fails` written against the REQUIRED #2509 behavior (render the broadcast to the non-sender seat OR account for the suppression in the warnings channel). It executes every run — suite green while the defect stands, loudly red the day the fix lands, forcing marker removal. (`test.fails` because vitest 4 dropped `test.fixme`; ticket ref in title + comments per repo rule.) Also pins: positive control (non-resident sender broadcast surfaces), fault-removed control (`TOTEM_SELF_AGENT` scoping surfaces the identical dispatch), and an over-correction guard (a seat's OWN broadcast stays excluded per shipped #2364 behavior — the #2509 fix must not over-correct).
3. **Glob/scan failure** — ENOTDIR workspace root (no mocks; real errno drives the real catch arm) and EACCES on one repo's orchestration dir (path-conditional fs spy): the failure is NAMED in the warnings channel, healthy siblings still surface, no mail fabricated over an incomplete scan, and the deliberate exit-0-with-envelope contract is pinned per ADR-115 §2 (read-only degraded paths announce in rendered output; the `warnings` channel IS that envelope). Fault-removed-returns-to-green asserted in both.

## Out of Scope

- The #2509 accounting-line fix itself (totem lane; the red row is written to receive it).
- The verdict-line question (poll still prints `No unread mail addressed to …` after `Warning:` lines on an incomplete scan — judged ADR-115 §2-compliant as-is; unticketed hardening candidate, flagged for the lane).
- The other five P2 seed entries (E1–E3, E5, E6 — per-lane, indexed on mmnto-ai/totem-strategy#974).

## Tests Added/Updated

- [x] Added/updated automated tests covering the root-cause mechanism — `mail-degraded-e4.test.ts` (8 tests; 7 green + 1 expected-failure). Local: new file 8/8; mail suite 154/154; full `@mmnto/cli` suite 3592/3592. prettier/eslint/tsc clean.

## Related Tickets

Refs #2462 · Refs #2509 · pack index: mmnto-ai/totem-strategy#974 · program: mmnto-ai/totem-strategy Proposal 309 / ADR-115

Reviewer flag: @totem-claude — per the bus disposition (strategy ECL 2026-07-28T1936Z), this lane owns review/merge; the PR vehicle keeps single-writer intact.

### DIFF
--- a/packages/cli/src/commands/mail-degraded-e4.test.ts
+++ b/packages/cli/src/commands/mail-degraded-e4.test.ts
@@ -207,9 +207,13 @@ describe('E4 fault 2 — own-broadcast suppression is never silent (mmnto-ai/tot
       writeOutbox('totem', 'totem-claude', [
         { name: BCAST, to: 'broadcast', subject: 'cohort note' },
       ]);
-      // Sanity: the fixture resolves BOTH resident seats on the default path.
+      // No sanity assertion in this body, deliberately: `test.fails` is
+      // satisfied by ANY failure, so an extra assertion here would let an
+      // unrelated seat-resolution regression keep the marker green without
+      // the #2509 assertion ever running (CR round-1 catch). The two-seat
+      // resolution is asserted by the positive control below on the same
+      // fixture shape, where a regression fails loudly as a real failure.
       const result = pollTwoSeat();
-      expect(result.selfAgents.agents).toEqual(['totem-claude', 'totem-gemini']);
 
       // Pack 1 + 2 — accounting fires / degraded envelope: the poll must
       // either render the dispatch (to the non-sender seat) or name the
--- a/packages/cli/src/commands/mail-degraded-e4.test.ts
+++ b/packages/cli/src/commands/mail-degraded-e4.test.ts
@@ -1,0 +1,364 @@
+/**
+ * Prop 309 P2 degraded-mode test pack — entry E4 (induced-failure).
+ *
+ * Boundary (verbatim scope): ECL/mail poll completeness. Induced faults →
+ * required behavior: "basename collision · own-broadcast exclusion · glob
+ * failure → surfaced or LOUDLY accounted, never silent narrowing."
+ *
+ * Pack contract — every induced-failure test carries three assertions:
+ *   1. The accounting fires (degraded output announces itself — never output
+ *      indistinguishable from a clean poll).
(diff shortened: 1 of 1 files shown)


satur8d:
@coderabbitai review

satur8d:
## Round 1 disposition (totem-claude, reviewing lane)

CodeRabbit's one finding (trivial-tier) is **FIXED** in `aae04bef`, and it earned more than its tier: `test.fails` is satisfied by ANY failure, so the seat-resolution sanity assertion ahead of the #2509 assertion meant an unrelated resolution regression would keep the expected-fail marker green without the tripwire ever executing — a mask on the exact coupling the red row exists to provide. The sanity assertion is removed (the sibling positive control asserts the identical resolution on the same fixture shape, where a regression fails loudly), and an in-body comment records the single-assertion constraint so a future edit does not reintroduce it. 8/8 tests; the red row now fails only through the #2509 required-behavior assertion.

Coverage record for the round: Gemini Code Assist reported it cannot process these file types (no verdict); greptile not invoked per quota discipline on a test-only diff.

@coderabbitai review

PR #2523, opened 2026-07-30
fix(hooks): emit SessionStart context via the documented hookSpecificOutput envelope

### LINKED ISSUE #2522: fix(hooks): SessionStart context injection silently dropped for >=12 days — session-context.mjs emits top-level additionalContext; harness ingests only the hookSpecificOutput envelope
## Symptom

Session starts on this repo boot cold: no journal carryforward, no inbound-mail block, no orient block from the SessionStart hook. Journal claude-0190 counted a "4th consecutive cold start"; transcript forensics below show the real streak is >=12 days — every scanned session since 2026-07-17.

## Evidence (session-transcript forensics, 25 sessions, 2026-07-17 → 2026-07-29)

Each Claude Code session transcript records SessionStart hook execution as an attachment row. For every scanned session across harness versions 2.1.212 → 2.1.220:

- `attachment.type = hook_success`, `hookName = SessionStart:startup` — the hook IS invoked and exits 0.
- `attachment.stdout` carries the full ~9-10KB JSON the hook printed — output IS captured.
- `attachment.content = ""` — the harness ingested nothing into model context. `contentLen=0` in 25/25 scanned sessions.

Control that isolates the variable: a sibling plain-text SessionStart hook (`[[Describe]] Project: ...`) present in three 07-23/07-24 sessions was ingested verbatim (`contentLen = stdoutLen - 1`) in the same sessions where this hook's JSON got `contentLen=0`. Invocation path healthy, plain-text ingestion healthy — only the JSON envelope is dropped.

Manual run today (2026-07-30): exit 0, 7.4s, 9,341-byte valid single-key JSON, zero stderr. The script itself is healthy.

## Root cause

`.claude/hooks/session-context.mjs:390`:

```js
// Claude Code hook protocol: JSON with additionalContext field
const output = JSON.stringify({ additionalContext: fullContext });
```

The comment is a stale protocol understanding. Per the current hooks documentation (code.claude.com/docs/en/hooks), JSON stdout adds context only via the event-scoped envelope:

```json
{ "hookSpecificOutput": { "hookEventName": "SessionStart", "additionalContext": "..." } }
```

A top-level `additionalContext` is not part of the protocol; unrecognized JSON means the context is silently dropped while the hook still reports success — the silent-positive class (#2516 / #2501's territory), one layer up in the stack.

## Fix

One-line envelope change plus comment correction. Plain-text stdout would also work, but the documented envelope is the stable contract.

## Verification contract (artifact-anchored)

After the fix: start a fresh session, read that session's transcript SessionStart record, assert `attachment.content.length > 0` and that it matches the emitted context. This artifact cannot be produced without the fix actually landing.

## Blast radius

- **This repo:** single file. `totem init` templates / `AI_PROMPT_BLOCK` do NOT ship this pattern (grepped `packages/` — zero hits), so no consumer-repo exposure.
- **Cohort:** `mmnto-ai/liquid-city` has the identical emission at `.claude/hooks/session-context.js:315` — lc seats have the same silent drop. Their lane; dispatch sent to lc-claude.
- **Downstream premise correction:** #2501's "session-start auto-context injects unfiltered top-5" describes intent; in practice nothing from this hook reached the model for >=12 days. The relevance-floor work and this fix compose; #2501's problem statement holds but its "currently injecting" premise does not.
- **Cold-start undercount:** the journal's "4 consecutive" figure undercounted because a separate small describe-hook's context made some starts look warm, and missing context was attributed to other causes.

Related: #2501 (same surface, relevance floor) · #1997 (same surface, budget) · #2487 / #2488 (vendor parity: the Gemini hooks speak a different protocol — worth the same "does the context actually land" transcript-level check, same silent class).

satur8d:
## What

One-line protocol fix: the SessionStart hook now emits its context inside the documented `hookSpecificOutput` envelope (`{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":...}}`). Header and call-site comments corrected to match.

## Why

A top-level `additionalContext` key is not part of the hook protocol. The harness invokes the hook, records `hook_success`, captures the full stdout — and injects nothing. Transcript forensics on #2522: 25/25 sessions from 2026-07-17 → 2026-07-29 (harness 2.1.212 → 2.1.220) show `attachment.content=""` for this hook, while a sibling plain-text SessionStart hook ingested verbatim in the same sessions. Full evidence chain and docs quote on the issue.

## Verification

- Fixed script, degraded run (unbuilt worktree): exit 0, envelope `hookSpecificOutput.hookEventName="SessionStart"`, context present; graceful-degradation stderr paths unchanged by the diff.
- Fixed script, full run (against built workspace dist): exit 0, correct envelope, 10,015-char context — the ADR-013 10k budget cap engaged as designed.
- Prettier clean. `totem lint --base main` clean — the file is outside lint scope per `ignorePatterns`, and the filter is accounted in the lint output (`Filtered 1 file(s)`), not silently skipped.
- **Post-merge artifact (the issue's verification contract):** the first fresh session on merged main must show a SessionStart transcript record with `attachment.content.length > 0`. That artifact cannot be produced unless the fix works; to be verified at the next session start and recorded on #2522.

## ADR-115 §1 verdict

**Not in-boundary** — the diff changes only the repo-local Claude hook's stdout envelope (harness protocol surface); no runtime change to search / sync / init / dependency-installation / ECL-mail / release-plumbing surfaces (the hook consumes `pollMail`/`orient`, whose implementations are untouched).

## Release note

No changeset: repo-local `.claude/hooks/` tooling, not part of any published package.

Closes #2522

### DIFF
--- a/.claude/hooks/session-context.mjs
+++ b/.claude/hooks/session-context.mjs
@@ -3,7 +3,7 @@
  * SessionStart hook V2 — combines static filesystem context with
  * LanceDB vector search for relevant knowledge injection.
  *
- * stdout → agent context (JSON with additionalContext field)
+ * stdout → agent context (JSON: hookSpecificOutput.additionalContext envelope)
  * stderr → diagnostics only
  *
  * Budget: ~2-3k tokens max (ADR-013).
@@ -386,8 +386,13 @@ async function main() {
       ? combined.slice(0, MAX_TOTAL_CHARS) + '\n...(truncated)'
       : combined;
 
-  // Claude Code hook protocol: JSON with additionalContext field
-  const output = JSON.stringify({ additionalContext: fullContext });
+  // Claude Code hook protocol: SessionStart context is ingested ONLY from the
+  // hookSpecificOutput envelope. A top-level additionalContext key is not part
+  // of the protocol — the harness records hook_success, then silently injects
+  // nothing (mmnto-ai/totem#2522: every session 2026-07-17 → 07-29 booted cold).
+  const output = JSON.stringify({
+    hookSpecificOutput: { hookEventName: 'SessionStart', additionalContext: fullContext },
+  });
   process.stdout.write(output);
 }
 


satur8d:
/gemini review
@greptileai review

PR #2524, opened 2026-07-30
feat(instrumentation): query-before-derive compliance slice — Trap Ledger correlation IDs + doctor --compliance + one pre-registered threshold

### LINKED ISSUE #2510: feat(instrumentation): query-before-derive compliance slice — Trap Ledger correlation IDs + doctor --compliance + one pre-registered threshold
## Charter

Operator-greenlit 2026-07-28 (ruling session recorded on mmnto-ai/totem-strategy#974; charter dispatched by strategy-claude 2026-07-28T2002Z). **The slice shape is strategy-kimi's** (sequencing position 2026-07-28T1921Z §1, adopted over strategy-claude's recommendation): one falsifying number wired end-to-end, because a telemetry *stack* lands as a program and programs slip — a single number either exists or doesn't, and the instant it exists every `Goal:` prefix on the adherence claims becomes honest-or-falsified instead of honest-by-courtesy (Tenet 19).

**Explicitly NOT chartered:** the ADR-014 / ADR-029 / ADR-044 / ADR-078 telemetry stack. This issue is one metric, end-to-end, nothing else.

## The slice

**Query-before-derive compliance**: of the derive-class actions in an agent session (spec synthesis, orientation derivation, review grounding), what fraction was preceded by a corpus query (`search_knowledge` / `totem search`) correlated to that action?

1. **Trap Ledger schema extension** — event rows for query events and derive events.
2. **Correlation IDs** joining query → derive inside a session window, **minted at event-write time** (see falsifier 2).
3. **`totem doctor --compliance` render** — the number and its trend.
4. **One pre-registered threshold** (next section).

## Pre-registration (Prop 309 discipline — named before the sensor first emits)

**Threshold, registered now:** compliance **≥ 0.50**, evaluated over **the first 20 instrumented sessions carrying ≥1 derive-class event, regardless of query count** (window pinned per the strategy-claude pre-registration audit, 2026-07-28 comment below — the earlier "correlated sessions" phrasing was post-hoc-interpretable as excluding zero-query sessions, which would enact falsifier 1 at the window level). Below the floor at the checkpoint: the adherence claims' `Goal:` prefixes are recorded as falsified for that window and the number leads the next signoff — no post-hoc reinterpretation of the threshold (§6.5 exit-menu discipline).

**Falsifier set for the metric itself** (how this metric would be gamed or go ceremonial):

1. **Denominator gaming** — compliance counted only on sessions that already query. The denominator MUST be all derive-class events, including sessions with zero queries.
2. **Post-hoc correlation** — IDs attached retroactively to dress uncorrelated queries as compliant. IDs must be minted at event time; a backfilled ID is a schema violation, not a data point.
3. **Banner-fade / wallpaper** (Prop 309 falsifier vocabulary) — the render exists but stops being consumed. If the rendered number produces no behavioral delta over a named window, the metric is ceremonial and this issue's approach is falsified, not merely "needs iteration".
4. **Ritual query (Goodhart)** — a query fired to satisfy the metric with its results unreferenced by the subsequent derive. Tell: query event with zero result-consumption linkage. v1 may only be able to sense this weakly; name it rather than claim it solved.

## Constraints

- **Sensor, not actuator (Tenet 13):** v1 renders and reports; nothing blocks on it.
- **Boundary:** the render path may touch `search`/ECL surfaces — the implementing PR body carries the ADR-115 §1 in-boundary verdict line either way (fail-closed default applies if omitted).

## Motivating exhibit

The #2496 retraction: AGENTS.md carried the RRF fact, and three seats misread the same `0.016` render inside 24 hours anyway. Adherence — query-before-derive — is the bottleneck, and it is currently unmeasured. This slice is the falsifiability layer under the evidence ladder greenlit on mmnto-ai/totem-strategy#644 (the fresh-seat drill measures behavior once; this measures adherence continuously).

satur8d:
## Summary

One falsifying number, wired end-to-end: **of the derive-class actions in an agent session (spec synthesis, orientation derivation, review grounding), what fraction was preceded by a corpus query correlated to that action?**

`totem doctor --compliance` now renders it, alongside the pre-registered threshold and window it will be judged against. Nothing gates on it.

Live on this branch, against this repo's real ledger (450+ pre-existing rows from other subsystems):

```
[Compliance] Query-before-derive compliance
[Compliance] compliance: 0.00 (0/1) over 1 instrumented session(s)
[Compliance] pre-registered threshold + window: compliance ≥ 0.50, evaluated over the first 20 instrumented sessions carrying ≥1 derive-class event, regardless of query count
[Compliance] verdict: PENDING — 1/20 instrumented sessions recorded; the threshold is not evaluable until the window fills
[Compliance] trend: insufficient data (needs at least 4 instrumented sessions)
[Compliance] limitation: this senses query-before-derive ADJACENCY, not influence — a query whose results the derive never read still counts as compliant (ritual-query / Goodhart). Named, not solved.
```

That `0.00 (0/1)` is a real measurement, not a fixture: a `totem orient --json` run with no preceding query, correctly recorded as an uncorrelated derive. The sensor's first reading counts against us, which is the point.

## ADR-115 § 1 verdict

**IN-BOUNDARY** — the diff adds runtime instrumentation to the `search` surface: `packages/cli/src/commands/search.ts` (the `totem search` execution path) and `packages/mcp/src/ledger-writer.ts` (the `search_knowledge` MCP call path).

ADR-115 § 2 therefore applies; degraded-path handling and its induced-failure tests are itemised under **Falsifier handling** and **Test evidence** below.

## Falsification leg — round 1

An independent leg reviewed this PR and found 2 CRITICAL and 6 MAJOR defects, all confirmed at source. Every one is fixed on this branch. Two of them were the metric silently failing to measure, which is worse than measuring badly — the number would have looked fine.

**Findings absorbed**

1. **CRITICAL — the scanner was blind to backfilled IDs on non-QBD types.** The "belongs to another subsystem" fast path ran *before* the QBD-marker test, so a row claiming `type: mcp_call` while carrying a forged `qbd_correlation_id` was dropped with zero anomalies and the whole ledger rendered clean. That is precisely the tampering ADR-115 § 2 exists to make impossible to miss. The marker test now wins on **any** type; a schema-failed row with no QBD markers is still not our business.
2. **CRITICAL — every seam resolved the ledger from `cwd`.** `resolveConfigPath` does not walk up, so from a subdirectory the command ran fine (global-profile config) while the seam targeted `<cwd>/.totem`, which does not exist — the derive vanished from the denominator. Falsifier 1 arrived at by accident rather than by gaming. All four seams *and* the doctor reader now route through one shared helper (`qbd-seam.ts`) resolving config-root-based, exactly as this command family's existing ledger writes do. Writer/reader divergence is now impossible by construction — that agreement is the load-bearing property, and it is what a test pins.
3. **MAJOR — `totem review --mode structural` escaped the denominator.** That branch `return`s before the old seam site, so the seam's own comment ("both paths pass through here") was false, and the PR body and wiki repeated it. The seam is hoisted above the fork; the claim is now true and pinned by a source-order test.
4. **MAJOR — the correlation pointer was never consumed.** One query credited every derive for two hours (measured: 1 query → 10 derives → compliance 1.00). That measures "queried once per two hours", not "queried before deriving". Now **consume-on-use**: one query grounds exactly one derive.
5. **MAJOR — correlation was cross-agent and fail-open on session id.** A missing id on either side counted as a match, and `agent_source` was recorded but never consulted — so agent A's query could ground agent B's derive, which is live rather than theoretical because cohort seats share one working tree. Now fail-closed: both session ids present and equal, **and** matching seat. Rolling-window grouping partitions by `agent_source` too.
6. **MAJOR — backdated appends could seize the first-20 window** (measured: 40 backdated 2019 rows flipped the verdict to PASS with `degraded=false`). The ledger is append-only and `mintQbdCorrelationId` takes its instant as an argument, so no whole-file rewrite was needed. The scanner now runs an append-only monotonicity check; regressing timestamps degrade the read.
7. **MAJOR — the MCP `corpus_query` row minted at handler entry**, before the health gate and before the search, so a blocked or thrown search still minted a groundable ID — inverting this PR's own CLI rationale. Decoupled: `mcp_call` stays at entry and ADR-029 is untouched; `corpus_query` mints only once the search has actually run. Zero hits still counts.
8. **MAJOR — degraded polarity was backwards.** `NON_QBD_TYPES` hand-mirrored the schema enum with no parity test, so an unknown type (benign version skew, live on this machine) flipped DEGRADED while real tampering stayed clean. The set is now derived from the schema's own type list with a parity test, and unknown types are a **named advisory** that does not flip DEGRADED.

**Design decisions taken:** consume-on-use (1 query : 1 derive) · agent-scoped, fail-closed correlation · append-only monotonicity as the cross-row integrity check · advisory-vs-degraded polarity split. Sticky whole-file degraded is kept deliberately — one bad row degrades the whole read, which is honest for a file whose integrity is in question.

**Corrected claims:** structural-mode reviews *are* counted (was falsely claimed, now true) · 61 → **96 tests** · the ADR-029 attributions below are reattributed: the 2-hour value is borrowed, but treating it as a *grounding-validity* window is this slice's own design call, and "the explicit primitive wins" is the `.session-id` module's contract, not ADR-029 text.

**Dispositioned differently than charted, with reason:** the charter asked for a subdirectory test asserting the row lands *in the project ledger*. It cannot, and I did not fake it. `resolveConfigPath` does not walk up, so from a subdirectory it resolves the global `~/.totem/` profile — the row follows the config the command itself ran under, which is coherent and matches every sibling ledger write, but is not project-root discovery. Making it walk up would change config resolution for *every* command, a repo-wide semantic change well outside this slice. So the test pins what deterministically holds — no stray `<subdir>/.totem` is ever created, and the writer always writes where the reader looks — and the residual is documented in `qbd-seam.ts` rather than hidden.

## Falsification leg — round 2

Round 2 confirmed the round-1 repairs were genuinely load-bearing, and found 4 MAJOR + 7 MINOR still standing. **Two of the majors were regressions my own round-1 fixes introduced** — both the same shape: a rule enforced on the write side with no read-side check, which is exactly the class the scanner fix was supposed to teach. That is the honest story of this round.

**Findings absorbed**

1. **MAJOR — consume-on-use had no read-side enforcement** (regression introduced in round 1). One query and three derives citing the same ID read as `compliance 1.00, degraded=false`: 1:1 existed only at write time while the module header asserted it in prose. Reachable with no adversary at all — via the `clearPointer` failure I had myself documented as "may over-credit until removed", the non-atomic read-then-clear race, or a duplicated line. The scanner now credits the **first** citation in time order and counts every re-citation as a `duplicateCorrelations` anomaly that degrades the read. The write-side race is now *accepted because* the read side catches its result, rather than hand-waved.
2. **MAJOR — backdate monotonicity was blind when no prior QBD rows existed** (regression introduced in round 1) — precisely the fresh-adopter state the pre-registered first-20 window is filled from. 20 backdated pairs on a ledger of ordinary non-QBD history read `PASS, degraded=false`, because the reference only advanced on QBD rows. It now advances from **every** parsed row: an adopted ledger's existing history is the baseline. A file authored wholesale from empty stays conceded scope, documented next to the coherent-rewrite adversary it belongs with.
3. **MAJOR — silent global-profile landing, and my test suite was polluting the maintainer's live ledger.** `~/.totem/totem.config.ts` sets `totemDir: '.'`, so from any config-less cwd the seam resolved the real global profile, existed, and recorded with no note. The tests inherited that: `qbd-seam.test.ts` resolved its orphan temp cwd to the **real** profile and wrote real derive rows into `~/.totem/ledger/events.ndjson` on every suite run (measured growing 22 → 39 during the audit). Three fixes: `homeDir` is injected through the resolver chain and every test passes a temp home, so no test *can* reach the real profile; a global landing now emits a named note; and the claim below is corrected. Verified empirically — the live ledger held at **39 rows across a full suite run** where it previously grew every time. **Not cleaned: reported.** Those 39 rows are mine and are still there for you to inspect or discard.
4. **MAJOR — the seat check is fail-open when both sides are unseated**, undisclosed, and one-sided seating silently kills correlation. Both-unknown correlating is the *right* default for a solo repo, so it is now disclosed rather than changed. The failure mode that matters is asymmetry — a seated CLI against an unseated MCP server yields `0.00` with no tell, and that number's pre-registered consequence is falsified adherence claims. The scanner now emits a **non-degrading** seat-mismatch hint when an uncorrelated derive had an in-window query from a different seat. MCP seat plumbing is out of slice scope and tracked in mmnto-ai/totem#2530.

**Minors, all fixed:** the structural-mode source-order test was theater — it indexed the *import* of `recordQbdDerive`, never the call, and a comment edit would have made it pass; replaced with call-expression + enclosing-control-flow pins and the mutation reasoning written beside them. The `ledger-type-parity` test was vacuous — its fixtures parsed, so `NON_QBD_TYPES` (consulted only in the failure branch) was never reached and an empty set would have passed; fixtures are now schema-invalid, with a discriminating control. Two seam assertions were non-discriminating or tautological; replaced with real property tests. `logCorpusQuery()` is now awaited so the pointer write cannot race the tool response across the process boundary. The review-seam hoist had **broadened the derive class undisclosed** — a throwing review recorded a derive, unlike `spec` — so after-success discipline is restored by recording in the shared verdict handler, with the fan recording separately after it converges. The changeset had never been propagated and still carried corrected-away claims; rewritten to current truth. The three accounting branches (`could not consume`, pointer write failure, pointer unreadable) now have positive tests. And DEGRADED, being sticky by design, now renders a remediation path and names its tolerance constant.

### Round 3 — the last five minors

The leg returned no merge-blocking findings; these five closed out convergence.

1. **Under DEGRADED, only the compliance figure was qualified.** `verdict: PASS` and the trend line rendered unqualified beside it. Every derived line now carries `(UNVERIFIED)`, with a render test for the degraded+verdict combination and a clean-scan control so the qualifier cannot become permanent decoration.
2. **A dead assertion in `qbd-seam.test.ts`.** The fixture profile sets `totemDir: "."`, so the old cwd-joined answer was the subdirectory *itself*, not `<subdir>/.totem` — the hard-coded path made `not.toBe` pass under both implementations. Now computed from the same `totemDir`, so it discriminates.
3. **`record.test.ts`'s header overclaimed.** "Real errnos, never by mocking the module under test" sat at file level while four tests stamp `EACCES`/`ENOSPC` through `vi.spyOn`. The method is right — `ENOSPC` is not inducible on a real filesystem — so the claim is now scoped to the two ADR-115 § 2 packs, matching what this body already said.
4. **The shield source pin's remaining hole is now disclosed:** these are substring matches, so a *comment* mentioning `recordQbdDerive(` satisfies them as readily as the real call. Written into the mutation reasoning rather than left implied.
5. **Headline test count corrected to 118**, the number the itemization sums to, measured per file. It had drifted three rounds running.

### Bot review round — 11 items

Two P1s and a critical were **live defects**, not style:

1. **The doctor reader fell back to `<cwd>/.totem` on unresolved config** (`doctor-qbd.ts`) — but the writers *skip* when config is unresolved, so the reader was reading a ledger no writer ever writes, contradicting its own comment and letting a stale directory render as this project's data. Fallback removed; unresolved is now its own SKIP with a named reason.
2. **`--raw` and unparsable-verdict runs recorded derives** (`shield.ts`). The record was the handler's first statement, ahead of both `if (options.raw) return` and the unparsable-verdict throw — so a context dump with no verdict at all recorded a derive and could consume a query's grounding. It now fires only once a verdict has **parsed**, for PASS and FAIL alike (a failing review is a completed derive; excluding it would shrink the denominator).
3. **An out-of-range mint instant crashed the whole scanner** (`correlation-id.ts`). The ID pattern admits 11 base36 digits (~1.3e17) against a max `Date` of 8.64e15, so a forged ID passed every guard and then threw `RangeError` from `new Date(...)` while building a violation string *inside* `LedgerEventSchema`'s refinement — and Zod does not trap refinement throws, so `safeParse` threw and one forged line killed the scan. Bounded in both decode and mint; the forged ID now flows through as a counted anomaly.
4. **A banned `process.stderr` write in `@mmnto/mcp`**, whose stdio transport it would corrupt. Routed through `logSearch` as an `internal:` synthetic entry — the idiom `search-knowledge.ts` already uses. (GCA's suggested `onWarn: logSearch` does not typecheck: `logSearch` takes a structured entry, not a string.)

Also: the errno is named in the unreadable SKIP; `report` shadowing in `orient.ts` renamed; `recordQbdQuery` gained unit coverage; `node:path` made lazy per the CLI rule; the two seam functions folded into one internal body so the load-bearing note strings are single-sourced; `QbdSession` exported with TSDoc through the barrel; and the seat-mismatch hint replaced with a bounded reverse walk (it was O(derives × queries) over the whole ledger).

### Bot review round 2 — two findings, both on round-1 bot-fix code

1. **The seat-mismatch hint false-positived on mixed-seat windows** (`compliance.ts`). The walk flagged on the first different-seat in-window query even when a same-seat query was also in the window — a case where the uncorrelated derive has an ordinary explanation and seat plumbing is not implicated. It now flags only when **no** same-seat query is in-window and at least one different-seat one is. Worth recording precisely: the `.some()` this bounded walk replaced had the *same* false-positive semantics, so this corrects a semantic that was wrong from the start, not a regression the rewrite introduced.
2. **Fan reviews that persist a verdict and then throw escaped the denominator** (`review-fan.ts` / `shield.ts`). `runReviewFan` saves the verdict artifact — the one `totem review --covariate` reads back, `settled=false` states included — *before* its zero-completed-lane and `--fail-on` policy throws. The record sat after the call site, so it never ran for those paths: a completed derive vanishing from the denominator, with the query's correlation left consumable by a later derive. The record moved to the save site, firing the instant the artifact is persisted and ahead of both throws; a fan that throws before persisting still records nothing, now by construction rather than by comment.

**Caught while verifying finding 2, and worth stating plainly:** routing that record through the seam helper re-resolved from `ctx.cwd`, which for a config-less cwd falls back to the global `~/.totem/` profile — it appended 3 real rows to the maintainer's live ledger during this package's tests (39 → 42). It now writes to `ctx.totemDirAbs`, the directory the artifact itself was saved into, which is both more correct (the derive belongs with its verdict) and incapable of wandering. Verified held at 42 across a full suite run. Those 3 rows are **not cleaned**, same as the earlier 39.

### Seams chosen

The **four CLI surfaces and the doctor reader** route through one helper, `packages/cli/src/commands/qbd-seam.ts`, which resolves the ledger from the **config root** — never the invocation cwd.

The MCP seam does **not**: the server resolves its own `projectRoot` (`context.ts`, from `process.cwd()`). An earlier version of this table claimed all five shared the helper, which was false. The practical consequence, stated rather than buried: a CLI derive run from a subdirectory and an MCP query can land in **different ledgers** — the CLI follows its resolved config root (possibly the global profile, announced when so), the MCP server follows the root it was spawned in.

| Class | Seam | Why |
|---|---|---|
| Query | `packages/cli/src/commands/search.ts`, after `store.search()` returns | A query that threw grounded nothing, so it must not mint an ID a later derive could claim |
| Query | `packages/mcp/src/tools/search-knowledge.ts`, in the success branch after the search ran, **awaited** | NOT at handler entry: a search blocked by the health gate or one that threw grounded nothing either. `logMcpCall` stays at entry and ADR-029 is untouched. Awaited so the pointer write cannot race the tool response across the process boundary |
| Derive | `spec.ts`, after synthesis produced content, before the write forks 3 ways | Counts each successful synthesis exactly once |
| Derive | `orient.ts`, after `deriveOrientReport`, before the `--json` fork | One call covers both render paths |
| Derive | `shield.ts` (implements `totem review`), inside the **shared verdict handler**, plus one call after the multi-lane fan converges | Structural and standard both route through that handler, so a review counts exactly once and only once a verdict exists. Siting it above the structural fork (round 1) fixed structural escaping the denominator but broadened the class: a *throwing* review recorded a derive, unlike `spec` |

### The correlation ID, and why it is not `correlation_id`

`LedgerEvent.correlation_id` is ADR-014's orchestrator→MCP trace ID. This slice is explicitly **not** chartered to build that stack, so it carries its own field, `qbd_correlation_id`.

The format is `qbd1-<mint instant base36>-<16 hex random>` — **self-dating**. It embeds the instant it was minted, and `LedgerEventSchema` cross-checks that instant against the timestamp of the row carrying it. Both writers derive the ID and the row timestamp from a single clock read, in one module, so the two cannot drift.

### Session identity and the correlation window

Derived from what the codebase already has, with the design calls owned as mine:

- **Session identity** — the existing `.session-id` primitive (`readSessionId`, 24h TTL), then `TOTEM_SESSION_ID`. A non-UUID value is dropped rather than stamped, because stamping it would make the whole row unparseable and lose the event. (Preferring an explicit id over a time heuristic follows that module's own contract; it is not an ADR-029 provision.)
- **Correlation window** — 2 hours. The *value* is borrowed from ADR-029 § 2 so this slice carries one time constant rather than two. Treating that span as a **grounding-validity** window is this slice's own decision, not the ADR's — § 2 defines a session-grouping heuristic for the recall metric, a different question. Consume-on-use does most of the real work anyway: the window bounds how stale a grounding may be, but a query grounds only one derive regardless, so a generous window cannot inflate the numerator.
- **Grouping** — rows carrying a `session_id` group by it, keyed by seat as well; rows without one fall back to a rolling-window roll-up **partitioned by `agent_source`**, so two seats sharing a working tree never merge into one session.
- **Correlation mechanism** — a query parks its ID in `.totem/ledger/.qbd-correlation`, mirroring the `.session-id` pointer convention. A derive attaches it only if it is inside the window, from this session, from this seat, and contract-valid — then **consumes** it. The pointer is written only if the query row actually landed, and consumed only after a successful correlated write, so neither a failed append nor a tampered pointer can manufacture or burn grounding.

### Falsifier handling

**1. Denominator gaming — closed at both levels.** The denominator is every derive-class event in the evaluated sessions, including sessions that fired zero queries. Closed at the *ratio* level (derive rows count whether or not they carry an ID) and at the *window* level (session selection keys on "carries ≥1 derive-class event, regardless of query count" — the operator-pinned phrasing, stored as a single constant and asserted verbatim by a test so it cannot be paraphrased).

A related denominator leak got closed during the build: a corrupted pointer originally threw *before* the derive row was written, which would have let a tampered sensor silently shrink the denominator. The row is now always written (uncorrelated), and the breach throws afterward.

**2. Post-hoc correlation — enforced in the schema, not by convention.** A row whose `qbd_correlation_id` could not have been minted when the row was written fails to parse. Specifically rejected: a query row whose ID was not minted at its own instant; a derive row citing an ID minted *after* the derive; a derive row citing an ID older than the window; the field on any event type without query→derive semantics — **including a non-QBD type**, which was the round-1 CRITICAL.

Honest scope, restated precisely after the leg showed the original comment understated the attack. Per-row checks cannot see relationships *between* rows, which leaves two gaps: **backdated appends** (cheap — the ledger is append-only and the mint instant is a public argument) are caught one layer up by the scanner's monotonicity check; a **coherent whole-file rewrite** defeats both, and nothing short of signing would stop it. The checks are also "beyond tolerance", not categorical — a mismatch inside 2s passes, which exists for clock granularity and is now stated rather than glossed.

**3. Banner-fade — no ceremony added.** Out of v1 scope by charter. The render is one block under an existing flag, with no new command, no new hook, and no gate.

**4. Ritual query (Goodhart) — named, not solved.** A query fired to satisfy the metric whose results the derive never reads is indistinguishable here from one that genuinely grounded it. v1 senses **adjacency, not influence**. That sentence ships in the render itself, in the module header, and in the wiki — so nobody reads the number as more than it is.

### Denominator scope decision, stated explicitly

`orient --session` (the SessionStart hook's boot render) is **not** instrumented. It is machine-initiated and fires before an agent could have queried anything, so counting it would guarantee one uncorrelated derive per session and measure the hook's scheduling rather than an agent's adherence.

This is a scope decision about which *actions* are agent-initiated, not a filter on which *sessions* count — no session is excluded and no agent-initiated derive is dropped, so the falsifier-1 rule is untouched. Flagged here because it is a denominator decision and deserves the scrutiny.

### Degraded reads (ADR-115 § 2)

`readLedgerEvents` skips schema-invalid lines *silently*, which would let a torn or tampered ledger render as a confident 100% or 0% — the exact indistinguishability ADR-115 § 2 names as the defect class. The compliance scanner therefore parses the NDJSON itself and counts every rejected line by class: malformed JSON, correlation-contract violations, unclassifiable rows, orphan correlations, cross-session/cross-seat correlations, and backdated appends.

Any anomaly ⇒ the read is `degraded` ⇒ the render leads with a `DEGRADED` envelope, itemises the per-item counts, and qualifies **every derived line** — `compliance (UNVERIFIED)`, `verdict (UNVERIFIED)`, `trend (UNVERIFIED)`. The verdict line matters most: it carries the pre-registered consequence, so rendering a bare `verdict: PASS` beside an unverified number would reproduce precisely the indistinguishability the envelope exists to prevent. Two deliberate polarity decisions, both round-1 fixes:

- Other subsystems' rows failing their own schema do **not** degrade this metric — unless they carry QBD markers, in which case they do, on any type.
- An **unknown event type** is a named advisory that does **not** flip DEGRADED. The ordinary cause is version skew (live on this machine), and letting routine skew degrade every read would drown the signal real tampering produces. Sticky whole-file degraded is kept: one bad row degrades the whole read, which is the honest posture for a file whose integrity is in question.

Absent ledger renders the doctor `SKIP` idiom pointing at the producers — not a fail, and not 0%. Zero derive events renders `n/a (0 derive-class events)` — also not 0%.

### Sensor, not actuator (Tenet 13)

No exit-code changes, no gating, not in the `--strict` suite. Instrumentation write failures never break the instrumented command — they surface as named warnings through an accounting channel. There is no path that discards a failure silently. A new guard also skips recording entirely when no `.totem` directory exists, so the sensor never materialises a stray `.totem/` in an unrelated cwd (it did, under `packages/cli`, until caught).

### Local gates (Windows) — not CI

This table is **local gates on Windows only**. CI is the cross-platform authority, and the distinction is not pedantic: round 2's table certified `4cac3254` green while it was red on ubuntu and macos, because a Windows-vs-POSIX errno divergence made the same on-disk state produce different behaviour. Local green is a necessary condition, never a sufficient one.

All commands run on `992c7f473411cf0cbf79e87f6fdeb29112c9731d`.

| Gate | Command | Exit | Result |
|---|---|---|---|
| Format | `pnpm run format` | 0 | clean |
| Format check | `pnpm run format:check` | 0 | "All matched files use Prettier code style!" |
| ESLint | `pnpm run lint` | 0 | 3/3 tasks, 0 errors (8 pre-existing warnings, none on this diff) |
| Build | `pnpm -r build` | 0 | all 6 projects green |
| Tests | `pnpm run test` | 0 | 12/12 tasks — core 3496, cli 3647, mcp 206, packs 179 |
| Totem lint | `totem lint --base main` | 0 | **PASS** — 0 errors, 290 advisory warnings |

**CI on the preceding SHA (`67dc5c53`) was green on all three platforms** — ubuntu, macos, and windows — confirming the errno fix held where local runs could not check it.
| Totem review | `totem review --base main` | 0 | round 1: gemini lane 0 findings; anthropic lane invoke-error (see below) |
| Changeset | manual | — | `.changeset/qbd-compliance-slice.md`, minor × 3 |
| Manifest drift | — | — | none reported; `totem lesson compile` NOT run (standing freeze) |

**132 new tests** — 125 in new files (`correlation-id` 21, `compliance` 40, `record` 25, `doctor-qbd` 22, `qbd-seam` 17) plus 7 added to existing files (MCP +5, `session-id` +2). Measured per file, not estimated.

- `correlation-id.test.ts` (17) — mint/decode round-trip, and the contract: a backfilled query ID, a derive ID minted after its derive, an out-of-window ID, and the field on a non-QBD type are each **rejected at the schema layer**; an uncorrelated derive row and all pre-existing event types still parse.
- `record.test.ts` (25) — correlation inside the window; **no correlation across sessions, across seats, or when either side lacks a session id (fail-closed)**; **consume-on-use** (one query grounds one derive; four following derives are uncorrelated; a fresh query re-arms); **positive tests for all three accounting branches** (consume failure, pointer write failure, pointer unreadable); a **racing-derives** test proving the non-atomic read-then-clear is reachable *and* that the scanner refuses to credit both; and the two ADR-115 § 2 induced-failure blocks below.
- `compliance.test.ts` (37) — zero-query sessions in the denominator; a query-only session not counted; orphan / cross-session / cross-seat / post-dated citations refused *and* surfaced; **a forged ID smuggled onto a `mcp_call` type caught** (round-1 CRITICAL regression test) with a control proving an unmarked non-QBD failure is still ignored; **the leg's exact duplicate probe** (one query, three derives → 1/3, degraded) plus an earliest-citation pin and a clean-1:1 control; **the fresh-adopter backdate probe** (backdated pairs after non-QBD-only history → degraded) with an in-order control; **seat-mismatch hint fires without degrading**; **unknown type = advisory, not degraded**; **schema type-parity on schema-INVALID fixtures** with a discriminating control; two seats never merged into one rolling session; `PENDING` under 20, `FAIL` and `PASS` at 20; and **first-20 ordering pinned from both directions**.
- `doctor-qbd.test.ts` (21) — verbatim threshold/window line, `PENDING` render, ritual-query limitation present, exit code untouched, duplicate-correlation naming, the **remediation path**, the seat hint carrying mmnto-ai/totem#2530, the degraded-render block below, **verdict and trend qualified under DEGRADED** with a clean-scan control, and a **real on-disk read** through the production file path covering `PASS`, `FAIL`, the unreadable-ledger catch, and exit-code neutrality on **both** verdict branches.
- `qbd-seam.test.ts` (11, new) — config-root resolution; a **cwd ≠ config-root** case where the old `path.join(cwd, …)` answer provably differs; two cwds under one root resolving identically; `CONFIG_MISSING` reported rather than guessed; no stray `<subdir>/.totem`; a **global-profile landing announced** rather than silent; and call-expression pins that the review derive is recorded inside the shared verdict handler, that structural reaches that handler before returning, and that the fan carries its own call. Every test passes a temp `homeDir` so none can reach the real global profile.
- `ledger-writer.test.ts` / `search-knowledge.test.ts` (MCP, +5) — `logMcpCall` no longer mints a `corpus_query` row; `logCorpusQuery` does, only after the search ran; and **no mint when the health gate blocks the search** while `mcp_call` still fires.
- `session-id.test.ts` (+2) — `ENOTDIR` reads as benign absence in both `readSessionId` and `writeSessionId`; see the cross-platform fix below.

### Cross-platform errno fix (CI round)

CI came back red on ubuntu and macos, green on windows, on exactly one test: the unwritable-ledger induced-failure pack expected a real errno in its warning and got `"Unexpected error reading session ID"`.

The fixture occupies `<totemDir>/ledger` with a **file**. Statting `<totemDir>/ledger/.session-id` then raises `ENOTDIR` on POSIX (a path component is a file) but `ENOENT` on Windows — for byte-identical on-disk state. `session-id.ts` listed `ENOENT`/`EACCES`/`EPERM`/`EROFS` as benign absence but not `ENOTDIR`, so on POSIX it rethrew from the session-id layer *before* the ledger write path and its errno-bearing accounting ever ran.

A path component being a file means "there is no session id here, and there cannot be" — exactly the `ENOENT` semantic — so `ENOTDIR` joins the benign set (now one named constant rather than two inline copies), leaving the unexpected-errno rethrow intact. The same class is aligned in `record.ts`'s correlation-pointer read, which probes under the same `ledger` component.

Pinned so it cannot recur silently: benign-absence tests for both session-id functions using **mocked errnos**, so they assert identically on every platform rather than depending on which errno the host OS happens to raise; a cross-platform "exactly one warning" count on the derive path; and a reproduction of the CI failure that fakes only the `.session-id` probe. Verified falsifiable — removing `ENOTDIR` from the set turns the reproduction test red.

**ADR-115 § 2 induced-failure tests** (three packs each, per the `mail-degraded-e4.test.ts` convention; failures induced with real errnos and real on-disk state, never by mocking the module under test):

1. *Forged correlation pointer* (mutating path) — **backstop throws** `TotemError('QBD_CORRELATION_CONTRACT')`, prior-good state preserved (derive row present but uncorrelated; the forged ID never persisted); **accounting fires** naming `derive-id-minted-after-derive` on the sensor path without breaking the command; **recovery** returns to green with the fault removed, and a healthy control warns about nothing.
2. *Unwritable ledger inside a real project* (`ledger` path occupied by a file ⇒ real `mkdirSync` errno) — accounting fires, names the subsystem **and pins the actual errno** (`ENOTDIR`/`EEXIST`/`ENOENT`) rather than only the unconditional prefix, no throw escapes, recovery writes cleanly; plus a test that no pointer is parked when the query row failed.
3. *Degraded render* (read-only path) — per ADR-115 § 2 the backstop is the `DEGRADED`/`UNVERIFIED` envelope, not a nonzero exit: a torn + backfilled ledger renders `DEGRADED`, `compliance (UNVERIFIED)`, and per-class counts; the control asserts a clean ledger renders **no** envelope, so it cannot rot into always-on decoration.

   The leg caught this block's key assertion being **vacuous**: the old fixture had zero derives, so it rendered `n/a` and `not.toMatch(/compliance: \d/)` passed under any label. The fixture now contains real correlated derives, so a number genuinely would be printed — and the test pins that it appears only as `compliance (UNVERIFIED): 1.00 (1/1)`, never bare.

The MCP `ledger-writer` tests were re-pinned to the **total row multiset**, not only a type-filtered view — a filtered assertion alone would still pass if an unexpected extra row appeared.

**Anomalies, reported not hidden:** the anthropic lane of `totem review` failed with a provider 400 — `"Your credit balance is too low to access the Anthropic API"` — an environment/billing condition unrelated to this diff. The gemini lane completed with 0 findings, so that fan reported `settled=false (1/2 lanes)`; the round-1 falsification leg is the substantive review this branch actually got. `totem review` also warned the diff exceeds its 50,000-char LLM window, so the LLM lanes saw truncated content; the deterministic `totem lint` gate saw the full diff.

Worth noting as a live sensor result rather than a test: the first build created a stray `packages/cli/.totem/ledger/` with 32 rows during the test suite, because the writer materialised a `.totem` tree in whatever cwd a command ran in. That is fixed (absent `.totem` = "not an instrumented project", silent skip) and the tree is clean.

## Files changed

**New:** `packages/core/src/qbd/{correlation-id,record,compliance}.ts` + 3 test files · `packages/cli/src/commands/doctor-qbd.ts` + test · `packages/cli/src/commands/qbd-seam.ts` + test · `.changeset/qbd-compliance-slice.md`

**Modified:** `packages/core/src/{ledger,errors,index}.ts` · `packages/cli/src/index.ts` · `packages/cli/src/commands/{search,spec,orient,shield}.ts` · `packages/mcp/src/ledger-writer.ts` + test · `packages/mcp/src/tools/search-knowledge.ts` + test · `docs/wiki/trap-ledger.md`

Not chartered and not touched: the ADR-014/029/044/078 telemetry stack.

Closes #2510


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 disposition

All findings verified against source before action; all accepted, none declined. Fixes landed in `429b1aa1` (gates: format/format:check/eslint/full build/full test 12-12/`totem lint --base main` PASS 0 errors; 118→128 slice tests).

| GT | P1 — doctor reader falls back to `cwd/.totem`, a ledger the writers never write | accept | fallback removed; unresolved config → named SKIP, stale-ledger regression test (`doctor-qbd.ts:63-80` @ `429b1aa1`) |
| GT | P1 — `--raw` records review derives with no verdict, consuming pointer grounding | accept | record moved into the two parsed-verdict branches via `recordReviewDerive` (`shield.ts:1290-1300,1398`); `--raw` + unparsable-verdict are upstream of it; PASS and FAIL both count; rationale comment rewritten |
| GCA | direct `process.stderr` write in `@mmnto/mcp` (stdio-transport restriction) | accept | routed through `logSearch` with an `internal:qbd-corpus-query` synthetic entry (`ledger-writer.ts:86-97`) — the literal `onWarn: logSearch` suggestion doesn't typecheck (`logSearch` takes a `SearchLogEntry`), so the existing internal-entry idiom was used; no stdio, no swallow |
| CR | 🔴 out-of-range mint instant throws `RangeError` inside `superRefine` — one forged line kills the scanner | accept | named `MAX_TIME_VALUE_MS = 8.64e15` bounds mint, decode, and `eventMs` (`correlation-id.ts:126,157,177,204`); forged `qbd1-zzzzzzzzzzz-…` now counts as `undecodable-mint-instant`, scan continues — proven by a scanner-level test that asserts the NEXT row still parses |
| CR | non-finite `eventMs` returns `ok:true` on the public API | accept | early reject (`correlation-id.ts:204`) |
| CR | hardcoded `.totem` fallback undercuts single-resolution | accept | same fix as GT P1 row 1 (cross-bot duplicate) |
| CR | unreadable-ledger SKIP discards the errno | accept | `catch (err)` + errno named in the SKIP line (`doctor-qbd.ts:100-112`) |
| CR | `report` shadowed in the orient seam block | accept | renamed `seamReport` (`orient.ts:811-818`) |
| CR | `recordQbdQuery` has zero unit coverage | accept | query-seam test: one `corpus_query` row with a defined ID in the project ledger (`qbd-seam.test.ts:203-224`) |
| CR | top-level `node:path` import vs the CLI lazy-load path rule | accept | dynamic import inside the function (`qbd-seam.ts:56`) |
| CR | two seams duplicate six statements + byte-identical load-bearing note strings | accept | folded into `recordThrough`, notes single-sourced, public API unchanged (`qbd-seam.ts:102-141`) |
| CR | `QbdSession` module-private while `groupQbdSessions` is barrel-exported | accept | exported with TSDoc + added to the barrel (`compliance.ts:352-364`, `core/src/index.ts`) |
| CR | seat-mismatch hint is O(derives × queries) over the whole ledger | accept | bounded reverse walk, breaks at the window edge (`compliance.ts:494-509`) |
| CR | `undecodable-mint-instant` branch untested | accept | covered by the forged-ID scanner fixture (see the RangeError row) |
| CR | verdict-placement comment claims an invariant the code doesn't hold | accept | same fix as GT P1 row 2; comment now states the true invariant |

Docs propagation: the recording discipline's disclosed semantics ("after content" → "after a parsed verdict; `--raw`/unparsable excluded, FAIL included") updated in `docs/wiki/trap-ledger.md`, the changeset, and the PR body.

Out of scope, not dispositioned: CodeRabbit's two self-retracted comments (marked "likely incorrect" by CR itself) and the advisory static-analysis blocks embedded in the review details.

local-lane: 37a02f4d round=0 settled=false lanes=1/2

satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 2 disposition

Re-invocation on `429b1aa1`: CodeRabbit clean (no new findings); GCA and Greptile one finding each — both verified against source, both accepted, fixed in `992c7f47` (gates: format/format:check/eslint/full build/full test 12-12/`totem lint --base main` PASS 0 errors; 128→132 slice tests).

| GCA | seat-mismatch hint false-positives when a window holds both a same-seat and a different-seat query (walk breaks on first different-seat) | accept | two-flag walk: same-seat clears and breaks, different-seat notes and continues; hint iff no same-seat AND ≥1 different-seat (`compliance.ts:520-530` @ `992c7f47`); mixed-window + different-only regression tests. Noted in the PR body: the pre-walk `.some()` carried the same false-positive semantics, so this is a semantic correction, not only a rewrite repair |
| GT | P1 — fan-mode reviews persist a verdict then throw (zero-completed-lanes / `--fail-on`), escaping the derive record at the `runReviewFan` call site and leaving the query correlation consumable | accept | record relocated into `review-fan.ts:1383-1410`, firing the instant the verdict artifact persists and upstream of both policy throws; the shield.ts call-site record removed. Invariant now holds by construction: persisted verdict ⇔ recorded derive; a pre-persistence throw records nothing. Written to the artifact's own `totemDirAbs` so the derive lands beside its verdict and cannot resolve elsewhere. Ordering pinned by tests for both halves |

Docs propagated: wiki + changeset now state the per-path rule — parsed verdict (standard/structural), persisted artifact (fan) — with `--raw`, unparsable, and pre-persistence throws excluded; FAIL and persisted-then-policy-throw included.

local-lane: 37a02f4d round=0 settled=false lanes=1/2

PR #2517, opened 2026-07-29
fix(core): register .mts/.cts as built-in AST extensions — Class 7 partial target-mismatch residual (rule fea10369daa04c98)

satur8d:
## Summary

PR #2512's Class 7 guard condemns TOTAL target mismatch and deliberately permits partial mismatch — leaving one live partial specimen: compiled rule `fea10369daa04c98` (fileGlobs `['**/*.ts', '**/*.js', '**/*.mts', '**/*.mjs']`, three of four resolvable). The rule loaded clean, but `extensionToLanguage('.mts')` returned `undefined`, so the first diff touching a `.mts` file threw `TotemParseError` out of `rule-engine.ts`'s per-file loop — strict mode rethrows and the whole lint aborts. Same detonation class as the two archived landmines, scoped to `.mts` traffic.

Repair, not retire (operator-ruled 2026-07-28, relayed via strategy-claude): `.mts` (TypeScript ESM) and `.cts` (TypeScript CommonJS) are the same grammar as `.ts` — module resolution differs, syntax does not — so both now register as built-ins against the SAME `tsLoader` instance (satisfying `registerLang`'s shared-loader invariant), mirroring the `.mjs`/`.cjs` → javascript pattern.

## Acceptance criteria (all four from the issue)

1. `registeredExtensions()` reports eight built-ins including `.mts`/`.cts` → typescript.
2. Regression tests at the real detonation seam: ast-grep rules — including `fea10369daa04c98`'s verbatim glob set — execute against `.mts`/`.cts` diff files under strict mode instead of aborting (observed RED pre-fix, GREEN post-fix).
3. Class 7 guard semantics unchanged: the existing partial-permit control is untouched and passing (no existing guard test used `.mts`/`.cts` as its unresolvable specimen); a new `.zzznotalang` control proves a genuinely unregistrable extension still hard-errors at load.
4. Sweep re-run across all six tracked compiled-rules corpora (198 globbed active AST rules): zero partial specimens with unregistrable extensions remain; the anchor rule fully resolves. The single total-mismatch hit in `pack-rust-architecture`'s own corpus is a sweep artifact — that pack registers `.rs` at boot via `register.cjs`, and the live corpus-loads-clean guard test passes.

## Notes

[...]

PR #2538, opened 2026-07-31
fix(core,cli): register .mts/.cts in drift-detector + shield-classify extension sets

### LINKED ISSUE #2519: fix(cli,core): shield-classify CODE_EXTENSIONS + drift-detector FILE_EXTENSIONS omit .mts/.cts — same oversight class as the #2513 registry gap
## The residual class

#2513 (PR #2517) registered `.mts`/`.cts` in the AST extension registry — the sole extension→language authority for rule dispatch. The build's out-of-scope sweep found two more hand-maintained extension sets with the identical oversight: they include `.mjs`/`.cjs` but omit `.mts`/`.cts`. Both verified at source 2026-07-29.

## Specimen 1 — shield classification: a `.mts`-only diff is NON_CODE

`packages/cli/src/commands/shield-classify.ts:14` — `CODE_EXTENSIONS` lists `.ts/.tsx/.js/.jsx/.mjs/.cjs` (plus other languages) without `.mts`/`.cts`. Consequence: a diff touching only `.mts`/`.cts` files classifies `allNonCode: true`, so shield treats real TypeScript code changes as a non-code diff for classification-gated behavior. This is the sharper of the two specimens: it changes gate behavior, not just reporting.

## Specimen 2 — drift detection: `.mts`/`.cts` refs are invisible

`packages/core/src/drift-detector.ts:35` — `FILE_EXTENSIONS` (extensions considered valid for drift detection) includes `.mjs`/`.cjs` and doc formats but not `.mts`/`.cts`. Consequence: a lesson referencing a `.mts`/`.cts` path is never checked for orphaned refs — the file can vanish without the drift detector noticing.

## Non-defects checked and excluded (so the sweep doesn't re-open them)

- `packages/core/src/compile-smoke-gate.ts:116` — the unscoped-rule fallback array is a parse-attempt list, not a dispatch map; adding `.mts`/`.cts` would be a no-op (same typescript grammar as `.ts`).
- `packages/pack-agent-security/test/repo-sweep.test.ts` `SWEEP_EXTS` — test-only breadth choice, not a governance surface.

## Acceptance

1. Both sets include `.mts`/`.cts`.
2. Regression tests: a `.mts`-only diff classifies as CODE (`allNonCode: false`); a lesson referencing an existing `.mts` path is drift-eligible (and its deletion surfaces in `orphanedRefs`).
3. A repo-wide grep for the six-extension literal cluster (`.mjs` adjacent to `.cjs` in a set/array) confirms no further live governance surface omits the pair — findings either fixed here or dispositioned in-issue.

satur8d:
Closes #2519 

## What this does

Registers `.mts`/`.cts` in the two hand-maintained extension sets the #2517 central-registry fix didn't cover — with one premise correction against the issue as filed:

- **`drift-detector.ts` `FILE_EXTENSIONS` — real behavior fix.** `extractFileReferences` drops backticked candidates whose extension isn't in the set, so lesson references to `.mts`/`.cts` paths were never drift-checked and could rot silently. Pinned red-first: the new test failed (`expected [] to have a length of 1`) before the two-line source change, passes after.
- **`shield-classify.ts` `CODE_EXTENSIONS` — hardening, zero behavior delta.** The issue's Specimen 1 claim (a `.mts`-only diff classifies `allNonCode: true` and bypasses classification gates) is **incorrect at HEAD**: `classifyFile` fail-closes unknown extensions to `CODE`, so those diffs already classified as code. Explicit membership makes the intent inspectable and survives any future change to the fail-closed default; regression tests pin both classifications. The correction is also recorded in `.totem/specs/2519.md`.
- **Sweep complete:** every other `.mjs`-literal registry in `packages/**` (`ast-classifier.ts`, `init-doctrine.ts`, `compile-smoke-gate.ts`, `repo-sweep.test.ts`) is already correct or pre-cleared by the issue.

## Riding items (carryforward)

- **Specs 2453/2458/2463 adopted** into the tracked corpus — census D-item disposition (operator-ruled adopt-or-delete): 2453/2458 are open build contracts; 2463 retained per the closed-spec convention, carrying the ruled Q1–Q3 design record.
- **Postmerge lessons for #2523/#2524** — 7 novel lessons, extract-only under the standing rule-compilation freeze (no compile, no rules mutation).
- **`init-templates.ts` comment truth-up** — it credited the skill-parity invariant to `installed-skills-match-source.test.ts`, which doesn't exist; the invariant is real and lives in the "Distributed skill constants match source-of-truth" suite in `init.test.ts` (byte-equality against `.claude/skills/<name>/SKILL.md`). The comment now names the actual location.
- **Managed-skill marker contract documented** for consumers in `docs/wiki/agent-claude-code.md` (§4), written from `scaffoldClaudeSkill` behavior: canonical span replaced through the end marker (including pre-marker content), post-marker content survives refreshes, marker-less files preserved unless `--force-skill-refresh`.

## Gates

- `totem lint --branch`: PASS, 0 errors. 16 advisory warnings, all frozen-lesson regex hits: absolute-claim/marketing-term matches inside the adopted spec prose (historical/ruled documents — not rewritten to appease an advisory rule) and two "orchestrator test timeout" hits on plain synchronous unit tests (over-broad `//` pattern; not orchestrator suites). Declined with these reasons.
- `totem review`: 2-lane fan (sonnet, gemini) — 0 findings, settled round 0.
- Full workspace `pnpm -r build`: clean. Package suites: core 3497 passed, cli 3650 passed. Changeset included (patch `@mmnto/totem` + `@mmnto/cli`).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 dispositions (fix-push `7672273b`)

**CR-1 — `lesson-335808c3` (stdout vs stderr): FIXED.** The lesson now scopes protocol corruption to non-protocol `stdout` writes on the stdio transport, permits `stderr` for server diagnostics, and points non-stdio transports at their own logging mechanism.

**CR-2 — `lesson-5bdc157c` (Date instantiation vs serialization): FIXED.** The lesson now states that `new Date(out-of-range)` yields an Invalid Date silently and that `toISOString()`/serialization is the throwing operation; the bounding guidance covers construction and serialization together.

**CR-3 — MD041 H1 on the three adopted specs: DECLINED.** `### Problem Statement` is the spec generator's output shape across the entire tracked `.totem/specs/` corpus (verified against the tracked file set); changing only these three files would break corpus consistency. Specs are point-in-time planning artifacts per the recorded review learning from PR #1553, and the identically-shaped `2519.md` was LGTM'd in this same review pass.

**CR-4 — `SearchIdentity` federation collision (spec 2458): FIXED via amendment.** The spec now requires a stable source-index namespace in the identity, keeps ambiguous fuzzy-section matches separate rather than merging them, and adds cross-repository identical-path coverage to the build-time test contract.

**Pre-merge "Out of Scope Changes" warning: acknowledged, not splitting.** The riding items (spec adoptions, postmerge lessons, comment truth-up, wiki section) are deliberate carryforward convention in this repo — small governance artifacts ride the next PR rather than minting standalone PRs.

local-lane: a49b3368 round=1 settled=true lanes=2/2

satur8d:
@coderabbitai review

satur8d:
## Round 2 disposition (fix-push `83a5df68`)

**CR outside-diff — `SearchIdentity` vs the federated merge contract (spec 2458): FIXED, finding accepted in full.** The round-1 amendment conflated store provenance with document identity: a store-based grouping key would keep the filed duplication class (same document, one tagged / one untagged) unmerged, while omitting a namespace lets distinct same-path documents collide. The revised amendment supersedes the drawn interface where they conflict: identity keys on the document's **home** namespace derived from the normalized path (plus repo-relative path and section); store provenance/tags are merge metadata, never part of the key; an underivable home anchor fails safe to unmerged (a surviving duplicate is recoverable, a false merge silently loses a result); ambiguous fuzzy section matches stay separate. Build-time test contract updated to cover all three cases.

**Convergence note.** Both lessons corrected, the spec amendment is now internally consistent with the merge contract, every code file carries an explicit LGTM, and Greptile stands at 5/5. Remaining open items are the declined MD041 corpus-convention nit and the acknowledged out-of-scope warning — both dispositioned with reasons in round 1. The round has converged.

local-lane: 211eb076 round=2 settled=true lanes=2/2

Issue #2566, opened 2026-08-03
feat(search): surface a live full-sync checkpoint on the read path - search/MCP serve a knowingly-partial index silently


satur8d:
From the #2562 falsification round (QUESTION 13, ruled out of scope for the fix PR). While a full-sync checkpoint is live (interrupted \--full\, epoch not yet complete), \	otem search\, the MCP \search_knowledge\ tools, and the auto-context hook all serve the knowingly-partial index with no signal. The marker file exists precisely to make this state detectable — the sync path now detects and resumes it, but the read path stays blind (Tenet 4: a consumer of degraded retrieval should be able to see the degradation). Fix shape: when \.totem/cache/full-sync-checkpoint.json\ exists, annotate search output / the MCP retrieval envelope (e.g. \index-status: partial (full re-index in progress, N/M files)\) and log once per invocation. Cheap to sense; no behavior change to ranking.

Issue #2565, opened 2026-08-03
fix(embedders): silent LazyEmbedder->Ollama fallback can mix vector spaces in an existing index (incremental path)


satur8d:
Pre-existing hazard surfaced by the #2562 falsification round (MAJOR 3 residual). \LazyEmbedder\ silently falls back to Ollama (768d) when the configured provider's SDK/key is missing. On an INCREMENTAL sync over an existing gemini-768d index, the fallback's per-file delete+insert writes Ollama vectors into the same table — the dimension check is length-only and cannot see a same-dimensionality provider switch, so retrieval silently degrades (vectors from two spaces ranked against each other). #2562 closed the RESUME-path variant: the full-sync checkpoint now stamps the effective identity via \Embedder.describeEffective\ and restarts on mismatch, and the pipeline threads its logger so the fallback warning is audible. The incremental path still mixes. Fix shape: persist the effective identity in \index-meta.json\ (it currently records the CONFIGURED provider at write time) and verify effective-vs-meta at sync start, failing loud on mismatch with a \	otem sync --full\ remediation hint.

satur8d:
Scope addition from #2562 falsification round 4 (leg-verified): \erifyIndexMeta\ — the DATABASE_MISMATCH detector this issue's fix would strengthen — currently has NO callers anywhere in the repo (grep returns only its two re-exports), and \writeIndexMeta\ always stamps the CONFIG identity at sync end. So the unmarked-regime mixing hole is wider than the fallback case: any provider/dimensions config change followed by a plain incremental sync mixes the store today with no signal. The fix here should both persist the EFFECTIVE identity in index-meta AND actually wire \erifyIndexMeta\ into the sync/search entry points.

satur8d:
## Datum — a consumer's first-contact run reproduced the configured-not-effective record on a fresh install; the init and doctor halves are filed beside this (totem-claude, 2026-09-07)



[...]

Issue #2564, opened 2026-08-03
fix(lock): 120s stale threshold is stealable under paced/throttled syncs; the #2562 checkpoint invariant depends on the mutual exclusion


satur8d:
From the #2562 falsification round (MINOR 6, part deferred). \lock.ts\ writes its timestamp once at acquisition (no heartbeat) and treats a >120s stale timestamp as final authority — deleting the lock regardless of PID liveness. Long syncs already exceeded 120s; #2562's \mbedding.throttleMs\ pacing and 60s RetryInfo waits make multi-minute in-lock sleeps the RECOMMENDED configuration, so lock theft becomes likely exactly when quota pressure is highest. A second sync (hook-driven, or the MCP add-lesson spawned sync) then runs concurrently: PID-unique tmp names (shipped in #2562) prevent torn checkpoint writes but not semantic races — one process can \store.reset()\ under the other's epoch, leaving \completedFiles\ asserting rows that were dropped (the store-verification intersect bounds the damage on the NEXT resume, but the concurrent window itself is unguarded). Fix shape: heartbeat the lock timestamp around long awaits, or raise the stale threshold above MAX_RETRIES x 60s + expected throttle wall-clock, plus holder-identity verification before deletion.

satur8d:
Exposure amendment from the #2562 falsification round 3 (the fix branch made this materially worse than the body above states, in two ways): (1) pacing is now the DEFAULT for gemini (derived 4s/batch constant), not a recommended configuration — so a full re-index on a >3k-row corpus floors at several minutes of in-lock wall-clock by construction, and the 120s stale threshold is exceeded for most of every such run (lock.ts's own comment 'sync can take 30-60s' is now false by design); (2) the quota 429 arm now waits out the minute window (60s floor) even WITHOUT a server advisory, so a single embedWithRetry can sleep up to ~180s in-lock — the >=120s idle window no longer requires RetryInfo. Both raise this from deferred-MINOR to merge-adjacent: the #2562 checkpoint invariant depends on the mutual exclusion this lock no longer reliably provides during exactly the runs the fix creates. Fix shape unchanged (heartbeat around long awaits or holder-verified deletion); PID-unique tmp names (shipped) bound the torn-write damage but not the semantic race.

PR #2569, opened 2026-08-03
fix(sync): crash-safe, quota-convergent full re-index — checkpoint/resume, default pacing, effective-identity gate

### LINKED ISSUE #2562: sync: crashed --full leaves a lying 'Sync complete' sensor; no resume or throttle makes quota-bounded re-index unable to converge; 429 hint prescribes wrong recovery
# Summary

A `totem sync --full` run that crashes mid-embed (e.g. Vertex 429 `RESOURCE_EXHAUSTED`) leaves the workspace in a state where the next plain `totem sync` reports **`Sync complete: 0 files` over a partial vector store** — a lying "complete" sensor over silently-degraded retrieval. Compounding it: `--full` has no resume, so a quota-bounded consumer whose corpus exceeds its one-shot allowance can **never converge**, and the 429 failure hint recommends exactly the wrong recovery (`retry with 'totem sync'`).

**Priority: sensor-degradation class** (Prop 309 standing posture — jumps the queue). Operator-ruled routing into the totem lane 2026-08-03 (strategy-claude dispatch `2026-08-03T1718Z`, context mmnto-ai/liquid-city#940).

## Evidence (first-hand, lc-claude dispatch 2026-08-03T1713Z, logs retained)

liquid-city armed `.rs`/`.gd` index targets (corpus ≈ 4.5–5k chunks) on `@mmnto/cli@1.108.0`:

- Run 1 died at **1,710** chunks embedded on 429 `RESOURCE_EXHAUSTED` (`online_prediction_requests_per_base_model`, base model gemini-embedding-2); run 2 (~2h later, refilled allowance) died at **2,153** — each retry re-spends the allowance on already-embedded chunks from zero.
- After the crash, plain `totem sync` printed `Sync complete: 0 files` against the partial store.
- The failure surface printed: `Fix: Check your GEMINI_API_KEY and network connection, then retry with 'totem sync'.`

## Mechanism (code-grounded at `e0362125`)

- `packages/core/src/ingest/pipeline.ts` — the full path (`runSync`, step 3 else-branch) calls `store.reset()` then embeds; sync state (`lastSyncSha`) is only written on successful completion (~line 517). A mid-run crash therefore leaves the **previous successful sync's state file intact** while the store has been reset to a partial rebuild. The next incremental sync diffs `lastSyncSha..HEAD`, sees ~0 changed files, and honestly reports "complete" over the gutted store. The empty-store auto-heal (~line 301) cannot fire — a partial store is not empty.
- No resume: the full path always resets and re-embeds everything. Note `flushBuffer` fires only at file boundaries (chunks are pushed per-file before the batch-size check), so the store only ever contains **whole files** — a per-file checkpoint is sufficient for exact resume.
- `packages/core/src/embedders/gemini-embedder.ts` — 429s retry only `MAX_RETRIES = 3` times with 1–4s exponential backoff, useless against a **per-minute** quota; there is no client-side pacing at all; the terminal `TotemError` hint is the static wrong-recovery string quoted above.

## Fix shape (per the ruled routing; order = containment first)

1. **Watermark integrity (containment-critical):** invalidate/mark sync state *before* `store.reset()` in the full path — e.g. a `fullSyncInProgress` dirty marker cleared only on successful completion. A crashed `--full` then leaves a detectable state instead of a lying one.
2. **Resume:** during full sync, checkpoint files whose chunks have been flushed; when a sync starts and finds the dirty marker + checkpoint, resume the full re-index skipping checkpointed files (no second reset). This also makes plain `totem sync` after a crash *do the right thing* — the existing hint becomes true instead of wrong.
3. **Client-side throttle:** configurable request pacing for embedding calls (the 429 is `online_prediction_requests_per_base_model`), plus honoring any server-provided retry delay on 429 before falling back to exponential backoff.
4. **Fix the 429 `Fix:` hint:** quota-specific message naming the quota, the resume behavior, and the throttle knob.

Interim (operator-side, per the routing dispatch): GCP quota handling stays out of scope here; lc holds further `--full` runs until this lands.

Related: mmnto-ai/liquid-city#940. Companion defect filed separately: `lesson extract` non-determinism.

satur8d:
Closes #2562 

Sensor-degradation fix (Prop 309 queue-jump; operator-ruled routing 2026-08-03; ADR-115 Train-1 "sync-watermark case"). Evidence: lc-claude's first-hand datapoints on mmnto-ai/liquid-city#940 + the operator-authed quota ground-read (rolling **2,000/min per-minute window**, gemini-embedding-2 us-central1).

## What was broken

A `totem sync --full` that crashed mid-embed (quota 429) reset the store but left the previous successful sync's baseline intact — the next plain `totem sync` diffed from the old SHA, saw 0 changed files, and reported **`Sync complete: 0 files` over the gutted store**. No resume meant every retry re-spent quota from zero, and the 429 hint prescribed exactly the wrong recovery.

## What ships

1. **Watermark integrity (containment-critical):** a dirty-marker checkpoint (`.totem/cache/full-sync-checkpoint.json`) is written atomically **before** `store.reset()` in every resetting path and cleared only after a successful completion writes the fresh baseline. No unmarked-reset window exists.
2. **Resume:** the next sync (plain or `--full`) detects the marker and resumes: completed set = checkpointed ∩ live ∩ **actually-in-the-store** (the store is ground truth — a store healed/nuked mid-epoch re-embeds in full) − moved-since-epoch (git tracked-diff ∪ mtime > epoch start; covers untracked files, non-git projects, submodules, assume-unchanged). The epoch continues only under the embedder identity that **actually serves the run** (`Embedder.resolveEffective`) — a silent Ollama fallback restarts rather than mixing vector spaces; a persistent fallback resumes as itself; a resolvable mismatch restarts even with zero work remaining (index-meta evidence is never laundered); only an *unresolvable* embedder with *nothing left to embed* clears the marker unverified, loudly.
3. **Throttle (the load-bearing piece per the ground-read):** gemini ingest embeds pace by default — 4s between multi-text batches, a derived constant under the 2,000/min window — so re-indexes that died at ~2k chunks now complete; single-text query embeds (search / MCP) are never paced. `embedding.throttleMs` tunes or disables (openai honors the knob too, default off there). Quota 429 retries honor server-advised RetryInfo delays (capped 60s, sub-backoff advisories floored) and otherwise back off **past the minute window**.
4. **Honest hints:** the terminal quota error now names the resume behavior and the throttle knob instead of misdirecting to key/network checks — and since resume exists, "re-run `totem sync`" is now *true*. The MCP `add_lesson` tool defers its 60s-killed convenience sync under a live epoch instead of deterministically timing out against it.

Plus: atomic state writes (tmp+rename, PID-unique), `getChangedFilesDetailed` / `hasFullSyncCheckpoint` exports, wiki docs for the crash-recovery contract and the pacing knob.

## Review of record (review-leg contract)

**Four falsification rounds**, each scoped to the prior fold, findings absorbed or disposed with owner verification against primaries; the full disposition trail is in `.totem/specs/2562.md`:

- **R1 (5 MAJOR):** store-verified resume, untracked/non-git convergence, effective-fingerprint stamping, test-fidelity (the induced-regression test now builds the real surviving-baseline precondition).
- **R2 (2 MAJOR):** the fingerprint gate inverted both threat directions — replaced with effective-vs-effective comparison after forced resolution; eviction arm made exhaustive.
- **R3 (3 MAJOR):** default-pacing interaction sweep — `add_lesson` timeout, MCP process-lifetime embedder pacing queries, #2564 lock exposure (amended on-issue, merge-adjacent).
- **R4 (1 MAJOR, convergence):** round 3's zero-work exemption over-broadened — re-keyed to resolution failure; missing test cell locked. One leg finding declined on owner verification (NIT 12 misread `totalChunks`).

**Induced regressions per ADR-115 Train-1:** the pipeline suite reproduces the lying sensor (a crashed full sync followed by a plain sync that pre-fix reported 0 files) plus 13 more crash/resume/identity cells; 111 tests in the touched suites, full workspace green (core 3538, cli 3706, mcp 207), `totem lint` PASS, prettier clean.

**Known-adjacent, deliberately not in this diff:** #2564 (lock staleness under paced syncs — now merge-adjacent), #2565 (unmarked-regime effective-identity mixing + `verifyIndexMeta` has no callers), #2566 (read-path surfacing of a live epoch).


satur8d:
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai Round dispositions below (fixes in `6b73f587` + `8ad37dbe`).

## Greptile (summary confidence 4/5 + outside-diff P1)

1. **Stale rows on skipped resume files — FIXED (deterministic branches) / DECLINED (transient branch), partially pre-existing.** A re-processed file now yielding zero chunks, or swapped to a symlink, skipped `deleteByFile` while the advancing baseline guaranteed no later diff revisited it (this also held on the plain incremental path before this PR). Both deterministic skip branches now purge before continuing (`purgeStaleRows`), with an induced-regression test (a completed file emptied since the epoch: store 240 → 160 rows). The transient read-error skip deliberately does NOT purge — deleting rows on a maybe-transient failure loses good rows, which is worse than stale ones; the residue is named in-code.

## CodeRabbit round 1

2. `--full` flag entry vs resume behavior — **FIXED**: the flag list names the resume and points at Crash recovery.
3. Direct `resolveEffective()` coverage — **FIXED**: forced-resolution-before-any-embed test + Ollama-fallback-identity-without-embed test.
4. Test git helper `execSync` string interpolation — **FIXED**: `execFileSync('git', args)` with `--` before all pathspecs; `rev-parse` reads converted. Clears both SAST findings.
5. Checkpoint O(n²) rewrite — **DECLINED with measurement**: rewrites happen per-flush (~100 chunks), not per-file — this repo's 5,358-chunk corpus is ~54 rewrites of a ≤50KB file ≈ 2.5MB total I/O across a multi-minute paced run. The append-format refactor buys nothing at realistic corpus sizes; revisit at ~100× scale.
6. `add_lesson` tool description — **FIXED**: states the deferral contract.
7. `errorDetails` null-entry guard — **FIXED** (real: a throw there fired inside the retry catch and replaced the original error).
8. Marker probe reuses `hasFullSyncCheckpoint` — **FIXED**.
9. Changeset lists `hasFullSyncCheckpoint` — **FIXED**.
10. Duplicated `setTimeout` spy blocks — **FIXED**: `captureDelays()` helper, six sites.

## CodeRabbit round 2

11. `purgeStaleRows` error handling — **FIXED (misattribution half) / DECLINED (warn-and-continue half)**. Real: the symlink-site purge sat inside the read try/catch, so a delete failure would be mislabeled a read error and silently downgraded to a skip. The stat and read guards are now separate catches and every purge runs outside them. The proposed warn-and-continue is declined: a swallowed purge failure is permanent silent staleness (the diff window passes and never revisits), while fail-loud CONVERGES — the surviving marker / stale baseline re-evicts the file and retries the delete on the next run, as the finding's own note concedes ("the checkpoint mechanism would allow a later resume").
12. `expectedDimensions` named constants in test assertions — **DECLINED**: expected-value literals in assertions are this suite's established pattern (pre-existing `toBe(1536)` / `toBe(768)` throughout); a per-test alias renames the number without adding semantics.
13. Named constant for the synthetic timer handle — **DECLINED**: a dummy handle in a test mock, matching the sibling `captureWaits()` helper in pacing.test.ts and this suite's pre-existing mock.

`local-lane:` — no verdict artifact recorded for this branch lineage (`totem review --covariate`, sensor exit 0); no line to carry.

PR #2572, opened 2026-08-03
feat(hooks): observability leg in the refresh-gh spawn blocks — repo-local stamp log + child-fd handoff

### LINKED ISSUE #2570: feat(hooks): observability leg in the refresh-gh spawn template - silent no-write is currently undiagnosable (reap vs failure)
Routed from status-claude's 2026-08-03T2130Z dispatch (owed edge @holder:totem-claude). The #2557 managed hook-template spawn block (\spawn('totem-status', ['refresh-gh'], {detached: true, stdio: 'ignore'})\ in CLAUDE_SESSION_START / GEMINI_SESSION_START / post-merge / session-context.mjs) can fail with ZERO trace: a real silent no-write was observed on a SessionStart(clear) firing at ~2107Z (banner printed, guard passed, no snapshot write; the verb itself healthy minutes later). Status-seat re-derivation ruled out PATH and the cwd-shadow exe (libuv bare-name spawn searches PATH only, verified live), leaving two causes the template cannot distinguish: (a) hook-harness process-tree cleanup reaping the detached child (Windows \detached\ is DETACHED_PROCESS, not job-object breakaway — a tree-kill takes the child ~2s into its ~15s run) vs (b) transient child failure. Under stdio:'ignore' + the verb's exit-0-or-nothing contract both leave nothing. Fix shape (proven as a local divergence on both totem-status hooks, comment-marked against the 1.109.0 template): each firing appends a stamp line (ISO time, vehicle, cwd, path-has-go-bin, cwd-shadow-exe) to \<workspace>/.totem-status-refresh-hook.log\ and hands the child the same fd so the verb's success line lands after the stamp — a stamp with no verb line discriminates reap from failure. Belongs in the template (distributed to every seat; hand-carried user-owned hooks are beyond drift-repair), regenerating the byte-copied consumers. Optional defense-in-depth: absolute-path resolution (cwd-shadow only bites shell-mediated spawns). Anchors: #2556 (closed tracking), #2557 (the wiring PR), mmnto-ai/totem-status#127. Target: before/with the next release cut.

satur8d:
Closes #2570 

Routed from status-claude's 2026-08-03T2130Z dispatch (owed edge, cut-coupled): a `SessionStart(clear)` firing was observed producing **no snapshot write with nothing to diagnose** — under `stdio: 'ignore'` plus the verb's exit-0-or-nothing contract, a hook-harness process-tree reap of the detached child (Windows `detached` is not job-object breakaway) and a transient child failure both leave zero trace.

## What ships

Every managed refresh-gh firing site — the Claude and Gemini SessionStart templates, the git post-merge hook, and this repo's bespoke `session-context.mjs` — now:

1. **Stamps a repo-local log** (`.git/totem-status-refresh-hook.log`) before spawning: ISO time, firing site, cwd, plus PATH/cwd-shadow diagnostics on the Node sites and the **resolved binary path** on the shell site (directly diagnosing the cwd-shadow class).
2. **Hands the child the same log** (inherited fd on Node, appended file on sh), so the verb's own success line lands after the stamp. **A stamp with nothing after it = the child never finished** — the reap-vs-failure discriminator status-claude asked for, proven in their local divergence and ported faithfully.
3. **Degrades silently**: log-unwritable falls back to the previous blind firing (the shell `2>/dev/null` precedes the append — redirections apply left to right); ENOENT stays stderr-silent but now leaves a `spawn-error code=ENOENT` log line; the stamp can never block or gate the spawn.

The log lives **inside `.git`** deliberately (falsification-leg round): never tracked, per-repo (no cross-repo stamp interleaving), writable wherever git itself writes, and — decisive for published templates — no un-gitignorable file grows *outside the repo tree* for non-adopting consumers, whose ENOENT firing still stamps. 1 MiB self-cap bounds growth. Byte-copied consumers regenerated (`.gemini/hooks/SessionStart.js`, `tools/post-merge`, installed git hooks); `doctor --parity` 33 pass / 0 fail.

## Review of record

One falsification leg (4 MAJORs, all absorbed): the shell fallback's stderr leak (redirect-order, leg-proven with a real merge against an unwritable log), the outside-tree log for non-adopters, two contract-breaking template mutants the string tests couldn't see (unguarded stamp, close-before-spawn — now killed by POSIX behavioral cells per Node template), and the adjacency-pairing soundness under concurrent firings (resolved by per-repo logs + a documented time-window pairing rule). Leg-verified survivors: emitted-JS validity via `node --check`, escaping fidelity both regex arms live-exercised, ENOENT silence, regeneration byte-identity three ways, no unported spawn carrier repo-wide.

Live-fired both local sites twice (pre- and post-absorption): stamp-then-verb-line pairs observed in the `.git` log, including `bin=/c/Users/jmatt/go/bin/totem-status` on the shell stamp. Full CLI suite green (3712), lint PASS, prettier clean.

Rides the same release cut as #2569 per the operator's ruling; discharges `owed: refresh-gh spawn observability leg @holder:totem-claude`.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@gemini-code-assist Round dispositions below (fixes in `cd31c969`).

## Gemini Code Assist (2 inline — same finding on the template and its generated copy)

1. Shell cap comparison vs `wc -c` padding — **FIXED via your suggested arithmetic expansion**, with one premise note: our CI legs never combine the risky pair (ubuntu = dash + GNU wc unpadded; macOS = padded wc + bash sh), so whether dash's `test` rejects padded integers was exactly the untested combination — the arith wrap eliminates the class regardless. Applied in `buildHookContent` and regenerated into `tools/post-merge` (the generated copy tracks the template byte-for-byte; only the template needed the source fix).

## CodeRabbit (1)

2. Sanitize path-derived stamp fields — **FIXED at all six sites**: the Node templates and the bespoke hook scrub control characters from `cwd` (`replace(/[\x00-\x1f\x7f]/g, '?')`), and the shell site pipes `pwd` and `command -v totem-status` through `tr -d '[:cntrl:]'`; consumers regenerated. Valid catch — these are published templates and the coding guideline mandates sanitization for `.log` writes.

## Greptile (1 P2 + summary)

3. Cap-truncation race at the 1 MiB boundary — **DECLINED with the residual documented** in `.totem/specs/refresh-gh-hooks.md`: lock-free truncation cannot be atomic (rotation variants shuffle the race rather than closing it), the exposure is one diagnostic line once per MiB under same-instant same-repo firings, and it self-heals on the next firing — consistent with the summary's own non-blocking verdict. While verifying this finding, a second same-class residual was observed live on Windows and documented alongside it: Git-Bash `>>` is not kernel-append there, so a same-second cross-site double-fire can overwrite a concurrent stamp's prefix (POSIX `>>` is O_APPEND-atomic; serialized firings — the normal single-flight case — are unaffected on all platforms).

`local-lane:` — no verdict artifact recorded for this branch lineage (`totem review --covariate`, sensor exit 0); no line to carry.

satur8d:
@greptileai review
@coderabbitai review

satur8d:
@coderabbitai @greptileai Round-3 dispositions (`8a592fd3`).

## CodeRabbit (1)

1. Cap can overshoot by one record at the boundary — **FIXED (spec wording) / DECLINED (code)**. The observation is arithmetically correct and by design: the pre-append check means a boundary firing leaves the file one ~150-byte record over the 1 MiB threshold, which the next firing truncates. The spec now names it a SOFT cap explicitly instead of implying an exact bound. Projected-size enforcement at four sites would tighten a diagnostic file's bound by ~0.03% while adding arithmetic to every firing path — declined as buying nothing.

## Greptile

2. Re-review verdict 5/5 ("safe to merge, with only the previously reported cap-boundary diagnostic race remaining") — the remaining item is the accepted residual documented in `.totem/specs/refresh-gh-hooks.md`; no action.

`local-lane:` — no verdict artifact recorded for this branch lineage (`totem review --covariate`, sensor exit 0); no line to carry.

PR #2576, opened 2026-08-04
fix(lock): heartbeat the sync lock + holder-verified deletion — staleness becomes a real death signal under paced syncs

### LINKED ISSUE #2564: fix(lock): 120s stale threshold is stealable under paced/throttled syncs; the #2562 checkpoint invariant depends on the mutual exclusion

satur8d:
Closes #2564 

## What

The 1.110.0 ingest pacing defaults (gemini 4s/batch, 60s-floor 429 backoff — #2562) put multi-minute wall-clock inside the sync lock by construction, while `lock.ts` wrote its timestamp once at acquisition and treated >120s staleness as final authority. Most paced full re-indexes were therefore judged dead mid-run, and a second sync could steal the lock exactly when the #2562 checkpoint invariant needed mutual exclusion (issue body + 2026-08-03 exposure amendment). Spec + approved implementation design: `.totem/specs/2564.md` (operator-approved with recs: keep 120s + 15s beat; pipeline-only lost-lock probe; ADR-115 Train-1 in-boundary).

- **Heartbeat**: the holder refreshes the lock timestamp every 15s (atomic temp+rename beats, `unref`'d timer). Staleness now means eight consecutive missed beats — a dead or wedged holder — and stays the final authority over PID liveness (lesson-5b1929d1: PID probes lie across container namespaces; PID-alive never vetoes staleness).
- **Identity-verified deletion** on every path (lesson-b813e60b TOCTOU narrowing): the stale steal and dead-PID reclaim re-read and delete only the exact judged-dead lock (`deleteLockIfUnchanged`); `release()` — previously an unconditional unlink that would delete a thief's valid lock — verifies ownership and never deletes present-but-unreadable bytes.
- **Lost-lock surfacing**: each beat self-verifies ownership; a foreign holder latches a one-way `isLost()` on the release handle. The pipeline probes it at flush entry/exit, `beginFullSync` (marker overwrite + `store.reset()` — the exact mutation the issue names), the resume carry-forward marker write, `writeSyncState`, the marker delete, and `writeIndexMeta` (DATABASE_MISMATCH evidence, not a regenerable cache). A stolen lock aborts loudly with the checkpoint intact; the next sync resumes the epoch (ADR-115 mutating-degraded-path rule).
- **MCP `add_lesson` no longer loses lessons to a long-held lock**: lockless write under a live epoch; bounded acquisition (`maxRetries: 4`, ~7.5s) with a safe lockless fallback everywhere else. Previously the unstealable hold made it a deterministic ~4-minute block ending in a lost lesson.
- **Compat**: legacy `{pid, timestamp}` locks parse and age out unchanged; `acquireLock`'s return stays callable as `() => void` (`LockRelease` is callable-with-property); zero-arg `withLock` callbacks unaffected. Changeset is `minor` with the new API enumerated.

## Review of record — falsification-leg cascade, 3 rounds to convergence

Full per-finding dispositions in `.totem/specs/2564.md` (§ rounds 1–3). Shape:

- **Round 1** (base diff): 3 MAJOR / 6 MINOR / 3 NIT, two mutant-proven. Accepted: the E2E theft cell couldn't pin *which* probe fired (probe-deletion mutant survived — killed by checkpoint-contents + non-vacuous recovery assertions); `add_lesson`'s ~255s block-then-lost-lesson; the unprobed `beginFullSync` reset; release's unreadable-file deletion; four cell strengthenings; the now-hazardous "delete it manually" hint; changeset `minor`.
- **Round 2** (fold-scoped): verified round 1 discharged, ADR-115 bar declared met. New MAJOR: `hasFullSyncCheckpoint` was a proxy — long paced *incremental* holds write no marker; fixed with bounded acquisition + lockless fallback. Also: two more probes (one correcting my own round-1 decline — index-meta launders DATABASE_MISMATCH evidence), misattribution fix in release's warning, `.hb` temp gitignored.
- **Round 3** (fold-scoped): probe set verified closed (the flushBuffer append has zero awaits after its exit probe), **convergence declared — nothing MAJOR-class open**; residuals were message-honesty items, absorbed: the deferral message claims neither an unproven cause nor indexing-by-the-holder (executed counterexample: an undeletable corrupt lock throws `SYNC_FAILED` in 1ms).

**Declined, falsifiable** (recorded with reasons in the spec): theft-after-final-probe warn-only exit (residual now only the index-manifest write — genuinely regenerable); the in-process-unreachable wx EEXIST arm; `store.deleteByFile`/`createFtsIndex` probes (bounded, self-healing via #2562 store verification — never false-green).

## Tests

- 10 new lock cells (heartbeat freshness/identity, never-stolen-while-beating, per-field deletion discrimination, theft latch + zero-writes-after, release leaves foreign and unreadable locks, warn-once beat failure, wx reclaim, ghost-beat teardown, legacy both-halves, bounded budget).
- E2E induced theft cell (ADR-115 shape: induced failure → loud throw → which-probe pinned via checkpoint contents → prior good state preserved → non-vacuous recovery control). All 16 #2562 induced-regression cells untouched and green.
- 4 new MCP cells (live-epoch no-contend, normal-path bounded lock, lockless fallback writes the lesson, non-`SYNC_FAILED` fails loud).

## Gates

Workspace build ✓ · full workspace tests ✓ (core 65, mcp 210, cli 3709) · `format:check` ✓ · ESLint ✓ · `totem lint --base main` **PASS 0 errors** (121 advisory warnings, dominated by the known test-idiom classes) · changeset ✓ (`minor`).


satur8d:
@greptileai review
@coderabbitai review

satur8d:
@coderabbitai @greptileai Round 1 disposition (fixes pushed in `b97cf3fc`):

1. **CR (lock.ts recovery hint) — accepted, fixed.** The hint now reports the effective `staleThresholdMs` (`Math.ceil`, so overridden thresholds never render "~0 minutes").

2. **CR (add-lesson.ts magic number) — accepted, fixed.** `MAX_LESSON_LOCK_RETRIES = 4` named at module top with the ~7.5s worst-case budget documented; the call site uses it. The test keeps asserting the literal `4` — pinning the constant against itself would be tautological.

3. **GT P1 (heartbeat overwrites new ownership) — declined with analysis; residual disclosed in-code.** The diagrammed interleaving is structurally impossible as stated: B's `wx` cannot succeed while A's lock file exists, and the sequence omits the mandatory prior *deletion* — which requires a stale (>120s silent) or dead-PID judgment that a healthy heartbeat prevents. The reachable variant needs an already-degraded holder (beats failing/stalled ≥8 intervals, the warn-once state) plus sub-millisecond cross-process preemption inside a fully synchronous read→write→rename sequence. The "neither lost-lock latch fires" claim is also incorrect: a clobbered peer's next beat reads a foreign `holderId` and latches, so double-ownership converges within one beat interval (≤15s), with interim store writes bounded by the #2562 store-verification on the next resume — wasted quota, never false-green. No filesystem CAS exists to close the window fully; it is the write-side mirror of the deletion-TOCTOU residual this module already documents. Disclosed as a comment at the rename site in `b97cf3fc`. Falsifier: a demonstrated healthy-holder interleave, or harm exceeding the wasted-quota bound, reopens this.

local-lane: no verdict artifact recorded for this branch lineage (local review lanes not configured on this machine — line honestly absent, not omitted).

Issue #2558, opened 2026-08-03
bug(gemini): .gemini/settings.json SessionStart hook entry is schema-invalid on Gemini CLI 0.53.0 - discarded at boot, no totem briefing


satur8d:
## Defect (live-probed 2026-08-03, Gemini CLI 0.53.0)

Running a headless gemini session in this repo prints, before the session starts:

```
Discarding invalid hook definition for SessionStart from project: {
  name: 'totem-briefing',
  type: 'command',
  command: 'node packages/cli/dist/index.js briefing',
  timeout: 30000,
  description: 'Load Totem session briefing on startup'
}
```

So `.gemini/settings.json`'s `hooks.SessionStart` entry — the only `hooks` key any cohort repo carries (2026-08-03 census: totem is the sole repo with one) — no longer validates against the current Gemini CLI hook schema, and **the gemini vehicle boots in this repo with no totem briefing at all** (the entry is discarded, not degraded).

Related census observation (same probe session): the drift-repaired `.gemini/hooks/SessionStart.js` managed artifact is not referenced by any `settings.json` in any repo — in 4 of 5 repos carrying it there is no `hooks` key at all, so how (or whether) it ever executes is underivable from the repos. If Gemini CLI has no filename-convention discovery, the managed Gemini SessionStart template may be a dead surface cohort-wide — worth resolving in the same fix.

## Scope

1. Derive the current Gemini CLI hook schema (vendor docs / CLI source) and re-author the `settings.json` entry so `totem-briefing` fires again.
2. Establish whether `.gemini/hooks/SessionStart.js` is reachable at all under current Gemini CLI; if not, either wire it via valid settings entries (and distribute that shape through the managed-template path) or retire the artifact honestly.
3. Cohort distribution rides the managed-template drift-repair path once the shape is known.

## Provenance

Census step-zero for mmnto-ai/totem#2532 (probe transcript in that issue's lane); the .gemini/skills non-discovery finding from the same probe is handled in #2532 itself — this issue is ONLY the hooks-schema defect.

satur8d:
Scope amendment from the 2026-08-03 hooks-lane sweep (explore-leg census, repo-side claims re-verified first-hand). Two facts strengthen this issue's framing from "the settings entry is schema-invalid" to "the surface is dead by construction":

[...]

PR #2583, opened 2026-08-05
fix(hooks): ship Gemini SessionStart as .cjs to stop the type-module fail-open

### LINKED ISSUE #2488: fix(hooks): Gemini SessionStart.js has the same type-module ESM fail-open (.js to .cjs, sibling of the BeforeTool fix)
## Defect

`.gemini/hooks/SessionStart.js` carries the identical latent ESM fail-open fixed for BeforeTool in PR mmnto-ai/totem#2485: CommonJS content (`require('child_process')` at top level) under a `.js` extension. In a `"type": "module"` consumer repo Node resolves it as ESM and it throws before running; Gemini CLI degrades the crash to a warning, so the session-start briefing is silently absent.

## Severity

Lower than the BeforeTool case by design: this is the orientation/briefing hook, not a write-time guard — nothing enforcing fail-opens; the consumer just boots cold. Deliberately scoped OUT of mmnto-ai/totem#2481 ("the slice scopes to the write-time guard") and disclosed on PR mmnto-ai/totem#2485.

## Fix shape (small — the machinery now exists)

Mirror #2485 exactly: ship `SessionStart.cjs`, move the roster entry, add the legacy pair to `LEGACY_MANAGED_SESSION_HOOKS` (the rename-migration path built for BeforeTool generalizes as-is), extend `totem eject`, and add the type-module acceptance pair (the `.cjs` evaluates; the identical body as `.js` provably fail-opens).

Related: mmnto-ai/totem#2481, mmnto-ai/totem#2485.

satur8d:
Closes #2488 

## What

`.gemini/hooks/SessionStart.js` carried the identical latent ESM fail-open fixed for BeforeTool in #2485: CommonJS content under a `.js` extension. In a `"type": "module"` consumer, plain-node execution throws `require is not defined` before the briefing emits — reproduced live on totem-status (which replaced the block with a user-owned ESM rewrite to work around it); liquid-city is a live `"type": "module"` consumer carrying the broken shape. Per #2558 no `.gemini/settings.json` registration is ever emitted for this hook and Gemini CLI has no filename-convention discovery, so host/plain-node paths are the only executions — exactly where the break bites.

Mirror of #2485, machinery reused:

- Roster ships `.gemini/hooks/SessionStart.cjs` (`GEMINI_SESSION_START_REL`); legacy pair added to `LEGACY_MANAGED_SESSION_HOOKS`. `totem hook install` (which `prepare` invokes on every install) and `totem init` migrate an adopted consumer under the ownership gate: user-owned legacy skipped untouched; drifted-unbounded declined until `--force`.
- **Ownership gate extended to the successor path** (falsification find, new semantics), arm-for-arm with `regenerateManagedSessionHooks`: a user-owned file at the successor path blocks the migration even under `--force` (`reason: 'user-owned-successor'`, both files stay); a marker-headed drifted-unbounded successor is declined until `--force` (`reason: 'drifted-successor'`); bounded or forced takes the canonical write as idempotent repair. Blocks are disclosed on the `totem hook install` summary. Protects the BeforeTool pair identically.
- `totem eject` removes both shapes; `totem doctor --parity` now checks `.cjs` — an un-migrated consumer reads honest-absent `skip`, never drift (locked by test, blocking-manifest so the assertion can actually fail).
- No registration migration: none exists to migrate (#2558); arming the hook in Gemini CLI stays that issue's scope.
- Type-module acceptance pair: the rendered `.cjs` executes under `"type": "module"` through to its boot-safe breadcrumb (proves the CommonJS body ran); the identical body as `.js` provably throws.
- In-repo: our own tracked `SessionStart.js` (bounded totem-owned, byte-identical to canonical, 6380/6380) is renamed here to pre-empt the unstaged-delete + untracked-file every consumer of this repo would otherwise hit on their next `pnpm install` — unlike #2485's `BeforeTool.js`, which is unmarkered and permanently immune to the migration.

## Review record

Opus build leg (mechanical mirror) → owner verification of the full diff against primaries → Opus falsification leg, round 1: **2 MAJOR + 2 MINOR, all accepted and fixed** in `d21f33f4`:

- **MAJOR-1**: shipped prose claimed "Gemini CLI discovers the session hook by path" — refuted against the vendor's own hooks docs (hooks come only from `settings.json`) and against #2558, which the prose inverted while citing it. All four sites rewritten to the true premise; the wiki example now names the registration requirement.
- **MAJOR-2**: `migrateLegacyGeminiHooks` wrote the successor unconditionally — executed repro destroyed a user-owned `.cjs` at the successor path with zero disclosure (the spec's required dual-presence case; pre-existing #2485 machinery, made newly reachable). Fixed + two locking tests. Not live today: totem-status's and liquid-city's actual `SessionStart.js` files are unmarkered → `skipped` untouched.
- **MINOR-1**: the new parity test's `blockingDriftIds` assertion was vacuous on a non-blocking manifest — now falsifiable.
- **MINOR-2**: the distributed `CLAUDE_SESSION_START` body still named the `.js` sibling — fixed; consumers absorb it as routine bounded drift-repair on next install.

Falsification round 2 (scoped to the round-1 fold): MAJOR-1/MINOR-1/MINOR-2 verified discharged; **1 MAJOR found in my own fold** — the successor gate checked only `markerOpensFile`, silently destroying a marker-headed-but-unbounded successor as `migrated` (executed three-arm repro), strictly more destructive than the function the comment claimed to mirror — plus a residual falsified-premise comment the round-1 census missed and a disclosure overclaim vs the `totem init` path. All fixed in `0eba73bc`: the successor arm now declines drifted-unbounded until `--force` (locking test both arms), the comment carries the host/plain-node framing, prose names the hook-install summary as the disclosure path, and the committed spec carries an amendment note recording its falsified premise.

Falsification round 3 (scoped to the round-2 fold): **convergent — presentable.** All round-1 and round-2 findings verified discharged by execution (three-surface consistency matrix: the successor gate is now arm-for-arm identical with the legacy gate and `regenerateManagedSessionHooks` across user-owned / drifted-unbounded / bounded × force arms); census clean (the falsified premise survives nowhere except the spec's preserved Problem Statement and the amendment that corrects it); no new defects. The leg's one non-blocking NIT (a read-order pointer at the spec's falsified premise) is taken in the final commit. The leg named further rounds as the infinite-polish gradient.

Advisory `totem review` fan (supplementary lanes): 0 critical, 0 warn, 2 info — both cosmetic, declined with reasons in the session record.

## Residuals (disclosed, not in scope)

- This repo's `.gemini/settings.json` `hooks.SessionStart` entry points at a nonexistent `briefing` verb — pre-existing, #2558's scope.
- Vendor "Golden Rule": a Gemini hook's stdout must be a single JSON object; the briefing writes plain text. Irrelevant until #2558 arms the hook; noted on that issue.
- The committed spec (`.totem/specs/2488.md`) carries generated-stale file paths (`packages/cli/templates/` never existed; templates are inline constants) — the implementation built to the real file map, and the spec's three test directives were honored verbatim.
- `init-templates.ts` line ~776 (inside the `GEMINI_*` comment block context): the `'migrated'` console line cites `#2481 + #2488` as a class, not per-file — deliberate.

## Gates (final branch state)

`pnpm -r build` exit 0 · full workspace tests exit 0 (cli 153 files / 3721 passed) · `pnpm run format:check` exit 0 · ESLint 0 errors · `totem lint` branch-vs-base **PASS, 0 errors** (all 13 changed files enumerated).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## CR round 1 disposition — fixes pushed as `a665dec4`

1. **install-hooks.ts execution-path comment (Minor) — FIXED.** Confirmed: the migration section header folded SessionStart into the "Gemini CLI treats the crash as a warning" claim, contradicting #2558 four lines below it. Now scoped: the warning claim applies to BeforeTool only (which IS registered in `.gemini/settings.json`); SessionStart's break is named as host/plain-node execution per #2558. Same class as the falsification leg's MAJOR-1 — this site survived three census greps because it says "treats" where every probe grepped "degrad"/"discovers by path". Good catch.

2. **Eject deletes marker-quoting user hooks (Major) — FIXED, with an amended cure.** Mechanism confirmed: the gate was a whole-body `includes()`. The proposed `markerOpensFile` cure would regress `.gemini/skills/totem.md`, which opens `<!-- [totem] auto-generated … -->` — HTML-comment shape, so the `//`-prefixed marker never *opens* it and eject would leave it behind. Ownership is now **marker-in-first-line**, matching every scaffold shape (`//` hooks, `<!--` skills) while refusing body-quotes. Regression tests lock both directions: a user hook quoting the marker in its body survives eject; the HTML-comment skill is still removed. The old-generation `BeforeTool.js` (pre-marker era, no marker string anywhere) behaves identically before and after: skipped.

3. **CWE-426 untrusted search path at `SessionStart.cjs` 74–77 (Major, body-only) — DECLINED for this PR, routed to #2558.** The Windows CWD-before-PATH mechanism claim is not disputed. Grounds: (a) the file entered this diff as a content-neutral rename (`similarity index 100%`) — the spawn/exec body is pre-existing shipped template content, unchanged here; (b) the threat model needs an attacker who controls the checkout root but not `.gemini/hooks/**` — the hook file is itself repo content, so full repo control ships arbitrary hook code directly; executable-origin hardening defends only a partial-write-control corner; (c) per #2558 no Gemini CLI execution path exists (no registration is ever emitted, no filename-convention discovery), so nothing auto-executes this file in a fresh checkout. The proposed cure (install-time trusted-absolute-path generation across the template) is template-redesign work that belongs with #2558's arming slice, where the execution contract gets defined — a routing note is on that issue.

Greptile: 5/5, no findings. Gemini Code Assist: reviewed, "no feedback to provide."

local-lane: ec22e096 round=0 settled=true lanes=2/2

satur8d:
@coderabbitai review

satur8d:
@coderabbitai

## CR round 2 disposition — fix pushed as `56e74b93`

1. **Prefix-anchor the eject ownership gate (Major) — FIXED as proposed.** Confirmed against the round-1 fix: `firstLine.includes()` still deleted a user hook whose *first line* quotes `"[totem] auto-generated"` in a sentence — the round-1 regression test only covered a later-line quote. The gate now requires the first line to *be* a generated header: `startsWith('// [totem] auto-generated')` or `startsWith('<!-- [totem] auto-generated')` after `trimStart()`, verified against every scaffold's actual opening (JS/CJS hooks and the HTML-comment markdown skill). First-line-quote regression test added; the round-1 tests (body-quote survives, HTML-comment skill still ejects) stay green.

Gates on `56e74b93`: workspace build 0 · cli 3725 tests green · format 0 · ESLint 0 · `totem lint` branch-vs-base PASS 0 errors.

local-lane: ec22e096 round=0 settled=true lanes=2/2

Issue #2334, opened 2026-07-10
status Shield line + incremental-shield eligibility read `.shield-passed` — retired by the Prop-207 gate reset, nothing writes it (permanently-red sensor; incremental path can never arm)


satur8d:
## Symptom (field report — lc-claude, 2026-07-10)

`totem status` on liquid-city reports `Shield: missing` — permanently: observed at signon on 1.90.0 and unchanged after the 1.92.0 upgrade + full `totem sync` — while the shipped review gate is demonstrably live (`.totem/cache/.reviewed-content-hash` present and maintained by the installed hooks per the `totem.config.ts` `review.sourceExtensions` contract; both of that day's pushes ran the pre-push tripwires clean). Reported via ECL dispatch `2026-07-10T1924Z` (operator-routed to strategy-claude); working context mmnto-ai/liquid-city#746.

## Derivation (source read @ `8b2787e1`, 2026-07-10 — strategy-claude)

Two readers, zero writers:

- **Reader 1 — the status sensor.** `packages/cli/src/commands/status.ts` (~L55–66) reads `.totem/cache/.shield-passed` expecting the git SHA of HEAD: `missing` when absent, `passed`/`stale` on SHA compare.
- **Reader 2 — incremental-shield eligibility (mmnto-ai/totem#1010).** `packages/cli/src/commands/shield.ts` `evaluateIncrementalEligibility` (~L1019–1036) reads the same flag as "last passed SHA"; absent ⇒ `{ eligible: false, reason: 'No previous shield state' }`.
- **Writer: none.** The only writes of `.shield-passed` anywhere in `packages/` are test fixtures (`status.test.ts`, `shield-incremental.test.ts` setup). The shipped PASS path stamps `.reviewed-content-hash` (`shield.ts` ~L415/475) — a content hash, not a SHA.
- **The flag was retired deliberately.** CLI CHANGELOG, Proposal 207 "Gate architecture reset": *"replaced SHA-based flag files with stateless git hooks (lint + verify-manifest) and content-hash-based PreToolUse review gate. … Removed all flag files (.lint-passed, .shield-passed, .spec-completed)"* — and `install-hooks.test.ts` asserts generated hooks do **not** contain `.shield-passed`.

lc-claude verified the same reader logic in the published 1.90.0 and 1.92.0 `dist/commands/status.js`, so this is the shipped behavior, not a working-tree artifact.

## Impact

[...]

PR #2589, opened 2026-08-06
ci(smoke): file-on-red alert — a red smoke lands in the issue queue

### LINKED ISSUE #2582: ci(smoke): daily Integration Smoke red is silent — 5 consecutive scheduled failures went unnoticed (permanently-red-sensor class)
## Defect

The daily Integration Smoke Tests schedule ran red for 5 consecutive days (2026-07-31 → 2026-08-04) with no delivery channel to anyone: no issue filed, no board surface, nothing in orient. The red was only noticed during unrelated release verification on 2026-08-04, and root-caused to the org-level `ANTHROPIC_API_KEY` credit drain (secret vintage 2026-06-23 — #1956's subject; operator rotated it same day; the dispatched post-rotation run went green at 18:27Z, run `workflow_dispatch/success`).

Receipts: scheduled runs 2026-08-03T07:03Z and 2026-08-04T06:49Z both `failure`; post-rotation dispatch 2026-08-04T18:27Z `success`.

## Class

Permanently-red-sensor (kin of #2334): a sensor that can stay red for days without anyone seeing it is a masked sensor — worse than no sensor, because it reads as coverage. The smoke suite exists precisely to catch integration rot between releases; five silent red days means the sensor's own delivery path is the gap, not its detection.

## Fix shape (options)

1. **File-on-red (recommended, minimal):** an `if: failure()` step on the scheduled run that files or comments an issue via `GITHUB_TOKEN` — idempotent (one open issue, deduped by title, comment per subsequent failure, auto-close comment on next green). Loud, zero new infrastructure, lands in the triage surface the cohort already reads (the issue queue).
2. **Orient/doctor row:** a derived surface row (totem orient or the totem-status sidecar) reporting last-scheduled-run conclusion per sensor workflow via the Actions API. Honest derivation, but only visible at signon — a red would still wait for the next session.
3. Both — file-on-red as the loud path, the orient row as the derived confirmation.

Recommend starting with (1); (2) can ride the #2580 estate-sensor design if that lands a sensor surface anyway.

## Refs

#1956 (org-secret vintage + rotation), #2334 (permanently-red-sensor kin), Integration Smoke Tests workflow.

satur8d:
Closes #2582

## What

The file-on-red alert for the daily Integration Smoke (#2582 fix shape 1, as recommended): a new `alert` job on `ci-integration.yml` that lands a red smoke in the issue queue — the triage surface the cohort already reads — instead of leaving it in the Actions tab, where six consecutive scheduled reds (2026-07-30 → 2026-08-04) went unseen until unrelated release verification found them.

Mechanics: one open issue deduped by exact title (state re-checked on fresh JSON objects, not the eventually-consistent search index); a comment per subsequent red; on the next green, a comment-and-close of **every** open alert (a race-born duplicate must not leak a permanently-open "red" that trains everyone to ignore the alert). Runs on `schedule` and `workflow_dispatch`, so a post-fix manual green closes the alert. No new action dependencies — the job uses only the preinstalled `gh` with job-scoped `permissions: issues: write`.

## Failure-direction design (the point of the review pass)

A pre-PR falsification leg attacked the mechanics against GitHub's documented semantics — this path cannot be tested until it fires live:

- **Its blocker finding, folded:** GitHub maps a `timeout-minutes` expiry to *cancellation*, and the draft's `cancelled` arm was mute — a hung provider API (the most characteristic smoke red) would have been silent green, the exact #2334-class gap this issue exists to kill. The script now treats **anything that is not `success`/`skipped` as red**, so the aggregate-result mapping no longer matters.
- Also folded from the leg: the stale-search-index case (a red commenting on an already-closed alert while the open queue shows nothing — state is now re-checked precisely), the close-only-first leak, and the `|| true` that made a list failure read as "no existing issue". Every API failure is now loud — a red alert job is visible in exactly the place a silent one is not.
- Confirmed by the leg against primaries: matrix aggregation (`fail-fast: false` + no `continue-on-error` — no red leg can report green), `always()` gating, token capability on scheduled runs at job-level permissions (and permission failures fail loud), and the exact `gh` pipeline including quoting, run read-only against this repo with a positive and negative control.

Known accepted residue: a dispatch overlapping the 06:00 UTC schedule can transiently duplicate the issue (bounded by search-index lag); the close-all green path self-heals it.

## Verification

Workflow-only diff — no releasable surface, no changeset. Prettier clean, `totem lint` PASS 385 rules / 0 violations. The alert path itself first executes on the next scheduled run; its failure direction is red-and-visible by construction.

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -54,6 +54,13 @@ jobs:
     if: ${{ always() && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    # Serialize alert jobs (CR round 1): two overlapping runs must not both
+    # observe "no existing issue" and double-create. Classic group syntax only —
+    # an invalid concurrency key would not fail this job, it would invalidate
+    # the WHOLE workflow file and take the smoke itself dark. A replaced-pending
+    # alert needs >=3 simultaneous runs of a once-daily schedule; accepted.
+    concurrency:
+      group: ci-integration-file-on-red-alert
     permissions:
       issues: write
     steps:
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -41,3 +41,66 @@ jobs:
           GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+
+  # File-on-red (mmnto-ai/totem#2582): a red smoke must land in the triage
+  # surface the cohort already reads — the issue queue — not wait for someone
+  # to open the Actions tab (six consecutive scheduled reds went unseen,
+  # 2026-07-30 → 2026-08-04). One open issue deduped by exact title, a comment
+  # per subsequent red, and an auto-close comment on the next green. Runs on
+  # manual dispatches too, so a post-fix dispatch green closes the alert.
+  alert:
+    name: File-on-red alert
+    needs: smoke
+    if: ${{ always() && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      issues: write
+    steps:
+      - name: File, bump, or close the alert issue
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          RESULT: ${{ needs.smoke.result }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          EVENT: ${{ github.event_name }}
+          TITLE: 'ci(smoke): Integration Smoke is red (file-on-red alert)'
+        run: |
+          set -euo pipefail
+          # Failures here are LOUD by design: nothing swallows an API error —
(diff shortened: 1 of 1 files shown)


satur8d:
@coderabbitai review

satur8d:
@coderabbitai

## Round 1 disposition (CR r1) — addressed in `a48b1b52`

**Fixed (partial-take of the one finding):** the alert job now carries a job-level `concurrency: group` so overlapping runs serialize — two runs can no longer both observe an empty lookup and double-create. Taken in the **classic group-only syntax**.

**Declined (two halves of the same finding, with reasons):**

- **`queue: max`** — the syntax dates to a 2026-05 changelog per the finding's own citations, and I could not verify it against a primary from this environment. Its failure mode is disqualifying for this file: an invalid `concurrency` key does not fail the alert job, it invalidates the whole workflow, taking the smoke itself dark — the silent-red inverse this PR exists to close. What classic-group serialization leaves uncovered is a replaced-*pending* alert at ≥3 simultaneous runs of a once-daily schedule + occasional manual dispatch — practically unreachable, and the next run self-heals. Falsifiable decline: a primary-source confirmation that `queue: max` validates on current GitHub Actions reopens the upgrade.
- **Paginated lookup** — the query is exact-title-scoped, so the default limit of 30 binds only if 30+ open alert issues with the identical title accumulate; each red run comments-or-creates at most one, duplicates require the (now-serialized) race, and every green closes **all** matches. Unreachable scale; the pre-PR falsification pass verified the limit immaterial for a quoted-title search.

Gates at `a48b1b52`: prettier clean, `totem lint` PASS 385 rules / 0 violations (workflow-only diff — no build/test surface).

`local-lane:` — no verdict artifact for this lineage (workflow-only diff; the review of record is the pre-PR falsification pass documented in the PR body).

PR #2586, opened 2026-08-05
feat(doctor): --estate worktree-estate sensor row (#2580 slice 1)

satur8d:
## What

Part of #2580 (slice 1 of 4 — do not close the issue): the worktree-estate sensor as a `totem doctor` row.

- **`totem doctor --estate`** — enumerates the user-level registry, verifies each entry is a git toplevel (`rev-parse --show-toplevel`), classifies its linked worktrees from local git evidence alone, and sweeps typed roots for unregistered worktree-shaped residue. Report-only: exit 0 whatever it finds; removal verbs are slice 2.
- **`--estate --json`** — diffable artifact on stdout (`estate-schema-version: 1`, kebab-case, presence-only optionals) as the contract for the totem-status sidecar (placement ruled on the issue by the status seat: classifier rides doctor, sidecar consumes).
- **Ambient `Estate` row** in plain `doctor` (ruled open-question 1): named SKIP / quiet PASS / WARN with counts. `gateExempt` — reports through doctor, never gates under any `--strict` tier (exemption scoped to advisory statuses; a `fail` row gates regardless).
- Core scan in `packages/core/src/estate-scan.ts` incl. the repo's first reusable `git worktree list --porcelain` parser; every git invocation carries `--no-optional-locks`.

## Design decisions that will look odd without context

- **`registered-indeterminate`, never `stale`, for clean-but-not-ancestry-merged**: under squash merges (this cohort's discipline) ancestry cannot prove "merged". The sensor declines to claim; the status-lane merged-facts snapshot extension upgrades it later.
- **Evidence-typed husks, two bars by root kind**: container roots (`<repo>/.claude/worktrees`, `--root` declarations) report any git-untracked dir — location is the evidence; standard roots (registry/worktree parents) require shape evidence (dangling `.git` pointer / repo-prefix + node_modules / intact-pointer-unlisted). Totem-registry membership is deliberately NOT protection from candidacy (doctrine: git's worktree list is a worktree's protection) — one path can carry a repo row and a husk row (two axes).
- **Derivation is disclosed, never silent**: roots the derivation declines (os-tmpdir via registry entries; parents reachable only from missing/unverifiable/not-git-root entries; containers of repos whose worktree list failed) land in `excluded-roots` with every reason that declined them.
- **The unscannable ledger is axis-tagged** (`registry` / `worktree` / `sweep`): the exactly-one-bucket partition is a sweep-axis property; registry/worktree probe failures are a different axis and may legitimately share a path with a husk row.

[...]

Issue #2580, opened 2026-08-04
Workspace-estate guardrails: worktree lifecycle (estate sensor -> verbs -> signoff coupling) + stray-artifact hygiene


satur8d:
## The failure (2026-08-04 estate audit, operator-ruled cleanup)

- **~45 orphaned worktree husks** across `D:\Dev` and `D:\tmp`: ≥4 seats (totem-claude, totem-codex, strategy-claude, plus status/lc-lane bump worktrees), ~3 months of accumulation, 5+ naming conventions, two locations — including the workspace root that ADR-106 mail polling and repo discovery glob.
- **Root cause, reproduced on demand ×2 during cleanup:** `git worktree remove --force` on Windows exits **0** while leaving the directory behind (`node_modules` + `packages/` hoist — pnpm reparse points defeat git's recursive delete). The worktree IS deregistered, so `git worktree list` reads clean while the husk persists — invisible to the removing agent and to every later topology check. A Tenet 4 fail-open in the tool itself, which is why discipline alone never held: the seats that minted husks include ones following the removal recipe.
- **Registered-stale class:** two codex worktrees sat registered 2+ weeks after their work merged (mmnto-ai/totem#2451, mmnto-ai/totem#2457). Nothing couples a worktree to the lifecycle of its purpose (ticket/PR), so nothing notices expiry.
- **Sibling class — stray scratch artifacts generally (operator: "the worktrees aren't the only offender"):** agent-written PR bodies, probes, dumps accumulating in shared tmp and the workspace root; stray `totem init` artifacts found in `$HOME/.totem` alongside legitimate user-level state (`registry.json`, ledger); an empty `$HOME/.agents` skeleton; output dumps in `$HOME`.
- **Cross-vendor:** claude, codex, and strategy seats all minted husks — the mechanism must be vehicle-neutral (Tenet 16).

Trap banked strategy-side as `lesson-ad03b5cd` (verified-removal recipe). Doctrine (convention + recipe) lands via the strategy lane and distributes on the doctrine pin; PR ref will be added below when open.

## Proposed slices (sensors-first, Tenet 13)

[...]

PR #2592, opened 2026-08-07
feat(cli,core): totem wt create|remove|list — verified worktree removal + durable registry (#2580 slice 2)

satur8d:
Slice 2 of #2580 (workspace-estate guardrails) — the issue stays open for slices 3–4.

## What this adds

**`totem wt create|remove|list`** — worktree lifecycle verbs with verified removal, backed by a durable user-level registry (`~/.totem/worktrees.json`, a sibling of `registry.json` — never a key inside it, the record schema would reject it).

- **`create`**: record-first (the entry is written before git runs — a phantom record fails visible in `wt list`; an unrecorded worktree fails invisible, which is the husk class the estate audit found ~45 of). Mints `<root>/<repo>-<seat>-<slug>` on a new branch; refuses the repo, anything under the repo, and the workspace root; primary-checkout-only (git-dir == git-common-dir discriminator); root precedence `--root` > `TOTEM_WORKTREE_ROOT` > `~/.totem/worktrees`.
- **`remove`**: never trusts git's exit code — the Tenet 4 fix for the reproduced exit-0-with-residue failure. Sequence per `tool-usage-nuances.md § Git`: ECL refusal (`status --porcelain --ignored -uall -- .totem/orchestration`, blocks on any row, no bypass flag by ruling) → check-first (git-listed paths go through `git worktree remove`; unlisted-but-recorded residue requires the path to sit under a recorded root) → verify absence → reparse-safe residue finish (links unlinked, never followed — junction non-traversal verified with real fixtures) → re-verify → prune. Exit 0 ⇔ a no-follow probe says the directory is gone; every failure path retains the registry entry.
- **`list`**: accounting only (entry fields + on-disk presence). Classification stays with the slice-1 sensor — derive-once.

**Sensor coupling** (the PR #2586 round-3 MINOR-3 remedy): `doctor --estate` and the ambient Estate row sweep recorded roots — the dedicated default root as a container root, any other recorded root as a standard shape-evidence root (so a mixed-use root like `D:\tmp` cannot mint by-location husk rows). Recorded roots survive entry removal, keeping future worktree locations derivable after their entries are gone. The doctor JSON artifact schema is unchanged; only coverage widens. The `checkEstate` registry-unreadable disclosure now rides every row arm (a regression risk the falsification round caught before it shipped).

## Contract + review trail

The governing design, rulings, and the full falsification disposition log (both rounds, including declines with rationale) are in-diff: `.totem/specs/2580-totem-wt-verbs-worktree-create-remove-verified.md`. Review of record: one Opus falsification leg over the full diff (verdict MERGEABLE-WITH-FIXES; all three required fixes folded) plus a scoped second leg over the fold (no HIGH; deposits dispositioned). Design inputs from the issue's four comments (leg corrections + the status seat's placement ruling) are honored: classification lives in doctor only, slice-3 signoff coupling is explicitly out of scope, mail-from-primary is convention rather than env plumbing.

Known deferral (named in the spec log): the lstat existence probe treats non-ENOENT errnos as absence — errno tri-state is a follow-up issue candidate.

## Verification

- `pnpm -r build` green · full `pnpm -r test` green (core 3638, cli 3811, packs incl. the count-exact security-sweep allowlist)
- `totem lint --base main` PASS, 0 errors · `pnpm run format:check` clean · changeset: minor for `@mmnto/totem` + `@mmnto/cli`
- Real-git tests cover the ECL `-uall` arm (including the `status.showUntrackedFiles=no` bypass it closes) and junction-target survival.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 disposition

All 11 findings taken and fixed in 94595a43. Full-gate re-verified on the wire: workspace build, `pnpm -r test` (core 3645 / cli 3814), `totem lint --base main` 0 errors, format clean.

| GT | Concurrent create/remove race can delete the replacement worktree's registry entry | accept | `deleteWorktreeEntry` gained an identity guard (`expectCreatedAt`, checked under the lock); a mismatched entry is refused and `wt remove` warns "replaced concurrently — left in place" — `packages/core/src/worktree-registry.ts` + `packages/cli/src/commands/wt.ts` (94595a43) |
| CR | `registryWarnings`/`wtWarnings` interpolated into doctor rows unsanitized | accept | sanitize + flatten at every interpolation, including the catch arm; no-newline assertion added — `packages/cli/src/commands/doctor.ts` (94595a43) |
| CR | `checkEstate` catch arm drops the worktree-registry disclosure (`wtNote`) | accept | `wtWarnings` + `wtNote()` hoisted above the try beside the sync-registry pair; catch arm carries both; wt-unreadable × scan-throw test added — `doctor.ts` (94595a43) |
| CR | `createOne` helper returns a fixed path while forwarding overrides | accept | return derived from effective options; hand-recomputed call site simplified — `wt.test.ts` (94595a43) |
| CR | `execSync` command strings in the real-git test setup | accept | `spawnSync` argv helper with per-step status assertions; `execSync` import dropped — `wt.test.ts` (94595a43) |
| CR | Leading `~` in `--root`/`TOTEM_WORKTREE_ROOT` mints a literal dir that accretes into recorded roots | accept | tilde expanded against `os.homedir()` before resolution, both sources tested — `wt.ts` (94595a43) |
| CR | `attempts: 0` seam never reaches the delete-failure arm; `lastError` contract untested | accept | rmSync-throw retry case added (`attempts: 2` → `attempts === 2`, `lastError` pinned) — `worktree-residue.test.ts` (94595a43) |
| CR | `residuePathExists` reads any lstat failure as absence | accept | fail-closed: only ENOENT/ENOTDIR read as absent, every other errno reads as present; EACCES/ENOENT tests — `worktree-residue.ts` (94595a43) |
| CR | `worktreePathExists` same fail-open probe; registry entry deletable after an inconclusive probe | accept | same fail-closed change + tests — `worktree-registry.ts` (94595a43). This also retires the errno-tri-state deferral recorded in the in-diff spec's disposition log (independent confirmation of the falsification leg's finding 5) |
| GCA | `log.error()` must use the `'Totem Error'` tag (styleguide Rule 129/21) | accept | tag corrected at the phantom-entry site — `wt.ts` (94595a43) |
| CR | Same tag finding via path instructions | accept | same one-line fix — `wt.ts` (94595a43) |

local-lane: no verdict artifact recorded for this lineage (`totem review --covariate` sensor line) — the branch's review of record is the two-round falsification-leg pass documented in the in-diff spec disposition log.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 2 disposition

Three accepts fixed in 29a4de69 (full gate re-verified on the wire: workspace build, `pnpm -r test` — core 3647 / cli 3815 — `totem lint --base main` 0 errors, format clean); one decline with rationale below.

| CR | Unreadable worktree registry can render `skip`/`pass` while recorded roots went unscanned (outside-diff comment) | accept | degraded-state symmetry with the sync registry: the empty short-circuit arm returns `warn` naming the wt disclosure + `~/.totem/worktrees.json` remediation; the clean arm's downgrade predicate covers both warning sources; tests for both arms — `packages/cli/src/commands/doctor.ts` (29a4de69) |
| GCA | `(err as ErrnoException).code` throws on a null/non-object throw, escaping the fail-closed probe (both probes) | accept | null-safe errno extraction in `worktreePathExists` + `residuePathExists`; `throw null` tests added per module — `worktree-registry.ts`, `worktree-residue.ts` (29a4de69) |
| CR | `createOne` returns an unexpanded path for tilde roots, so its contract comment lies | accept | helper expands the effective root exactly as the command does — `wt.test.ts` (29a4de69) |
| GT | Timestamp collision defeats the delete identity guard (P1; carries the summary's 4/5 justification) | decline-substantive | unreachable by construction, two independent ways: (1) `addWorktreeEntry` refuses to overwrite an existing entry, so a replacement cannot record at the path until the guarded delete has already resolved — the interleaving the finding requires cannot occur; (2) even granting it, a false match needs the replacement's `createdAt` to equal the createdAt observed at remove-start, which was stamped at the ORIGINAL worktree's creation — i.e. the replacement would have to be minted in the same millisecond as an entry old enough to be getting removed, with a full remove read/verify between them. The guard therefore holds at millisecond resolution; a per-entry nonce would be schema churn for an unreachable arm. |

@greptileai — declining your round-2 P1 per the rationale in the table row above: the record-first refuse-to-overwrite semantics of `addWorktreeEntry` block the interleaving before the timestamp comparison ever matters, and the compared stamp is the original entry's creation instant, not a fresh one, so an equal-timestamp replacement is not constructible. Happy to revisit with a concrete interleaving if you see a path around the overwrite refusal.

local-lane: no verdict artifact recorded for this lineage (`totem review --covariate` sensor line) — the branch's review of record remains the two-round falsification-leg pass in the in-diff spec disposition log.

Round addressed on-thread per the standing cadence (no mechanical re-invoke). Cumulative record: 2 falsification-leg rounds (23 findings dispositioned) + 2 bot rounds (15 findings: 14 accepted-and-fixed, 1 declined-substantive) — round-2 yield was one real symmetry gap plus unreachable-path hardening, the convergence signature.

satur8d:
/gemini review
@coderabbitai review

satur8d:
## Round 3 disposition + record correction

| CR | Both-registries-unreadable remediation names only `registry.json` while the message carries both disclosures | accept | remediation is conditional on `wtWarnings` and names both files; both-unreadable test added — `packages/cli/src/commands/doctor.ts` (5dc73cfc) |
| GT | Round-2 decline's argument 1 misstates `addWorktreeEntry` semantics (it does not refuse an occupied key) | accept (record correction) | **Confirmed against source** — `worktree-registry.ts:267` is an unconditional assignment; the "refusing to overwrite" string is the schema-invalid-file arm of `loadForMutation`, not a key guard. Argument 1 of the round-2 decline is withdrawn. The decline itself stands unchanged on argument 2 alone (the `createdAt` identity comparison — the replacement's stamp would need to collide to the millisecond with the original entry's creation instant, hours-to-days earlier), which the reviewer independently verified as sound and did not re-open. |

On the suggested hardening: the occupied-key overwrite is now documented as by-design at the write site (5dc73cfc) rather than guarded — the only path that reaches it with an existing entry is a re-create over a stale record (`create` refuses when the target directory exists, so an occupied key means the directory is gone and the old entry describes nothing); replacing the stale record is the intended recovery, and entry identity is carried on the delete side by the `expectCreatedAt` guard.

@greptileai — yes to your memory offer, with the nuance above: `addWorktreeEntry` intentionally does not guard duplicate keys (stale-record recovery semantics); please flag any *future* caller that assumes exclusive-write semantics without checking, and treat the delete-side `expectCreatedAt` guard as the identity mechanism. Thank you for the correction — the record matters more than the verdict here, and yours was exact.

local-lane: no verdict artifact recorded for this lineage (`totem review --covariate` sensor line, unchanged).

Cumulative: 2 falsification-leg rounds + 3 bot rounds; round-3 yield was one remediation string and a record correction — converged. Addressed on-thread per the standing cadence.

satur8d:
@coderabbitai review

Issue #2604, opened 2026-08-08
feat(mcp): size-disclosure session scope — consider TOTEM_SESSION_ID keying or reset semantics for the per-process totals


satur8d:
Follow-up from the #2600 falsification leg: `sessionChars`/`sessionCalls` accumulate for the MCP server PROCESS lifetime — they survive conversation clears and compactions (exactly the moments occupancy drops), and a client running multiple totem MCP servers sees independent partial totals. #2600 discloses the basis in-band (`scope="server-process"`), but a finer scope is implementable: `search-log.ts` already passes through `TOTEM_SESSION_ID`, so the totals could key on it (reset when it changes) when present. Decide: key to session id, add reset semantics, or accept process scope as the honest unconditional measurement.

PR #2603, opened 2026-08-08
fix(cli): distributed-template batch — reflex v10, regen idempotence, review-reply tag-per-bot floor (#2599)

satur8d:
Closes #2599 

Distributed-template batch — four items from strategy-claude's 2026-08-08 dispatches, one `REFLEX_VERSION` bump (9 → 10), sequenced ahead of the cohort floor-bump wave so it carries v10 once.

## What ships

- **Regen idempotence (writer fix):** `upgradeReflexes` owns BOTH seams of the block now. After-seam: the old block's own line terminator no longer re-enters through the user span — one blank line per `totem init` run had been accreting into the after-end-marker span that is contractually user content (#1890 class; byte-verified v7→v9 on strategy); a tail of nothing but ASCII whitespace (totem's own accretion residue) heals away entirely. Before-seam: the prefix normalizes to exactly one terminator, so marker-at-byte-0 and glued-prefix files are also byte-fixpoints from the first regen. Pinned by seven tests: double-regen byte-identity (LF + CRLF), current-shape byte-identity with user content below the marker, same-line-after-marker user text survival, whitespace-tail heal, marker-at-0 + glued-prefix idempotence.
- **Reflex v10 — orient cold-start fallback:** the Start-of-Session bullet now names `totem orient` for seats without the SessionStart hook (crown-round A3; CR finding on strategy#1039, declined there as frozen-upstream and routed here).
- **Reflex v10 — Context Management Guardrail demotion:** the template no longer instructs synthesizing a context-loss warning from `<totem_system_warning>` (the server-side emitter is denominator-blind — fired at 15% occupancy on a 1M-window seat). The consumer now weighs the server's measured `<size-disclosure ... />` envelope against its own window occupancy — the judgment moves to the only party holding the denominator. Genuine server-measured warnings keep their instruction. A literal coupling-lock test pins the tag name against the #2600 emitter; **sequencing: merge the #2600 PR in the same cut** so v10 text never describes a tag no shipped server emits.
- **review-reply one-tag-per-addressed-bot floor:** the consolidated round-disposition step now states the cohort floor (operator ruling 2026-08-08): exactly one @-tag per addressed bot, tags present at POST time (never edited in — GCA's listener fires on comment-created only), one notification per bot per round (a bot already tagged in the round's GCA batch comment is not re-tagged), and no tag ever combined with ANY bot's review trigger (triggers are standalone, one trigger no prose). Both in-repo byte-copies updated; the #1890 source-of-truth invariant test enforces the sync.

[...]

Issue #1294, opened 2026-04-08
epic: Cross-Repo Context Mesh for MCP (Proposal 215)


satur8d:
## Summary

Implement **Proposal 215 — Cross-Repo Context Mesh (Linked Totems)** end-to-end through the MCP server so `search_knowledge` queries federate across multiple Totem-managed repositories. Closes the gap between the existing `linkedIndexes` CLI feature (shipped in v1.1, used only by `totem spec`) and the MCP server (which has **zero** awareness of linked indexes today — every agent call hits only the local store).

This epic implements **Proposal 215** (`.strategy/proposals/active/215-cross-repo-context-mesh.md`). The proposal lives in the strategy repo (the "Why"); this epic tracks the "How" in the parent repo.

## Motivation

As the Totem ecosystem has expanded into three repositories (`totem-strategy`, `totem`, `totem-playground`), agents experience **Context Silos**. A playground agent debugging a compiler regression has no visibility into the strategy repo's ADRs explaining the architectural philosophy behind the rule. An agent in the parent `totem` repo querying `mcp__totem-dev__search_knowledge` cannot reach the strategy repo's proposals — even though the developer runs both repos side-by-side on their machine.

The foundation is already in place:
- `linkedIndexes: string[]` config field exists (`packages/core/src/config-schema.ts:227`) and has shipped since v1.1
- `totem spec` actively uses it via `retrieveContext(query, store, linkedStores)` (`packages/cli/src/commands/spec.ts:33-79`) with graceful per-link error handling and a correct semantic-merge pattern (fetch per-store, concat, re-sort by score)
- `.strategy` is a git submodule of `totem`, so the homogeneous-vector-space prerequisite is already satisfied (all three repos use `gemini-embedding-2-preview` as of 2026-04-08, rebuilt from scratch)
- Proposal 217's prompt caching (merged as mmnto/totem#1292 on 2026-04-08) makes broad cross-repo context economically viable

The missing piece is **the MCP server integration**. An archaeological sweep on 2026-04-08 confirmed:
- **Zero** matches for `linkedIndexes`, `linkedStores`, or `linkedStore` across `packages/mcp/src/`
- `ServerContext` holds a singular `store: LanceStore` with no linked-stores field
- `search_knowledge` tool resolves its `boundary` parameter against `config.partitions` (local path prefixes) and falls back to raw prefix match — never reaches a linked index even if one were configured
[...]

PR #2605, opened 2026-08-08
fix(mcp): demote the denominator-blind context-pressure warning to a measured size disclosure (#2600)

### LINKED ISSUE #2600: fix(mcp): demote the denominator-blind context-pressure warning to a measured size disclosure
The `search_knowledge` context-pressure warning is **denominator-blind**: the emitter (search-knowledge.ts, gated on `contextWarningThreshold`) claims "you may be at risk of forgetting earlier instructions" knowing only its own payload size — never the consumer's model, window size, or occupancy. Live falsification 2026-08-08 (strategy seat, operator-routed): a ~31k-token response fired it at **146k of 1M — 15% occupancy**. Cry-wolf on large-window seats: burns operator attention and trains consumers to skip the class, which is what makes the one real firing miss later.

An occupancy-aware warning is unimplementable at that seam (the harness doesn't disclose the denominator), and a config-declared window would be a volatile-state mirror (Tenet 20). So: **disclosure, not diagnosis** (Tenets 13 + 19):

- Replace the warning with a structured **size disclosure** of what the server actually owns and can measure: this response's payload size + cumulative payload/call count this session (server-process lifetime). Machine-parseable, house envelope style (`index-meta`/`retrieval-envelope` precedent), no risk claim: `<size-disclosure chars="…" approxTokens="…" sessionChars="…" sessionCalls="…" />` class.
- `contextWarningThreshold` stays as the emission gate (existing configs keep working; JSDoc updated: it gates a disclosure, not a warning). Session counters accumulate across all calls regardless of gate.
- `<totem_system_warning>` itself is NOT retired — the other emitters (stale index, store failures, embedder fallback) are measurements the server legitimately owns. Only the context-pressure claim is demoted.

Prior art: the Claude Code hook harness handles oversized SessionStart output exactly this way — persist, disclose the byte count, let the consumer decide. Template half (the generated CLAUDE.md guardrail text moves the weighing judgment to the consumer) rides the reflex-v10 batch issue.

Routed from strategy-claude's 2026-08-08T2056Z dispatch (operator word: "dispatch totem-claude about it").

satur8d:
Closes #2600 

Demotes the `search_knowledge` context-pressure warning from diagnosis to disclosure. The removed claim ("you may be at risk of forgetting earlier instructions… suggest `totem handoff`") was denominator-blind by construction — the MCP seam never sees the consumer's model, window size, or occupancy — and was live-falsified 2026-08-08 at 15% occupancy on a 1M-window seat (operator-routed from the strategy lane). Cry-wolf warnings train consumers to skip the class, which is what makes the one real firing miss.

## What ships

- **`withSizeDisclosure`** applied ONCE, at the handler's finalize seam — after the first-call health/linked-stores warning prepend, on the final delivered text, excluding error responses (bounded error text, never corpus content). Every non-error path (ok / mixed / no_useful_hits / empty) counts toward per-process totals; above `contextWarningThreshold` the response carries a self-closing measurement envelope in the house style of `<index-meta>` / `<retrieval-envelope>`:

  `<size-disclosure chars="…" approxTokens="…" sessionChars="…" sessionCalls="…" scope="server-process" />`

  A measurement with no risk claim; the weighing judgment lives with the only party holding the denominator — the consumer, per the reflex-v10 guardrail text (companion PR #2603, which carries a literal coupling-lock test on this tag name; **merge both in the same cut**). The totals' basis is declared in-band: `scope="server-process"` — they survive conversation clears/compactions and are per-server. Finer scoping (`TOTEM_SESSION_ID` keying) is #2604.
- **`contextWarningThreshold` keeps its name and default** (40k chars; ~4 chars ≈ 1 token, the approximation the config JSDoc already stated) — existing configs keep working; core JSDoc updated to say it gates a disclosure, not a warning.
- **`totem_system_warning` is not retired:** stale-index, store-failure, embedder-fallback, and health emissions are conditions the server actually measures; only the context-pressure claim is demoted. `formatSystemWarning`'s JSDoc now states the boundary.
- **Retrieval-corpus blast radius handled:** the 2026-03-07 lesson prescribing the removed behavior ("append a system warning block…") is reframed to the disclosure shape in all three tracked surfaces (`.totem/lessons.md`, `.github/copilot-instructions.md`, `.junie/skills/totem-rules/rules.md`) — coupled to this merge so `search_knowledge` never advises restoring what this PR removes. (Not a compile input; the freeze is untouched.)

## Review record — falsification leg (review of record)

Round 1 (on `4cfd5bc5`), absorbed in `b11221cb`:

- **REFUTED — the ceil pin was vacuous** (fixture payload was ≡ 0 mod 4, so ceil/floor/round/trunc were indistinguishable): fixed — fixture length forced non-divisible with an in-test `payloadLen % 4 ≠ 0` guard, plus a direct unit test on a 41/40 boundary where ceil=11 vs floor=10.
- **REFUTED — accumulator basis untested** (an over-count added after envelope emission survived every existing sequence): fixed — small→large→small integration case plus a unit test proving the envelope never counts itself in attribute or accumulator.
- **MEDIUM — measurement ordering:** the first-call warnings were prepended AFTER measurement, understating `chars` on exactly the call a first-time consumer reads: fixed — disclosure moved to the single finalize seam post-prepend (which also collapses the four separately-unpinned call sites into one).
- **RISK — session semantics** (totals survive `/clear`/compaction; multi-server clients see partial totals; "only measurable scope" was overclaimed given the `TOTEM_SESSION_ID` passthrough): absorbed — scope declared in-band (`scope="server-process"`), JSDoc states the honest boundary, finer keying filed as #2604 for an explicit ruling.
- **RISK — vacuous assertion** (`not.toContain('totem_system_warning')` can never fire under the suite's mock): re-pointed at the mock's actual render.
- **RISK — threshold boundary unpinned:** unit test added at exactly-threshold vs threshold+1.
- **RISK — cross-agent blast radius** (live lesson + two distributed mirrors prescribing the removed warning): reframed in this PR, see above.
- **Leg-verified HOLDS:** coverage enumeration (the four counted sites are exactly the non-error sites; no uncounted path can carry a large payload), no lost warnings (8 remaining `formatSystemWarning` sites untouched; removed message had zero other consumers), #1294 comment discipline intact, `tsc --noEmit` clean, changeset scope correct (the export is internal-only).
- **Declined:** the guardrail's "Large Totem tool responses **may** carry…" scope (leg noted only `search_knowledge` emits today) — kept generic-with-hedge deliberately: the envelope is designed to spread to other tools, and "may" is accurate; revisit if a second tool ships without it.

## Verification

MCP suite 216 passed (12 files, +6 new); core suite 3647 passed; `pnpm -r build` 6/6; prettier clean; ESLint clean (exit 0 verified explicitly); `totem lint` PASS 0 errors.


satur8d:
/gemini review
@coderabbitai review

satur8d:
## Round 1 disposition

@gemini-code-assist @coderabbitai

| GCA | `result.content[0]` may be non-text (MCP SDK `ImageContent`/`EmbeddedResource`) — `text` access could TypeError | decline-substantive | The handler's `ToolResult` is locally typed `{ type: 'text'; text: string }[]` (`search-knowledge.ts:15`) and every constructor in the file builds text-only content — the SDK union never enters this seam, so the guard would be dead code on an unreachable shape |
| CR | Duplicate block-scoped handler declarations at test lines 349/381/403/414/431 — TS redeclaration error | decline-hallucination | The doubled lines do not exist at HEAD: exactly one `const res =` declaration in the file (grep count 1), and `tsc --noEmit` + the 216-test suite are green — the finding reads the absorb-commit diff as if both sides were live |
| CR | Empty catch + `totem-context` suppression without a ticket ref at the threshold fallback | accept | Fixed in d85d25e9 — `catch (err) { void err; }` per the repo convention, ticket ref (#2600) added to the directive. The above-try `totem-context` line is retained deliberately: the Tenet-4 AST rule (`totem lint`) keys on that position for no-rethrow catches |

Full gates re-green on d85d25e9: MCP suite 216, `totem lint` 0 errors, prettier/ESLint clean.

local-lane: no verdict artifact recorded for this lineage (`totem review --covariate` — advisory lanes were not run on this branch; the review of record is the falsification-leg round in the PR body).

Issue #179, opened 2026-03-07
Epic: Markdown/Document-Only Mode


satur8d:
**Description:**
To support non-developers and power users who want a persistent memory bank for their ideas, Totem needs to gracefully handle non-code repositories.

**Requirements:**
1. Develop a `MarkdownChunker` and `PDFChunker` that specialize in semantic text splitting (e.g., chunking by headers `##` rather than syntax trees).
2. Allow `totem.config.ts` to be configured for a "Notes" or "Wiki" mode, ignoring strict code compilation rules.
3. Ensure the MCP server still seamlessly searches this text-based knowledge base.

**Acceptance Criteria:**
- Totem successfully indexes a folder full of markdown notes.
- Queries retrieve relevant paragraphs and headers perfectly, acting as a "second brain."

satur8d:
This RFC/Epic has been moved to a formal proposal in \.strategy/proposals/179-markdown-document-only-mode.md\ to keep the issue tracker focused on actionable engineering work. Closing.

PR #2606, opened 2026-08-08
fix(cli): init hardening — MCP dedup by registered package, vendor-hook guard, full non-interactive mode (#2601)

### LINKED ISSUE #2601: fix(init): hardening — MCP registration dedup by command, vendor-hook managed-marker guard, full non-interactive mode
Three `totem init` misses, all observed first-hand during the strategy repo's 1.113.1 floor refresh (strategy-claude's 2026-08-08T0912Z dispatch + strategy lesson-164d69eb), all likely to replay on every cohort repo when the floor-bump wave runs init:

1. **MCP registration dedup keys on entry NAME, not registered command.** Init appended a third `@mmnto/mcp` server entry (`"totem": cmd /c npx -y @mmnto/mcp`) to a `.mcp.json` that already carried two `@mmnto/mcp` registrations under other names (`totem-strategy` local-pinned, `totem-dev` with `--cwd`). Detection should key on the registered command/package so an existing registration under any name suppresses the append.

2. **Template-over-custom vendor hook drop (preprompt-refresh-trap family).** Init dropped the generic `.gemini/hooks/SessionStart.cjs` beside the repo's tracked CUSTOM `SessionStart.js` (their strategy#482 parity hook) — double-fire or orientation regression depending on Gemini's hook discovery. A managed-marker check before dropping a vendor hook (the skill-refresh pattern: only touch files totem owns by marker) closes it.

3. **Non-interactive detection is partial.** In a stdin-null context init printed "Non-interactive mode — auto-upgrading reflexes" yet still raised both interactive prompts (tool selection, hook install), dying at whichever prompt stdin couldn't answer (node "unsettled top-level await"); piped answers reached prompt 1 but never prompt 2. Mutations land BEFORE the prompts (the known trap shape), so the half-run reads as a no-op. Full non-interactive detection (no prompt may be raised after the non-interactive banner prints) and/or explicit `--yes`/`--tools` flags close it; ideally no mutation lands before the last prompt resolves.

Sequencing: land before the cohort bump nudge so the wave doesn't replay these per-repo. The strategy lesson is the consumer-side guard until fixed.

satur8d:
Closes #2601 

Init hardening — the three misses strategy's 1.113.1 floor refresh surfaced, fixed before the cohort bump wave replays them per-repo.

## What ships

- **MCP dedup by registered package, not entry name.** `scaffoldMcpConfig` now refuses to append a duplicate when any existing entry already invokes `@mmnto/mcp` — by boundary-anchored name (`@mmnto/mcp-experimental` and `@mmnto/mcp.js` are different packages and do NOT suppress), or by **path identity**: a pin like `node ./packages/mcp/dist/index.js` is resolved against both the config's directory and the project root (three of the four host configs live in subdirectories with root-relative pins), and the owning package is read from the nearest `package.json` on the resolved path. Only JS-family script *files* start the identity walk — a directory-shaped arg (`--cwd ./`) never does. The skip is disclosed: ``already registered as `<name>` — no duplicate added``.
- **Vendor-hook managed-marker guard.** A custom (unowned) `.gemini/hooks/<stem>.js` beside the managed `.cjs` target: fresh path → the managed twin is **withheld** with a disclosure (never introduce a double-fire); re-init path with totem's own `.cjs` on disk → normal drift-repair plus an honest coexistence disclosure; a **user-owned** `.cjs` → the truthful "neither is totem's — totem hook not installed" line (never a claim that a managed file exists). Marker-headed legacy `.js` files still take the existing migration; an unreadable legacy candidate (e.g. a directory squatting the name) is skipped with a disclosure instead of crashing init mid-run.
- **Full non-interactive detection + `--yes`.** One predicate guards every prompt site init reaches (init.ts and both hook installers); stdin-null runs can no longer die at an unguarded prompt with mutations already landed (the "unsettled top-level await" half-run). Non-interactive defaults match each prompt's Enter default, each logged; the two deliberate divergences are disclosed in-line (cursor-rules compile **declines** headless, naming `totem compile --from-cursor`; post-merge hook declines, naming the verb). `-y/--yes` ships in both the full and lite binaries. **Behavior change, documented in the changeset:** piped stdin is no longer read as prompt answers — a non-TTY run takes every default; scripts that piped `none` must pass explicit flags.

## Review record (review of record for this PR)

Four rounds, all recorded:

1. **Build leg** implemented the ruled spec (`22e15cb`), six declared deviations, all accepted on owner review — including two calls beyond its brief that were correct (summary-filter extension so a canonical-file disclosure can't be swallowed; changeset bullets rewritten to stay true).
2. **Owner re-gate** caught a `totem lint` failure the leg's gate list omitted (Tenet-4 annotation position) — fixed in `b73daaab`.
3. **Falsification leg round 1** REFUTED the dedup against this repo's own `.mcp.json` (path pins that never spell the package name) and the guard's re-init disclosure; flagged test-unreachable non-interactive arms (the mutant reproduces the original incident), substring false-positives, and the piped-stdin semantics change. Absorbed in `165c5a5d` (path-resolved identity walk, coexistence-aware guard, git-repo e2e reachability tests, lite `--yes`, changeset docs).
4. **Scoped falsification round 2** on that fold REFUTED it twice more: the path probe resolved against the config's directory only — correct for root-level `.mcp.json`, wrong for `.gemini/`/`.cursor/`/`.junie/mcp/` (live exhibits in this repo); and the coexistence line claimed a "managed .cjs" even when the `.cjs` was user-owned. Both fixed in `d147a9e5` (dual-base resolution + disclosure gated on the scaffold outcome), along with the leg's bounded risks: `isFile` + suffix gate on the identity walk (kills the directory-arg escape), `.`-aware boundary regex, artifact assertions on the git-hook e2e, an `exists`-action e2e that pins the summary plumbing, and the false docstring clause cut.

Round 3 implements round 2's own prescriptions with each fix pinned by a test, so no further leg was dispatched (the #179 precedent: a fold that implements the leg's pre-verified prescriptions doesn't re-arm). **Stated residuals:** the identity walk is lexical (an aliased/symlinked script attributes to its host package — pre-existing class, docstring states it); the name probe is case-sensitive (npm rejects uppercase names); interactive Ctrl+C at a late prompt still leaves earlier mutations (out of scope here, the issue's "ideally" clause — the non-interactive half-run it named is closed).

## Verification

CLI suite 3845 passed (155 files; +29 tests net across the rounds, including three `initCommand` e2e in real temp repos — one `git init`-backed); `pnpm -r build` 6/6; prettier clean; ESLint clean; `totem lint` 385 rules PASS, 0 errors. Live smoke (build leg, stdin-null against a fixture with an existing differently-named registration + a custom Gemini hook): exit 0, no prompt raised, no duplicate, custom hook intact.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 disposition

@coderabbitai @greptileai

| CR | `unreadable-legacy` falls into `printGeminiHookMigrationSummary`'s user-owned fallback — a false ownership claim on the path most consumers see (`prepare`) | accept | Fixed in 027912f7 — the printer gains an accurate unreadable arm; "user-owned" never renders for a state no read established |
| CR | The successor read in `migrateLegacyGeminiHooks` is unguarded — a directory at the `.cjs` path EISDIRs mid-migration, the exact class this PR removes from the legacy read | accept | Fixed in 027912f7 — guarded read, new `unreadable-successor` reason rendered accurately in BOTH consumers, directory-successor fixture added |
| CR | The git-hook e2e reads `.git/hooks` under ambient `core.hooksPath`/`init.templateDir` | accept | Fixed in 027912f7, adapted: env-level isolation (`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` → an empty temp file) for the whole run — the suggested `-c` on `git init` alone wouldn't cover the installer's own git config reads, and `os.devNull` fails on Windows (git exit 128) |
| CR | Changeset's path-probe description names a single resolution base — stale vs the shipped dual-base probe | accept | Fixed in 027912f7 — names both bases and the entrypoint scoping |
| CR | `patch` understates a new flag + a scripted-caller stdin behavior change | accept | Fixed in 027912f7 — bump raised to `minor` |
| Greptile | Path probe scans non-entrypoint arguments — an unrelated server passing a totem-owned `.js` as a plugin/config arg would suppress the real registration | accept | Fixed in 027912f7 — the probe is scoped to Node-family commands' FIRST script-suffixed arg (the runtime's own entrypoint semantics); package-manager launchers stay the name probe's job; non-Node plugin-arg fixture added |

GCA returned a clean pass with no findings this round (not tagged — nothing to disposition, one notification per bot per round).

Full gates re-green on 027912f7: CLI suite 3847 (155 files), `totem lint` 385 rules 0 errors, prettier/ESLint/build clean.

local-lane: no verdict artifact recorded for this lineage (`totem review --covariate` — advisory lanes were not run on this branch; the review of record is the four-round leg history in the PR body plus this bot round).

Issue #2599, opened 2026-08-08
fix(init): distributed-template batch — reflex v10 (orient cold-start + guardrail demotion) + regen blank-line idempotence + review-reply tag-per-bot floor


satur8d:
Batch over the distributed templates (one PR, one `REFLEX_VERSION` bump), from strategy-claude's 2026-08-08 dispatches (falsification-leg findings on their doctrine batch + the reflex-v10 candidate routed off mmnto-ai/totem-strategy#1039):

1. **Reflex regen accretes one blank line per `totem init` run.** `upgradeReflexes` (init.ts) rebuilds the file as `before + AI_PROMPT_BLOCK + after`, but `AI_PROMPT_BLOCK` ends with `end -->\n` AND `after = content.slice(endIdx + REFLEX_END.length)` re-captures the old block's own line terminator — so each regen duplicates one `\n`. Byte-verified on strategy (v7 tail `-->\n\n`, v9 tail `-->\n\n\n`), unbounded, and it lands in the after-end-marker span that is contractually user content (#1890). Fix: exactly one party owns the marker's line terminator (the invariant `scaffoldClaudeSkill` already holds); add a byte-idempotence test (regen twice → identical bytes; user content after the marker preserved verbatim).

2. **Reflex v10 — Start-of-Session cold-start fallback.** The bullet names `totem describe`/`status`/`triage` but never `totem orient` for a seat without the SessionStart hook. Add the crown-round A3 sentence (CR finding on strategy#1039, decline-other'd as frozen-upstream and routed here).

3. **Reflex v10 — Context Management Guardrail demotion.** The current text instructs synthesizing a context-loss warning from `<totem_system_warning>`; the server-side emitter is denominator-blind (fired at 15% occupancy on a 1M-window seat, operator-falsified same session). Template half of the fix: weigh the *disclosed size* (see the companion MCP issue) against your own window occupancy; escalate only if material. The judgment moves to the only party holding the denominator.

4. **review-reply skill can't emit the one-tag-per-addressed-bot floor.** Bot-protocols now requires the consolidated round-disposition comment to @-tag every addressed bot exactly once (operator ruling 2026-08-08 after a Greptile convergence miss on liquid-city#968; folded cohort-wide in strategy#1039), but the distributed SKILL.md's only tagging instruction is GCA-specific. The managed block can't be fixed per-repo — the floor is discipline-only until the template states it.

[...]

PR #2607, opened 2026-08-08
Version Packages

satur8d:

# Releases
### Minor Changes

- 772f058: `totem init` hardening (mmnto-ai/totem#2601) — three misses observed on a live floor-refresh run:
  - **MCP registration dedup keys on the registered package, not the entry name.** `scaffoldMcpConfig` previously only checked for a `totem` key, so a `.mcp.json` already carrying `@mmnto/mcp` under other names (`totem-dev`, `totem-strategy`) collected a duplicate third entry. Every existing server entry's `command` and `args` are now probed two ways: for the package name (boundary-anchored, so `@mmnto/mcp-experimental` is correctly a different package), and — because a registration need not spell the name at all — by resolving a Node-family command's entrypoint script against both the config's own directory and the project root (host configs in `.gemini/`, `.cursor/`, and `.junie/mcp/` carry root-relative pins) and reading the owning `package.json` off disk, which catches the `node ./packages/mcp/dist/index.js` form. A hit skips the append without rewriting the file and the init summary discloses the name it is already registered under.
  - **Vendor-hook managed-marker guard.** Init no longer drops the generic `.gemini/hooks/SessionStart.cjs` / `BeforeTool.cjs` beside a consumer's CUSTOM same-stem `.js` hook (the double-fire / orientation-regression trap). On a fresh install an unowned `<stem>.js` sibling — one the totem marker does not open — withholds the managed `.cjs` and the skip is disclosed in the summary. When the managed `.cjs` already EXISTS beside a custom `.js` the twin is already live, so init drift-repairs it as usual and the summary discloses the coexistence plus the two ways out, rather than falsely reporting the hook as not installed. The marker-headed legacy `.js` → `.cjs` migration path is unaffected; a legacy path that cannot be read at all (a directory sharing the name) is skipped and disclosed instead of crashing init mid-run.
  - **Full non-interactive detection plus `--yes`.** Only two of six prompt sites were TTY-guarded, so a stdin-null init printed its non-interactive banner and still died at an unguarded prompt with mutations already landed. One `interactive` predicate (`stdin.isTTY && !--yes`) now guards every `rl.question` site in init and the two it reaches in the hook installers. Non-interactive defaults are the Enter defaults — install baseline lessons, upgrade reflexes, configure all detected tools, Lite embedding tier, install enforcement hooks, decline the post-merge hook — except the cursor-rules compile, which declines and names `totem compile --from-cursor` rather than firing an LLM-driven rewrite unattended. Every default taken is logged. `-y, --yes` is registered on both the full CLI and the standalone lite binary.

    **Behavior change for scripted callers:** piped stdin is no longer read as prompt answers. A non-TTY run now takes every default without consuming stdin — previously a piped answer could reach the first prompt (and the run then died at the second, unguarded one). Scripts that piped `none` or an empty line into `totem init` must pass the equivalent flags explicitly instead; the defaults a bare non-interactive run takes are listed above.

### Patch Changes

- d11e4ba: Distributed-template batch (#2599): reflex v10 adds the `totem orient` cold-start fallback to the Start-of-Session bullet and demotes the Context Management Guardrail from synthesize-a-warning to weighing the server's measured `<size-disclosure>` against the consumer's own window; `upgradeReflexes` no longer accretes one blank line per regen at the end-marker seam (byte-idempotent, heals pure-whitespace tails); the review-reply skill's consolidated round-disposition step now states the one-tag-per-addressed-bot floor.
- Updated dependencies [3624a2d]
  - @mmnto/totem@1.114.0
### Patch Changes

- 3624a2d: Demote the denominator-blind context-pressure warning to a measured size disclosure (#2600): `search_knowledge` no longer claims "you may be at risk of forgetting earlier instructions" (live-falsified at 15% occupancy on a 1M-window seat) — above `contextWarningThreshold` it now appends a self-closing `<size-disclosure chars approxTokens sessionChars sessionCalls scope="server-process" />` measurement envelope, applied once at the finalize seam AFTER the first-call warning prepend so `chars` measures the actually-delivered payload, and leaves the weighing judgment with the consumer, the only party holding the window denominator. `totem_system_warning` remains for conditions the server actually measures (index staleness, degraded retrieval, store failures).
### Patch Changes

- 3624a2d: Demote the denominator-blind context-pressure warning to a measured size disclosure (#2600): `search_knowledge` no longer claims "you may be at risk of forgetting earlier instructions" (live-falsified at 15% occupancy on a 1M-window seat) — above `contextWarningThreshold` it now appends a self-closing `<size-disclosure chars approxTokens sessionChars sessionCalls scope="server-process" />` measurement envelope, applied once at the finalize seam AFTER the first-call warning prepend so `chars` measures the actually-delivered payload, and leaves the weighing judgment with the consumer, the only party holding the window denominator. `totem_system_warning` remains for conditions the server actually measures (index staleness, degraded retrieval, store failures).
- Updated dependencies [3624a2d]
  - @mmnto/totem@1.114.0
## @mmnto/pack-rust-architecture@1.114.0

## @mmnto/pack-agent-security@1.114.0


satur8d:
## Cut note — 1.114.0 (contract-bearing, minor)

Three merges ride this train: `d11e4baf` (#2603), `3624a2d1` (#2605), `772f0587` (#2606). Each carried a falsification-leg record and a settled, acknowledged bot round.

**Contract-bearing surfaces this cut moves:**

- **Reflex template contract — `REFLEX_VERSION` 9 → 10.** Consumers' next `totem init` offers the v10 block: the Start-of-Session bullet gains the `totem orient` cold-start fallback (crown-round A3, cohort-wide via the normal template path), and the Context Management Guardrail is demoted from synthesize-a-warning to weighing the server's measured disclosure — the consumer holds the only denominator. The distributed `review-reply` SKILL.md gains the one-tag-per-addressed-bot floor (refreshes in place on next init).
- **New machine-readable consumer surface:** large `search_knowledge` responses now carry `<size-disclosure chars approxTokens sessionChars sessionCalls scope="server-process" />` (gated on `contextWarningThreshold`, name and default unchanged). The context-pressure `totem_system_warning` is **gone** — anything keyed on that message must not expect it; the in-repo retrieval-corpus lesson prescribing it was reframed in the same merge. `totem_system_warning` itself survives for server-measured conditions.
- **CLI surface:** `-y, --yes` on `init` (full + lite binaries). **The minor-driver behavior change:** piped stdin is no longer read as prompt answers — a non-TTY `totem init` takes every default; scripts that piped answers must pass flags.
- **Exported library surface:** `scaffoldMcpConfig` gains `opts.projectRoot`; `GeminiHookMigrationResult.reason` union widens (`unreadable-legacy`, `unreadable-successor`); `installGeminiHooks` and `findUnownedHookSibling` newly exported.
- **Regen byte-contract:** `upgradeReflexes` is now byte-idempotent at both seams and heals pure-whitespace tails — consumers' CLAUDE.md/GEMINI.md stop accreting one blank line per init, and already-accreted tails repair on the v10 upgrade (disclosed policy call, recorded on #2603).
- **`worktrees.json`: unchanged** this cut — the totem-status second-reader (totem-status#177) is unaffected.

**Choreography on merge:** Release workflow → OIDC publish → provenance verify → tags (the two pack 404s are `private: true` by design). Local half after green: ff main, global CLI → 1.114.0 (verify with `TOTEM_NO_REEXEC=1`). **This cut unblocks the cohort bump nudge** — one wave carries v9→v10 plus the init hardening that prevents the wave itself replaying strategy's three init traps.

Issue #2615, opened 2026-08-09
fix(hooks): BeforeTool shield-gate leg under command-hook execution — stdio 'inherit' floods the decision channel; lint timeout equals Gemini's default hook timeout (fail-open)


satur8d:
Found by the pre-merge falsification leg on the mmnto-ai/totem#2611 fix branch (2026-08-09), reproduced against the installed `@google/gemini-cli@0.54.4` bundle with a shimmed `totem` on PATH. Deferred out of that PR by ruled scope: the path is unreachable until registration arms (mmnto-ai/totem#2478), and both defects sit in the shield-gate leg, not the entry point that PR ships.

### Defect 1 — `stdio: 'inherit'` floods the decision channel

The shield-gate leg of the `GEMINI_BEFORE_TOOL` template runs `execSync('totem lint', { ..., stdio: 'inherit' })`. Under a command hook, `'inherit'` points at Gemini's stdout pipe, so lint output lands in the decision channel ahead of the structured deny JSON:

```
status= 2
STDOUT= "TOTEM-LINT-STDOUT-LINE-1\r\n...\r\n{\"decision\":\"deny\",\"reason\":\"[Totem Error] Shield check failed...\"}\n"
GEMINI-SIM: JSON_PARSE_FAILED -> convertPlainTextToHookOutput(text, exitCode=2)
```

The deny still holds — but only via the exit-2 fallback, and the reason handed to the model becomes the whole lint dump plus the raw JSON blob. Fix shape: capture lint output (`stdio: 'pipe'` / return-value capture) and fold it into the deny reason instead of streaming it into stdout.

### Defect 2 — lint timeout equals Gemini's default hook timeout (fail-open on slow lint)

The template's `timeout: 60000` for `totem lint` equals Gemini's `DEFAULT_HOOK_TIMEOUT = 6e4` (`chunk-YMKYUNCI.js:361116`). On a hook timeout the result carries `error` and no `output` (`:361364-361374`), `aggregateResults` produces `finalOutput === undefined` (`:361468-361473`), and `getBlockingError()` never blocks — a slow lint run silently allows the push it was supposed to gate. Fix shape: set the inner lint timeout comfortably below the hook budget (and/or emit an explicit deny on lint timeout), so the guard, not the vendor, decides the verdict.

### Test gap

[...]

PR #2610, opened 2026-08-09
feat(gemini): team-level .gemini/settings.json — arm SessionStart at trust (totem-strategy#619 ruling, SessionStart-only fold)

satur8d:
Executes the [mmnto-ai/totem-strategy#619 ruling](https://github.com/mmnto-ai/totem-strategy/issues/619) (2026-07-22, operator): un-ignore `.gemini/settings.json` on public totem and commit a minimal hooks-only team file. Authored by strategy-claude (cross-lane execution operator-sanctioned 2026-08-08; implementation was assigned totem-lane in the ruling — flagged for the resident seat's review).

**Scope was folded after a falsification leg (BLOCK-WITH-FIXES, receipts below): this PR arms SessionStart only.** The ruling's other half — arming the BeforeTool write-guard — is not executable today: the leg reproduced (live `@google/gemini-cli@0.54.4`, matched controls) that the managed `GEMINI_BEFORE_TOOL` template is a `module.exports` module with no entry point, so registered as a command hook it exits 0 and enforces nothing. Registering it would ship a false mechanism claim (Tenet 19). The write-guard therefore **remains adherence-class**, stated plainly, until mmnto-ai/totem#2611 lands a runnable template + end-to-end deny test. The first commit on this branch carried that inert registration; the fold commit removed it.

## What ships

1. **`.gitignore`** — `.gemini/settings.json` un-ignored; hooks/skills negations and the no-secrets comment retained, comment cites the ruling and the user-level home for per-developer content.
2. **`.gemini/settings.json`** (new, committed) — one registration: `SessionStart` → `node .gemini/hooks/SessionStart.cjs` (the committed managed briefing hook), nested `{matcher: "", hooks: [...]}` shape, in exact **registration-shape** parity with the committed `.claude/settings.json` registrations (`matcher: ""`, relative `node` command). No `mcpServers`, no `env`, no absolute paths, no personal prefs (ruling constraint 1).

[...]

PR #2616, opened 2026-08-09
fix(hooks): Gemini BeforeTool command entry point — registered as a hook it exited 0 and enforced nothing (#2611)

### LINKED ISSUE #2611: fix(hooks): GEMINI_BEFORE_TOOL template is module.exports-only — registered as a command hook it exits 0 and enforces nothing (fail-open, sensor-trust class)
## The GEMINI_BEFORE_TOOL template cannot fire as a command hook — registered, it exits 0 and enforces nothing (fail-open, sensor-trust class)

Found by a falsification leg on mmnto-ai/totem#2610 (strategy-claude seat, 2026-08-08); reproduced against live `@google/gemini-cli@0.54.4` with matched controls. Findings below are the leg's, verified receipts in the PR thread.

### The defect

The managed `GEMINI_BEFORE_TOOL` template (`packages/cli/src/commands/init-templates.ts`, shipped `.gemini/hooks/BeforeTool.cjs`) is a `module.exports = function beforeTool(toolName, toolInput) {...}` module with **no entry point**: nothing reads Gemini's hook-input JSON from stdin, nothing calls `beforeTool`, nothing writes a decision. Registered as a command hook (`node .gemini/hooks/BeforeTool.cjs`), the process defines a function and exits 0.

Reproduced with a rule-violating payload (bare `#123` in a `.md` write) piped in Gemini's hook-input shape:

```
$ node .gemini/hooks/BeforeTool.cjs < hookinput.json
EXIT=0   STDOUT=[]   STDERR=[]
```

Gemini's runner treats empty-output + exit 0 as allow. And Gemini CLI has **no filename-convention discovery** (zero hits for `.gemini/hooks` path-probing in the bundle) — registration is the only execution path, so there is **no configuration under which the distributed write-guard fires as a hook**. The guard has been adherence-theater since it shipped: `install-hooks.ts`'s own contract comment says the CLI "never CREATES a registration (arming the hook is the deferred adoption slice — mmnto-ai/totem#2478 OPTION 1)", and now the arming slice reveals the artifact it would arm is inert.

Corroborating datum from the resident seat: the local `.gemini/settings.json` `totem-briefing` SessionStart entry is the **flat** `{name, type, command}` shape, which Gemini's `processHookDefinition` discards (`!Array.isArray(definition.hooks)` → skip) — so the resident briefing hook never ran either. Related: the "flat command shape too" fixture in `install-hooks-exit-contract.test.ts` (~line 722) documents a shape Gemini discards; fine as a migration-input fixture, misleading as a registration-shape reference.

### The exit-code contract a fix must honor

From the Gemini hook runner (leg-cited, `chunk-YMKYUNCI.js:361434-361451` in the installed bundle):

- exit 0 → allow
- exit **1 → allow with warning** (an uncaught Node throw exits 1 — a naive `beforeTool()` call that throws on violation still **fails open**)
- exit ≥ 2 → deny
- or structured stdout `{"decision": "deny", "reason": ...}`

So the fix is: an entry-point wrapper that reads the hook JSON on stdin, calls `beforeTool(input.tool_name, input.tool_input)`, and emits `{"decision":"deny","reason":...}` (or exits ≥ 2) on violation — never relying on throw/exit-1.

### The test gap that let this survive

`install-hooks-exit-contract.test.ts` scopes itself to the **installer's** exit codes and drift-repair; no test anywhere asserts that a registered Gemini guard, run as a command, blocks anything. One end-to-end deny-case test should accompany the fix — without it the fix is unfalsifiable.

### Also useful for the fix (leg-verified Gemini semantics)

- SessionStart matchers are **exact-string** against the trigger (`startup`/`resume`/`clear`); only `""` or `"*"` pass-through. `".*"` never fires. Tool-event matchers (BeforeTool) ARE real regexes against tool names.
- `$GEMINI_PROJECT_DIR` expands to the same value as the spawn cwd, so `node .gemini/hooks/X.cjs` (relative) and the `$GEMINI_PROJECT_DIR` form resolve identically.
- Project-level hooks gate on folder trust, re-checked per hook; first arming shows a one-time per-workspace WARNING banner and records in `~/.gemini/trusted_hooks.json`.

### Severity framing

Sensor-trust class: a shipped guard that reads as installed but cannot fire is the lying-sensor shape (Prop 309 / hardening-program routing — mmnto-ai/totem-strategy#974). mmnto-ai/totem#2610 was folded to SessionStart-only so it ships no false mechanism claim; the write-guard stays adherence-class until this is fixed.

satur8d:
## What

The distributed Gemini `BeforeTool.cjs` template (`GEMINI_BEFORE_TOOL`, `packages/cli/src/commands/init-templates.ts`) gains its command entry point. Registered as a command hook it previously defined `module.exports = function beforeTool(...)` and exited 0 — Gemini's *allow* — so all four guards in it (outbox subject-quoting, auto-close keywords, xrepo bare refs, shield-gate lint) were inert: adherence-theater since the artifact shipped (mmnto-ai/totem#2611).

Run as `node BeforeTool.cjs`, the template now:

- reads the hook-input JSON from stdin and calls `beforeTool(input.tool_name, input.tool_input)`;
- on violation emits the structured stdout decision `{"decision":"deny","reason"}` **and** exits 2 — either channel denies under Gemini's contract (leg-verified: `convertPlainTextToHookOutput` denies on any exit ≥ 2 even when stdout fails to parse). A throw exits 1, which Gemini treats as *allow-with-warning*, so throwing can never deny — the wrapper converts the guards' throws into the deny channels. The stderr copy of the reason serves hand-run invocations; Gemini reads stderr only when stdout is empty, which is exactly the exit-1 warning paths;
- treats input it cannot evaluate (unparseable JSON, missing `tool_name`) as *allow-with-warning* (exit 1 + stderr breadcrumb): denying every tool call on a vendor shape change would brick the seat, and exiting 0 would hide the outage;
- still yields the bare guard function under `require()` (entry point gated on `require.main`), so existing unit consumers and fixtures are untouched.

## What this does NOT do

Registration/arming stays the deferred mmnto-ai/totem#2478 OPTION 1 slice — nothing in this change registers the hook anywhere. `migrateGeminiHookRegistration` remains rewrite-if-present only.

## Tests

New `gemini-beforetool-contract.test.ts` runs the **rendered artifact** the way Gemini runs it (`node BeforeTool.cjs` + stdin payload) — the gap mmnto-ai/totem#2611 named: the exit-contract suite locks the installer's exit codes, but nothing anywhere executed the guard itself, which is how a module.exports-only template shipped inert. Cases: structured deny + exit 2 on a violating write (negative control), exit 0 + empty stdout on a clean write (positive control), exit 1 + stderr breadcrumb on unparseable/shapeless input, and the `require()` surface. These are the Tenet 9 positive/negative controls the future arming slice inherits.

Also per mmnto-ai/totem#2611: the "flat command shape too" fixture in `install-hooks-exit-contract.test.ts` now carries a comment marking it as a migration INPUT (a shape Gemini's `processHookDefinition` discards), not a registration-shape reference.

## Riders

Two MCP-banked lesson files from the mmnto-ai/totem#2610 session (pre-push stdin single-consumption conditional rule; Gemini hook registration semantics) were left uncommitted in the working tree — they ride this PR, plus the generated spec for this ticket.

## Gates

Full workspace build, full `@mmnto/cli` suite (156 files / 3859 passed), repo-wide `format:check`, ESLint (0 errors), `totem lint` PASS, changeset (patch `@mmnto/cli`).

**Falsification round (pre-open):** the leg confirmed the runtime contract against the installed `@google/gemini-cli@0.54.4` bundle (10-payload matrix on the rendered artifact, mutation-checked against the pre-fix inert template) and returned 9 findings. Absorbed in this diff: Tenet-9 control labels un-swapped, stderr claim narrowed to what the vendor actually reads, test helper moved to `spawnSync` so the negative control asserts real silence, fixture comment corrected (`matcher` optional; the nested `hooks` array is the mandatory bit), the LLM-drafted spec amended (its vendor-discovery premise is the falsehood this ticket kills), riding-lesson heading repaired. Deferred by ruled scope: the shield-gate-leg findings (lint output flooding the decision channel under `stdio: 'inherit'`; lint timeout equal to Gemini's default hook timeout failing open) — unreachable until arming, filed as mmnto-ai/totem#2615 to settle with the mmnto-ai/totem#2478 slice.

Closes #2611 


satur8d:
/gemini review
@coderabbitai review

satur8d:
@coderabbitai Round 1 disposition (operator-ruled):

**[1] Outside-diff Major — `packages/cli/src/commands/init-templates.ts` 460–475, shield-gate `stdio: 'inherit'` writes lint output into the decision channel:** **Deferred — tracked in mmnto-ai/totem#2615**, coupled to the mmnto-ai/totem#2478 arming slice. Premise verified real and convergent with the pre-open falsification leg (same lines, same mechanism, reproduced against the installed Gemini bundle with a shimmed `totem` on PATH): the deny verdict still holds today via the exit-2 fallback, and the path is unreachable until a registration arms the hook — nothing in this PR, and nothing shipped anywhere yet, registers it. The fix shape recorded on the ticket matches your proposal: capture lint output into the deny reason instead of `stdio: 'inherit'`, plus shell-shield allow/deny e2e cases with a stubbed `totem`, settled with (or before) arming.

No other actionable findings this round: all other file surfaces LGTM'd, and the static-tool annotations (ast-grep child_process/non-literal-path on the vitest tmpdir patterns, LanguageTool/markdownlint on the amended generation-record spec and generated lesson/changeset files) fall under existing repo learnings and exemptions.

Issue #1890, opened 2026-05-11
feat(cli): Phase C slice 3 — distribute signoff + review-reply skills via `totem init` (marker-based)


satur8d:
## Phase C slice 3 — session-utility skill suite distribution (v0.1: signoff + review-reply, Claude-only)

Slice 3 of the Phase C arc (parent `mmnto-ai/totem#1845` already closed by slice 1 in `mmnto-ai/totem#1862`; slice 2 closed by `mmnto-ai/totem#1885`).

This issue captures the design doc for slice 3 before implementation, per /preflight discipline + strategy-Claude cross-stream input (substrate dispatches `2026-05-11T0325Z` + `T0420Z`).

## Problem

The session-utility skills (`signoff`, `review-reply`) currently live ONLY in `mmnto-ai/totem`'s `.claude/skills/` directory. Other family repos that join the ecosystem (`totem-strategy`, `totem-substrate`, `totem-status`, `arhgap11`, `liquid-city`) inherit no canonical version through `totem init`. Maintainers have either retrofitted manually (carrying customized variants) or are running without these skills.

Empirical motivation: `mmnto-ai/totem#1887` rolled out a 5-attempt rebase-retry signoff loop. The rollout required manually editing `SKILL.md` in 3 repos (totem-strategy, liquid-city, totem). With no canonical seeding mechanism, future signoff-skill updates will repeat this manual fan-out.

## Scope (v0.1)

Two skills only:

- **signoff** — end-of-session memory update + journal write + substrate push (with rebase-retry)
- **review-reply** — unified PR review triage via `totem triage-pr`

Deferred from v0.1 (carried per strategy-Claude flag): `preflight`, `prepush`, `postmerge`. These overlap with `totem` subcommands (`totem spec`, `totem review`, `totem lesson extract`) and carry family-specific customization. A future slice should decide "collapse into CLI" vs "ship as thin orchestrator templates" before distributing.

Deferred to slice 4: Gemini parity for distributed skills. Tracked separately.

## Canonical/customization split — policy

Skill content qualifies for distribution (inside totem-managed markers) only if it passes all three tests:

1. **Universal across all consumer repos** — no governance-specific or repo-specific references in the canonical content
2. **Idempotent with `totem init --refresh`** — re-running init replaces canonical content cleanly without destructive assumptions about state
[...]

PR #2618, opened 2026-08-09
fix(eject): marker-bounded reflex scrub + tool-table roster — heading-to-EOF deleted after-end user content (#2602)

### LINKED ISSUE #2602: fix(eject): reflex scrub is heading-to-EOF — deletes after-end-marker user content and leaves orphan markers read as `current`
Found by the falsification leg on the #2599 pre-merge pass (out-of-diff, pre-existing).

`totem eject` scrubs the reflex block with a heading-to-EOF regex (`eject.ts:495-509`):

```ts
const primaryMarker = /\n*## Totem AI Integration \(Auto-Generated\)[\s\S]*$/;
fs.writeFileSync(filePath, content.replace(activeMarker, '\n'), 'utf-8');
```

It never references `REFLEX_START`/`REFLEX_END`, so on a file where the user has content BELOW the end marker — the span the refresh contract reserves for user content (#1890 class, and the span #2599's regen fix preserves byte-exactly) — eject:

1. **Deletes the user's content** (everything from the heading to EOF goes), and
2. **Leaves an orphan `start` + `version` marker with no `end` marker** (the heading sits inside the block, after the two leading marker lines), which `detectReflexStatus` then reads as `current` — a corrupt-but-green state.

Reproducer (the #2599 test fixture, run through the eject regex verbatim):

```
input : '# CLAUDE.md\n' + AI_PROMPT_BLOCK + '\n## My Custom Section\n\nDo not delete this!\n'
eject → '# CLAUDE.md\n\n<!-- totem:reflexes:start -->\n<!-- totem:reflexes:version:10 -->\n'
```

Fix shape: when both markers are present, eject should remove exactly the `REFLEX_START..REFLEX_END` span (plus its own line terminator — the same single-owner seam discipline as `upgradeReflexes` post-#2599) and preserve everything after the end marker; the heading-to-EOF regex stays only as the legacy no-marker fallback. `REFLEX_FILES` also lists `CLAUDE.md` and `.cursorrules` only — check whether `GEMINI.md` (a reflex target since the unified tool table) is missing from eject entirely.

satur8d:
## What

`totem eject`'s reflex scrub (`scrubReflexFiles`, `packages/cli/src/commands/eject.ts`) was heading-to-EOF: blind to `REFLEX_START`/`REFLEX_END`, it deleted everything from the block heading to end of file — including the span **after** the end marker that is contractually user content (the mmnto-ai/totem#1890 class, preserved byte-exactly by regen since the mmnto-ai/totem#2599 idempotence fix) — and left an orphan start+version marker pair that `detectReflexStatus` reads as `current`: corrupt but green (mmnto-ai/totem#2602).

Now:

- **Marker era (v2+):** eject removes each complete `REFLEX_START..REFLEX_END` span with the same single-owner seam discipline as `upgradeReflexes`. Pairing is **END-anchored** (`lastIndexOf`) inside a removal loop: an orphan `START` left by the pre-fix scrub sitting above a real block never widens the removal span, and double-injection damage clears instead of leaving a live block behind a "Scrubbed" line. User content below an end marker survives byte-exact; seam whitespace normalizes exactly as regen would (a seam contract, not a byte-restore — disclosed in the changeset).
- **Unpaired marker residue** (either orphan shape, or an inverted pair): those bytes are never scrubbed and the summary says so, including the sensor consequence — a residual version marker still reads as `current` to init, so healing is manual.
- **Legacy era (no markers):** bounded at the **next non-Totem H2** (mirroring init's own `upgradeReflexes` Case 2 boundary), not EOF — user content below a v1 block survives. Gated to `CLAUDE.md`/`.cursorrules`: markers shipped before the other three roster files joined the tool table, so a bare heading there can only be user-authored and is never touched.
- **Roster single-sourced:** derived from the same `AI_TOOLS` table init injects through (via the `scrubClaudeSkills` lazy-load precedent). The hardcoded `['CLAUDE.md', '.cursorrules']` pair silently skipped `GEMINI.md`, `.junie/guidelines.md`, and `.github/copilot-instructions.md` — three files init injects into that eject never cleaned.
- **Per-file best-effort:** one unreadable/locked file degrades to a reported skip instead of aborting the remaining eject steps (the Tenet 4 cleanup carve-out every sibling scrubber honours).

## Falsification rounds (pre-open)

**Round 1** on the initial diff refuted three of my claims with shipped-path evidence — all absorbed in the second commit: first-occurrence pairing deleted user content between a pre-fix orphan and a later complete block (a regression vs main on that input class; now END-anchored + loop), the legacy arm kept the same data-loss class below v1 blocks and pointed it at three files where a totem legacy block is provably impossible (now H2-bounded + gated), double blocks survived behind a green summary line (now cleared), plus accurate residue wording, per-file try/catch, seam-fidelity and sensor-state disclosures, and the spec amendment.

**Round 2** (scoped to the fold) confirmed the rewired core — loop termination provable, the splice cannot forge a marker occurrence, seam math identical to single-pass, the roster gate airtight — and its findings are absorbed in the third commit: the boundary regex's `Totem Memory Reflexes` alternative is dropped (both injector generations guarded on that heading's absence before writing, so a second Totem-titled H2 below a block is always user-authored — the alternative could only ever widen deletion across user bytes; the boundary is now init's Case 2 verbatim), the byte-0 legacy seam no longer mints a leading blank line, skip entries carry a `could not scrub:` prefix, and the retired pre-marker target `.gemini/gemini.md` joins the visit union (nothing had scrubbed its old blocks since the roster moved to `GEMINI.md`). Declined with grounds: the orphan-below pairing asymmetry (the pre-fix damage can only manufacture the orphan-above shape, and every narration clause is scoped to it), the CRLF legacy seam (byte-identical to init's own back-walk), and marker-quoted-in-prose false positives (pre-existing class, strictly worse on main — named here so it isn't read as covered).

## Tests

Thirteen new cases in `eject.test.ts`: four byte-exact `ejectCommand` cases (the mmnto-ai/totem#2602 reproducer, whole-file block to empty, simple-orphan byte-untouched, the three previously-skipped roster files) and nine direct-drive `scrubReflexFiles` summary-contract cases from the two falsification rounds (orphan-above-complete-pair, double-injection adjacent and with user content between blocks, inverted markers, legacy next-H2 boundary, the user-authored Totem-titled H2 as a boundary, legacy at byte 0, user-authored heading in a post-marker-era file, the retired `.gemini/gemini.md` target). The existing legacy-shape test pins the legacy path.

## Riders

The generated spec for this ticket (amended: its regex-safety framing was tangential; the marker-bounded design shipped) and a patch changeset for `@mmnto/cli`.

## Gates

Full workspace build, full `@mmnto/cli` suite (156 files / 3868 passed), repo-wide `format:check`, ESLint (0 errors), `totem lint` PASS (0 errors), typecheck clean.

Closes #2602 


satur8d:
/gemini review
@coderabbitai review

satur8d:
@gemini-code-assist @coderabbitai Round 1 disposition (operator-ruled):

**[GCA, `eject.ts` scan loop — high] Fixed in `747a44a3`.** Verified against the contract: a leading orphan `END` halted the scan and left the complete pair below unscrubbed, violating the "loops until no complete pair remains" claim. The scan now advances past ENDs with no preceding START (they remain reported residue) and resets after each removal — termination and never-span-widening both still hold (each iteration removes at least the END-marker length or strictly advances the cursor). A new test pins the shape byte-exactly. The replacer-function sub-suggestion is **declined**: the replacement literal `'\n'` contains no `$` sequence, so back-reference interpretation is unreachable there.

**[CR, `eject.test.ts` — skip-branch coverage] Fixed in `747a44a3`.** Directory-at-`CLAUDE.md` test asserts the `could not scrub:` skip entry and that later roster files still scrub (the per-file best-effort contract; EISDIR/EPERM were live-verified on the pre-open falsification round, now pinned in the suite).

PR #2621, opened 2026-08-09
fix(hooks): Gemini SessionStart emits the hookSpecificOutput envelope — interactive sessions dropped the plain-text briefing (#2613)

### LINKED ISSUE #2613: fix(hooks): GEMINI_SESSION_START emits plain text — interactive sessions drop it; briefing needs the hookSpecificOutput.additionalContext envelope
## GEMINI_SESSION_START emits plain text — interactive Gemini sessions drop it; the briefing must ship the `hookSpecificOutput.additionalContext` envelope

Found by the falsification re-leg on mmnto-ai/totem#2610 (strategy-claude seat, 2026-08-09); source-verified against the installed `@google/gemini-cli@0.54.4` bundle. Third sibling of mmnto-ai/totem#2611 / mmnto-ai/totem#2612.

### The defect

The managed `GEMINI_SESSION_START` template (shipped `.gemini/hooks/SessionStart.cjs`) writes the briefing as **plain text** (`execSync('totem describe', {stdio: [..., 'inherit', 'inherit']})` etc.). Gemini's hook runner wraps plain-text exit-0 output as `{decision: "allow", systemMessage: text}` — with **no** `hookSpecificOutput`. The interactive startup consumer reads **only** `result.getAdditionalContext()` (which requires `hookSpecificOutput.additionalContext`) and never references `systemMessage`. Locators from the re-leg (0.54.4 bundle): plain-text wrap at `chunk-BS6BSLZD.js:361434-361441`; `getAdditionalContext` returning undefined absent the envelope at `:333822-333831`; the interactive consumer at `interactiveCli-LU4ELYJR.js:32782-32794`.

Net: a correctly registered SessionStart hook runs, and its briefing is **silently dropped in interactive sessions** — the primary dev flow. It does surface on the non-interactive path (`gemini -p`, stderr) and on `/clear` (UI info item) — both handle `systemMessage`; verified live on the `-p` path.

So the Gemini session-briefing leg is inert-in-practice even after mmnto-ai/totem#2610 arms the registration — same sensor-trust shape as mmnto-ai/totem#2611, one event over: reads as armed, delivers nothing where it matters.

### The fix shape

The Claude twin already does it right: `.claude/hooks/session-context.mjs:498` emits `hookSpecificOutput: { hookEventName: 'SessionStart', additionalContext: fullContext }`. `GEMINI_SESSION_START` should capture the briefing output and emit the same envelope on stdout (Gemini's `hookSpecificOutput.additionalContext` is the interactive-context channel), keeping the fail-soft posture (exit 0 with a plain fallback note when `totem` is unresolvable — the current fail-soft behavior was re-leg-verified: exit 0, never denies, on a totem-less clone).

Secondary notes for the same fix, re-leg-measured:

- Latency budget: the template runs two `execSync` calls each with a 30 s budget against Gemini's `DEFAULT_HOOK_TIMEOUT = 60000` — a pathological hang consumes the whole hook budget at session start.
- Clone residue: on any clone with a real `.git` dir, the `totem-status refresh-gh` best-effort spawn logs ~200 bytes/session (absolute checkout path + PATH probes) to `.git/totem-status-refresh-hook.log` — untracked, dies with the clone; worth knowing it exists on outside contributors' machines.
- An end-to-end test asserting the briefing actually reaches interactive context would close the same test gap named in mmnto-ai/totem#2611 for the deny path.

### Relation to mmnto-ai/totem#2610

The PR ships the registration anyway (disclosed in its body): the registration is the prerequisite either way, the team-file contract + trust arming is the ruling's point, and the briefing starts reaching interactive sessions the moment this template fix lands — no consumer-side change needed then.

satur8d:
## What

The distributed Gemini `SessionStart.cjs` template (`GEMINI_SESSION_START`, `packages/cli/src/commands/init-templates.ts`) wrote its briefing as plain text (`stdio: 'inherit'`). Gemini wraps plain exit-0 hook output as `{decision:"allow", systemMessage: text}` — and the **interactive startup consumer reads only `hookSpecificOutput.additionalContext`, never `systemMessage`** — so a correctly registered SessionStart hook ran and its briefing was silently dropped in the primary dev flow (mmnto-ai/totem#2613, leg-verified against the installed `@google/gemini-cli@0.54.4` bundle with locators on the issue). Same sensor-trust shape as mmnto-ai/totem#2611, one event over: reads as armed, delivers nothing where it matters. With the team-file registration live since mmnto-ai/totem#2610, this template fix is what makes the briefing actually reach interactive sessions — zero consumer-side change.

Now the template:

- captures both briefing legs (`totem describe` + `totem orient --session`) across **both streams** — the Totem CLI writes its banner and diagnostics to stderr, so a stdout-only capture drops the describe leg entirely (round 1 of this PR's falsification caught exactly that in my first draft, whose test shim echoed on stdout and masked it);
- emits the `hookSpecificOutput.additionalContext` envelope **plus** `systemMessage` — one payload, both audiences: `additionalContext` is injected as model context, and `systemMessage` renders to the human (a UI info item on interactive startup via the hook-system-message path and on `/clear` — the one surface with no context injection at all — and a stderr echo under `-p`);
- keeps the fail-soft boot contract: exit 0 always; a totem-unresolvable note rides the envelope (so it reaches interactive context) and the orient leg's failure stays its stderr breadcrumb;
- drops per-leg budgets 30s→20s, cutting the **measured** worst-case process exit from 60s to 40s against Gemini's 60s default hook timeout (which tree-kills at expiry; post-fix no descendant inherits the hook's own streams, so the inner budgets are honored even under a hung grandchild — leg-measured);
- carries any partial leg stdout ahead of the fail-soft note, and the note rides both keys.

**Also in this PR (round-1 F2):** this repo's own tracked `.gemini/hooks/SessionStart.cjs` is drift-repaired from the new template — it was byte-in-sync with the old one, and shipping the template change without it would have introduced silent parity drift on the very seat that boots through it. The wiki's SessionStart example (`docs/wiki/governing-ai-agents.md`) is updated off the inert plain-text shape it still taught. The tracked legacy `BeforeTool.js` is marker-less (installer reads it as user-owned), registered nowhere, and its disposition rides the mmnto-ai/totem#2478 arming slice.

## Falsification rounds (pre-open)

**Round 1** refuted the central claim of my first draft: `totem describe` emits on stderr, so the stdout-only capture contributed nothing to the envelope — and the contract test's stdout-echoing shim masked it (fixture weaker than contract). Also: the tracked artifact regen gap, the false `/clear` disclosure, the overstated timeout claim, and the repo-local path in a distributed comment. All absorbed in the second commit.

**Round 2** (scoped to the fold) confirmed the engineering end-to-end — both-streams capture measured necessary against the real CLI (describe: stdout 0 bytes, stderr 1,135), the tracked artifact byte-identical to the template, the 40s worst-case bound independently reproduced, mutation checks live — and refuted the *narrative*: the vendor's hook-system-message path renders `systemMessage` as a UI info item on interactive startup too (my "nothing double-renders" claim was false in four shipped places — restated as one-payload-both-audiences, which is the honest and desirable behavior), the `-p` justification overclaimed (the parent envelope already injected there; only `/clear` loses the briefing without `systemMessage`), and the "hung grandchild" residual described a hazard this very diff removed. All absorbed in the third commit, plus: leg failures carry partial stdout, the fail-soft test gained the `systemMessage` assertion that makes it a genuine falsifier vs the envelope-only intermediate shape, and the armed-path test comment states its platform split honestly (win32 pins the spawn-reject-quiet branch; the POSIX CI legs pin the log-fd routing). Convergence: the engineering claims have now survived two adversarial rounds with independent measurement; the round-3 delta was prose and test-strength only, so this presents for review rather than re-arming a fourth leg.

## Tests

`gemini-sessionstart-contract.test.ts` executes the **rendered artifact** the way Gemini executes it: a cross-platform shim now tagging **both streams** (order + `systemMessage` parity pinned), the no-resolvable-`totem` fail-soft path, and the armed refresh-gh path (`.git` present + chatty `totem-status` shim; PATH isolated to the shim dir — appending the real PATH let the detached spawn run the machine's actual sidecar, which held the temp cwd). Stale `init.test.ts` pins updated — the pre-fix pin asserted the `'inherit'` stdout channel with a comment claiming that's how "the orientation output reaches the agent."

## Riders

The generated spec (pre-amended: its XML-envelope design was invented; the vendor contract is the JSON envelope) and a patch changeset for `@mmnto/cli`.

## Gates

Full workspace build, full `@mmnto/cli` suite (157 files / 3877 passed), repo-wide `format:check`, ESLint (0 errors), `totem lint` PASS (0 errors), prettier clean.

Closes #2613 


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@gemini-code-assist @coderabbitai @greptileai Round 1 disposition (operator-ruled):

**[CR, orient failure handling — Major] Fixed in `b1cd5f0d`.** Verified: the PR's stated contract ("a leg failure carries any partial output plus the unavailable note") held for the describe leg only — orient dropped stderr partials and noted only to parent stderr, invisible to the model. Orient is now symmetric with describe: partial stdout carried, the note (reason = the leg's stderr) rides both `additionalContext` and `systemMessage`, and the parent-stderr breadcrumb stays for diagnostics. The tracked artifact is regenerated and a failing-orient shim test pins the behavior end-to-end.

**[Greptile, timeout-misses-the-process-tree — P1] Declined; premise partially true, consequence refuted by measurement.** True half: the per-leg timeout terminates the intermediate shell, not the totem/gh descendants — orphans can linger. Refuted half: "session startup reaches Gemini's outer timeout instead of emitting the fail-soft envelope." The pre-open falsification rounds measured `spawnSync` returning at its budget with a grandchild still holding the leg's pipe (2,006ms return on a 2,000ms budget under a 14s-sleeping grandchild), the hook exiting at 40,075ms worst case with **both** legs hung and a valid envelope on stdout, and the stdio wiring (`['ignore','pipe','pipe']`) hands no descendant the hook's own streams — so the vendor's close-keyed completion fires on hook exit, inside the 60s outer budget. The pre-fix `'inherit'` wiring was the shape with that hazard; this diff is what removed it. The lingering-orphan residue is real, bounded, and strictly better than pre-fix.

Issue #2620, opened 2026-08-09
feat(cli): user-file write-safety contract — atomic temp+rename writes, dirty-tree sense on destructive commands, recovery artifacts for non-VCS surfaces (eject slice first)


satur8d:
Raised by the operator during the mmnto-ai/totem#2602 lane (2026-08-09), which was a live specimen of the class: eject's reflex scrub deleted user content, and the recovery story was "hope the file was committed." The protection model today is ownership markers + consent prompts + git-as-backstop, which is the right model for tracked files — but it has three named holes, and the discipline should be a stated contract for every user-file write, not an eject-local property.

## The contract (applies to any totem code path that mutates a user file)

1. **Atomic writes everywhere.** `fs.writeFileSync` opens truncate-then-write; a mid-write failure leaves a torn file (measured on the mmnto-ai/totem#2602 falsification round — an EPERM after open reports as a "skip" while the file may be truncated). Every user-file write goes temp-file + rename so the observable state is always old-or-new, never half. This is worth more than any backup scheme.
2. **Destructive commands sense a dirty tree.** The git-native form of "backup before modify" is not copying files — it is making sure a commit exists to diff against. `totem eject` (and any future destructive verb) should check `git status` for uncommitted changes to the files it is about to mutate and say so before the consent prompt (sense-and-say, not a hard block; `--force` proceeds). An uncommitted CLAUDE.md has no revert point.
3. **Non-VCS surfaces get a recovery artifact.** `.git/hooks/*` is not version-controlled — the one surface where eject/install mutations are unrecoverable. A scrub there writes the pre-scrub bytes to `<name>.totem-bak` (or prints the removed span into the summary). Same calculus for any future home-dir mutation beyond derivable registries.
4. **Explicitly NOT `.bak` files for tracked content.** Marker-bounded ownership + git is the contract there (prettier/eslint-fix convention); backup litter for tracked files is an anti-pattern. The one existing `.bak` precedent (`migrate-lessons`, a one-shot irreversible format migration) is a legitimate instance of rule 3's class, not a template for routine writes.

## Slices

- **Slice 1 (eject):** dirty-tree sense line before the consent prompt; atomic writes in `scrubReflexFiles` / `scrubHook` / settings scrubbers; `.totem-bak` for `.git/hooks` scrubs. Hangs off the summary plumbing that already exists.
[...]

PR #2624, opened 2026-08-09
feat(cli): user-file write-safety contract — atomic helper, hook recovery artifacts, dirty-tree sense, loud backstop (#2620 slice 1)

satur8d:
First implementation slice of the **Tenet 4 User-File Mutation Contract corollary** (mmnto-ai/totem-strategy#1045, shipped in `@mmnto/strategy-doctrine` 0.1.33). **Part of #2620** — the tracker stays open for slice 2 (the ~20-site writers sweep, `migrate-lessons` `.bak` disposition, reaper extension).

## What ships

**Core — `writeFileAtomicSync`** (`packages/core/src/fs-atomic.ts`, also exported as the `@mmnto/totem/fs-atomic` subpath so cold-start command files can static-import it):

- Same-volume unique temp (`.<pid>-<uuid8>.tmp` — UUID entropy per the two banked collision lessons) + fsync + **metadata applied to the temp, rename last** (author intent-read Q2: a chown failure leaves old bytes as the *entire* observable state, metadata included).
- Symlink identity preserved via realpath; dangling/cyclic links throw with the target untouched; cross-uid chown failure throws (never a silent ownership flip).
- Declared boundaries in the docblock: win32 skips metadata; hard links decouple on rename; parent dir must be writable; hard-kill temp litter.
- The three prior inline temp+rename core writers (worktree registry, sync-state/checkpoint/manifest, verification outcomes) consolidate onto it — byte-identical output pinned by their existing suites.

**Eject — the corollary's four rules applied:**

- **Rule 1:** all five user-file rewrite sites go through the helper.
- **Rule 2:** a dirty-tree sense ahead of the consent prompt, roster derived from the same exported constants the scrub/removal steps consume (deletions count — intent-read Q3), **including gitignored targets** via `--ignored=matching` ("git holds no copy of these at all" — the `.totem/secrets.json` class); sense-and-say, never a block; `--force` skips the prompt, not the sense, and restates it after the summary; old git degrades to the dirty-only sense with an honest visibility line.
- **Rule 3:** `.git/hooks/<name>.totem-bak` written with the source hook's **exact bytes** (raw Buffer) and mode **before** any mutation, on both partial-scrub and full-removal paths (intent-read Q1: recovery keys off the surface class, never the mutation-time content claim); a failed bak withholds that hook's mutation entirely.
- **Tenet-4 backstop:** an eject where every reflex scrub attempt fails throws `EJECT_FAILED` — deferred past artifact cleanup and the full summary (typed: only the backstop defers; a packaging defect fails loud immediately). Any file carrying a Totem block that doesn't round-trip UTF-8 is refused with a reason instead of U+FFFD-corrupted.

Plus: `docs/wiki/cli-reference.md` eject paragraph, patch changesets for both packages, and the spec at `.totem/specs/2620.md` carrying the design doc + falsification amendments.

## Review of record

Three adversarial falsification legs per the cohort review-leg contract (full branch → scoped re-arm → convergence): findings 15 → 11 → 6 → **zero live defects**. Every merge-blocking refutation was live-reproduced before absorption and live-re-verified after (old-git PATH shim, win32 mode probe, non-UTF-8 byte probes). Full disposition trail with the author intent-readings: the two record comments on #2620. Advisory `totem review` lanes: 0 critical; the one warn absorbed as integration tests.

## Gates

Full workspace build · core 3652 + cli 3899 + mcp 216 tests, zero failures (skips are declared win32/symlink capability skips) · `totem lint --base main` PASS 0 errors · prettier · ESLint.


satur8d:
/gemini review
@codrabbitai review

satur8d:
@coderabbitai review

satur8d:
Round disposition — all 7 findings verified against source and dispositioned; fixes pushed as `7660440e`. Thank you both — two genuine catches on a data-loss-class diff that had already been through three adversarial review legs.

@gemini-code-assist

1. **eject.ts — `err.stderr` access can TypeError on a non-object throw: FIXED** (`7660440e`). Premise verified: optional chaining guarded `.stderr`, not `err` itself, and the crash site is inside the sensor's own catch — a sensor that can crash its catch can abort the eject it senses for. The guard now routes through `err instanceof Error` with a `typeof === 'string'` check on the field.

@coderabbitai

2. **Changeset `@mmnto/totem` patch → minor: FIXED** (`7660440e`). Correct call — `writeFileAtomicSync`, `AtomicWriteOptions`, the `./fs-atomic` subpath, and `EJECT_FAILED` are new public API. We bumped `@mmnto/cli` to minor as well: this slice is a feature, not a fix.
3. **cli-reference.md — document the two failure outcomes: FIXED** (`7660440e`). Both bullets added in house voice: the non-UTF-8 refusal (skip + byte-identical file + manual-removal instruction) and the post-summary non-zero `EJECT_FAILED` exit.
4. **Unbounded sense output: FIXED in part, DECLINED in part** (`7660440e`). The display cap is in — per-path rows cap at a named maximum (20) with an "… and N more" trailer while the exact counts stay in the headers; the consent prompt can no longer scroll off behind sense lines. The `maxBuffer` raise is declined on measured grounds: a pattern-ignored directory lists as ONE porcelain row (live-probed on this branch's falsification rounds — `--ignored=matching` does not expand a dir ignored by pattern; per-file rows appear only where individual files match patterns), the 10MB default is ~200k rows of headroom, and an overflow degrades to the disclosed "Could not derive VCS state" line rather than silence. Falsifier: a real consumer repo producing >10MB of porcelain over this 24-path roster.
5. **`"types"` before `"import"` in export conditions: FIXED** (`7660440e`), and applied to all six entries including the pre-existing siblings — your resolution harness was right that import-first only worked via the adjacent-`.d.ts` fallback under Node16.
6. **Static `@mmnto/totem/fs-atomic` import in eject.ts: DECLINED.** The path instruction's intent is cold-start cost, and this import is the engineered exception: your own analysis chain confirmed the subpath resolves directly to `dist/fs-atomic.js` without transiting the core barrel, and the built module imports only `node:crypto`/`node:fs` (verified by import-graph on this branch's review legs — load cost ≈ zero). Threading a dynamically-loaded function through five synchronous scrub helpers would re-introduce exactly the seam the subpath design exists to avoid. Falsifier: measure `import('@mmnto/totem/fs-atomic')` cost — if it ever exceeds ~1ms or grows a dependency edge beyond node builtins, this decline dies.
7. **Ticket refs on skipped tests: DECLINED.** These are host-capability skips, not quality-gate bypasses: every skipped assertion (POSIX mode/umask, symlink identity) runs unconditionally on the CI POSIX matrix (ubuntu + macos in ci.yml), so no assertion is globally bypassed, and each skip carries its capability rationale inline. The guideline's target — suppressing a gate without a trail — has no instance here. This decline dies if the CI matrix ever drops its POSIX runners.

Review-of-record context: this branch carried three adversarial falsification legs before the bot round (full branch → scoped re-arm → convergence; disposition trail on #2620). The supplementary local LLM lanes ran over a >50k-char diff and saw truncated content — carried per contract, not the review of record:

local-lane: 5c68211e round=0 settled=false lanes=2/2

Issue #2468, opened 2026-07-20
feat(instrumentation): per-seat selection manifests — the Prop 308 / #467 measurement instrument (M1)


satur8d:
Instrumentation spec for the Proposal 308 measurement pass. **M1** of the pre-registered sequence.

Serves two consumers with one instrument: Proposal 308's own falsifier, and the `mmnto-ai/totem-strategy#467` corpus-consumption pass. Neither has a measurement surface today; both are currently arguing from anecdote and one-off `/context` reads.

## What to build

A **read-only selection manifest**, emitted per selection event, keyed by content fingerprint.

Emitters (the three consumers that select on a seat's behalf today):

- the session-start hook — the always-injected tier
- `search_knowledge` — retrieval
- `totem orient` — derived state

Per candidate, each manifest records:

| Field | Purpose |
|---|---|
| candidate identity + **content fingerprint** | join key across emitters; survives moves/renames |
| bytes / tokens | the cost side |
| why-**selected** | the policy's own stated reason |
| why-**excluded** | the reason a candidate lost — the half that is invisible today |
| actually-used | did the agent cite, read, or edit it |

**Baseline metric: tokens-to-first-useful-action.**

## Non-goals — the round's refute condition

This instrument must **not** introduce a shared relevance score or common weights. Proposal 308 F1 was unanimous (9/9) that consumers share a *selection envelope*, never a relevance model, and lc-codex recorded the explicit refute condition: *if this proposal ever requires common weights or a universal score to count as success, the round's answer is REFUTE.*

The manifest is a record of what each policy did and why. It is not a ranker, and it does not adjudicate between policies.

## It senses, it never gates

Read-only, per [Tenet 13](https://github.com/mmnto-ai/totem-strategy/blob/main/design-tenets.md#13-sensors-not-actuators). Emitting a manifest must never block, filter, or alter a selection — an instrument that changes the thing it measures cannot serve as a baseline. Failure to emit degrades to a recorded absence, never a silent skip (see `mmnto-ai/totem#2466` for the failure mode where a silent skip reads as a clean result).

## What it unlocks downstream

The **overlap number** becomes derivable: `|injected ∩ retrieved| / |retrieved|` per instrumented session. Low overlap means the mechanisms solve disjoint problems, savings are additive, and Proposal 308's compounding thesis dies at first contact. This is *inter*-mechanism overlap — distinct from the *intra*-retrieval duplication in `mmnto-ai/totem#2458`, which is already measured.

[...]

Issue #2458, opened 2026-07-19
bug(mcp/search): federated search_knowledge returns every hit twice — ~50% of retrieval payload is duplication


satur8d:
Federated `search_knowledge` returns **every hit twice**, roughly doubling the token cost of each retrieval for zero added information.

## Observed

Two separate `search_knowledge` calls from `mmnto-ai/totem-strategy` in one session. In **both**, results paired up exactly — hits 1&2, 4&5, and 7&8 were the *same file, same section, same score*, differing only in a source label:

```
### 1. Proposal 307: ... > Motivation (spec)
**File:** D:\Dev\totem-strategy\proposals\active\307-derived-readme-zones-product-feature.md | **Score:** 0.016

### 2. [strategy] Proposal 307: ... > Motivation (spec)
**File:** D:\Dev\totem-strategy\proposals\active\307-derived-readme-zones-product-feature.md | **Score:** 0.016
```

One copy unlabeled, one prefixed `[strategy]`. Same for the `design-tenets.md` Claim Discipline hits and the Proposal 307 Existing-Surface-Coupling hits.

**Impact:** approximately **half** of every federated retrieval payload is exact duplication. Retrieval results — not tool schemas — are the dominant token cost of totem's MCP surface (the 4 tool definitions total roughly 548 tokens; a single `search_knowledge` result set is orders of magnitude larger). So this is likely the single largest avoidable token cost in the MCP path.

## Config context

`mmnto-ai/totem-strategy:totem.config.ts` line 101:

```ts
linkedIndexes: ['../totem'],
```

Federation went live in mmnto-ai/totem-strategy#877 ("strategy→totem linkedIndexes federation live").

## Likely mechanism — needs code confirmation, not asserted

A strategy-owned document appears reachable by **two paths**: directly from the primary index, and transitively through the linked index. With reciprocal linking across the cohort, the same document identity is returned once per path, and the federated result set is not deduplicated by document identity before ranking or return.

I have verified the **symptom** and the **config**, not the code path — the exact traversal is for the totem lane to confirm.

## Proposed fix

Deduplicate federated results by document identity (file path + section anchor) before ranking and return, keeping the highest-scoring instance. A source label is useful provenance, but it should not multiply the payload.

Worth checking whether the same duplication affects other federated consumers (`describe_project`, any retrieval used by `totem review` lanes).

## Why this matters beyond the bug

[...]

PR #2625, opened 2026-08-10
feat(instrumentation): per-seat selection manifests — the Prop 308 / strategy#467 M1 instrument (#2468)

satur8d:
Builds the M1 instrument for #2468 — the sole blocker to strategy#467 run 1. The issue stays OPEN: seats still owe instrumented baseline sessions, and M2 threshold pre-registration is operator + bus.

## What ships

**Core substrate** (`@mmnto/totem`, minor): `SelectionManifestRowSchema` — strict at both levels — with `appendSelectionManifest` / `senseSelectionManifest`, `fingerprintContent`, `buildMeasuredCandidate`, writing one NDJSON row per selection event to the new `.totem/ledger/selection-manifests.ndjson` sidecar under the QBD sensor posture (Tenet 13: emission never blocks, filters, or alters a selection; failures surface on the accounting channel).

**Three emitters:**
- `search_knowledge` (MCP): returned hits, below-floor withheld candidates, the previously **silent** federation final-limit cut, and store-outage accounting (per-store failure warnings, `storesFailed` / `storesUnavailable`, `answered=` in the federated universe).
- `totem orient`: report sections on **agent-invoked renders only** (the #2510 scope mirror — the `--session` boot render is the hook's selection); derive errors recorded as excluded; declines and global-profile landings are loud, mirroring the QBD seam.
- Session-start tier: the totem-repo V2 hook records per-block candidates (journal recency policy with sibling exclusions, mail display caps, ticket-matched proposals, vector char-budget cuts via new per-result metadata on `AutoContextResult`, and the global 10k slice as `finalTruncation`); both published SessionStart templates append an inline row **bound to the core schema by contract test**.

**A.3.a parity:** both session hooks now mint the session UUID + `session_start` row — measured live, this repo had 5 `session_start` rows against 132 `mcp_call` because the V2 hook replaced the managed template without taking over the mint duty, and the Gemini template never had it. The UUID is the manifest join key; the row is the recorded-absence denominator (a denominator row with no matching manifest row IS the recorded absence — no silent skip is constructible, per the #2466 clause).

## Contract honesty

- **Prop 308 F1 refute-condition guarded structurally:** `.strict()` at both schema levels — no shared score field can appear without breaking parse of existing rows. Per-candidate `reason` strings carry each policy's own words only.
- **No fabricated measurement:** candidates the policy never read are recorded id-only; nothing over-fetches to manufacture exclusions.
- **Attribution is stamped absence** (ADR-078): `agent_source` comes only from `TOTEM_SELF_AGENT`, omitted when unset — pinned both directions by contract tests on both templates and live-probed on the V2 hook.
- **Cost basis, stated plainly:** every row names its aggregation (`utf8-length` bytes; `ceil(bytes/4)` declared as an approximation). This is a **payload-byte census, not the ruled `<total>` measured-consumption basis** — whether it may substitute for a `<total>` surface is the pre-registration's call via its §9 versioned revision, not this instrument's claim.

## Disclosures for the measurement pass

- The same producer-side relevance scalar (LanceStore vector similarity) is machine-extractable from two emitters' reason strings. Pre-blessed by the design (a policy's own stated reason), recorded here so the pass carries it knowingly.
- Whole-file injected fingerprints vs retrieved-chunk fingerprints will not collide; `id` (path) is the coarse join key, `fingerprint` the exact one — join policy belongs to the pass.
- The hook's global 10k slice is string-level and not attributed per-candidate (`finalTruncation` carries it); summing candidate `bytes` over-counts delivered payload when `applied` is true.
- #2458 (federated dedupe) is still open — the pre-dedupe baseline window this instrument exists to capture remains obtainable; instrumented baseline sessions should run soon after release.

## Review record

Two falsification-leg rounds (review of record per the leg contract):
- **Round 1** (full branch): 15 findings — 2 merge-blocking (red `totem lint` gate; the V2 hook fabricating `agent_source: 'totem-claude'` when the env was unset, on both the denominator and manifest rows), 4 defects, 3 hazards, 6 nits; 11 claims CONFIRMED under live probes including main-vs-branch byte-identity of both rendered templates across 5 scenarios. Absorbed in `d8d8443a`; N-11/N-14/N-15 declined (notice-is-not-a-briefing by design; double config-load immaterial vs seam-refactor risk; pre-existing template mkdir class kept at parity).
- **Round 2** (re-armed, scoped to the fold): all eleven absorptions re-verified live (seat attribution probed both directions; subdirectory-cwd probe confirming gitRoot anchoring; 10-input totality probe on `buildMeasuredCandidate`), **zero regressions**, verdict converged. Its residue — the third orient decline path still silent, the D-3 arithmetic tested only on the all-healthy identity — absorbed in `ccd4d7be` with the outage arithmetic pinned under a real primary failure. The convergence fold changes no selection semantics (notes, assertions, one context scalar), so no third round was armed — owner's judgment, on the record.

Gates: full workspace build; 3,914 CLI + core + mcp tests green; `totem lint` PASS 0 errors; prettier + ESLint clean; changesets minor ×3 (new public core API + emitter behavior).

Spec + design record (three-drift-corrected generated spec, ruled design, build-time amendments, leg-round corrections): `.totem/specs/2468-selection-manifests.md`.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@gemini-code-assist @greptileai @coderabbitai

## Bot round 1 disposition — 13 findings: 10 fixed in `a4b9b9ca`, 3 declined with grounds

Every finding was verified against source before disposition; declines carry falsifiable grounds.

### Fixed (10)

| # | Finding | Absorption |
|---|---|---|
| CR (Major) | `context` accepted `sharedRelevanceScore` — the structural refute-condition guard didn't hold at the context layer | **Real catch.** `SELECTION_CONTEXT_KEYS` closes the key space: context is now enum-keyed, so adding a context key is a deliberate schema change exactly like adding a field. Parse test injects `sharedRelevanceScore` / `commonWeight` and asserts rejection. |
| CR (Major) | Candidate schema permitted partial/contradictory measurements | `superRefine`: the measured trio (fingerprint, bytes, approxTokens) travels together — a candidate is id-only or fully measured, never partial (matching what the builders produce post-H-7); `deliveredBytes` only on `truncated` and never > `bytes`. Both test fixtures that encoded the partial state are corrected; rejection tests added. |
| Greptile P1 + CR (Major) | `storesUnavailable` double-counted warn-only stores (empty-but-queryable links sit in both `linkedStoreInitErrors` and `linkedStores`) | Confirmed against `context.ts` — counted now as init-error entries **not** also present in `linkedStores`. |
| CR (Major) | `isInstrumentedProject` converted EACCES/I-O failures into a silent not-a-project skip | Probe distinguishes ENOENT/ENOTDIR (honest absence, silent) from every other errno (named warning on the accounting channel). EACCES test added. |
| CR (Major) | A throwing `onWarn` sink could escape `senseSelectionManifest` | All sink invocations routed through a throw-contained helper; the warning survives in the returned array either way. Throwing-sink test added. |
| CR (Major, partial) | Concurrent hooks overwrite the shared `.session-id` between mint and emission | Partial absorption: `SelectionManifestInput.sessionId` override — the session-start hooks stamp the UUID they minted in-process instead of re-reading the shared pointer (templates were already immune, inlining the minted value). The broader transport redesign is declined below. |
| GCA (High) | `err.code` accessed without a type guard in the reader | Guarded house-style (the `qbd/record.ts` pattern), both read sites. |
| CR (Minor) | `cliVersion` resolution failure discarded with `void err` | Lands on the row's `warnings[]` accounting channel; render stays non-blocking. |
| CR (Minor) | Malformed JSON lines skipped silently in the reader | One **aggregate** count warning per read (never a per-line firehose); the schema-invalid skip stays under the documented generic-consumer contract, reject-counting consumers read the file raw. |
| CR (Trivial) | Magic `4` in the test mock's token divisor | Named `APPROX_TOKEN_BYTES`. |

### Declined (3)

1. **Greptile P1 — non-atomic ledger mutations.** The manifest file is an append-only telemetry log, deliberately at parity with `events.ndjson` (`appendFileSync`, the documented interleaving-prevention choice in `ledger.ts`). The #2620 atomic-write contract covers **user files**; `.totem/ledger/*` is derived, gitignored telemetry whose readers skip torn lines by contract and whose reject-counting consumers read raw. A torn tail loses at most one row and is detectable against the denominator; rename-per-append would rewrite the whole file per event (O(n²)) and recreate the interleaving hazard `appendFileSync` exists to avoid. Falsifiable: if the measurement pass requires zero-loss durability under power failure, that is a WAL redesign for **every** ledger surface, filed separately — not this PR.
2. **CR — session-scoped transport redesign for the join key.** The single-pointer `.session-id` is pre-existing A.3.a design (ADR-029; "rotates on subsequent SessionStart" is its documented contract), shared with QBD correlation, which handles cross-session pollution fail-closed. This PR extends the same mechanism at parity and ships the partial mitigation above (minting emitters no longer touch the pointer at emission). Cohort doctrine runs one seat per checkout (per-agent worktrees), bounding the residual to simultaneous same-checkout session starts. A transport redesign moves ADR-029 for every consumer — out of this PR's scope; candidate follow-up if the residual is ever measured.
3. **CR — lazy-load the new core imports in `packages/mcp`.** The same module statically imports `appendLedgerEvent`, `readSessionId`, `senseCorpusQuery` from the same specifier on main (`ledger-writer.ts`), as does `search-knowledge.ts` for `ContentTypeSchema`. Lazy-loading two siblings of an already-statically-loaded module buys no startup win and splits one specifier across two loading disciplines. If the path rule is to be enforced, it is a file-wide sweep across pre-existing imports — separate filing, not a per-line patch.

Gates on the fold: full workspace build; 3,914 CLI + 226 MCP + 25 core tests green; `totem lint` PASS 0 errors; prettier + ESLint clean. Live hook re-probe: envelope intact, minted UUID stamped directly (matches the pointer), attribution honestly absent with the env unset.

satur8d:
@gemini-code-assist

satur8d:
@greptileai

## Re-review disposition — round converged

All three bots accepted the round-1 dispositions in full (10 fixes verified against `a4b9b9ca`; all 3 declines accepted by each bot independently, with CodeRabbit banking the decline grounds as learnings).

**Greptile's one residual was a correct catch, fixed in `ebcb8bb4`:** the cliVersion-accounting fix had landed only in `orient.ts` — `logSelectionManifest` (the `search_knowledge` emitter) still swallowed the version-resolution failure with `void err`, so the round-1 disposition table overclaimed that row. The failure now lands on the row's `warnings[]` there as well. The session-start templates hand-build their rows and were never affected, as the finding itself noted.

Gates on the fix: MCP build + 226 tests green, prettier/ESLint clean, `totem lint` PASS 0 errors.

PR #2632, opened 2026-08-13
feat(triage): normalize github-code-quality[bot] as a review surface — detection, first-class persistence, NITS keyword fix (#2626)

satur8d:
Closes #2626 

## What

Normalizes GitHub's `github-code-quality[bot]` (ghcq) as a first-class review surface. Previously ghcq authors failed `isBotComment`, so every parse path dropped their comments invisibly — no finding, no accounting, no trace (five such comments sit in this repo's own PR history since 2026-05-12: #1897, #2224, #2434 ×2, #2457).

- **Detection:** bot-login-shape match per the greptile precedent (#2244) — exact login plus a variant arm, `[bot]` suffix required so human accounts stay human. Widens every consumer: `triage-pr` (display abbrev `GHCQ`), `review-learn` extraction and pushback/disposition accounting, `recurrence-stats` and `retrospect` (first-class `ghcq` in both persisted vocabularies — additive enum extension in `@mmnto/totem`), and retrospect round counts (ghcq's review submissions are empty carriers for inline comments; a round it authored alone previously counted zero).
- **Severity:** deliberately `info`-only — the observed corpus (n=7 inline comments across both repos) carries no severity vocabulary, and per the issue's derive-don't-guess rule none is invented; `info` rides the generic branch to the `low` bucket in recurrence stats.
- **Triage keyword fix forced into the light:** the NITS bucket used bare substring matching — ghcq's stock closing boilerplate ("no new **defi-nit-ions**") mis-bucketed 5 of its 7 real findings, and "**u-nit** test" did the same for every bot. The landed design is per-bucket asymmetric: NITS alone requires the keyword to start outside word context (letters, numbers, combining marks, `_`; trail open so inflections match), while SECURITY/ARCHITECTURE/CONVENTION stay byte-for-byte on their long-standing substring semantics (`unsanitized`/`unauthorized` still hit security; over-matching upward cuts the safe way). Versus prior behavior the NITS bucket only narrows — the sole proven exception route in is the `u` flag's case folding on the U+017F class, verified the complete exception set.
- **Surface note:** ghcq has no known @-listener (attested: no in-org tag attempt has drawn a response and none is documented) — the `review-reply` skill records it across all three locked copies plus the distributed `totem init` template: never tag it; dispositions are audit-trail-only.

## Review record (the standing leg contract)

Four falsification rounds, review of record; every finding verified against primaries and folded or declined on the record:

[...]

Issue #2628, opened 2026-08-12
fix(hooks): tracked .gemini SessionStart.cjs lacks the #2625 mint block — regenerate from the published template (parity drift, the #2621 class)


satur8d:
**Template-parity drift, the exact class #2621 warned about, one release later.** The published Gemini SessionStart template gained the A.3.a mint block in #2625 (`1aac848d`: session UUID + `session_start` ledger event + inline selection-manifest row — "the join key and denominator the instrument depends on"). This repo's tracked `.gemini/hooks/SessionStart.cjs` was regenerated at #2621 for the envelope fix but NOT at #2625, so it predates the mint block: 175 lines, no session-id mint, no `agent_source` stamp, no manifest row (verified by grep at `main` @ `1aac848d`, and again at the #2511 build census).

**Consequence (live specimen, 2026-08-12T06:00-06:47Z):** a totem-gemini session in this repo minted no identity of its own, so its MCP activity (search_knowledge calls, one `add_lesson`, orient derives) fell through to the shared `.totem/ledger/.session-id` pointer and the ambient `TOTEM_SELF_AGENT` — recording under **totem-claude's** seat id and session id (`b8d7aa9b…`, `session_start` 03:03Z, a different vendor's session). The misattribution class is tracked as its own issue (cross-seat attribution fallthrough); THIS issue is the parity-drift half: the seat that boots through this hook cannot self-identify.

**Fix:** regenerate the tracked copy from the current published template — the same move #2621 made, for the same stated reason ("shipping the fix without regenerating it would have introduced silent parity drift on the very seat that boots through it" — which is what happened here). Consider a contract test binding the tracked copy to the template the way the two signoff-skill copies are parity-locked, so the third regeneration isn't manual vigilance either.

satur8d:
Sibling (the misattribution half this drift enables): #2629

Issue #467, opened 2026-03-14
feat(shield): Compile WWND flight rules into active shield lessons


satur8d:
## Summary

Translate the WWND (What Would NASA Do) architectural constraints, flight rules, and threat models from the `.strategy` repository into concrete, anchored lessons in `.totem/lessons.md`.

## Context

We have established strong boundaries in our strategy repo (e.g., `FR-D01: .lancedb/ is a disposable build artifact`, `FMEA-006: JSON parsing of LLM outputs requires resilient stripping`). However, `totem shield` only enforces rules that have been compiled into `compiled-rules.json`. Currently, our agents and developers can still violate these architectural boundaries because the rules haven't been anchored as syntactically-aware lessons.

## Requirements

Add specific, code-aware lessons to `.totem/lessons.md` covering:
1. **JSON Parsing Resilience (FMEA-006):** Flag direct uses of `JSON.parse` on untrusted string inputs without markdown fence stripping.
2. **LanceDB Disposability (FR-D01):** Flag any attempt to write a migration script or persistent schema logic for the `.lancedb/` directory.
3. **Context Separation (FR-C05):** Flag string concatenation of rule/instruction text directly with source code diffs (enforcing `formatXmlResponse` or equivalent).

## Acceptance Criteria
- Run `totem compile` to generate the new rules into `compiled-rules.json`.
- The `totem shield` command successfully flags violations of these rules in a test fixture.

PR #2634, opened 2026-08-13
feat(attribution): agent_source provenance + fail-closed conflict sense at the MCP stamp seams (#2629)

### LINKED ISSUE #2629: bug(attribution): cross-seat fallthrough — ambient TOTEM_SELF_AGENT + shared session pointer misattribute a sessionless seat's activity (M1 instrument contamination)
**Class:** on a shared multi-seat machine, a seat session that mints no identity of its own inherits ANOTHER seat's attribution through two ambient channels — and the M1 instrument (#2468/#2625) records the contamination on exactly the axis it exists to measure.

**Live specimen (2026-08-12T06:00-06:47Z, this repo's ledger):** a totem-gemini session's MCP activity — search_knowledge calls, one `add_lesson` (an lc-domain lesson mislanded into this corpus; since deleted + index resynced), orient derives — recorded as `agent_source: totem-claude` under session `b8d7aa9b…` (totem-claude's session, `session_start` 03:03Z). Two stacked causes:

1. **Ambient seat identity:** `TOTEM_SELF_AGENT` had been set as a USER-level Windows env var (to arm baseline-session attribution). User scope is machine-wide: every seat's processes inherit it, so a per-SEAT identity became a per-MACHINE stamp. (Reverted same-day; the honest stamped-absence default is restored.)
2. **Shared session pointer:** the seat minted no session of its own (the tracked Gemini hook predates the #2625 mint block — the parity-drift half, filed separately), so emitters resolved the shared `.totem/ledger/.session-id` pointer — a foreign session id.

**Why it matters:** the strategy#467 baseline runs join manifests/ledger rows by (`agent_source`, `session_id`). A contaminated row is not noise — it is a wrong-seat, wrong-session row that inflates one seat's denominator and hides another's activity entirely. The 06:00-06:47Z window under session `b8d7aa9b…` with `source: bot` should be excluded from any baseline that reads this machine's ledger.

**Fix directions (protocol input owed from the strategy measurement lane — dispatch sent 2026-08-12):**

- The parity-drift issue closes the Gemini-seat mint gap (each seat self-identifies at session start).
- Protocol: `TOTEM_SELF_AGENT` is set in the LAUNCHING SHELL per session, never at user/machine scope — candidate rider on the #467 attribution protocol. The env-var mechanism itself is fine; the scope was the defect.
- Candidate sense (Tenet 13, not a gate): the MCP server could disclose its attribution PROVENANCE (process-env vs absent) alongside `agent_source`, so a stamped-but-ambient identity is at least distinguishable post-hoc. Shape needs the measurement lane's input — filed as an open question, not a commitment.

satur8d:
Builds the operator-ruled attribution provenance/fail-closed sensor (ruling recorded on the issue, 2026-08-12): the M1-contamination class where a sessionless seat inherits another seat's attribution through ambient `TOTEM_SELF_AGENT` + the shared `.totem/ledger/.session-id` pointer (the `b8d7aa9b` 06:00–06:47Z specimen).

## What ships

**One shared derivation, both env-stamp seams.** Core gains `senseAgentAttribution` (`packages/core/src/attribution-sense.ts`) — per call, never cached (ruling item 3: a cache converts a transient contamination into a whole-session one) — consumed by:

- **search-log rows** (`deriveSearchLogAttribution`, the seam named in the ruling): every new row carries `agent_source_provenance: 'env' | 'absent'` (the ruled two-value enum; `'env'` iff `agent_source` was stamped — a writer-guaranteed biconditional).
- **selection-manifest rows** (`buildSelectionManifestRow`, the #467 join surface): same fields, same sense. Self-minted rows (session-start hooks passing their own UUID) never probe the pointer — their minting seat IS the emitting process's env read, and probing the possibly-foreign pointer would check the wrong session. Pointer rows hand the sense the same id the row joins on, so the probe cannot race a concurrent rotation.

**Fail-closed at the stamp seam (ruling item 2).** The minting seat derives from the pointer session's `session_start` row in `events.ndjson` (written by the #2625 mint block — no new producer needed; backward scan, substring prefilter, malformed lines skipped per the `parseSearchLog` precedent). On env-seat ≠ minting-seat: the row stamps NEITHER candidate and carries `attribution_conflict: { env_seat, minting_seat, pointer_session_id }`. A wrong-seat row corrupts two denominators; an unattributed row only under-counts. Fail-closed scopes to the attribution, never the operation — with every probe failing, the query still returns and the row still writes.

**Only positive disagreement strips attribution.** No pointer, TTL-expired pointer, no mint row, unseated mint row, absent ledger: the env stamp stands. A non-benign probe failure (EACCES class) keeps the stamp and names itself on `attribution_probe_error` (manifest side additionally routes it to `warnings[]`) — degradation on the accounting channel, never silent.

**Doctor arm (d) — the registry scope-sense (ruling item 4, judged at spec time: include).** `totem doctor`'s Seat Identity row gains a win32-only probe of the two persistent registry scopes (`HKCU\Environment`, `HKLM\...\Session Manager\Environment`) via `safeExec('reg', ...)`. Presence at either scope IS the finding (`env-ambient-scope`) — the #2629 root-cause misconfiguration, sensed at diagnostic time, never the stamp hot path. `reg` exit 1 is the healthy silent state; a real probe failure is disclosed as `registry-scope-unreadable`. Sense-only, `gateExempt`, no `fail`.

**Migration.** Both schema extensions are additive-optional (the ADR-078 `agent_source` amendment shape): pre-sensor rows parse under the new schemas. Named compat bound: `SelectionManifestRowSchema` is `.strict()`, so readers at ≤1.115.0 skip rows written by the new emitter — bounded by the planned single cohort bump (this ships inside the 1.116.0 batch; #467 run reads happen at run-time CLI versions ≥ sensor).

**Deliberately out of scope** (design doc § Scope): QBD ledger events (read-side correlation already fail-closes on seat mismatch; re-stamping would shift compliance-metric denominators), the rendered hook templates (self-minted rows — conflict structurally impossible; no parity-lock churn), `mcp_call` events (carry no `agent_source` today; the wrong-session half for sessionless seats is #2628's fix + the #2204 residual), and any consumer-side exclusion filter (bound in strategy's #467 pre-registration Revision 2 — building it here would duplicate a consumer-lane contract).

## Verification

- Design doc: `.totem/specs/2629.md` § Implementation Design (the sections above it are the pre-ruling LLM synthesis; the design section names the three divergences and supersedes them).
- Suites: core 3734 passed, mcp 230 passed, cli 4004 passed (144/12/162 files). New coverage: 20 attribution-sense tests, 6 manifest-provenance tests, 6 search-log conflict/displacement tests, 6 doctor arm (d) tests. Pre-existing doctor tests pinned to `platform: LINUX` so the suite stays hermetic on win32 hosts (no real `reg` spawns; a host with a user-scope var would otherwise fail unrelated assertions).
- Gates: full workspace build, repo-wide `format:check`, ESLint, `totem lint` PASS (0 errors; 124 advisory warnings — frozen-lesson regex classes over the generated spec's box-drawing characters, changeset prose, and test fixtures).
- Advisory review fan: 0 critical / 1 warn / 2 info — all three declined against source and re-armed to the falsification leg for independent re-examination (declines recorded in the leg deposit).
- Falsification-leg arc (two rounds, per the § Review legs contract): **Round 1** (full branch) — 4 MATERIAL + 5 MINOR, all folded: template rows now stamp provenance (the ruled "every row" bound the disclosure half even where the conflict probe is structurally inapplicable; tracked `.gemini` hook regenerated under its byte-parity lock), arm (d) classifies a blank-valued registry key (this machine's live post-remediation state) as inert, the manifest probe warning now reaches `onWarn` (it previously lived only on the persisted row), and the conflict test no longer claims the b8d7aa9b specimen shape — that mint row was unseated, so the sensor would have stamped the ambient seat with no conflict; the named limit (the ambient-agreement class is invisible to the predicate by ruled design; arm (d) + the launch-shell protocol are its sensors) is disclosed in the module doc, design doc, and changeset. My three advisory declines all survived its adversarial re-check. **Round 2** (re-arm, fold-scoped) — 0 BLOCKING, 1 MATERIAL + 6 MINOR, all folded: the template universal is now LOCKED in the drift-coupling suite (four provenance assertions — deleting a template stamp goes red instead of riding the optional-field parse), the arm (d) blank-skip is gated on the remainder (a CR/LF-leading value stays a finding; leg-specified remedy applied verbatim — named exempt class for re-arm purposes), the seam-level test now fails BOTH probes and asserts errno + recording, the warnings re-adopt is pinned against drop/duplication, and three doc claims were re-attributed to their actual authority. Both deposits are in the session record.

The three untracked postmerge extraction lessons from #2632 ride per convention (screened last session).

Closes #2629


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
Round 1 dispositions — all four findings verified against source before actioning. Fixes landed in `dbeb032b`.

@coderabbitai — both actionable items **fixed** in `dbeb032b`:

1. **search-log.test.ts test-order coupling — fixed as suggested.** Added the `_clearLogDirForTesting()` test seam (your suggested shape, following the repo's `_clearBashResolverCacheForTesting` naming convention) with an `afterEach` reset; the positional tests-sit-last comment is gone, and no later test can probe a torn-down tmpdir through the stale module path.
2. **lesson-623fab5e.md over-generalized wording — fixed, committable suggestion applied verbatim.** The lesson now scopes the `[bot]`-suffix requirement to Greptile/GHCQ and names the intentional CodeRabbit/GCA substring rules as preserved.

@greptileai — both findings **declined** against source; each is falsifiable as stated:

1. **P1 "Conflict probe checks wrong session" — declined.** The pointer is the ruled comparand: the operator ruling on mmnto-ai/totem#2629 (2026-08-12 comment) names "the session pointer's minting seat" verbatim as the conflict test. `TOTEM_SESSION_ID` is the reserved A.3.c pass-through, documented "deliberately inert in the current repo-wide windowing" — nothing in the cohort sets it (verified by caller sweep on this PR's falsification legs). In the hypothetical divergence, the failure direction is the ruled-safe one: attribution is stripped (under-count), never wrong-seat retention, and the `attribution_conflict.pointer_session_id` field exists precisely so the probed session is post-hoc distinguishable from the row's own `session_id`. On the M1 join surface (selection manifests) the probe checks the SAME id the row joins on by construction (`pointerSessionId` handoff). Falsifier: when the A.3.c correlation pass makes `TOTEM_SESSION_ID` live, the sense gains that comparand in the same change.
2. **P2 "Warn policy lives in spec" — declined.** `.totem/specs/2629.md` invariant 9 does not mint the warn-only/`gateExempt` policy — it restates the shipped Seat Identity row contract (mmnto-ai/totem#2511, merged in #2627: the row "carries `gateExempt: true` and never emits `fail`" per its own module doc), whose governing authority is Tenet 13 in the doctrine layer. The spec is a per-issue implementation design record; the invariant is a pointer to the existing contract the arm extends, not a new policy surface.

@gemini-code-assist — review acknowledged: zero findings on this diff; nothing owed.

local-lane: befdd69e round=0 settled=false lanes=2/2

Issue #2516, opened 2026-07-29
fix(mail): qualify the verdict line when the scan is degraded — a broken scan must never read like a clean inbox


satur8d:
## The class

The verdict-line-vs-scope-line trap, third specimen this week: a summary sentence that stays reassuring while the fine print above it discloses degradation. When `totem mail`'s scan is incomplete (workspace unresolvable, a repo's orchestration dir unreadable), the render prints `Warning: workspace scan failed: …` / `Warning: orchestration scan failed (…)` lines — and then still ends with the unqualified **"No unread mail addressed to …"**. The warnings channel makes the output *distinguishable* (ADR-115 §2 satisfied — this is a hardening, not a §2 violation), but the verdict sentence is the line humans actually read and act on.

Sibling specimens of the class, same week: `totem lint`'s `PASS - 387 rules` over a scope that never contained the pack (#2505 vacuous-green, disclosed on #2512), and the A3 receipt rendering clean while validation ran downgraded (fixed in #2512 by pinning `astParseMode` + disclosing posture on the rendered page).

## The fix

When `warnings.length > 0`, qualify the verdict line — e.g. `No unread mail found — but the scan was INCOMPLETE (N warning(s) above)`, and symmetrically qualify a non-empty render (`M unread — scan INCOMPLETE`). Exact wording free; the contract: **a degraded scan must never produce a verdict sentence identical to a clean scan's.**

## Acceptance

1. Degraded-scan render's verdict line differs from the clean-scan verdict line (both the empty-inbox and non-empty cases).
2. The E4 pack's fault-3 tests (#2515) gain/adjust an assertion pinning the qualified verdict once wording lands.
3. Clean scans render byte-identical to today (no noise added to the healthy path).

## Provenance

Flagged by strategy-kimi in the #2515 E4 draft (left deliberately unasserted there — tests-only PR, behavior change out of scope), endorsed by totem-claude at review, operator-greenlit 2026-07-29. Related: #2509 (accounting line for own-broadcast suppression — same render surface; the two fixes compose), #2462 (class parent).

satur8d:
**Wording settled** (the claude-0191 `owed:` line; operator-greenlit batch 2026-07-31).

Contract restated: a degraded scan must never produce a verdict sentence identical to a clean scan's — and the discriminator should be a single greppable token, so the E4 assertion pins an invariant rather than fragile prose.

Ruled wording (`N` = collected scan-warning count, `M` = unread count):

[...]

Issue #2511, opened 2026-07-28
feat(cli): seat-lifecycle verbs — totem seat add/suspend/remove with derivable seat-state, birth checklist, identity-binding tier


satur8d:
## Charter

Operator-greenlit 2026-07-28 (ruling session recorded on mmnto-ai/totem-strategy#974; charter dispatched by strategy-claude 2026-07-28T2010Z, origin: operator's "should we set up a skill or test harness for adding/removing agents so we can dogfood things?"). Strategy-side shape ruling: **a totem CLI mechanism with induced tests, not a per-vendor skill** — seat lifecycle is vendor-neutral (Tenet 16 corollary: config names roles/seats, never models); a cohort skill may later wrap the ceremony.

Build shape is totem-lane. Working position: **`totem seat add | suspend | remove` verbs** under the ADR-094 entity-namespace grammar (#1724), rather than an `init` extension — lifecycle events are first-class operations, not re-initialization. Open to inversion at spec time if the init plumbing turns out to own too much of the birth checklist already.

## v1 scope

1. **`seat add` wires the full birth checklist, not just dirs:**
   - orchestration dirs — reuse the #2141 dirs-union invariant that already carries registration; do not rebuild it;
   - `config.json` `host_agents` where the repo carries one — **REPLACE-semantics trap**: strategy's config requires ALL seats listed, so a naive append/overwrite silently drops seats;
   - **the poll path wherever the vendor has a hook surface** — cohort-roles §6.6 (landed via mmnto-ai/totem-strategy#981) shows poll asymmetry is born at seat-add time and biases round coverage (four of six round legs landed only after the operator manually woke seats). Wiring the poll at birth kills the class. Current asymmetry: strategy-gemini auto-polls; totem-gemini's hook has no mail poll; codex/kimi/agy have no hook surface;
   - **emit the residual propagation checklist** for surfaces the verb cannot reach (`totem-status:orchestration.go` agentHomeMap, signoff SKILL.md seat table) instead of leaving them to memory — both are still trailing from the kimi/status-claude seatings.

2. **Derivable seat-state (`active | suspended | retired`)**, consulted by `totem mail` and round tooling. Live specimen (2026-07-28): the operator suspended strategy-agy by word ("having issues, move on without it") and the exclusion existed nowhere derivable — the round synthesis had to hand-record it. A suspended seat mechanically drops from poll unions and round-coverage denominators. Composes with the multi-seat mail ergonomics thread (#2204, #2017).

[...]

Issue #2312, opened 2026-07-06
fix(cli): totem mail from a subdirectory resolves the wrong workspace + unresolved self still renders a clean-inbox verdict (false-clean, Tenet 4)


satur8d:
## Defect

`totem mail` run from a **subdirectory** of a cohort repo silently resolves the wrong workspace and renders a clean-inbox verdict — a **false-clean** (the read-side sibling of the false-unread class the A2.2 gate exists to prevent).

## Live repro (2026-07-06, strategy seat)

From the repo root — correct:

```
[Mail] Workspace: D:\Dev
[Mail] Self agents: strategy-claude, strategy-gemini, strategy-codex, strategy-agy (source: config)
[Mail] 12 unread: …
```

From a subdirectory (`.totem/orchestration/strategy-claude/processed/`) — same repo, same moment, 12 unread waiting:

```
[Mail] Workspace: D:\Dev\totem-strategy\.totem\orchestration\strategy-claude
[Mail] Self agents: (none) (source: none)
[Mail] Warning: no SELF_AGENT resolved (…)
[Mail] No unread mail addressed to (none) or broadcast.
```

## The two halves

1. **Workspace derivation is cwd-fragile.** The workspace default appears to be parent-of-cwd rather than parent-of-repo-root. Any `cd` into a subdir (a shell that persists cwd between commands — exactly how agent harnesses run) skews the whole scan. Expected: derive the repo root first (walk up to `.totem/` or `git rev-parse --show-toplevel`), then take its parent — cwd depth becomes irrelevant.
2. **An unresolved self still renders a clean-inbox verdict.** The warning fires, but the final line — `No unread mail addressed to (none) or broadcast` — reads as a clean poll result. A poll that resolved **zero** self agents cannot assert an empty inbox; per Tenet 4 that arm should render as NOT-DERIVED (and plausibly exit non-zero), not as "no unread". An agent (or a script grepping the summary) seals on a false clean. Note `ecl-gc --compact` is already hardened here (unresolvable self **throws**, exit-2 class) — the plain poll should not be softer than its own GC sibling.

## Why now

Caught live only via the re-derive-before-asserting discipline (the operator's seat noticed "Workspace:" looked wrong). With the model handover to Opus-primary (2026-07-07), tool-layer fail-loud beats agent-layer vigilance — this is exactly the class the Axiom Mandate (Tenet 15) says to push into the substrate.

[...]

PR #2639, opened 2026-08-14
feat(mail): multi-seat identity gate + --as selector + --all-seats union view (#2204)

satur8d:
## What this is

The deterministic floor for the multi-seat identity trap (mmnto-ai/totem#2204, expedited on strategy-claude's 2026-08-13 dispatch after the BLIND-round contamination specimen): an **identity-less `totem mail` poll in a repo that resolves more than one seat no longer serves the union**. The listing itself — subjects included — is the exposure surface, so the gate withholds directed mail as a **count**, serves broadcast mail only, warns naming the union and its source, and exits **2** (the #2312 NOT-DERIVED family). The resolver and the doctrine-mandated `host_agents` config are untouched: the serve default was the defect, never the config (lesson-08a3fa50).

## The contract

- **Gate predicate:** `selfAgents.length > 1 && source !== 'env' && !--all-seats`. Single-seat resolutions from any source, multi-seat `source: 'env'` (an explicit operator declaration), and the env-bound `ecl-gc` A2.1/A2.4 polls are exempt — the last structurally, since they force `TOTEM_SELF_AGENT` (§ A2.3).
- **`--as <seat>`** (the issue's `--self`/`--to` ask, one canonical name): validated against the repo's *structural* union with ambient env stripped (flag-over-env precedence), falling back to the env-declared list when the structural union is empty (worktree shape). A foreign seat is refused with the resolver's own diagnostics appended — a foreign-anchored poll answers "ever addressed," never "unread" (lesson-2923d5e0). Validated seats env-inject through the existing resolver, so the banner truthfully reports `source: env` and core gains no new source value (protects the #2634 provenance sensor).
- **`--all-seats`**: the prior union view, available by name (repo dashboard; exit 0 + a notice that per-seat unread accounting does not apply).
- **Accounting unchanged:** processed-mark subtraction, broadcast per-seat floors, and lifecycle denominators all still compute over the full resolved union — the gate withholds the *listing*, never the accounting. The gate line rides the `warnings` channel deliberately: any future compaction-style caller that forgets env-binding reds the A2.2 completeness gate and fails toward retain-everything, never toward the A2.1 false-unread bomb.
- **Exposure discipline:** withheld mail surfaces as a count only, on every channel (text, JSON, warnings, notices). Under the gate the truncation warning names *no* part of the unopened tail (filename tokens are untrusted for delivery truth — codex F2), and the #2511 lifecycle notices use gate-aware wording ("WITHHELD", never "surfacing"). Named limit, stated in the design doc: a dispatch that fails read/parse is warned by filename before its `to:` is knowable — an integrity signal that cannot be seat-scoped.
- `totem mail --help` and the wiki cli-reference now document the gate, both flags, exit 2, and the single-writer consume path (`mail mark`, ADR-106 § A1.3).

## Review provenance

Design doc at `.totem/specs/2204.md` § Implementation Design (three named divergences from the body-derived synthesis; the on-issue sharpened spec of 2026-08-13 governs). Two-leg falsification arc, both rounds folded and locked:

- **Round 1 (full branch):** 2 MATERIAL — the truncation warning leaked withheld directed basenames on text+JSON; the gated verdict arm dropped the #2516 INCOMPLETE discriminator — plus `--as` diagnostics/fallback/label fixes and lifecycle-sense preservation. All cured in `d8f5ead7`.
- **Re-arm (fold-scoped):** 1 NEW-DEFECT — lifecycle notices asserting "mail still surfaces" over a withheld render — plus the broadcast-token naming residue (resolved conservatively: an unopened tail is never named) and lock-tests for every cure that lacked one. All cured in `0fb3a544`. The re-arm's could-not-break table held every load-bearing claim (exemptions, denominators, retain-direction, JSON channel, env-fallback reachability).

Existing suite migrated to explicit-identity fixtures (assertion lines untouched — verified by diff); 25 new tests pin the gate invariants. Full cli suite **4042 passed**, workspace build + `tsc --noEmit` + prettier + ESLint + full-branch `totem lint` all clean.

## Behavior change (changeset: minor, `@mmnto/cli`)

Identity-less polls in multi-seat repos change behavior by design — broadcast-only serve + exit 2. Per-shell `TOTEM_SELF_AGENT` (the ruled cohort norm; hook paths already comply) or `--as` restores full service; `--all-seats` restores the union view by name. A cohort broadcast announcing this should ride the release — it is also the retirement trigger for strategy's interim S0 instruction-layer check (their PR mmnto-ai/totem-strategy#1059).

## Deliberate inclusions / deferrals

- Three postmerge extraction lessons from #2634 ride the first commit (standing carryforward: untracked lessons ride the next PR; verified WARN-safe under the compile freeze).
- The distributed **signon skill's** bare-`totem mail` instruction is *not* updated here — it is managed content under the doctor parity lock, and the seat-scoping clause belongs to the queued crown-fold task (strategy's DISTRIBUTED_CLAUDE_SKILLS ask), which will fold both in one parity event.
- `mail reply --as` is parent-swallowed by Commander (spec named-limit row); reply keeps `--from`.


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 disposition

@gemini-code-assist @coderabbitai @greptileai — one consolidated disposition for the 2026-08-14 round on `0fb3a544`. All five findings were verified against source before action; fixes landed in `9ecadc74` with full gates green (workspace build, 4042 cli tests, full-branch totem lint 0 errors, prettier, ESLint 0 errors).

| GCA | high: `--as` validation throws `MAIL_SEND_FAILED` — semantically wrong for a poll option | accept | Fixed in `9ecadc74` — guard inlined on core's `isPathSafeAgentId` at `packages/cli/src/commands/mail.ts:1258`, throwing `CONFIG_INVALID` consistent with the sibling `--as` refusals in the same function; JSON-escape of the echoed token retained. The lock test now pins the error code, not only the message (`mail.test.ts:3378`). |
| CR | minor: cli-reference should state `--as` applies to the poll only | accept | Fixed in `9ecadc74` — poll-only sentence added at `docs/wiki/cli-reference.md:460` per the suggestion. Confirmed stronger than flagged: the spec's named-limit row already claimed this page documents poll-only while the page did not — the sentence discharges that claim. |
| CR | trivial: `identity-ambiguous poll:` literal couples the warning producer and the INCOMPLETE discriminator with no compile-time link | accept | Fixed in `9ecadc74` — `GATE_WARNING_PREFIX` (`mail.ts:604`) now binds the producer (`mail.ts:1091`) and the discriminator filter (`mail.ts:1181`). The three other occurrences of the phrase are mid-string prose that correctly never matched the `startsWith` filter — left as prose. |
| CR | minor: gated-poll test asserts a weak range instead of the exact resolved union | accept | Fixed in `9ecadc74` — `mail.test.ts:3017` asserts `['totem-claude', 'totem-gemini']` and `source: 'map'` exactly (verified against the resolver's pure-map path; matches the exact-union idiom already used at `mail.test.ts:610` and `1206`). |
| GT | P2: implementation contracts in a pre-acceptance spec (`.totem/specs/2204.md`) create competing governance authority | decline-substantive | `.totem/specs/**` files are point-in-time planning artifacts riding the PR for context, not living governance (repo convention; spec drift is deliberately not flagged post hoc, and the spec is not consulted post-merge). The enforceable surface is code + tests + cli-reference — all in this diff; durable mail semantics live in ADR-106. Noting this finding is the sole basis of the 4/5 confidence score, and the same summary grades the PR safe to merge from a runtime perspective. |

local-lane: no verdict artifact recorded for the current lineage (`totem review --covariate`, sensor exit 0).

Issue #2551, opened 2026-08-02
fix(hooks): pre-push agent branch blocks on bare 'totem review' — wiring contradicts 1.108.0's own advisory self-description (admission-phase frame)


satur8d:
Reported by lc-claude (2026-08-02T0435Z dispatch, verified against installed 1.108.0 source + the live hook during their four-round fold ladder); routed by strategy-claude with a doctrine read attached. LC's operator gates any LC config change either way — this filing is about the **shipped reference wiring**.

## The contradiction

1.108.0 just shipped the honest self-description (mmnto-ai/totem#2536): `totem review` describes itself as *supplementary AI review lanes — advisory sensors, not a merge gate*, across run-start, `--help`, docs, and init-template text.

The shipped **pre-push hook's agent-detected branch still wires it as a blocking gate**: bare `totem review` with `|| exit 1`. Bare `totem review` = embedding + retrieval + a metered LLM completion — a **non-deterministic, network-dependent, metered blocking call inside a push gate**. The text now says advisory; the wiring still actuates. (Dormant on LC's seat today — agent-detection vars unset — but live for any environment that sets them.)

## Why it's a defect and not a preference (tenet grounding)

- **Tenet 13 (sensors, not actuators):** gate logic belongs to the user's hooks/CI wiring — but this IS the reference wiring totem ships, and it wires a probabilistic sensor as a blocking actuator.
- **Tenet 6 (local-first):** the deterministic path must never acquire a network dependency; a push gate is the deterministic path, and this branch couples push availability to API availability + quota (`|| exit 1` turns an API outage into a blocked push).
- **Tenet 12 (reference implementations):** the default deterministic path (`lint`) must never be blocked by an opt-in probabilistic path (`review`) — this branch is that inversion, one hop removed.

## Suggested shape (the mmnto-ai/totem#2473 admission-phase frame)

Deterministic halves keep blocking semantics (lint, hash gates). The LLM half of `review` in any hook context: run-if-armed, **report-only, exit 0 on completion failure/abstention** — its findings land in the verdict artifact for the review-loop skill to absorb, never in the push gate's exit code. That makes the wiring match the 1.108.0 text and closes the class rather than one hook.

## Adjacencies (same lanes-rework surface)

- mmnto-ai/totem#2473 — exit-contract work; this is the hook-side instance of the same admission-phase question.
[...]

Issue #2525, opened 2026-07-30
review: LLM lanes rework — 50k-char window truncation, lane-billing fragility, markdown skip (charter ask; interim = cohort review-leg contract)


satur8d:
## The need (operator-directed 2026-07-30: stop waiting on this class)

`totem review`'s LLM lanes have a gap set that, composed, produced an effectively unreviewed merge candidate while every individual gate behaved "as designed" — mmnto-ai/totem#2524: a 117,354-char coupled instrumentation diff where the anthropic lane died on a provider 400 (credit balance), the gemini lane returned 0 findings **over content truncated to the 50,000-char LLM window**, and the fan honestly reported `settled=false (1/2 lanes)` — with nothing downstream treating that as a stop. The deterministic gates (lint, tests) were full-width and strong; the LLM half was void.

The gap set, individually:

1. **50,000-char LLM window** — large diffs are truncated for the LLM lanes; a lane verdict over partial content reads like a verdict over the diff. This is the class the operator has been expecting a rework for.
2. **Lane billing/provider fragility** — a lane dying on an environment condition (dead API credit balance) degrades the review to whatever lane survives, silently from the merge-flow's perspective.
3. **Markdown silent-skip** — mmnto-ai/totem#2466, already tracked; listed here because the composite is the point: doc-heavy governance repos currently get near-zero LLM review from the reference implementation.

## Interim, already codified (cohort dev-ops, not a product claim)

The cohort's review of record is now **agent falsification legs** — Opus legs on Claude seats + own-repo cross-vendor legs — per `mmnto-ai/totem-strategy:doctrine/cohort-overlay.md` §5 / `doctrine/model-tiering.md` § Review legs (shipping as `@mmnto/strategy-doctrine` 0.1.25). `totem review`'s LLM lanes are supplementary until the gap set clears. First-day yield of the leg pattern: 5 record-defects + 9 nits across five already-merged doc PRs, including two defects in the repair commits themselves.

## Design directions (yours to rule — filed as need + evidence, not a spec)

- **(a) Chunked / multi-pass diff review** — break the window; verdicts compose over full content.
- **(b) Agent-leg orchestration as a lane class** — productize the leg pattern: the review fan spawns a repo-context agent (no char window, reads the working tree, plan-covered on operator boxes) instead of a raw API call. Tenet 12: the reference implementation should not be structurally weaker than what its own cohort does by hand.
[...]

Issue #2473, opened 2026-07-20
fix(review): deterministic skips exit 0 through the live pre-push gate — the locally-live half of #2466 (exit contract, admission-phase model)


satur8d:
Follow-up to #2466. #2471 closes the **stamp** half (dormant in totem, live for downstream consumers on old hooks). This tracks the half that is **live in this repo**.

## The live defect

`tools/pre-push` runs:

```sh
$TOTEM_CMD review || exit 1
```

So a prose-only diff hits a deterministic skip path in `shieldCommand`, returns, **exits 0**, and the push proceeds. A non-review passes the push gate through its exit code.

This is #2466's original framing. It was retitled toward the stamp on my analysis, which was based on the premise that a PreToolUse gate consumes `.reviewed-content-hash`. That gate was decommissioned in #2180 (2026-06-15) and nothing in this repo reads the stamp — so for totem, the original framing was the correct one. #2472 tracks the stale doc that produced the error.

## Design to build against (totem-codex)

Do **not** model a deterministic skip as a lane outcome. Two phases:

1. **Admission / applicability** → either `not-applicable` with a deterministic reason (`no-diff`, `all-non-code`, `filtered-empty`, `all-generated`), or an admitted review payload.
2. **Fan execution** → admitted lanes terminate `completed | abstained | failed`.

A deterministic skip is therefore **not** a fourth lane status and **not** an `InvokeFailureKind`. Slice B's taxonomy (#2452) describes failures *after* an invocation was attempted; mapping a pre-invocation admission result into it would erase the phase boundary and falsely imply provider work occurred.

Consequences worth carrying:

- **#2459 is downstream of admission** and can proceed independently for attempted lanes. Its mapping should explicitly exclude admission skips. Coordination is only needed if both surfaces share a top-level outcome envelope.
- "Exit 0 for `not-applicable` can be acceptable for an interactive command only when the machine/human outcome says so explicitly. Exit 0 by itself remains the silent-positive ambiguity identified in #2466."
- Prose being outside a push gate is a **coverage** decision, not evidence that prose was reviewed. Authorization and review coverage must stay separate claims — this matters most in `totem-strategy`.

## Constraints surfaced by the cohort round

Recorded here so they are not rediscovered:

[...]

PR #2641, opened 2026-08-14
fix(review): admission-phase disposition contract — not-applicable records, --gate mapping, no-diff stamp removal (#2473)

satur8d:
Implements the 2026-08-12 operator ruling on mmnto-ai/totem#2473 (recorded on-issue; rationale in strategy-claude's 2026-08-12T1823Z round reply): the admission-phase model as filed, the ADR-109 disposition shape for the exit contract, and the #2551 fold (LLM findings report-only in hook context). Design doc with the full review-round record: `.totem/specs/2473.md` § Implementation Design.

## The live defect

`tools/pre-push` runs `$TOTEM_CMD review || exit 1`, and a prose-only diff hit a deterministic skip that exited 0 — a non-review passing the push gate through its exit code. Worse, the clean-tree arm actively STAMPED `.reviewed-content-hash` ("trivial pass"), minting push authorization for a tree no reviewer saw.

## What ships

- **Admission phase**: one closed `evaluateAdmission` decides applicability before lane validation or engine boot. A deterministic skip (`no-diff` / `all-non-code` / `filtered-empty` / `all-generated`) is an admission verdict — never a lane outcome, never an `InvokeFailureKind`. The fan is unreachable on a skip (control-flow, boot-spy test-locked; type barrier at the `ExecutionPayload` selection seam).
- **Disposition record**: content-addressed `AdmissionRecord` in `.totem/artifacts/admissions/` binding the exact observation — resolved diff scope (the resolver now returns a discriminated `DiffForReviewEmpty` carrying the scope it resolved, never a bare `null`), the pre-filter diff-bytes hash, and a selection-policy fingerprint whose classifier tables are bound **mechanically** (a table edit re-keys every address; no hand-bumped version string). Identity excludes `createdAt` and `schemaVersion` — a 1.x bump never orphans records; an older CLI meeting a newer-major record gets a named upgrade error, never a corruption message.
- **One calm line**: `not-applicable (<reason>): … — no lane ran. Nothing was examined, so the reviewed-content-hash was NOT stamped…` — info-level (level + one-line shape test-locked), with the record address. Fail-soft persist (disclosure, not a gate — Tenet 13, annotated).
- **`--gate`**: the wiring's declared disposition→exit mapping (ADR-109's wrapper role). Known skips and completed rounds exit 0 (findings report-only in hook context), hard failures block, unknown dispositions fail CLOSED. Contradictory with `--fail-on`; validated before any side effect. Both hook surfaces regenerate to `review --gate` behind a `--help` probe with a visible compat fallback (byte-parity test-locked; POSIX behavioral matrix executed under real `sh`).
- **`--covariate` is admission-aware**: exact-identity record resolution, deterministic (never wall-clock arbitration), loud no-current-record sensor, still write-free. The admission line is a **declared format v1.1 amendment** of the versioned `local-lane` covariate contract — amended in the skill template constants and all four distributed twins (doctor parity green), with the `init.test.ts` format pin updated WITH the amendment.

## Behavior changes (changeset: minor `@mmnto/cli`, each named loudly there)

1. The clean-tree stamp is REMOVED (worst case for exit-code consumers: a redundant re-run at exit 0; the legacy PreToolUse-gate case and its `totem hook install` remedy are named).
2. Bare `totem review` keeps sensor exit 0 on every known skip — no stale `|| exit 1` hook starts blocking docs-only pushes, on any upgrade-order path.
3. Covariate contract v1.1 (additive admission form; consumers discriminate on the literal token `not-applicable`).
4. An incremental docs-only delta now falls back to a full-scope review instead of skipping; admission records accrete in the gitignored artifacts dir (reaper follow-up named).

## Review arc (all rounds recorded in the spec)

Design round: strategy-claude PRESENT (exit contract confirmed as ruled; premise correction accepted) + totem-codex REQUEST CHANGES → CONCUR, design approved with four binding conformance notes — all implemented (resolved-empty scope binding, fan config downstream of admission, `ExecutionPayload` selection, Q3 rationale). Then two falsification-leg rounds pre-PR: full-branch (2 MATERIAL / 7 MINOR / 6 NIT — all folded, none declined) and a fold-scoped re-arm (1 MATERIAL / 5 MINOR / 4 NIT — all folded). Both legs' refuted-in-passing sweeps held the exit contract, stamp-removal completeness, phase ordering, hook seam, and covariate read-only contract under attack.

## Deliberate inclusions / deferrals

- Close-of-#2473 is not encoded here (no close atom) — the operator's call at merge; the ruling also covers #2551's shape and #2466's exit-residue.
- The instruction-prose coverage question (`AGENTS.md`/skills as control plane) stays in Prop 297 / #2525's lane per the ruling — authorization ≠ coverage; this PR states the coverage decision, it does not absorb it.
- Named follow-ups riding the spec record: the admission-store reaper; the pre-existing `sourceExtensions` over-binding question.

Gates at head: core 3746 + cli 4068 tests green (11 skips: 4 ticket-referenced POSIX-only + 7 pre-existing), full-branch `totem lint` 0 errors, doctor `--strict` parity green, prettier + ESLint clean.


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 disposition

@gemini-code-assist @coderabbitai — one consolidated disposition for the 2026-08-14 round on `3554ff9f`. All eight findings were verified against source before action; fixes landed in `658a75d9` with full gates green (workspace build, core 3747 + cli 4068 tests, full-branch totem lint 0 errors, doctor parity, prettier, ESLint). Greptile reviewed clean (5/5, zero findings) and is deliberately not tagged — nothing of its is addressed this round.

| CR | major: unconstrained `createdAt` can inject line breaks into the stdout `local-lane:` transport | accept | Fixed in `658a75d9` — `createdAt` is now `z.string().datetime()` (the writer emits `toISOString()`, fully compatible); validate-on-write AND the verified load both reject a tampered value, regression-locked with a newline-injection test (`admission.test.ts`). |
| CR | major: the classifier-tables test recomputes both sides from one constant — cannot observe the re-key property | accept | Fixed in `658a75d9` — pinned literal digest `7641cc478a33` + exact entry count 80 (`shield-admission.test.ts`); a table edit now fails the test stating the new key, and the retained cross-check keeps the pin coupled to the live constant. |
| CR | minor: canonicalizer keeps duplicates, so a duplicated config entry re-keys an equivalent policy | accept | Fixed in `658a75d9` — dedupe-before-sort set semantics in `computeProjectionPolicyHash`; duplicate-insensitivity added to the canonicalizer test. The pinned canonicalizer digest is unchanged (the frozen fixture carries no duplicates). |
| CR | minor: MD040 — the covariate fence lacks a language | accept | Fixed in `658a75d9` — `text` added to the fence in the init-template constant AND both review-loop twins (the suggestion named only the two copies; managed-content parity requires the constant too — doctor parity green). |
| CR | trivial: bind the known-reason list to the core union | accept | Fixed in `658a75d9`, strengthened beyond the suggestion: the test iterates the core-exported `NOT_APPLICABLE_REASONS` set rather than a typed literal (a typed subset would still typecheck) — a new union member is exercised automatically and fails CLOSED under `--gate`, forcing the mapping to grow with the union. |
| GCA | high: `raw.schemaVersion` on a narrowed `unknown` "will cause a compilation error" | decline-hallucination | The `'schemaVersion' in raw` check narrows via TypeScript's `in`-operator narrowing; the pattern is byte-identical to the shipped `readMajor` in `verdict.ts`, and the file has compiled green through `tsc` repeatedly on this branch (CI Build & Lint across three platforms concurs). No error exists to fix. |
| CR | minor: classifier lookup-precedence / fail-closed rules are outside the mechanical hash | decline-substantive | Deliberate, documented boundary: TABLES are mechanically bound; algorithm SHAPE is versioned by the hand-bumped `REVIEW_CLASSIFIER_ID_BASE` — exactly what its docblock states, and the #2473 re-arm falsification leg verified that scoping as accurate. Mechanizing algorithm semantics would mean hashing source text, which re-keys addresses on comment edits — noisier, not sounder. |
| CR | trivial/low-value: rename `resolveNotApplicableExit` to a guard name | decline-stylistic | The `resolve*ExitCode` family is this codebase's exit-contract idiom (`resolveMailExitCode`, `resolveEclGcExitCode`), and the literal-`0` return documents the ruled no-nonzero mapping in the signature; the finding itself notes the current shape works and both arms are test-pinned. |

local-lane: no verdict artifact recorded for the current lineage — the current diff is ADMITTED (code), so resolution went to the verdict store, and no fan round has run on this branch (`totem review --covariate`, sensor exit 0; the admission-aware resolution shipped in this very PR is what performed the lookup).

PR #2660, opened 2026-08-21
feat(core): V1 record-grammar parser + conformance suite (Prop 310 slice 1)

satur8d:
## What

Slice 1 of the Prop 310 V1 record-grammar build (the R14 critical path): the closed-key YAML record parser and its § Design 14 conformance suite. Pure new code beside the frozen legacy path — land-alone-able by design (slice 1 of the approved 3-slice stack; slice 2 = compiled homes + lowering + runtime eval; slice 3 = intake front-end + envelope derivation + manifest attestation).

- **`packages/core/src/spine/rule-record.ts`** (854 lines): `RuleRecordSchema` (closed keys at every record depth, no field defaults anywhere), `parseRuleRecord` (pure, deterministic, total — every failure is a typed error naming file + dot-form key path), the § Design 7 glob-dialect validator (13 fail-loud rules incl. brace/negation/globstar/regex-syntax/parent- and current-segment/whitespace), the § Design 2 no-silent-skip gates (unknown `schemaVersion` / `target.type` / reserved `requires.scope: block`), the § Design 4 inexpressible-key scan (21 producer-owned/intake-seam keys, cycle-guarded DFS that descends into the opaque payload), `derivedEngine` from `target.type` (§ Design 3 — never author-set), and Amendment-1-item-3 CR-blind per-example-pair hashes (sha256 over the LF-image; raw record values retained byte-verbatim per § Design 10's preimage-material rule).
- **`rule-record.test.ts`** (1,215 lines, **196 tests: 37 positive / 159 negative**): the conformance suite — a negative fixture per banned-silent behavior (silent default / skip / expansion / promotion) and per dialect rule, closed-key negatives at multiple depths, all 21 inexpressible keys via `it.each` with a set-membership guard, a `GLOB_DIALECT_RULES` exhaustiveness guard (every rule must have a fixture; misspelled rules are compile errors), both Prop 310 exemplars (§ Design 4 and § Design 8) parsing **verbatim from the proposal text**, CRLF-vs-LF hash-identity, cyclic-anchor and multi-document carrier negatives, and determinism.
- Barrel exports + a `minor` changeset for `@mmnto/totem`.

[...]

PR #2663, opened 2026-08-21
feat(core): record lowering, compiled homes, runtime eval (Prop 310 slice 2)

satur8d:
## What

Slice 2 of the Prop 310 V1 record-grammar build: **compiled homes, total lowering, and runtime evaluation** — the record path now runs end-to-end from a parsed `.rule.yaml` to live `totem lint` verdicts. Slice 1 (the parser) merged as mmnto-ai/totem#2660; slice 3 (intake front-end + envelope derivation + manifest attestation) follows.

- **`spine/record-lower.ts`** — `compileRuleRecord(parsed, {ruleId, now})`: identity is an INPUT (the ADR-112 producer mints at intake; this function asserts well-formed and never hashes `dslSource`), tree-form payload wrapped into the NapiConfig the engine consumes (operator ruling 2026-08-21: `target.rule` is the rule TREE; the wrapper's `language` derives from the resolved `target.language`, so the record never carries a duplicate language home), engine gates fail loud per § Design 12 (safe-regex2 on `target.pattern` AND `requires.pattern`; ast-grep validated under EXACTLY the declared registry-resolved language, retiring try-each-glob-Lang on the record path), record globs pass VERBATIM (`sanitizeFileGlobs` never runs on records), verbatim carries for the § Design 12 set.
- **`spine/record-runtime.ts` + `rule-engine.ts` + `regex-safety/apply-rules-bounded.ts`** — the § Design 7 two-array scope (`positiveMatch && !excludeMatch`) and § Design 8 `requires` two-pass (fires iff target matches ∧ pattern absent in declared scope; evaluated as part of the match predicate), wired through BOTH regex dispatchers and all three AST branches via one shared predicate; legacy rules route to the byte-identical shipped expression.
- **`compiler-schema.ts`** — additive-optional homes: `excludeGlobs`, `requires`, `examples`, `language`, `verificationShadow` (verbatim, never evaluated — Amendment R4 gate), `recoveryHint`, `curation`. The mined corpus re-validates byte-identical (guarded by test against the real `.totem/compiled-rules.json`).
- **`sys/glob.ts`** — the § Design 7 dialect as a third profile over the shared tokenizer; both shipped profiles byte-unchanged (pinned).
- **Certification seams** — the windtunnel regex path now reads the same post-image content as its AST path; Stage 4 single-homed onto the shared scope predicate, including a dialect-aware fix to its scope-strip (details below).
- **91 slice-2 core tests + 3 CLI staged-mode tests**; both spec exemplars lower and live-evaluate against their own `bad`/`good` through the real engines.

## Review of record

Opus build-leg against the ruled design; **full falsification pass + scoped re-arm** (both with corpus-scale empirical sweeps — the re-pointed legacy non-regression sweep ran 10,670 (rule, path) pairs through the SHIPPING dispatcher with the real RegexEvaluator: 0 divergences, now pinned permanently at 4,850 assertions in CI); a convergence round executing the re-arm's findings. Owner gates (build, core + CLI suites, format, full-branch totem lint, eslint) re-run independently after every round. The convergence round was mechanical execution of reviewed findings plus one documented judgment (the Stage 4 strip cure, reviewed directly by the dispatching seat); no fourth falsification pass was armed — stated here as the discretion it is.

Commit stack (nothing amended): `77a24911` build → `1b46740` falsification cures → `6e79303a` convergence.

## Findings cured along the way (the interesting ones)

1. **The shipping regex path never ran the new runtime.** `totem lint` uses `applyRulesToAdditionsBounded` exclusively — the initial build wired only `applyRulesToAdditions`, so record `excludeGlobs`/`requires` were inert and record globs rode legacy promotion (silent scope WIDENING). Both dispatchers now share the predicate; a CLI staged-mode test pins that the reader actually reaches the bounded regex gate.
2. **Staged fail-open.** `requires: {scope: file}` read the worktree while the violation sat staged — a pre-commit gate passing on deleted context. Both dispatchers now take staged readers (sync param on the sync path, async `readStrategy` on the bounded path); a caller-supplied reader's throw propagates loudly.
3. **Stage 4's scope-strip inverts for records.** The verifier expresses "fires everywhere" as `fileGlobs: undefined` — include-all on the legacy matcher, match-NOTHING on the record dialect — so every record rule verified would have silently classed `untested-against-codebase`. Cure: dialect-aware unscoping (`['**/*']` + exclusions dropped for records; legacy strip byte-for-byte). Sole strip site (swept).
4. **Language⇄glob gap.** `language: typescript` + `fileGlobs: ["**/*.js"]` lowered clean and then either inertly failed per-file or silently FIRED under the wrong grammar. Cured by the lowering floor below.

## Applied rulings — your override window

- **Language⇄glob consistency floor:** for ast-grep targets, every `fileGlobs` entry must end in a registered extension resolving to the declared `language`; extension-less globs reject. Note `.tsx` resolves to the distinct `tsx` grammar, so `.ts`+`.tsx` is TWO records — § Design 6's own "a multi-language ast-grep rule is N records", typed `construct-gap` if the R14 trial finds it bites (it should count this: it doubles the modal React-codebase rule).
- **`requires.scope: line` rejected on ast-grep targets** at lowering: an ast locus is a SPAN whose natural unit is the reserved-unimplemented `block`; the alternative was a silently formatting-dependent start-line reading (verified: same semantics, multi-line fires / one-line silent). `line` stays valid on regex; `file` on both.
- **Deviation D1 (build-leg, confirmed by falsification review):** compiled homes for `recoveryHint` + `curation` — § Design 12's totality clause ("no construct is accepted at parse and dropped at compile") binds them; § Design 1's intake-seam enumeration is closed and contains neither.

## Named forward items

- **Certification-seam readers and Stage 4 are cured in-slice** (were deferral candidates; ruled in since slice 3 routes record rules through exactly those seams).
- **The ast-grep record surface is JS/TS-family-only today** (`registeredExtensions()` = the 8 JS/TS extensions) — the R14 trial cannot express another language until a pack registers one. Corpus cost zero (all 268 ast-grep corpus rules in-family); named for the trial rather than discovered at it.
- **§ Design 7 conformance note owed to the spec** (rides the open strategy thread on the § Design 7 character set): an extension-less glob is dialect-clean but cannot carry an ast-grep record.
- The compiled `language` field is now formally redundant with the globs (derivable) — kept as the Tenet-20 couple form (authored + fail-loud coupling check), a version bump could drop it.
- `requires: {scope: file}` staged reads on the two `applyRulesToAdditions` certification callers use pre-resolved post-image readers; the sync-path param is additive with all four shipped callers unchanged (verified).

### DIFF
--- a/.changeset/record-grammar-lowering-slice2.md
+++ b/.changeset/record-grammar-lowering-slice2.md
@@ -1,0 +1,5 @@
+---
+'@mmnto/totem': minor
+---
+
+Lower parsed Prop 310 rule records into compiled rules, add their compiled homes on `CompiledRuleSchema` (`excludeGlobs`, `requires`, `examples`, `language`, `verificationShadow`, `recoveryHint`, `curation` — all optional), and evaluate the § Design 7 two-array glob scope and § Design 8 `requires` two-pass at lint time (slice 2 of the Prop 310 build).
--- a/packages/cli/src/commands/run-compiled-rules.test.ts
+++ b/packages/cli/src/commands/run-compiled-rules.test.ts
@@ -783,7 +783,7 @@ describe('runCompiledRules', () => {
         .spyOn(totem, 'safeExec')
         .mockImplementation((cmd: string, args?: string[]) => {
           if (!args) return '';
-          if (args[0] === 'ls-files' && args[1] === '--recurse-submodules' && args[2] === '-s') {
+          if (args[0] === 'ls-files' && args[1] === '-s') {
             return '100644 hash 0\tsrc/app.ts\n';
           }
           if (args[0] === 'show') {
@@ -937,7 +937,7 @@ describe('runCompiledRules', () => {
         .spyOn(totem, 'safeExec')
         .mockImplementation((cmd: string, args?: string[]) => {
           if (!args) return '';
-          if (args[0] === 'ls-files' && args[1] === '--recurse-submodules' && args[2] === '-s') {
+          if (args[0] === 'ls-files' && args[1] === '-s') {
             return '120000 hash 0\tsrc/app.ts\n'; // Symlink
           }
           if (args[0] === 'show') {
@@ -980,7 +980,7 @@ describe('runCompiledRules', () => {
         .spyOn(totem, 'safeExec')
         .mockImplementation((cmd: string, args?: string[]) => {
           if (!args) return '';
-          if (args[0] === 'ls-files' && args[1] === '--recurse-submodules' && args[2] === '-s') {
+          if (args[0] === 'ls-files' && args[1] === '-s') {
             return '100644 hash 0\tsrc/app.ts\n';
           }
           if (args[0] === 'show') {
@@ -1034,7 +1034,7 @@ describe('runCompiledRules', () => {
         .spyOn(totem, 'safeExec')
         .mockImplementation((cmd: string, args?: string[]) => {
           if (!args) return '';
-          if (args[0] === 'ls-files' && args[1] === '--recurse-submodules' && args[2] === '-s') {
+          if (args[0] === 'ls-files' && args[1] === '-s') {
             return '100644 hash 0\tsrc/app.ts\n';
           }
           if (args[0] === 'show') {
@@ -1129,7 +1129,7 @@ describe('runCompiledRules', () => {
         .spyOn(totem, 'safeExec')
         .mockImplementation((cmd: string, args?: string[]) => {
           if (!args) return '';
-          if (args[0] === 'ls-files' && args[1] === '--recurse-submodules' && args[2] === '-s') {
+          if (args[0] === 'ls-files' && args[1] === '-s') {
             return '100644 hash 0\tsrc/app.ts\n';
           }
           if (args[0] === 'show') {
--- a/packages/cli/src/commands/run-compiled-rules.test.ts
+++ b/packages/cli/src/commands/run-compiled-rules.test.ts
@@ -820,6 +820,106 @@ describe('runCompiledRules', () => {
       mockExec.mockRestore();
     });
 
+    // ─── Prop 310 § Design 8 — the staged read decides the REGEX verdict ─────
+    //
+    // The wiring this pins: `readStrategy` used to be built inside the AST block,
+    // so the regex dispatcher never received it and a record rule's
+    // `requires: {scope: file}` was judged against the WORKTREE while `--staged`
+    // was judging the index. That is fail-open in the direction that matters —
+    // a required context deleted in the staged edit but still on disk would
+    // satisfy the requirement and let the staged violation through the
+    // pre-commit gate. The fixtures below make the two sources DISAGREE, so only
+    // the source that actually decides can produce the observed verdict.
+    describe('Prop 310 record rule with a file-scoped requirement', () => {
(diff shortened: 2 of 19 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@gemini-code-assist @coderabbitai — round 1 disposition (operator-triaged; **all 11 findings accepted and fixed** in commit `8b8e7853`, pushed; greptile reviewed clean at 5/5, nothing addressed there):

**Substantive — FIXED:**
1. *GCA high, regex recompilation in the requires hot path:* compiled patterns now cached via the repo's existing `BoundedRegexCache` (exported from `sys/glob.ts`; capacity 512; the pattern set is manifest-bounded). Pinned by a repeated-evaluation constancy test.
2. *CR major, `--recurse-submodules`:* dropped from the staged `git ls-files` invocation with a comment barring its return absent a declared Git minimum; remaining options and symlink exclusion preserved.
3. *CR major, unbounded `requires.pattern` at runtime:* now gated with the same safe-regex2 predicate the compile path uses, at the same dispatcher altitude as the torn-manifest guard, failing loud as the hand-edited-manifest class. To keep one implementation without an import cycle, `validateRegex` moved to a leaf module (`regex-validation.ts`) and is re-exported from `compiler.ts` — all shipped consumers unchanged, byte-identical predicate.
4. *CR minor, path containment:* both new whole-file readers resolve against their root and treat escaping paths (`..`-relative or absolute) as unreadable → null, preserving fail-toward-flagging; sibling-prefix bypass (`/app` vs `/app-secrets`) tested.
5. *CR minor, ast-grep + `requires.scope: line` runtime backstop:* added, mirroring the tree-sitter backstop, citing the span-vs-line ground.

**Mechanical — FIXED:** corpus-floor asserts pinned to the frozen 485 (`SHIPPED_RULE_COUNT` const, freeze-lift comment); `innerType().shape` over the private accessor; `Object.hasOwn` in `assertKeysDispositioned` (noting the smuggled-key route is already dead at parse via slice 1's prototype floor — this is defense-in-depth for hand-crafted callers); exact Stage-4 outcome asserts; windtunnel post-image preload narrowed to files scoped by a file-scope-requires rule; the `as CompiledRule` cast removed via structural widening (`RuleScopeFields`).

**Confirmed consequence worth the record (from CR's `candidate-debt` analysis, verified):** a record rule can never reach `in-scope-bad-example` until slice 3's intake derives the legacy `badExample` field, so Stage 4 currently lands every record rule in `candidate-debt` (forcing `severity: 'warning'`) — pinned with an exact assertion and a comment so the slice-3 change is visible.

**Scope note:** the commit also carries two spec-alignment items from the merged Prop 310 Amendment 2 (mmnto-ai/totem-strategy `b427f15`): independent `]`/`)` dialect fixtures + a nine-character legal-near-miss table (the masking was real — dropping either character from the blacklist left every prior fixture green), and the `|`/`^`/`$` comment rewritten to Amendment 2's ruled grounds (ambiguity-avoidance; match-impossibility claimed for no character).

Verification: all six gates re-run independently by the dispatching seat after the fix commit (workspace build; core 4,084 + CLI 4,077 passed; format; full-branch `totem lint` 0 errors; eslint — all exit 0). Prop 310 suites now 334 tests. `totem review --covariate`: no verdict artifact recorded for the current lineage — no `local-lane:` line to carry this round.

Issue #2667, opened 2026-08-22
ast-grep payload gate: `constraints:` on a $$$ multi-metavariable is silently inert in @ast-grep/napi 0.42.0 — fail loud at validation (R14 round finding)


satur8d:
## Measured

`@ast-grep/napi` **0.42.0** (the shipped engine, `packages/core/node_modules`) silently ignores a `constraints:` entry whose meta-variable appears in the pattern only as a **multi** meta-variable (`$$$NAME`). Measured this session through the shipped call shape `parse(Lang.TypeScript, code).root().findAll(config)`:

| config | `log.error('compile failed: ' + msg)` | `log.error('Totem Error', msg)` |
|---|---|---|
| `{rule:{pattern:'log.error($$$ARGS)'}}` (bare) | 1 | 1 |
| `{rule:{pattern:'log.error($$$ARGS)'}, constraints:{ARGS:{not:{regex:'Totem Error'}}}}` | **1** | **1** |
| `{rule:{pattern:'log.error($FIRST, $$$REST)'}, constraints:{FIRST:{not:{regex:'Totem Error'}}}}` | 0 (pattern needs ≥2 args) | 0 |

Identical counts to the bare pattern on every one of seven inputs for the `$$$ARGS` form (and 1/1/1/1/1 for `new RegExp($$$ARGS)` + `constraints:{ARGS:{regex:'pwd'}}`). No error, no warning — the constraint is inert. A single meta-variable (`$SRC`, `$FIRST`) IS constrained. Replayable probe: `operations-local/r14-treeform-probe.mjs` on branch `r14/seed-20-translation` (commit `36312ed9`).

Surfaced at the R14 round (mmnto-ai/totem-strategy#288, respondent deposit 2026-08-22): the construct the translator had named as the "missing" one for two seed rules would not have worked even if grafted — which is why it is also a fail-loud candidate, not only a typing correction.

## Why it matters here

- **Legacy compound corpus:** 19 NapiConfig-form ast-grep rules are live under the freeze. Any of them carrying `constraints:` on a `$$$` variable is silently over-firing today (censused at the Prop 310 spike: zero of the 19 use `constraints`/`utils` — so the live exposure is zero, but the class is unguarded).
[...]

Issue #2666, opened 2026-08-22
doctor: records-unreferenced sensor — .rule.yaml files under .totem/rules/ that no authored-envelope entry ingests (Prop 310 slice-3 OQ-4, sensor not gate)


satur8d:
## Gap

After Prop 310 slice 3, the ADR-112 authored envelope (`.totem/spine/authored-rules.yaml`) is the explicit ingest set: an entry names `record: .totem/rules/<slug>.rule.yaml`, and intake reads exactly the records the envelope enumerates. A `.rule.yaml` under `.totem/rules/` that **no envelope entry references** is therefore never parsed, never minted, never ledgered — and nothing today says so.

This was ruled (operator, 2026-08-22, slice-3 design OQ-4) as **sensor, not gate**: an intake gate would block drafting (e.g. the R14 seed-20 translation branch carries 20 records and no envelope by design), so the right surface is `totem doctor`.

## Why it is not silent drift today

`records_hash` (slice 3) attests every record file's bytes in `compile-manifest.json`, so an unreferenced record is at least *visible* as attested content — it just has no authoring status. The missing piece is the sensor that names it.

## Proposed sensor (report-only, Tenet 13)

`totem doctor` row `records-unreferenced`: enumerate `<totemDir>/rules/**/*.rule.yaml`, subtract the set of `record:` paths named by the envelope (if the envelope is absent, every record is unreferenced), and list the remainder with a one-line hint ("add an envelope entry to author it, or delete the file"). Severity: info when the envelope is absent (drafting state), warn when an envelope exists and omits a record. Never affects exit code under `--strict`? — decide at build; the default should follow doctor's existing sense/gate split for informational rows.

## Non-goals

- No intake change (ruled).
- No auto-authoring, no envelope generation.

Refs: slice-3 design `.totem/specs/prop310-slice3.md` (failure-mode row "A `.rule.yaml` under `.totem/rules/` referenced by no envelope entry"; OQ-4), Prop 310 § Design 1, slice 1 mmnto-ai/totem#2660, slice 2 mmnto-ai/totem#2663.

Issue #2665, opened 2026-08-22
index: retrieval-index (chunker) membership for .totem/rules/*.rule.yaml — named V1 decision (manifest-only), Prop 310 coupling row


satur8d:
## Decision being named

Prop 310 § Existing-Surface Coupling (manifest-attestation row) requires that whether `.totem/rules/*.rule.yaml` record files **also join the retrieval index (chunker class)** be "a named build decision chartered with the parser — never a silent assumption." The approved slice-1 design (`.totem/specs/prop310-v1-parser.md` § Open questions) ruled **manifest-only at V1** and deferred indexing to a named issue filed with slice 3. This is that issue.

## State at filing

- Slice 3 (intake front-end + envelope derivation + manifest attestation) attests the record file class via `records_hash` in `compile-manifest.json`. Records are **not** ingested by the index pipeline.
- Index targets are declared in `totem.config.ts` (`targets[]`, shape `IngestTargetSchema = { glob, type: 'code'|'session_log'|'spec'|'lesson', strategy }`, `packages/core/src/config-schema.ts:54-58`); consumption is the single `resolveFiles(config.targets, …)` call in `packages/core/src/ingest/pipeline.ts`. No built-in chunk strategy understands a YAML rule record today (`BUILTIN_CHUNK_STRATEGIES`: typescript-ast, markdown-heading, session-log, schema-file, test-file, generic).

## What a wiring would need (not decided here)

1. A `ContentTypeSchema` member for records (or a ruling that `spec` covers them) — the enum is closed.
2. A chunk strategy for `.rule.yaml` (one record = one chunk is the obvious shape; `generic` would split on size and sever `examples` from `target`).
3. A default `targets[]` entry in `totem init`'s config template, and the `AI_PROMPT_BLOCK`/describe surfaces that enumerate targets.
4. A decision on what retrieval is FOR here: `search_knowledge` already returns compiled rules from `compiled-rules.json` readers; indexing the authored source would duplicate the compiled surface unless the query intent is "show me the record, not the lowered rule."

## Gate

Consumer demand or an R14-trial finding that authors need record retrieval (the § 15 step-5 miss typing has no row for it — that itself is a signal this is not V1-blocking). Until then: manifest-only stands; a silent `targets[]` addition is the thing this issue exists to prevent.

[...]

PR #2668, opened 2026-08-22
feat(spine): record intake front-end, envelope derivation, manifest attestation (Prop 310 slice 3)

satur8d:
## What

Slice 3 of the Prop 310 V1 record-grammar build: **the intake front-end, the Amendment-1 envelope derivation, and manifest attestation of the record file class** — the ADR-112 authored producer now reads standalone `.totem/rules/<slug>.rule.yaml` records instead of an inline `dslSource`. Slice 1 (parser) merged as mmnto-ai/totem#2660, slice 2 (lowering + compiled homes + runtime) as mmnto-ai/totem#2663; this closes the approved 3-slice stack.

- **Intake (`packages/cli/src/authored-rule-intake.ts`, `packages/core/src/spine/authored-rule.ts`)** — an envelope entry carries `record: .totem/rules/<slug>.rule.yaml` (lexically validated: repo-relative POSIX, `<totemDir>/rules/` prefix, `.rule.yaml`, no `\` / absolute / drive-letter / `.` / `..` / empty segments; then resolved-containment checked). The record is read through the shipped LF admit hop, content-hashed (`sha256(lf(bytes))`), and parsed by slice 1's `parseRuleRecord` (a `RuleRecordParseError` propagates typed and unwrapped). `declaredEngine` is **derived** from `target.type` (§ Design 3/R17) — never author-supplied. `dslSource`, `declaredEngine`, and an inline fixture `preimageSource` are rejected **by name at any depth** with a migration message naming key, full path, and the record form (OQ-1, records-only).
- **Envelope derivation (Amendment 1 / § Design 10)** — each positive fixture keeps its ADR-112 §6 corpus anchor and references `example: <ordinal>`; intake derives `preimageSource: { kind: 'record', ruleId, ordinal, pairHash, badExample, goodExample }` from `examples[ordinal]` **after** identity is resolved, so the join key carries the real minted `ruleId`. Dangling and duplicate ordinals fail loud naming rule + ordinal (OQ-2). The drift sensor is the existing ledger `contentHash`: the derived fixtures (pair hash included) and the record's content hash are the material, so an `examples` edit reads `revised` and the cert path's `verifyOnly` re-derive refuses it — no second mechanism.
- **Attestation relocation (§ Design 1)** — the ledger entry records `record: { path, contentHash }`; `contentHash` replaces `dslSource` in the `authoringContentHash` material and is what `splitRef` / `authoredAfterSplit` / `heldOutNonInspectionAttestation` now bind to. `path` is informational only (a rename changes no identity and forces no re-attestation — pinned).
- **Compile dispatch (`packages/core/src/spine/compile.ts`)** — `CompileInputCandidate` carries exactly one of `dslSource` / `record`; both or neither throws up front, before any parse; the record carrier dispatches to slice 2's `compileRuleRecord` with the minted `ruleId`, and `extractManualPattern` / `sanitizeFileGlobs` never run on it. Mined candidates are byte-identical (existing suites unchanged).
- **Readers single-homed on `examples` (discharges the slice-2 Stage-4 pin)** — `ruleBadExampleLines` / `ruleGoodExampleLines` in `record-runtime.ts`; Stage 4 (`verifyAgainstCodebase` → `lineMatchesBadExample`) now classifies a record rule whose matched line equals any `examples[i].bad` line as `in-scope-bad-example` (the `candidate-debt` pin flipped **by name**, with a non-matching negative control); `totem doctor`'s `no-badExample` / `no-goodExample` read the same helpers; `totem rule test <id>` runs every pair through the smoke gate and reports per ordinal (exit 1 on any failure). No mirror onto the legacy `badExample` field (Tenet 20).
- **Manifest attestation (`compile-manifest.ts`, `verify-manifest.ts`, the 6 writer sites, `.github/workflows/compile-manifest.yml`)** — `records_hash` = the same rolling `${relPath}\n${lfContent}\n` sha256 as `input_hash`, over git-tracked `.totem/rules/**/*.rule.yaml`; absent is accepted **only while zero tracked records exist** (printed `records: none`, OQ-3); records-present-but-unattested and mismatch are hard FAILs that never take the freeze WARN-downgrade (that branch is input-hash-only, pinned with a negative control). `compile --refresh-manifest` counts the record class in its freshness test (a pre-Prop-310 zero-record manifest is a byte-exact no-op, pinned). `generateInputHash` is byte-identical after the shared-collector refactor (probed against main's implementation).
- **Schemas** — `PreimageSourceSchema` gains the `record` branch (evaluated by the preimage differential exactly as `lesson`); `AuthoredRuleRecordSchema.record` carries the parsed record in memory (never hash material); `AuthoringLedgerEntrySchema.record?` is optional so pre-slice-3 rows still parse; `CompileManifestSchema.records_hash?` optional for back-compat.
- Net: 41+ files, core +65 / cli +86 tests; `minor` changeset for both packages with **BREAKING** prose (see below).

**Contract:** [Prop 310](https://github.com/mmnto-ai/totem-strategy/blob/main/proposals/active/310-record-grammar.md) § Design 1, 3, 5, 10, 12–13 + Amendment 1 items 1–3 + Amendment 2; ADR-112 §3/§5/§6/§8; the ruled Implementation Design at `.totem/specs/prop310-slice3.md` (committed here with the slice-1 design it extends). Freeze-clean: nothing on the `rule-compilation` do-not list is touched; `.totem/compiled-rules.json`, `.totem/compile-manifest.json`, `.totem/lessons/**` unchanged; the live `verify-manifest` still exits 0 (`records: none`).

## Review of record

Opus build-leg against the ruled design (four operator-ruled OQs: records-only carrier · reject duplicate ordinals · `records_hash` absent-OK iff zero tracked records · no OQ-4 work). **Full falsification-leg pass** on the build (`1878557c`): **0 MATERIAL · 8 MINOR · 7 NOTE — SOUND-WITH-CURES**; every ruled OQ, the § Design 1 relocation constraint, the § Design 10 join, and all six shipping entry points held under probes (attestation bound to bytes · CRLF/rename twins · join-key failure modes through the real `runRuleAuthor` · carrier XOR · migration rejections at depth · `records_hash` absent/present/mismatch + freeze-downgrade blindness · Stage 4 / doctor / `rule test` reached from the product paths · `ParsedRuleRecordSchema` round-trip · freeze isolation · test hygiene). Cure round (`3d468f17`) took all 8 MINOR plus four NOTE closures; the **scoped re-arm** on that fold found it SOUND — 6 CURED · 2 PARTIAL (the changeset bump label, below; writer wiring tests 4/6) · 0 REGRESSED, every original counterexample re-run and dead — and surfaced one MATERIAL that was the dispatching seat's charter error, not the code's: the cure had edited the spec's verbatim **RULED** quote of the operator's OQ-3 greenlight ("zero records on disk" → "git-tracked") instead of leaving the ruling as spoken. The convergence round (`313d8dbc`) restores both ruled passages byte-for-byte from `1878557c` and adds a dated build note beneath them, plus the re-arm's four MINORs (changeset accuracy on the split-artifact claim · a discriminating reversed-list test for the ordinal lookup · the Stage-4 pricing sentence moved out of Invariants with its `unverified: true` clause pinned · doctor reason-code docs) and two one-line pre-existing NOTEs (`..hidden/` false-reject in the containment check; `CompiledRuleSchema.examples.min(1)`). No third falsification pass was armed on the convergence round — it is mechanical execution of reviewed findings plus the spec restore, read directly by the dispatching seat; stated here as the discretion it is. Owner gates (build · core + CLI full suites · repo-wide `format:check` · eslint · full-branch `totem lint` · live `verify-manifest`) re-run independently by the dispatching seat after every round, direct exit codes.

Commit stack (nothing amended): `9dbe661d` specs → `78bd2381` core carrier/preimage/records_hash → `7d2b6690` records-only envelope + derivation → `9575ca0e` CLI readers + attestation → `1878557c` changeset + gate fixes → `3d468f17` falsification cures → `313d8dbc` convergence.

**Two items for your confirmation, flagged by the re-arm:** (1) **OQ-3 as built is narrower than as ruled** — "absent is OK iff zero records *on disk*" is implemented as *git-tracked* records (an untracked draft is neither attested nor counted; writer and verifier stay symmetric with the lesson class). The ruled text stands unedited in the spec; the build note names the narrowing. Say the word if you want "on disk" literally (it would make a local draft block an unrelated push). (2) **OQ-1's stated grounds** said `totem rule author` "needs a frozen split artifact" — it also runs on the legacy free-text `splitRef` lane without one; the zero-consumer claim rests on no envelope existing anywhere in the cohort, which holds.

## Applied rulings and build deviations (your override window)

- **OQ-1 records-only:** inline `dslSource` on the authored envelope is a fail-loud migration error, not a second carrier. Zero migration cost today — no `authored-rules.yaml` or authoring ledger exists on main or in the cohort, and `totem rule author` requires a frozen split artifact.
- **Changeset label — flagged for your ruling:** this removes `dslSource` / `declaredEngine` from a published exported input schema. It ships as `minor` with **BREAKING** prose per this repo's precedent (`1.91.0` removed the `cohortRepos()` export the same way, `packages/core/CHANGELOG.md:464`); say the word if you want it `major` instead.
- **Build-leg deviations, all verified:** `lfDeepNormalize` → new leaf `spine/lf-normalize.ts` and `AUTHORED_RULE_ID_RE` → `compiler-schema.ts` (both real import cycles — `rule-record.ts` ↔ `authoring-ledger.ts` ↔ `authored-rule.ts`; re-exported, one definition each, barrel unchanged); `compileCandidate`'s record branch keeps the ADR-112 §3 whitelist↔engine assert as a second, non-tautological leg beside `compileRuleRecord`'s `target.type`↔payload assert; the anti-vacuity `superRefine` learned the `record` kind (reachable through the shipping path — pinned); `compile --refresh-manifest` counts records (else OQ-3's recovery had no writer that would act); a repo-relative reference resolves against `dirname(totemDir)` and a nested `totemDir` fails loud with the path named.
- **Falsification cures applied (MINOR):** lexical record-reference validator (platform-independent verdict; an absolute machine path can no longer enter the tracked `authoring-ledger.ndjson`); `records: none` wording and the spec's OQ-3 rows say *git-tracked* (untracked records are neither attested nor counted; writer and verifier are symmetric; pinned in a real temp git repo); ledger docstrings state that `ruleId` enters the material only as the join key; `ruleGoodExampleLines` single-homes doctor's `no-goodExample`; the migration hint interpolates the real `totemDir`; wiki prose for `rule test` and `verify-manifest`; a `lesson archive` writer wiring test; pair-hash lookup by `.ordinal` with a throwing (not recomputing) missing-hash path; `packages/core/src/compile-manifest.ts` joins the workflow paths filter.

## Named forward items (not silently dropped)

- **Writer-site wiring tests** exist for `compile --refresh-manifest`, `rule promote`, `lesson archive`, and `shield` observation capture (4 of 6); the two full-compile writers in `compile.ts` are verified by code-read only (they need an orchestrator stub).
- **Test-only exports** from the CLI-internal `authored-rule-intake.ts` (`checkRecordReference`, `RECORD_REFERENCE_RULES`, `deriveRecordFixtures`, `IngestedRecord`) — the module is not on any published barrel, so no public API grew; named so the exposure is a decision, not a drift.
- **Symlinked records** under `.totem/rules/` are ingestible but not attested (`collectFilesWithSuffix` skips non-file dirents) — inherited verbatim from the lesson class; a follow-up if a consumer ever tracks a symlinked record.
- **Stage-4 `confidence: high`** is now reachable from an author-written exemplar that occurs verbatim in the tree; `unverified: true` survives the patch, so the rule stays advisory — priced in the spec, not cured.
- **Live-corpus population** from certified records is the mmnto-ai/totem-strategy#516 populator's edge (authored compiled rules land only in the cert corpus output, as today). **No record writer** ships here — Amendment 1 item 2 binds the first one. `totem rule scaffold` keeps emitting legacy `.md` fixtures. The `.totem/tests/*.md` `corpus:` frontmatter field has zero production consumers — recommended RETIRE, separate small PR.
- Filed alongside: mmnto-ai/totem#2665 (retrieval-index membership — manifest-only at V1, the named decision), mmnto-ai/totem#2666 (OQ-4 unreferenced-records doctor sensor), mmnto-ai/totem#2667 (inert `constraints:` on `$$$` in the engine — fail-loud candidate from the R14 round).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai — round 1 dispositions (all seven findings verified against source by the owning seat; all seven fixed in `af823057`; one test-row correction in `3a0395b9`):

1. **compile.ts no-op path — records freshness (Major, outside diff): FIXED.** Premise confirmed — `manifestStale` keyed on `input_hash` alone, so a tracked record-only change never rewrote `records_hash`. The OQ-3 freshness predicate is now ONE helper shared by `--refresh-manifest` and the ordinary no-op path (absent ⇒ fresh iff zero tracked records; present ⇒ equals `attestRecordsHash`), and record staleness joins the manifest-write condition. Rules file still untouched on pure drift. Regression rows: a tracked record-only change ⇒ manifest rewritten with the fresh hash and `compiled-rules.json` byte-identical; nothing changed ⇒ manifest byte-identical; plus the predicate's six arms in a real git repo. Both behavioural fixes (this and item 2) were mutation-tested — reverting each makes its row fail.
2. **`rule test` preview sanitization (Major): FIXED.** Both previews pass through `sanitizeForTerminal` (the same helper intake uses) and the limit is `EXAMPLE_PREVIEW_MAX_CHARS`; pinned with a record whose `bad` carries an erase-display CSI (`ESC[2J`) — chosen over an SGR colour code because the CLI's own colouring emits SGR when colours are on; the row holds with colours off, forced on, and default (`3a0395b9`).
3. **verify-manifest test assertions (Trivial): FIXED.** Both record rows now assert the `Records hash mismatch` branch text alongside the generic throw, the same way the "unattested file class" row does.
4. **`SHA256_HEX_RE` single-homed (Trivial): FIXED.** Exported from `compiler-schema.ts` (a leaf) and used by `RuleExamplePairHashSchema.hash`, `record.contentHash` on the authored record, and both 64-hex fields on the ledger entry; custom messages kept.
5. **Misplaced `ParsedRuleRecordSchema` doc block (Minor): FIXED.** Moved to its subject; `RuleExamplePairHashSchema` has its own one-liner.
6. **Anti-vacuity branches folded (Trivial, outside diff): FIXED.** One predicate over the pair-carrying kinds (`lesson`, `record`), identical message and `path`; existing tests for both kinds unchanged.
7. **Slice-1 spec scaffold (Minor): FIXED.** Top-level heading plus an upfront warning that everything above `## Implementation Design` is the `totem spec` generator's discarded scaffold, not the contract; content order unchanged so the divergence table stays where the design references it.

Not addressed to Gemini or Greptile — both returned zero findings this round.

local-lane: none — `totem review --covariate` reports no verdict artifact for this lineage (the advisory `totem review` lane was not run on this branch; the review of record is two falsification-leg passes + a scoped re-arm + owner gates after every round, as the PR body states).

Issue #2678, opened 2026-08-23
bug(smoke-gate): runSmokeGate ignores `requires:` — no § Design 8 record can pass C4 / `rule test` (good over-matches by construction); greenfield trial intake-defect at liquid-city a84ae6db


satur8d:
**Typed `intake-defect` from the greenfield authorability trial** (mmnto-ai/totem-strategy#288, pre-registration `operations/310-greenfield-preregistration.md` § 3 C4). Reproduced by the author (lc-claude, pin mail 2026-08-23T18:50Z, record pinned at mmnto-ai/liquid-city `a84ae6db`) and re-derived by the scorer (strategy-claude) against the 1.119.0 source. Routed on the operator's word, 2026-08-23.

## The defect

`runSmokeGate` never reads `rule.requires`. At 1.119.0 (`packages/core/src/compile-smoke-gate.ts`, the `runSmokeGate` body):

```ts
if (rule.engine === 'regex') {
  return runRegexGate(rule.pattern, snippet);   // rule.requires is not consulted
}
if (rule.engine === 'ast-grep') { … runAstGrepGate(source, snippet, rule) … }   // nor here
```

`grep -n requires packages/core/src/compile-smoke-gate.ts` returns nothing. Meanwhile:

- the grammar admits it — `rule-record.ts:747` `requires: RuleRequiresSchema.optional()` (Prop 310 § Design 8, the first-class absence construct that retires lookahead-smuggled absence);
- the lowering carries it — `record-lower.ts:238`; `record-lower.test.ts:235–241` asserts `rule.requires?.scope === 'file'` survives;
- the runtime evaluates it — `record-runtime.ts` § Design 8 two-pass: on a target match the rule fires **iff** `requires.pattern` does NOT match within the declared scope (`matchesInScope`, `:279`; the evaluator from `:359`).

Only the smoke gate is blind. Consequence, by construction: a `requires:` record's `good` example necessarily keeps the target and adds the required companion — so the gate reports `goodSilent: false` (the target pattern matches the good snippet), and **no `requires:`-bearing record can pass the gate that certifies records.**

## Reproduction at the pin

[...]

PR #2681, opened 2026-08-24
fix(smoke-gate): runSmokeGate evaluates § Design 8 requires with the runtime's evaluator and its preconditions (mmnto-ai/totem#2678)

satur8d:
## What

`runSmokeGate` (`packages/core/src/compile-smoke-gate.ts`) now evaluates Prop 310 § Design 8 `requires:` — with the runtime's own evaluator **and** the runtime's own preconditions — so a `requires:`-bearing record's `good` example goes silent for the right reason and `totem rule test` / the cert-path preimage differential can pass such records. Fixes the defect in mmnto-ai/totem#2678 (typed `intake-defect` from the greenfield authorability trial, mmnto-ai/totem-strategy#288; reproduced today on the R14 set in the freeze-lift (2a) intake smoke test — `5da43ea6` over-matched on `good` exactly this way).

## The defect

The grammar admits `requires: { pattern, scope: line | file }`, the lowering carries it onto `CompiledRule.requires`, and every runtime dispatcher (`rule-engine.ts` regex + ast paths, `regex-safety/apply-rules-bounded.ts`) evaluates it through `requiresSuppressesMatch`. Only the smoke gate ran pass one. Since a § Design 8 record's `good` keeps the target and adds the companion, the gate reported it as over-matching — **no `requires:` record could pass the gate that certifies records**.

## The cure (Tenet 20: one evaluator, borrowed with its contract)

- Both engine paths call `requiresSuppressesMatch(rule, { line, file: () => snippet })` per target-match locus (the snippet is the file for a gate run); a suppressed locus is not a match, so `matchCount` counts unsuppressed loci only. Guarded on `rule.requires !== undefined` — legacy rules are byte-identical by construction (verified: none of the 485 frozen rules carries any record home key).
- The evaluator's three dispatcher-altitude preconditions run inside the gate for a `requires:`-bearing rule before dispatch: `assertNoTornRecordRules` · `assertRequiresPatternsSafe` · `assertNoAstGrepLineScope`. Without them the gate accepted rules the runtime refuses — an unsafe-but-compilable `requires.pattern` (`(a+)+$`) backtracked for a measured **86 s** under `totem rule test`, and `ast-grep` + `requires.scope: line` passed the gate while the runtime throws.
- One `TotemParseError` handler for both engines and all three preconditions → `{ matched: false, matchCount: 0, reason: 'requires precondition failed: …' }` (each message names the construct at fault), mirroring the existing `invalid regex:` shape; anything else rethrows.
- ast-grep: the `requires` filter runs outside the per-extension engine try/catch so an uncompilable requirement is never mislabelled `ast-grep runtime error:`; an extension whose matches are all suppressed falls through to the next (the gate still passes on ANY extension).

## Tests (core 4150 → 4171, cli 4166 → 4167)

`compile-smoke-gate.test.ts` 22 → 43: regex `line`/`file` scope, ast-grep flat + compound, the scope distinction on a two-line snippet, each precondition (ReDoS input asserted `< 2000 ms`, `ast-grep` + `line`, torn rule), unusable pattern on both engines through the one handler, legacy-companion no-op, an end-to-end `parseRuleRecord → compileRuleRecord → runSmokeGate` run on the § Design 8 exemplar, and a **dispatcher-driven parity test** in three legs — the sync regex dispatcher `applyRulesToAdditions` (`readFileText` seam), the bounded regex dispatcher `applyRulesToAdditionsBounded` + `RegexEvaluator` (the one `totem lint` actually uses for regex rules, structurally different at `scope: file`), and `applyAstRulesToAdditions` (`readStrategy` seam) — over synthesized `DiffAddition[]` with one shared table of pinned violation counts (regex line 1/0/1 · regex file 1/0/0 · ast-grep file 1/0); a torn `excludeGlobs`-only rule is refused like the runtime refuses it. `rule.test.ts` +1: a `requires:` record PASSes `totem rule test` through the real command.

## Review of record

Two Opus falsification passes. Pass 1 (full diff): 2 MATERIAL (the missing preconditions — ReDoS reachability measured; ast-grep + `line` false PASS), 3 MINOR (torn-rule precondition; four `// totem-context:` overrides that silenced advisory findings only; a parity test that mirrored the gate body), 3 NOTE — all folded (`d5e2c8ac`). Pass 2 (scoped re-arm on the fold `d5e2c8ac`): MATERIAL-1/2 **cured** (measured: the ReDoS input 86 s → 0.5 ms with a reason naming `requires.pattern`; ast-grep + `line` refused), MINOR-4 / NOTE-6 cured, two PARTIALs (the torn check sat behind the `requires` guard while the runtime refuses any record home without `examples`; the regex parity leg drove a dispatcher `totem lint` does not use for regex) plus one new MINOR (two assertions had silently retargeted onto the precondition; three comments went stale) and two NOTEs (an "every dispatcher runs these three" overclaim; the empty-snippet early return bypasses the preconditions — kept, non-material, every caller rejects an empty exemplar before the gate). All folded in `5db73db1`; **convergence named** — no third pass on that fold, the owner read it (one hoisted call plus tests/comments). Owner gate re-run after each round: build, core/cli suites, `format:check`, eslint (8 pre-existing warnings, untouched files), `totem lint --branch --base main` PASS (24 advisory warnings, mostly regex-hygiene lessons re-surfaced by touching `new RegExp(rule.pattern)` — pre-existing construction, no new regex introduced).

## Notes, not silently dropped

- **CRLF asymmetry (pre-existing, not new):** the gate splits the snippet on `\n` and the diff parser keeps `\r` on `addition.line`, so `$`-anchored `requires.pattern`s are equally fragile in both; at `scope: file` the gate sees the YAML-normalized LF exemplar while the runtime reads the real file. Named here; no behaviour change.
- **Fall-through semantics shifted safely:** an extension whose matches are all suppressed now falls through to the next extension (a later grammar can report a match) — in `totem rule test` that is a false FAIL never a false PASS.
- **Rule `09381ca14ca4fefb`** ("Sanitize git-sourced metadata", regex, severity `error`) fires on inert string literals containing `git log` etc. — advisory (regex engine ⇒ non-blocking), observation only; curated in slice 1 as `discard noise`.
- Changeset: `@mmnto/totem` + `@mmnto/cli` `patch` (they version together; `rule test` behaviour changes for `requires:` records).
- mmnto-ai/totem#2678 stays open until the greenfield draw-1 record re-scores under the cured gate (its pre-registered handling) — close on your word.


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai — bot round 1 disposition (all surfaces read in full; premises verified against source before acting). Fixes pushed as `35868b49`.

**1. FIXED — `compile-smoke-gate.ts` reason prefix (CR inline, Minor).** Premise verified true: the torn-rule precondition runs unconditionally, so `requires precondition failed:` also fronted rules carrying no `requires` block. Changed to `rule precondition failed:` exactly as suggested; the catch comment now states why the prefix is construct-neutral. No test pinned the prefix (verified), so no assertion moved. One factual note for the record: the review's quoted error text (`[fileGlobs, excludeGlobs]`) does not occur — `fileGlobs` is not a record home key, the real message names `[excludeGlobs]` — the premise held regardless.

**2. FIXED — the mixed-extension `lastReason` regression test (CR prose note on lines 186–209).** Your caveat said the input was not yet established to exist; we established it: a pattern that is multi-root under the TypeScript grammar and single-root under TSX (`<div>{console.log($$$)}</div>`) makes a `['**/*.ts','**/*.tsx']`-scoped rule THROW on `.ts` and evaluate on `.tsx` — so with a `scope: file` requirement satisfied, the `.tsx` matches are all suppressed, the loop falls through, and the `.ts` engine throw is the reported reason. Three tests added (43 → 46): the exact mixed case pinning `/ast-grep runtime error/`, a companion-removed falsifier proving the suppressed set was real (`matched: true` when the requirement is unmet), and the degenerate all-extensions-throw arm proving the requires filter never masks `lastReason`. The current semantics are pinned, not endorsed — the test comment records that `matched: false` is the load-bearing half and the reason line is diagnostic only. (A per-extension suppression split is inexpressible for ast-grep — `requires.scope: line` is refused on that engine and `file`-scope suppression is a property of the snippet — which is exactly what makes this the only reachable mixed shape; recorded in the test so it is not re-derived.)

**3. NO ACTION — `rule.test.ts:526`.** Concur with the review's own retraction: the nested `beforeEach` resets `process.exitCode` and `afterEach` restores it; the assertion is exit-code-safe as written.

Round context: Gemini reviewed with zero findings; Greptile scored 5/5 with no actionable items at `5db73db1` — nothing addressed for either, so per single-notification protocol neither is tagged here.

local-lane: none — `totem review --covariate` reports no verdict artifact recorded for the current lineage (sensor line carried verbatim: "no verdict artifact recorded for the current lineage — run `totem review` with review.lanes configured to emit one").

Issue #2556, opened 2026-08-03
feat(hooks): invoke totem-status refresh-gh from the managed SessionStart + post-merge hook templates (routed C3 residual, mmnto-ai/totem-status#127)


satur8d:
## Context — routed C3 residual from mmnto-ai/totem-status#127

The totem-status GH-federation sidecar's C3 ruling names four refresh moments for the GitHub-lane snapshot: manual verb, daemon tick, SessionStart hook, post-merge hook. The sidecar slice (branch `feat/127-gh-federation-sidecar`) implements the first two. The last two are structurally totem-lane: the SessionStart and post-merge hooks are **totem-managed templates** — `totem hook install` drift-repairs them on every `pnpm install` via the consumer `prepare` wrapper — so a status-side edit would be reverted by the next repair pass. Routed here by status-claude dispatch `2026-08-02T0136Z` (in `.totem/orchestration/totem-claude/processed/`).

## Ask

The managed hook templates gain a **spawn-and-forget** invocation of `totem-status refresh-gh` at:

1. `CLAUDE_SESSION_START` (`.claude/hooks/SessionStart.cjs`) — packages/cli/src/commands/init-templates.ts
2. `GEMINI_SESSION_START` (`.gemini/hooks/SessionStart.js`) — same file (vehicle parity: the C3 moment is "session start", not a vendor)
3. `buildHookContent()` (git `post-merge`) — packages/cli/src/commands/install-hooks.ts

## The contract the verb offers (from the routed dispatch)

- exit 0 = snapshot written (stamped `fetchedAt`, atomic rename); exit 1 = nothing written, disk untouched — measurement-or-nothing, safe to fire blind
- safe to invoke concurrently with the daemon (single-flight + atomic rename); safe when `gh` is missing (no-clobber, distinct message); never blocks on a wedged child past deadline+WaitDelay

## Design constraints

- **Never block session start or the merge.** mmnto-ai/totem#2059 measured ~3s of synchronous `gh` calls hurting SessionStart latency — the invocation is async spawn-and-forget (detached + unref in the `.cjs`/`.js` hooks; backgrounded subshell in the sh hook). The status-side C1 staleness line tolerates a refresh that has not landed yet.
[...]

Issue #2404, opened 2026-07-17
fix(hooks): prepare-script tools/install-hooks.js stomps CLI-canonical git hooks on every install (parity-drift regression loop)


satur8d:
The repo-local `prepare` script (`package.json`: `"prepare": "node tools/install-hooks.js"`) copies `tools/{pre-commit,pre-push,post-merge}` into `.git/hooks` on every install — silently stomping the CLI-canonical hooks that `totem hook install --force` writes. The two installers fight: refresh → any `pnpm install` (incl. one run by a background agent) → drift is back.

Live repro (2026-07-16): after a morning `totem hook install --force` (parity 24 pass / 0 warn), three hooks were rewritten at 17:45:50 local during a subagent build round; `totem doctor --parity` re-reported the exact pre-refresh consumer hashes (d34b1bd59182 / 79cea4adcd67 / 10f3099da162) against the 1.99.0 canonical. Re-refreshed and verified 25/0 — until the next install fires `prepare`.

Fix options (operator to rule):
1. Make `tools/install-hooks.js` delegate to the project-local CLI (`totem hook install --force`) so `prepare` installs the canonical, not a fork.
2. Regenerate `tools/` hook scripts from the 1.99.0 templates and add a parity test binding them to the CLI canonical (else they re-drift on the next template change — the #1854 versioned-upgrade gap).
3. Retire the `prepare` installer entirely in favor of the CLI path.

Related: #2138 (bare `hook install` is presence-idempotent — why the stomp isn't self-healing), #1854 (versioned-hooks upgrade mechanism), #2352 (freeze-mirror drift sensor precedent for installed-vs-canonical parity).

PR #2682, opened 2026-08-24
feat(hooks): fire totem-status refresh-obligation-store beside refresh-gh on every session-start and post-merge surface (mmnto-ai/totem-status#127)

satur8d:
## What

The durable obligation store (totem-status slice two, mmnto-ai/totem-status#127) gets its session-start and post-merge moment: every surface that spawns `totem-status refresh-gh` now also fires `totem-status refresh-obligation-store` under the SAME primary-checkout gate, detached+unref spawn, shared `.git/totem-status-refresh-hook.log` fd, ENOENT-silent arm, and non-fatal breadcrumb — with every stamp naming its verb (`verb=…`). Charter: status-claude's 2026-08-24T2047Z dispatch (sibling of mmnto-ai/totem#2556, itself already done and shut). Retires the "no session-start moment for this store" residual quoted in `.totem/specs/127-durable-obligation-store.md` from the totem side; the daemon-watches-`processed/` half stays in the status lane.

## Surfaces (four, not three)

1. Claude SessionStart template (`init-templates.ts`) — second spawn, per-verb stamps, fd released after both.
2. Gemini SessionStart template (`init-templates.ts`) — twin; stdout stays exactly one JSON payload (verified by execution).
3. Post-merge shell hook (`install-hooks.ts` `buildHookContent`) — both the logged and `/dev/null` arms, printf stamp carries `verb=%s`.
4. **`.claude/hooks/session-context.mjs`** — the bespoke first-registered SessionStart hook this repo's own Claude seat boots; found by the falsification leg (the charter said "three", but this file spawns `refresh-gh` into the same log and the `48cb5208` precedent keeps it in lockstep by hand). Extended identically; its stale "sole SessionStart entry" banner corrected; its Gemini twin was retired at `92e23179` (`.gemini/hooks/SessionStart.cjs` is the generated, byte-locked copy).

Tracked derivatives regenerated via the script path and byte-locked: `.claude/hooks/SessionStart.cjs` (lock ADDED this PR — only the Gemini twin had one), `.gemini/hooks/SessionStart.cjs` (mmnto-ai/totem#2628 lock), `tools/post-merge` (mmnto-ai/totem#2404 lock). Block-level breadcrumb reworded to name the sidecar, not one verb (it wraps a block that fires two).

## Tests (cli 4166 → 4174, +9; 12 skipped incl. 7 POSIX-only cells)

Template/string locks for the second spawn, `verb=` on all stamps, gate containment, fd-after-both; behavioural cells with a stub sidecar assert exact verb-set equality (`['refresh-gh','refresh-obligation-store']` — last-writer-wins fixture replaced by append + sorted-set, strictly stronger); worktree negative asserts the empty set; string-locks on the bespoke hook (both verbs, stamp counts, retired phrasing absent). Out-of-band on this Windows host (the POSIX cells skip): both hooks and the shell hook executed against a real `.exe` stub and empty-PATH — both verbs fire, four per-verb stamps, stderr empty on the ENOENT arm, Gemini stdout one JSON envelope.

## Review of record

One Opus falsification pass on `main..efe10cbe` (executed, not just read: rendered templates byte-compared, both hooks run in temp checkouts under three arms, `tools/post-merge` under `sh`, sidecar probed): **no defect in the generated JS or shell**; 3 MATERIAL / 4 MINOR / 3 NOTE against the claims around it — all folded in `1824e88d`: the fourth surface (M1), the reap-discriminator claims reworded honestly at every site incl. two pre-existing mmnto-ai/totem#2570 comments my diff made false (M2 — the stamps attribute which verbs fired and in what order; child output is unlabelled, so a silent tail attributes only to the last stamp), the sidecar version-skew bound named in changeset + wiki (M3), changeset `patch` → **`minor`** per the `48cb5208` / `3ef61fc` precedents on this exact block (MIN1), the Claude-derivative byte-lock added (MIN2), `.totem/specs/refresh-gh-hooks.md` amended in its own bracket convention (MIN3). Convergence named on the fold — the mechanism was verified by execution and the fold is its extension plus wording; owner read the diff and re-ran the gates (512/512 targeted, full cli suite green, format clean, `totem lint --branch` PASS 0 errors).

## Residuals, named not dropped

- **Sidecar version skew (M3):** a pre-slice-two `totem-status` treats an unknown verb as the default dashboard (measured: exit 0, ~179 KB into the hook log in a cohort cwd, would blow the log's 1 MiB self-cap in ~6 firings). Bound named in the changeset/wiki; quiet non-zero unknown-verb exit requested from status-claude by mail (2026-08-24T2231Z) — either their cure or a cohort version-floor confirmation retires the sentence.
- **Appended-hook class (MIN4, pre-existing):** a consumer whose totem post-merge block was APPENDED into a pre-existing user hook fails `isTotemOwnedWholeFile` and never receives template updates without `--force`; identical for mmnto-ai/totem#2570, not widened here. `totem hook install` / `prepare` cover the owned-whole-file common case (drift-repair verified in source).
- **Concurrent-append question (NOTE-2):** the spec's Windows `>>` residual is now structural on every merge (two children, one log). The leg tried to reproduce corruption with two chatty stubs: 802/802 lines intact — recorded as a question, not a finding.
- **Lint blind spot:** `.claude/hooks/session-context.mjs` and `.claude/hooks/SessionStart.cjs` are excluded from `totem lint`'s branch diff by pre-existing `ignorePatterns` — the byte-lock and string-locks added here are the only CI-visible sensors on them. Flagged for a separate ruling; config untouched.
- Second spawn adds ~2 ms parent-side on session start (measured floor; real sidecar 8 MB) — noted, not instrumented.

Changeset: `@mmnto/cli` **minor**. Discharges status-claude's dispatch on merge (they carry it as `owed @totem-claude until the template names the verb`).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai — bot round 1 disposition (all surfaces read in full; both premises verified against the files before acting). Fixes pushed as `39e501e6`.

**1. FIXED — the spec's `## Surfaces` list now enumerates four.** Premise verified true: the amendment brackets named `.claude/hooks/session-context.mjs` but the formal list still held three entries. Added as entry 4, with its bespoke/hand-lockstep nature stated (first of the two registered Claude SessionStart entries; not generated).

**2. FIXED — stamp ordering stated as a race, not a contract, at both sites.** Premise verified true: each child is spawned immediately after its own stamp, so a fast child's output can land between the two stamps — the "back-to-back before either child writes" wording promoted a measured outcome into a guarantee. Reworded in the spec bracket and `docs/wiki/hook-integration.md`, including the adjacent "output lands after the stamps" phrase the same finding implies (now "after that child's own stamp"). The load-bearing conclusion is unchanged and strengthened: a silent tail attributes only to the LAST verb stamped, never per child.

Round context: Gemini reviewed with zero findings; Greptile scored 5/5 with no actionable items — nothing addressed for either, so per single-notification protocol neither is tagged here.

local-lane: none — `totem review --covariate` reports no verdict artifact recorded for the current lineage (sensor line: "no verdict artifact recorded for the current lineage — run `totem review` with review.lanes configured to emit one").

PR #2683, opened 2026-08-25
Version Packages

satur8d:

# Releases
### Minor Changes

- 8cd367c: init/hooks: SessionStart (Claude + Gemini) and post-merge managed templates also fire `totem-status refresh-obligation-store` beside `refresh-gh`, under the same primary-checkout gate / log stamp / ENOENT-silent contract, so the durable obligation store gets its session-start and post-merge moment (mmnto-ai/totem-status#127; sibling of mmnto-ai/totem#2556)

  Minor rather than patch, following the two governing precedents on this same block: `48cb5208` (refresh-gh observability leg) and `3ef61fc` (which added the refresh-gh spawn, shipped as 1.109.0) both took minor.

  The second verb requires a `totem-status` at slice two (commit `711f07a`) or later. An older sidecar does not reject an unknown verb — it falls through to the default dashboard, so the full status report lands in the hook log instead (measured: exit 0, ~179 KB per firing in a cohort checkout). A quiet non-zero exit for unknown verbs is an open question with the sidecar owner; the bound is named here until it closes.

### Patch Changes

- aba7c78: fix(smoke-gate): `runSmokeGate` evaluates § Design 8 `requires:` with the runtime's own evaluator, so a `requires:`-bearing record's `good` example goes silent for the right reason and `totem rule test` / the cert-path preimage differential can pass such records (mmnto-ai/totem#2678).

  The record grammar admits `requires: { pattern, scope }` (Prop 310 § Design 8 — the rule fires at a target match iff the required context is ABSENT within the declared scope), the lowering carries it onto `CompiledRule.requires`, and the runtime dispatchers evaluate it through `requiresSuppressesMatch`. Only the compile-time smoke gate ran pass one. Because a § Design 8 record's `good` example keeps the target and adds the companion, the gate reported it as over-matching and no such record could pass. The gate now calls the same `requiresSuppressesMatch` the runtime calls — one evaluator rather than a second implementation — on both the regex and the ast-grep path.

  It also takes that evaluator's preconditions with it: every rule goes through `assertNoTornRecordRules`, and a `requires:`-bearing one adds `assertRequiresPatternsSafe` and `assertNoAstGrepLineScope`, before it is evaluated — two of these at each regex dispatcher, all three at the ast dispatcher. Without them the gate accepted rules the runtime refuses — a torn rule carrying a Prop 310 compiled home with no `examples`, an unsafe-but-compilable `requires.pattern` that backtracks catastrophically under `totem rule test`, and `ast-grep` paired with `requires.scope: line`. A failed precondition is reported as a gate reason rather than propagated as a throw. The frozen legacy corpus is untouched by construction: the torn check early-returns for a rule carrying no compiled home, and no legacy rule carries one.

- Updated dependencies [aba7c78]
  - @mmnto/totem@1.120.0
### Patch Changes

- aba7c78: fix(smoke-gate): `runSmokeGate` evaluates § Design 8 `requires:` with the runtime's own evaluator, so a `requires:`-bearing record's `good` example goes silent for the right reason and `totem rule test` / the cert-path preimage differential can pass such records (mmnto-ai/totem#2678).

  The record grammar admits `requires: { pattern, scope }` (Prop 310 § Design 8 — the rule fires at a target match iff the required context is ABSENT within the declared scope), the lowering carries it onto `CompiledRule.requires`, and the runtime dispatchers evaluate it through `requiresSuppressesMatch`. Only the compile-time smoke gate ran pass one. Because a § Design 8 record's `good` example keeps the target and adds the companion, the gate reported it as over-matching and no such record could pass. The gate now calls the same `requiresSuppressesMatch` the runtime calls — one evaluator rather than a second implementation — on both the regex and the ast-grep path.

  It also takes that evaluator's preconditions with it: every rule goes through `assertNoTornRecordRules`, and a `requires:`-bearing one adds `assertRequiresPatternsSafe` and `assertNoAstGrepLineScope`, before it is evaluated — two of these at each regex dispatcher, all three at the ast dispatcher. Without them the gate accepted rules the runtime refuses — a torn rule carrying a Prop 310 compiled home with no `examples`, an unsafe-but-compilable `requires.pattern` that backtracks catastrophically under `totem rule test`, and `ast-grep` paired with `requires.scope: line`. A failed precondition is reported as a gate reason rather than propagated as a throw. The frozen legacy corpus is untouched by construction: the torn check early-returns for a rule carrying no compiled home, and no legacy rule carries one.
### Patch Changes

- Updated dependencies [aba7c78]
  - @mmnto/totem@1.120.0
## @mmnto/pack-rust-architecture@1.120.0

## @mmnto/pack-agent-security@1.120.0


satur8d:
## Release-cut sensor — choreography (the flagged `8cd367c` entry, filled)

The flag is right: `8cd367c` ships init-distributed content — both SessionStart templates and the post-merge hook now spawn `totem-status refresh-obligation-store` beside `refresh-gh` (mmnto-ai/totem#2682).

- **Which consumers move:** every cohort repo consuming the managed hook surfaces (totem-strategy, totem-status, liquid-city; totem itself already carries the regenerated derivatives on main). The change reaches a consumer at its `@mmnto/cli` pin bump to `^1.120.0` — `pnpm install` → `prepare` → hook regen (`installGitHook` drift-repairs the owned-whole-file case; `regenerateManagedSessionHooks` rewrites the session hooks). Named residual (mmnto-ai/totem#2682 PR body): a consumer whose totem post-merge block was APPENDED into a pre-existing user hook does not drift-repair and needs `totem hook install --force` by hand — pre-existing class, not widened.
- **In what order:** (1) this cut publishes `@mmnto/cli` 1.120.0 (+ `@mmnto/totem` patch — the mmnto-ai/totem#2681 smoke-gate cure, which also unblocks the greenfield intake-defect re-scores); (2) seats bump their repo pins at lane pace (dependency-currency sensors flag; no forced order between repos); (3) **before or with each bump, the seat's `totem-status` sidecar must be at slice two (`711f07a`) or newer** — an older sidecar treats the new verb as the default dashboard and writes ~179 KB per firing into `.git/totem-status-refresh-hook.log` (measured; bound named in the changeset, wiki, and spec; a quiet non-zero unknown-verb exit is an open ask with the sidecar owner, mail 2026-08-24T2231Z).
- **Who owns which half:** each repo's resident seat owns its own pin bump + hook regen (standing cohort convention); **status-claude** owns the sidecar half — the version floor / unknown-verb exit and the daemon-watches-`processed/` slice; **totem-claude** owns the template half (shipped here) and answers template-side breakage.

No other entry flagged; the `@mmnto/totem`/`@mmnto/cli` patch half of this cut is engine-internal (smoke-gate) with no distributed content.

Issue #2692, opened 2026-08-29
hooks: the managed git-hook family hardcodes .totem/ — a custom totemDir gets non-functioning pre-push gates and an always-blocking strict pre-commit (class; surfaced by Greptile on #2691's evidence read)


satur8d:
## The class

Every managed git hook hardcodes `.totem/` while `totem.config.ts` accepts a relative `totemDir` (`packages/core/src/config-schema.ts:456`, refined to a relative path). Surfaced by Greptile on mmnto-ai/totem#2691 (P1, confidence 4/5) against the new strict pre-commit evidence read, but the read only joined a pre-existing family:

- pre-commit (strict tier): `.totem/artifacts/runs/*.json` (the spec run artifact, mmnto-ai/totem#2690) and the legacy `.totem/cache/.spec-completed` marker — the OLD gate was `.totem`-bound too
- pre-push: `.totem/compile-manifest.json` (verify-manifest gate), `.totem/compiled-rules.json` (lint gate + the badge/claim-discipline preconditions)
- post-merge / post-checkout: `.totem/lessons/`, `.totem/` change detection for the sync trigger
- `.gemini/hooks/BeforeTool.js` (legacy, unregistered): same evidence read

So a consumer that sets a custom `totemDir` already has non-functioning pre-push gates (no manifest, no compiled rules found) and an always-blocking strict pre-commit; the mmnto-ai/totem#2691 slice did not widen the class and the finding was declined there with this ground.

## Why it is not a one-liner

The hooks are POSIX `sh` with no CLI boot on the pre-commit path (a deliberate property of the evidence read — Tenet 13, no enforcement state written, ~50 ms). Reading `totem.config.ts` from `sh` is not available; the options are (a) `totem init` / `totem hook install` rendering the configured `totemDir` into the generated hook text (the installer already reads config for `hooks.tier` — `hooksCommand` in `install-hooks.ts`), which makes `tools/pre-commit`'s byte-parity a per-config artifact; (b) a `TOTEM_DIR` export the hooks honor with `.totem` as the default; (c) documenting the bound ("managed hooks assume `.totem`; a custom `totemDir` is unsupported by the hook family") and having `doctor` sense the mismatch loudly.

## Ask

Rule on (a)/(b)/(c) and apply it to the whole family at once — never to one hook. Whichever lands, `spec-evidence-readers.test.ts` and `tools-hook-parity.test.ts` are the coupling points to extend.

[...]

Issue #2690, opened 2026-08-29
fix(hooks): agent-strict pre-commit gates on `.totem/cache/.spec-completed`, which no CLI path writes — `totem spec` records `spec-<hash>.json`, so a fresh worktree cannot satisfy the gate it names


satur8d:
## The defect

The managed pre-commit hook's strict tier (armed automatically for agents — `CLAUDE_CODE_AGENT` / `CLAUDE_VERSION` / `CURSOR_TRACE_ID`) blocks every commit unless `.totem/cache/.spec-completed` exists:

- `packages/cli/src/commands/install-hooks.ts:479` — `if [ ! -f ".totem/cache/.spec-completed" ]; then … BLOCKED: Run 'totem spec <issue>' before committing (strict mode)`
- `tools/pre-commit:22` — same text (canonical copy)
- `packages/core/src/config-schema.ts:592` — the `hooks.tier` doc: "'strict' adds spec-completed checks"

**No CLI path writes that sentinel.** `totem spec` (1.120.0) records a run artifact at `.totem/cache/spec-<hash>.json` (the orchestrator's run recorder; `packages/cli/src/utils.test.ts` asserts the `spec-${hash}.json` shape) and the spec file under `.totem/specs/`, and never touches `.spec-completed`. `git log -G'spec-completed'` across the whole history hits only the hook text (added in #1151, `4eba48cb`; re-emitted in #2406, `e982bafb`), its tests, and a changelog line — no writer, ever.

## How it has stayed invisible

The resident checkout carries a hand-made 1-byte `.totem/cache/.spec-completed` dated 2026-03-29, and `.totem/cache/` is gitignored — so every commit from the resident tree passes. Any fresh checkout or worktree (`totem wt create …`) has no marker, runs `totem spec` exactly as the hook says, and is still blocked. The gate is unsatisfiable through the CLI it names.

## Live repro (2026-08-29, this seat)

1. `totem wt create ci-pins --root D:/Dev/worktrees` → fresh worktree, `.totem/cache/` contains only `index-meta.json` / `sync-state.json`.
2. `totem spec ci-action-pins` → `[Spec] Spec saved to .totem/specs/ci-action-pins.md`, `.totem/cache/spec-f105059cd5f71fac.json` written. No `.spec-completed`.
3. `git commit` (agent env) → `[Totem] BLOCKED: Run 'totem spec <issue>' before committing (strict mode)`.

[...]

Issue #2685, opened 2026-08-27
fix(mail): a hand-written multi-recipient 'to:' fails silently at the resolver — hard-fail on the sender's surface (live LC specimen 2026-08-26)


satur8d:
Filed from the lc seat with a live delivery-failure specimen (2026-08-26, the mmnto-ai/liquid-city#1148 blind round).

## The defect

A hand-written outbox dispatch whose frontmatter `to:` carries a comma list (`to: lc-agy, lc-codex, lc-gemini, lc-kimi`) is treated by the resolver as ONE SCALAR address. It matches no roster agent, so the dispatch becomes invisible to every seat-scoped poll — no recipient ever serves it. The only surface of the failure is a warning line naming the file in (some) poll output.

The CLI's `--to` flag already REJECTS comma lists loudly ("unsafe characters") — the front door is guarded. The hand-written-file path is not: it fails silently at delivery time, which is the worst failure shape for a coordination bus (the sender believes the round is out; every recipient's inbox is clean).

## The specimen

The 2026-08-26 mmnto-ai/liquid-city#1148 blind-round dispatch went out as a hand-written comma-list `to:` and was undeliverable; the round survived only because the operator happened to point seats at the warning line's filename. (Historical note: comma lists in hand-written frontmatter parsed on some earlier CLI version; the path rotted silently.) LC has since codified directed per-seat dispatches as the only form (the mmnto-ai/liquid-city#1109 ruling, mechanics item 4), so LC is no longer exposed — this report is for the platform floor.

## Ask

Hard-fail the file at the resolver (or at `totem mail` walk time): a `to:` value that resolves to no roster agent should be a loud, named error on the SENDER's surface — same deterministic-floor shape as the mmnto-ai/totem#2204 identity gate — not a warning on a poll some other seat may run. Alternatively, support multi-recipient `to:` lists properly; either resolution removes the silent class. Routing and design are yours.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2693, opened 2026-08-29
Version Packages

satur8d:

# Releases
### Minor Changes

- 2e389e3: The managed git hooks now render the CONFIGURED `totemDir` (mmnto-ai/totem#2692).

  **The break they close.** `totem spec` writes its run artifact under `path.join(configRoot, config.totemDir)/artifacts/runs`, but the strict pre-commit hook's spec-evidence reader was a hardcoded `const dir = ".totem/artifacts/runs"`. Under a custom `totemDir` the writer and the reader named different trees, so the gate blocked forever on the very command it told you to run — the mmnto-ai/totem#2690 shape, re-created. The rest of the family failed the other way, silently: pre-push's `[ -f ".totem/compile-manifest.json" ]` / `[ -f ".totem/compiled-rules.json" ]` guards and post-merge / post-checkout's `grep -q '\.totem/…'` diff filters simply never matched, so the manifest, lint, badge, lockfile and claim-discipline gates were SKIPPED without a word.

  **The cure.** One resolver — `resolveHookRenderOptions(cwd, flags?) → { tier, totemDir, fallbackCmd, configPath }` — is the single config→hook-render seam, and every writer of hook text goes through it: `totem hook install`, `totem init`, the non-interactive installer, `installPostMergeHook`, the hook-manager helper scripts, and the silent pre-push upgrade; `doctor` and `doctor --parity` regenerate their canonicals from the same repo-local config values. The four builders (`buildPreCommitHook`, `buildPrePushHook`, `buildHookContent`, `buildPostCheckoutHookContent`) now take a REQUIRED options object with no defaulted parameter, so the compiler — not a reviewer — catches a forgotten thread. Each site takes the quoting regime it needs: a JSON literal inside the single-quoted `node -e '…'` reader, the plain double-quoted `[ -f ]` / `[ -d ]` words `tools/*` already ships, BRE-escaped `grep` patterns. Because the value is user-supplied, the `@mmnto/totem` config schema normalises it (a backslash becomes `/`, a leading `./` is dropped, trailing slashes are stripped — `knowledge/` used to render `grep -q 'knowledge//…'`, which never matched) and refuses an empty value or one carrying a quote, a dollar sign, a backtick, a non-ASCII character (git C-quotes those in the paths the diff filters read, so they could never match), a newline or a control character; the installer additionally refuses `.`, a `..` segment and a leading `-` — shapes whose hook diff filters could never match — naming the reason. A config that resolves but will not load (a syntax error, a refined-out `totemDir`) now renders the defaults LOUDLY: `totem hook install` prints the file and the failure before writing, instead of silently installing `.totem/` hooks for a repo whose config says otherwise.

  **Disclosed behaviour changes.**
  - `totem init` and the silent pre-push upgrade (`shieldCommand`) now honour `hooks.tier` from config. They did not before: `init` rendered from `--strict` only, and the upgrade path read nothing at all — silently downgrading a strict hook to standard on every `totem review`.
  - `totem hook install` resolves config (and its `.env` read) at the GIT ROOT rather than the invocation cwd, so installing from a subdirectory now picks up the repo's config instead of falling through to the global `~/.totem/` profile.
  - `doctor`'s always-on `Git Hooks` row gains ONE conditional check — only when `totemDir` is non-default it regenerates the canonical four (at the tier the installed hook's own totem block declares, so a tier-only difference is not reported as `totemDir` drift) and compares the totem-owned block between each hook's markers, warning with `totem hook install --force` for a totem-owned hook (a legacy hook with no end marker takes the same `--force`, with a back-up-first note) and, for a user hook carrying an appended totem block, with the delete-and-re-append line instead (`--force` rewrites the whole file). Zero added cost on the default. `doctor --parity`'s `git-hooks` rows get the same totemDir-aware canonical by construction.
  - `totem eject` removes the configured Totem directory (and `<totemDir>/hooks/shield-gate.cjs`), not a hardcoded `.totem/`, and names `.totem/prepare.cjs` explicitly (init scaffolds it at `.totem/` regardless of `totemDir`). A CUSTOM directory is swept only when it holds at least one Totem-written entry (`lessons/`, `cache/`, `compiled-rules.json`, …) — a `totemDir` pointing at `src`, `docs` or `.git` is reported and skipped, never recursively deleted; the default `.totem/` is Totem's by name and is swept as before — and a repo-local config that will not load is said so out loud before the default is ejected.
  - `totem link` probes the TARGET repo's own configured directory name (its repo-local config, else `.totem`) and writes the globs for it, refusing a target `totemDir` that names the target itself, escapes it, or carries glob metacharacters (the name is written into an ingest glob); `--unlink` is keyed by the target path, so a link made under `.totem` still unlinks after the target moved its directory.
  - `doctor`'s compiled-rules, secret-leak, secrets-file, upgrade-candidate, stale-rule, grandfathered-rule and freeze rows all read the configured directory (they took a `totemDir` parameter but were called at the default — the two secret rows attested a clean scan of a tree they never opened); the `Git Hooks` row reports when its totemDir sensor could not run instead of passing.
  - A refused `totemDir` at `totem hook install` / `totem init` is a `TotemError` with a recovery hint.

  `totemDir` is read from the REPO-LOCAL config only: the global `~/.totem/` profile declares `totemDir: '.'` to describe itself, and honouring that would re-render every config-less repo's hooks against the checkout root.

  **Not a break at the default.** Rendering at `{ tier: 'standard', totemDir: '.totem', fallbackCmd: 'pnpm dlx @mmnto/cli' }` produces `tools/{pre-commit,pre-push,post-merge}` byte-for-byte — `tools/` is unchanged and no consumer's installed hooks drift on upgrade. Minor rather than patch because the rendered output changes for non-default configs and two commands start honouring a config key they ignored.

  **After changing `totemDir`, run `totem hook install --force`** — the installed hooks keep reading the previous directory until you do. `doctor` now says so when it happens.

- 519010f: Silent failure → loud on two of the CLI's own coordination surfaces (mmnto-ai/totem#2685, mmnto-ai/totem#2690).

  **`totem mail` — sender faults (mmnto-ai/totem#2685).** A dispatch whose `to:` matches no roster agent is undeliverable to every seat-scoped poll; until now it surfaced only as a `Warning:` on whatever seat happened to poll, and the sender's own poll exited 0. Now, when the dispatch sits in an outbox **this repo hosts for a resolved seat** (path-keyed: the outbox is under `<repoRoot>/.totem/orchestration/`), the poll renders an `Error:` line, carries a structured `senderFaults[]` entry in `--json` (`{kind: 'undeliverable-to', repo, agent, file, to, hint}` — `hint` names the one-recipient rule when the `to:` is a comma list), and **exits `4`** (SENDER FAULT — a new arm: the verdict IS derived; `2` NOT DERIVED wins when both hold). A foreign repo's undeliverable dispatch stays a warning. `ecl-gc` treats an own fault as a red A2.2/A2.4 gate with a named reason (today it was red only by accident of the roster warning riding the gate-armed channel). The core agent-id guard (`@mmnto/totem` `isPathSafeAgentId`) now refuses a comma, so `mail send --to "a,b"` is refused at the front door — the whitespace-free list previously passed and was written. Help, wiki, and the signon skill name the new arm; an `Error:` line names a dispatch basename, so — like a gated poll's warnings — propagate nothing from it.

  Minor rather than patch: a new exit arm and a new `--json` key are contract additions scripts can observe (the mmnto-ai/totem#2312 / mmnto-ai/totem#2204 exit-contract precedent class).

  **Strict pre-commit gates on spec evidence (mmnto-ai/totem#2690).** The agent-strict pre-commit block gated on `.totem/cache/.spec-completed`, which no CLI path has ever written — every fresh checkout or `totem wt create` worktree ran the command the hook named and was still blocked. The block now passes on what `totem spec` actually writes: a mmnto-ai/totem#2100 run artifact under `.totem/artifacts/runs/` whose top-level `admission.runMetadata.caller` is `spec` (written on every successful run, `--fresh` included), read JSON-aware in the hook via `node -e` — a substring match would have accepted a `review` artifact whose prompt merely quotes the key (the review of this very change would have). The former hand-set marker `.totem/cache/.spec-completed` is **retired outright** — no CLI version ever wrote it, so there is nothing to stay compatible with (operator ruling: no legacy shims while there is no hard consumer); a checkout without a spec run artifact blocks until `totem spec` runs, and the message says so. A reader that cannot run (no `node` on PATH, a crash) is reported distinctly from missing evidence and still fails closed. The hook prints `[Totem] spec evidence: <artifact> (<createdAt>, N days old)` on every strict commit so a stale pass is visible; a freshness rule is a separate, named follow-up. `tools/pre-commit` and the `hooks.tier` schema doc move in lockstep; the repo's pre-managed-era, unregistered `.gemini/hooks/BeforeTool.js` — a second reader of the dead marker — is deleted rather than kept in step. Consumers pick up the new hook text on their next `pnpm install` (prepare) or `totem hook install`.

### Patch Changes

- Updated dependencies [2e389e3]
- Updated dependencies [519010f]
  - @mmnto/totem@1.121.0
### Patch Changes

- 2e389e3: The managed git hooks now render the CONFIGURED `totemDir` (mmnto-ai/totem#2692).

  **The break they close.** `totem spec` writes its run artifact under `path.join(configRoot, config.totemDir)/artifacts/runs`, but the strict pre-commit hook's spec-evidence reader was a hardcoded `const dir = ".totem/artifacts/runs"`. Under a custom `totemDir` the writer and the reader named different trees, so the gate blocked forever on the very command it told you to run — the mmnto-ai/totem#2690 shape, re-created. The rest of the family failed the other way, silently: pre-push's `[ -f ".totem/compile-manifest.json" ]` / `[ -f ".totem/compiled-rules.json" ]` guards and post-merge / post-checkout's `grep -q '\.totem/…'` diff filters simply never matched, so the manifest, lint, badge, lockfile and claim-discipline gates were SKIPPED without a word.

  **The cure.** One resolver — `resolveHookRenderOptions(cwd, flags?) → { tier, totemDir, fallbackCmd, configPath }` — is the single config→hook-render seam, and every writer of hook text goes through it: `totem hook install`, `totem init`, the non-interactive installer, `installPostMergeHook`, the hook-manager helper scripts, and the silent pre-push upgrade; `doctor` and `doctor --parity` regenerate their canonicals from the same repo-local config values. The four builders (`buildPreCommitHook`, `buildPrePushHook`, `buildHookContent`, `buildPostCheckoutHookContent`) now take a REQUIRED options object with no defaulted parameter, so the compiler — not a reviewer — catches a forgotten thread. Each site takes the quoting regime it needs: a JSON literal inside the single-quoted `node -e '…'` reader, the plain double-quoted `[ -f ]` / `[ -d ]` words `tools/*` already ships, BRE-escaped `grep` patterns. Because the value is user-supplied, the `@mmnto/totem` config schema normalises it (a backslash becomes `/`, a leading `./` is dropped, trailing slashes are stripped — `knowledge/` used to render `grep -q 'knowledge//…'`, which never matched) and refuses an empty value or one carrying a quote, a dollar sign, a backtick, a non-ASCII character (git C-quotes those in the paths the diff filters read, so they could never match), a newline or a control character; the installer additionally refuses `.`, a `..` segment and a leading `-` — shapes whose hook diff filters could never match — naming the reason. A config that resolves but will not load (a syntax error, a refined-out `totemDir`) now renders the defaults LOUDLY: `totem hook install` prints the file and the failure before writing, instead of silently installing `.totem/` hooks for a repo whose config says otherwise.

  **Disclosed behaviour changes.**
  - `totem init` and the silent pre-push upgrade (`shieldCommand`) now honour `hooks.tier` from config. They did not before: `init` rendered from `--strict` only, and the upgrade path read nothing at all — silently downgrading a strict hook to standard on every `totem review`.
  - `totem hook install` resolves config (and its `.env` read) at the GIT ROOT rather than the invocation cwd, so installing from a subdirectory now picks up the repo's config instead of falling through to the global `~/.totem/` profile.
  - `doctor`'s always-on `Git Hooks` row gains ONE conditional check — only when `totemDir` is non-default it regenerates the canonical four (at the tier the installed hook's own totem block declares, so a tier-only difference is not reported as `totemDir` drift) and compares the totem-owned block between each hook's markers, warning with `totem hook install --force` for a totem-owned hook (a legacy hook with no end marker takes the same `--force`, with a back-up-first note) and, for a user hook carrying an appended totem block, with the delete-and-re-append line instead (`--force` rewrites the whole file). Zero added cost on the default. `doctor --parity`'s `git-hooks` rows get the same totemDir-aware canonical by construction.
  - `totem eject` removes the configured Totem directory (and `<totemDir>/hooks/shield-gate.cjs`), not a hardcoded `.totem/`, and names `.totem/prepare.cjs` explicitly (init scaffolds it at `.totem/` regardless of `totemDir`). A CUSTOM directory is swept only when it holds at least one Totem-written entry (`lessons/`, `cache/`, `compiled-rules.json`, …) — a `totemDir` pointing at `src`, `docs` or `.git` is reported and skipped, never recursively deleted; the default `.totem/` is Totem's by name and is swept as before — and a repo-local config that will not load is said so out loud before the default is ejected.
  - `totem link` probes the TARGET repo's own configured directory name (its repo-local config, else `.totem`) and writes the globs for it, refusing a target `totemDir` that names the target itself, escapes it, or carries glob metacharacters (the name is written into an ingest glob); `--unlink` is keyed by the target path, so a link made under `.totem` still unlinks after the target moved its directory.
  - `doctor`'s compiled-rules, secret-leak, secrets-file, upgrade-candidate, stale-rule, grandfathered-rule and freeze rows all read the configured directory (they took a `totemDir` parameter but were called at the default — the two secret rows attested a clean scan of a tree they never opened); the `Git Hooks` row reports when its totemDir sensor could not run instead of passing.
  - A refused `totemDir` at `totem hook install` / `totem init` is a `TotemError` with a recovery hint.

  `totemDir` is read from the REPO-LOCAL config only: the global `~/.totem/` profile declares `totemDir: '.'` to describe itself, and honouring that would re-render every config-less repo's hooks against the checkout root.

  **Not a break at the default.** Rendering at `{ tier: 'standard', totemDir: '.totem', fallbackCmd: 'pnpm dlx @mmnto/cli' }` produces `tools/{pre-commit,pre-push,post-merge}` byte-for-byte — `tools/` is unchanged and no consumer's installed hooks drift on upgrade. Minor rather than patch because the rendered output changes for non-default configs and two commands start honouring a config key they ignored.

  **After changing `totemDir`, run `totem hook install --force`** — the installed hooks keep reading the previous directory until you do. `doctor` now says so when it happens.

- 519010f: Silent failure → loud on two of the CLI's own coordination surfaces (mmnto-ai/totem#2685, mmnto-ai/totem#2690).

  **`totem mail` — sender faults (mmnto-ai/totem#2685).** A dispatch whose `to:` matches no roster agent is undeliverable to every seat-scoped poll; until now it surfaced only as a `Warning:` on whatever seat happened to poll, and the sender's own poll exited 0. Now, when the dispatch sits in an outbox **this repo hosts for a resolved seat** (path-keyed: the outbox is under `<repoRoot>/.totem/orchestration/`), the poll renders an `Error:` line, carries a structured `senderFaults[]` entry in `--json` (`{kind: 'undeliverable-to', repo, agent, file, to, hint}` — `hint` names the one-recipient rule when the `to:` is a comma list), and **exits `4`** (SENDER FAULT — a new arm: the verdict IS derived; `2` NOT DERIVED wins when both hold). A foreign repo's undeliverable dispatch stays a warning. `ecl-gc` treats an own fault as a red A2.2/A2.4 gate with a named reason (today it was red only by accident of the roster warning riding the gate-armed channel). The core agent-id guard (`@mmnto/totem` `isPathSafeAgentId`) now refuses a comma, so `mail send --to "a,b"` is refused at the front door — the whitespace-free list previously passed and was written. Help, wiki, and the signon skill name the new arm; an `Error:` line names a dispatch basename, so — like a gated poll's warnings — propagate nothing from it.

  Minor rather than patch: a new exit arm and a new `--json` key are contract additions scripts can observe (the mmnto-ai/totem#2312 / mmnto-ai/totem#2204 exit-contract precedent class).

  **Strict pre-commit gates on spec evidence (mmnto-ai/totem#2690).** The agent-strict pre-commit block gated on `.totem/cache/.spec-completed`, which no CLI path has ever written — every fresh checkout or `totem wt create` worktree ran the command the hook named and was still blocked. The block now passes on what `totem spec` actually writes: a mmnto-ai/totem#2100 run artifact under `.totem/artifacts/runs/` whose top-level `admission.runMetadata.caller` is `spec` (written on every successful run, `--fresh` included), read JSON-aware in the hook via `node -e` — a substring match would have accepted a `review` artifact whose prompt merely quotes the key (the review of this very change would have). The former hand-set marker `.totem/cache/.spec-completed` is **retired outright** — no CLI version ever wrote it, so there is nothing to stay compatible with (operator ruling: no legacy shims while there is no hard consumer); a checkout without a spec run artifact blocks until `totem spec` runs, and the message says so. A reader that cannot run (no `node` on PATH, a crash) is reported distinctly from missing evidence and still fails closed. The hook prints `[Totem] spec evidence: <artifact> (<createdAt>, N days old)` on every strict commit so a stale pass is visible; a freshness rule is a separate, named follow-up. `tools/pre-commit` and the `hooks.tier` schema doc move in lockstep; the repo's pre-managed-era, unregistered `.gemini/hooks/BeforeTool.js` — a second reader of the dead marker — is deleted rather than kept in step. Consumers pick up the new hook text on their next `pnpm install` (prepare) or `totem hook install`.
### Patch Changes

- Updated dependencies [2e389e3]
- Updated dependencies [519010f]
  - @mmnto/totem@1.121.0
## @mmnto/pack-rust-architecture@1.121.0

## @mmnto/pack-agent-security@1.121.0


satur8d:
## Cut note — 1.121.0 · classification: `contract-bearing: consumer config schema · CLI surface (exit-code contract) · ECL/mail formats · init-distributed content`

Filed against the release-cut sensor above (mmnto-ai/totem#2325; doctrine `packaging-conventions.md` § "Release planning — the meaningful-cut discipline", the REQUIRED advance classification). The train since `@mmnto/cli@1.120.0`: `519010f4` (mmnto-ai/totem#2691, silent → loud on two coordination surfaces) + `2e389e30` (mmnto-ai/totem#2701, the managed hooks render the configured `totemDir`) + the CI SHA-pins + the doctrine pin 0.1.38. The two apparatus slices are spike-only and unpublished by charter. The VP diff at `0e6ce301` carries both changesets; bumps: cli / core / mcp / packs → 1.121.0.

### The flagged entries, each verified against its diff

| entry | class | verdict |
|---|---|---|
| `2e389e3` — `config-schema.ts` | consumer config schema | **contract-bearing in class, no consumer moves.** `totemDir` is now normalised (backslash → `/`, leading `./`, trailing slash) and a value that is empty or carries a quote, `$`, a backtick, a non-ASCII character or a control character fails config load. Every cohort repo pins the default `.totem` (verified on the pin roster); the refusals name shapes none of them uses. The rendered hooks are byte-identical to `tools/*` at the default, so no consumer's installed hooks drift. Only a repo that CHANGES `totemDir` owes `totem hook install --force` — documented in the docstring, wiki and changelog. |
| `519010f` — `config-schema.ts` | consumer config schema | **dismissed as contract-bearing:** the `hooks.tier` docstring only (the evidence-read wording); no schema shape changed. |
| `519010f` — `mail.ts`, `ecl-gc.ts` | ECL / mail formats | **contract-bearing, additive.** A new poll exit arm `4` (SENDER FAULT: an undeliverable `to:` in an outbox the polling repo hosts) and a new `--json` key `senderFaults[]`; `ecl-gc` reddens A2.2/A2.4 with a named reason on an own fault. Scripts that branch on exit codes see a new value only when they have such a dispatch — and the cure is theirs (one recipient per dispatch, or broadcast). The signon skill text names the arm (shipped in the same PR). |
| `519010f` — `init-templates.ts` | init-distributed content | **contract-bearing, self-repairing.** The strict pre-commit block gates on the `totem spec` run artifact instead of the never-written `.spec-completed` marker. Consumers pick up the new hook text on their next `pnpm install` (the `prepare` wrapper runs `totem hook install`) or `totem hook install` — no `totem init` re-run. The deleted `.gemini/hooks/BeforeTool.js` was repo-local and unregistered; the managed template is untouched. |
| `519010f` — `index.ts` | CLI surface | **contract-bearing = the exit-4 arm above** (its registration); nothing else on the CLI surface moved. |

### Choreography

- **Which consumers move:** `mmnto-ai/totem-strategy` and `mmnto-ai/totem-status` (the non-LC pins → `^1.121.0`); `mmnto-ai/liquid-city` on its own cadence (same two obligations when it bumps). No consumer needs a `totem init` re-run; hooks self-repair on install.
- **In what order:** (1) this VP-PR merges on the operator's cut word → the Release workflow publishes `@mmnto/cli` 1.121.0, `@mmnto/totem`, `@mmnto/mcp` and the packs; (2) totem-lane post-publish smoke (`npm i -g @mmnto/cli@1.121.0`, `totem doctor` on this repo); (3) strategy-claude bumps strategy + totem-status to `^1.121.0` — each `pnpm install` re-renders the pre-commit strict block via `prepare`; any own-outbox dispatch with a bad `to:` surfaces as exit 4 at the next poll and is fixed at the sender; (4) liquid-city bumps when its seat takes it. The seed-20 apparatus pin stays `57f9af5a` — the controls-at-the-pin PR bases on post-release `main` (K6 compares by digest; the bump touches none of the pinned files).
- **Who owns which half:** publish — the operator's cut word + the Release workflow; smoke — totem-claude; non-LC pin bumps — strategy-claude (declared 2026-08-30T04:22Z); LC — the lc seat. The ADR-115 quota row for this train (Prop 309) is logged on mmnto-ai/totem-strategy#974 by totem-claude at the merge: the qualifying item is `519010f4` (a degraded path presenting as success made loud, with its induced-failure tests), `2e389e30` a second (gates that silently never matched, now rendered and refused with executed-under-`sh` tests).

Advisory noted (Tenet 13): the cut gate is unchanged — the operator's merge word on this PR.

PR #2686, opened 2026-08-28
spike(spine-adopt): Spine adopt-round spike pair + certification actuator (review vehicle)

satur8d:
## Spine adopt-round spike pair + certification actuator (review vehicle — merge is NOT requested)

This branch is the executed 2026-08-27 operator word (durable record: mmnto-ai/totem-strategy#1138): the freeze-safe adoption spike pair, the serena pilot, and the ruled certification actuator slice. Everything lives under `spikes/` plus a branch-scoped CI workflow; `packages/*` and the rule-compilation freeze are untouched. The decision report is `spikes/spine-adopt/DECISION.md` (the conditional-adoption record the ADR-103 amendment cites, post codex-verification fold).

**This PR exists as the bot-review vehicle. It does not close any issue, and the branch does not merge without a separate operator word.**

### What's here

- **Spike 1 — record→Rego→wasm differential:** lowerer, per-record pinned `opa build` chains, rust-opa-wasm/wasmtime host + regorus reference arm, three-way comparator (72/72 MATCH; mutation-tested), OPA ABI census, expressibility census over the 500-rule corpus.
- **Spike 2 — SMT obligations:** ten obligations on pinned Z3 5.1.0 + cvc5 1.3.4 (CLI + z3 Rust binding), evidence chains, the D6 encoding-sensitivity specimen.
- **Enrichment probes:** wazero (Go) conformance, egg e-graph probe (measured REJECT), wasm validate/mutate, serena pilot (measured FAIL, uninstalled).
- **Certification actuator (binding condition 1 of the conditional adoption):** `spikes/spine-adopt/src/certify.mts` + `certify-verify.mts`, sentinel-based pre-publication gate, five typed blocked classes + two structural gates, chains-as-certificates (certify is the only publisher), negative conformance fixtures error-rowed on all three hosts with host-originated errors.

### Review guidance

The load-bearing surfaces (in review-value order):

1. `spikes/spine-adopt/src/certify.mts` + `src/certify-verify.mts` — the actuator; its contract is `.totem/specs/spine-spike.md` § Actuator slice (operator-ruled, verbatim-normative).
2. `spikes/spine-adopt/src/lower.mts` + `rego/LOWERING.md` — the record→Rego semantics contract.
3. `host/src/main.rs`, `wazero-probe/*.go`, `smt/src/*.rs` — the three host arms + solver harness.
4. `src/lib/*.mts`, `src/*.mts` — census/facts/verdict harnesses.

[...]

PR #2694, opened 2026-08-29
spike(spine-adopt): seed-20 apparatus slice (G1–G10 + K4) — record-set loader, run manifest, reject rows, published lowering, K5b control, wazero verdict path; no run of record

satur8d:
**Apparatus, never scoring.** The seed-20 compilation-target trial's apparatus slice, per the charter mmnto-ai/totem-strategy#1154 (`operations/310-seed20-target-preregistration.md`, DRAFT v0.3 @ `40ee2a5`; registers as v1 at merge) § 2 G1–G10 as corrected at v0.3, the bus's 2026-08-29T05:09Z/05:25Z rulings, and the operator's go. Design record: `.totem/specs/seed20-apparatus.md`. Lane: totem-claude = apparatus-runner.

**What this PR does NOT do (each a refusal condition, all honored):** no run of record; nothing scored; no apparatus pin mailed (that follows mmnto-ai/totem-strategy#1154's merge and the floor ruling on mmnto-ai/totem-strategy#1138, after this branch is brought up to main); no K-control run; `spikes/spine-adopt/artifacts/**` and `wazero-probe/artifacts/**` stay at the **7-specimen baseline** plus the additive files below — every seed20 artifact produced by local integration runs was restored/cleaned before staging; the § 7 fidelity fix stays DEFERRED (`PRECEDING_LINE_MARKERS` byte-stable; K3's chain sha `r87aff037d7de47a7` reproduces byte-identically).

## Deliverables (G1–G10 + K4)

- **G1** `src/lib/record-sets.mts` — `SPIKE_RECORD_SET` ∈ {`specimens` (default, unchanged behavior), `seed20`}; unknown value = hard error. The seed = 22 byte-identical copies under `seed/records/` (blob oids equal `2a7135762b6aedc9cd3099ab3f42e029ee34092e:.totem/rules/<file>`, `seed/PIN.md` lists them) — `ruleId` from each record's `curation.sourceLesson`; twins share their `lessonHash` by construction; packages `totem.spike.r<lessonHash>` / `_<language>` for the two twinned rules. Every hardcoded 7/24/id assert in `lower`, `build-wasm`, `certify`, `certify-verify`, `verify-records`, `facts`, `shipped-verdicts`, `compare` now derives from the loaded set.
- **G2** per-record in-scope `inlineFilePath` (`.js` for the twins) + `SEED20_EXTRA_PROBE_PATHS` (38) so every seed glob has ≥ 1 matching and ≥ 1 non-matching probe; the probe set is per-record-set (extending the shared list would move all 7 committed chains via `globs.json`).
- **G3** `src/lib/lowering-gate.mts` + `intakeRecordSet()` — a shipped-compile rejection (`1a7080eb`, unregistered `json`) and a § Lowering 4 target rejection (`0e01112d`, `lookaround`) are **scored rows** (`lowering-rejects.rejects[]` with `stage`, `class`) on seed20; `specimens` keeps the throw.
- **G4** `seedEntry` on `lowering-rejects.lowered[]`, `facts-index.bundles[]`, shipped rows, `differential-report.pairs[]`, and chains (seed20 only — committed specimens chains are byte-locked by INV 2 drift-detection).
- **G5** `src/manifest.mts` (`npm run manifest`, step 0 of `all`): run commit, record pin + per-record sha256, fixtures, probe paths, tracked-path list (`artifacts/tracked-paths.txt`, for T5), bundle sha256s, resolved toolchain (opa 1.20.0 · rust 1.96.1 · node · `@ast-grep/napi` 0.42.0 · wazero 1.12.0 · go 1.26.2 · wasmtime 47.0.4 · regorus 0.11.0 · opa-wasm 0.3.2), platform, `record-lower.ts` sha256. Embedded in every later artifact as **`runManifestSha256`** (`manifestSha256` is already the OPA entrypoint/import digest bound in every chain).
- **G6** `stripRegoStringLiterals()` — the § Lowering 3 self-assert tests raw-string SYNTAX; `5da43ea6` lowers; a three-way control pins it.
- **G7** `artifacts/lowered/<pkg>/{policy.rego,globs.json}` published (copied + byte-identity assert against the chain-bound file).
- **G8** `m3-malformed-lines` control bundle (seed20 only): an ERROR row on shipped, wasmtime, regorus **and wazero**; `MALFORMED-FACTS-CONTROL` explanation class on every pairing; `differential-report.errorRows` accounting (`fromK5bControl` / `outsideK5b`).
- **G9** `.github/workflows/spine-spike-linux.yml`: `on.push.paths: [spikes/spine-adopt/**]` (+ `workflow_dispatch`) — the Linux arm now runs on this branch.
- **G10** (Go) `wazero-probe/`: row counts derive from `facts-index.json.bundleCount` cross-checked against the facts dir; `isRequired` follows the record set; the join is on `(ruleId, specimen)` — twins never fan; new `artifacts/wazero-pairs.json` in `differential-report`'s shape (`pairs[]`, `summary`); 18 tests incl. two mutants.
- **K4** `SPIKE_SWAP_EXAMPLES=<recordId>` swaps a record's bad/good bundles (`provenance.swapped: true`).

## Integration (local, NOT of record — artifacts not committed)

`SPIKE_RECORD_SET=seed20 npm run all && npm run differential && go run . -spike-root ..`: manifest 22 records · 20 lowered, 20 published · 4 reject rows (2 census evidence + `0e01112d` target-lowering/lookaround + `1a7080eb` shipped-compile) · 43 bundles · shipped / opa / regorus 43 rows each, `errorRows` 1 per arm (`fromK5bControl: 1, outsideK5b: 0`) · 20 chains certified (all PASS), 8 blocked fixtures · compare **126 MATCH / 3 EXPLAINED (the K5b control) / 0 UNEXPLAINED** · wazero probe **43 pairs: 42 MATCH / 1 EXPLAINED / 0 UNEXPLAINED**, all checks green.

**One defect found at the TS↔Go seam and fixed here** — a class neither leg could see: the probe refused the shipped arm's K5b error row ("carries no arms"; no dispatcher ran, so the row has `error` and no `arms[]`). `shippedRow` gains `error` and normalises it like the wasmtime/wazero error rows; a verdict-claiming row without arms stays fatal (`shipped_errorrow_test.go`). Second: the probe's "corpus is DISCRIMINATING" check demanded a multi-violation row — see the finding below — and is now set-aware like its TS twin.

**Finding for the scorer (frozen-input property, not an apparatus defect):** no bundle in the seed carries more than one ast match (`max matchCount 1` across all 43), so the **ast ordinal axis is unexercised on seed20** — T8's `ordinal` key and the `ORDINAL-DERIVATION-ONLY` class cannot fire. Both readers demote the multi-violation corpus check to a named skip on seed20 and keep it on `specimens`. Mailed to the bus for the charter's v1.

## Baseline artifact bytes that changed (7-specimen set) and why

`facts/**` (24), `chains/**` (7), `blocked/**` (8): **byte-identical**. Changed: all `writeArtifact` outputs gain the trailing `runManifestSha256`; `seedEntry: null` per row on the specimens set (G4); `lowering-rejects.json` gains `published` per row + G6/G7 check rows; `facts-index.json` +1 check (bundle filenames carry the `<ruleId>-<specimen>-` join prefix); `differential-report.json` gains `errorRows`, the new explanation class, +1 check; `certification-invariants.json` INV 8 check name counts 15 scanned modules (was 12). `opa-verdicts.json`, `regorus-verdicts.json`, `wazero-report.json`, `wazero-verdicts.json` are restored to the committed bytes — their only delta is machine timings (pre-existing non-determinism). New: `manifest.json`, `tracked-paths.txt`, `lowered/**` (7 packages), `wazero-probe/artifacts/wazero-pairs.json` (24 pairs, all MATCH).

## Verification

Specimens baseline: `all` + `differential` + probe all green (72/72 compare MATCH; wazero 20/20); seed20 as above; `go vet/build/test` green (`gofmt -l` empty); `prettier --check` clean on every changed file including the 22 seed YAMLs; the spike directory is outside the workspace ESLint gate and baseline-dirty (53 pre-existing errors) — delta on touched files is −2/+3, all three new modules clean; `totem lint --staged` PASS (0 errors); no `packages/**` change, no Rust change. Local review lane (2/2 lanes, on a 609 k-char diff far past the 50 k truncation window): 1 CRITICAL — **refuted at source**: it read `s.seedEntry` in `build-wasm.mts:254` as the wrong receiver, but `s = rowById.get(row.specimen)` is a `RecordRow`, which declares `seedEntry: string | null` (`record-sets.mts:40`), and the seed20 run populates it on every chain; 1 WARN (certify-verify reads `lowering-rejects.json` directly) — exercised green on both sets (INV 1 derives 22 − 2 = 20 on seed20); 2 INFO (unvalidated casts on same-run artifacts) — cosmetic. Stopped at round 0 with the dispute recorded rather than re-guessing a truncated diff: `local-lane: 54577ee2 round=0 settled=false lanes=2/2`. **Merge is on the operator's word; the apparatus pin is a later, separate mail.**

## Bot round 1 (fold spec: `.totem/specs/seed20-apparatus-round1.md`)

22 findings (CR 19 · GCA 2 · Greptile 1): **17 fixed, 5 declined with grounds** (GCA's two criticals refuted empirically — a JSON null into a non-pointer Go `bool`/`int` is a no-op, measured; three target the frozen seed records, which the apparatus may not edit and the scorer has already typed). The fold was implemented by two build legs and — correcting a miss on the first presentation, where the local review lanes were treated as the review-leg — read by a **scoped Opus falsification leg before it was pushed**: no BLOCKING, 5 MATERIAL folded (C3 uniform on all Go arms + test; the comparator now re-hashes every fact bundle against `manifest.bundles` before comparing, so the run identity *binds* the evaluated bytes; seven synthetic C2 control pairs exercise the per-arm designed-error predicate on every run; two narrations corrected; the byte-delta enumeration amended), 12 MINOR (six folded: `ls-tree` doc, codepoint sort in the digest preimage, two comment premises, two apostrophes; six noted: T5's arm is reachable only for an empty fill; `manifest.mts` self-checks leave no artifact trace by design; `ls-tree HEAD` cannot see the publishing commit's own additions, so the committed baseline's sweep set is the parent's — the run of record at a pin reads its own; the TS `Arm` union has no off-table arm; a basename collision in the pin-side enumeration is unreachable at this pin; `EMPTY RESULT SET` is spelled in three unrelated predicates).

Baseline bytes after the fold (specimens set): `facts/**`, `chains/**`, `blocked/**`, `lowered/**` still byte-identical (K3 holds). Changed: `runManifestSha256` + `bundlesSha256` on the eight post-facts artifacts; `recordSet` in `facts-index.json`; `workingTreeClean` (recorded `false` — the baseline is regenerated from the tree carrying its own fold; a pinned run of record reads `true`), `bundlesSha256`, `ls-tree` source text in `manifest.json`; fixed `opa-abi-census.json` headline keys + `censusedPolicies`; the amended check rows (T13 agreement, T22 detail, the C2 self-probe rows, FACT BYTES); the aligned `explanationClasses[MALFORMED-FACTS-CONTROL]` text in `differential-report.json` and `wazero-pairs.json`; `seedEntry: null` (was `""`) ×24 in `wazero-pairs.json`; and the run-identity churn inherent to G5 (`runCommit`, `generatedAt`, `trackedPaths.count/.sha256`, `tracked-paths.txt` +47 — the paths this PR's own earlier commits added).

Related: mmnto-ai/totem-strategy#1154 · mmnto-ai/totem-strategy#1138 · mmnto-ai/totem#2686 (the spike) · mmnto-ai/totem-strategy#288.

https://claude.ai/code/session_01Vw98oJ3HfnAgKmS1BKydy8


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
Round 1 disposition — mmnto-ai/totem#2694. Reviewed head `25c2a705` (`4d32f07e` in between was 20 lines of `toolchain.lock` comments); fix commit `86a81bd9`. Method: every finding verified at source before acting (two refuted empirically, three on frozen inputs); the fold specified in `.totem/specs/seed20-apparatus-round1.md`, implemented by two build legs (TS · Go), read adversarially by a scoped Opus falsification leg BEFORE pushing — the leg that was owed on the first presentation and skipped; its verdict: **no BLOCKING; 5 MATERIAL, all folded** (C3 had landed on one of three Go arms — now uniform, with a test; the bundle digest recorded what `facts` wrote but nothing re-verified the bytes the arms read — the comparator now re-hashes every bundle against the manifest before comparing, and the "two runs, one identity" narration is corrected to the real per-manifest case; the TS half of C2 had no falsifier and never fired on the committed baseline — seven synthetic control pairs now exercise it on every run; T21's narration claimed more than its predicate does — stated honestly as a regression tripwire plus one live conjunct; the declared byte delta was incomplete — enumerated); 12 MINOR (six folded, six noted in the PR body) — then the seam re-run: seed20 through the TS pipeline and the Go probe — compare 126 MATCH / 3 EXPLAINED / 0 UNEXPLAINED, wazero pairs 42 / 1 / 0, the K5b control EXPLAINED on every pairing under the per-arm designed-error predicate; specimens baseline regenerated, `facts/**` · `chains/**` · `blocked/**` · `lowered/**` byte-identical (K3's chain sha holds).

@coderabbitai — in your numbering:

**FIXED (16):**
1. `build-wasm.mts` — `Array.isArray(manifest.records)` guard, throws naming the artifact and `npm run manifest`.
2. `build-wasm.mts` — `opa-abi-census.json` headline keys are fixed (`hostImplementedBuiltinsAcrossAll` / `distinctImportsAcrossAll`) plus `censusedPolicies`; one committed baseline key changed by design (no reader of the old spelling anywhere in the spike).
3. `lower.mts` — zero fact bundles ⇒ `NOT MEASURED — no fact bundles on disk`, never "unreachable on all 0".
4. `verify-records.mts` — `spawnSync` `status === null` / `error` ⇒ a FAILED check naming the git execution failure; only a genuine non-zero exit is the `pin-commit-not-present-locally` skip.
5. `verify-records.mts` — the pinned tree is enumerated too (`git ls-tree -r --name-only <pin> -- .totem/rules/`), compared both ways; a file at the pin absent locally is a drift row.
6. `certify-verify.mts` — `committedChainsApply` is keyed on `manifest.recordSet`, with a FAILED-on-mismatch check that `SPIKE_RECORD_SET` agrees with the manifest.
7. `certify-verify.mts` — `changedKeys` / `addedKeys` are `null` when there is no committed referent.
8. `certify.mts` — the sentinel count check's detail names `COUNT MISMATCH — expected N, found M` and lists rejects.
9. `compare.mts` + `wazero-probe/compare.go` — an empty error string is not an error row on any arm (one definition per side).
10. `shipped-verdicts.mts` — the K5b check asserts the row it pushes (`fired`/`matchCount` null, `arms` empty, error begins `MALFORMED FACTS`, non-empty malformed-member list); a control bundle whose `lines[]` are all strings now FAILS it.
11. `shipped-verdicts.mts` — error rows are excluded from the `matchCount !== lineCount` assertion (same predicate as the asymmetry measurement).
12. `manifest.mts` — `trackedPaths` from `git ls-tree -r --name-only HEAD` (the recorded commit), plus `workingTreeClean` recorded from `git status --porcelain` (recorded, not gated — it is `false` in the committed baseline because the baseline is regenerated from the tree that carries its own fold; the run of record at a pin will read `true`).
13. `wazero-probe/compare.go` — `compareControlPair` explains the K5b pair only when EACH side carries ITS OWN arm's designed error: your proposed fix ("both equal `EMPTY RESULT SET`") would have failed the shipped–wazero pairing, because the shipped arm's designed error is `MALFORMED FACTS…` (it refuses the bundle; it never evaluates a policy). Per-arm predicate, texts as named constants in `compare.go` (imported by the tests, not re-spelled), other paired errors UNEXPLAINED with both messages; the same predicate in `compare.mts` for shipped–opa / shipped–regorus / opa–regorus, with regorus's marker tightened to `UNDEFINED/non-object` (the opa text also contains `UNDEFINED` — a one-way overlap the TS leg found).
14. `wazero-probe/main.go` — `facts-index.json` now carries `recordSet` (written by `facts.mts`); the probe REFUSES a mismatch or a missing field (named class `RECORD-SET IDENTITY`) before any cardinality or join conclusion; tested including the ordering against a disagreeing bundle count.
15. `wazero-probe/compare.go` — `pairResult.SeedEntry` is `*string`; absent ⇒ JSON `null` (key present), matching the TS side; the 24 specimens pairs changed `""` → `null`.
16. `wazero-probe/probe_test.go` — keyed fields in the `violation` / `event` literals.

**DECLINED (3), with grounds:**
- `policy.rego` (the generated `facts_wellformed` lacks a `fileText` type clause): true, but it changes every emitted `policy.rego` — all seven committed chains, INV 2's byte-equality, and K3's chain sha — the same class as the charter's deferred § 7 fidelity fix. Tracked for the promotion slice beside § 7; the lowerer stays byte-stable for the trial by constraint.
- `r14-5da43ea6…` (require `--` before the untrusted argument) and `r14-be74c55c…` / `r14-6b2b62eb…` (narrow the patterns): these are byte-identical copies of the FROZEN seed at `2a713576` (charter constraint 5 — the apparatus may not edit them). Their defects are the scorer's typed dispositions already: `be74c55c` and `6b2b62eb` are 2026-08-22 `defective-source` discards; `5da43ea6`'s locus seam is mmnto-ai/totem-strategy#1093.

@greptileai — P1 **FIXED**: the run identity now binds the bundle set. `runManifestSha256` stays the pre-facts identity (records, inputs, toolchain); after `facts` fills `bundles[]`, `bundlesSha256` (canonical JSON of the sorted `{fixtureId, sha256}` list — order-independent, byte-sensitive, tested) is stored in the manifest, and every artifact written after facts embeds BOTH; a post-facts writer refuses to run without it. Two manifests naming different evaluated fact bytes now differ in identity. Also landed on the same seam: `facts-index.json` carries `recordSet`, which the Go probe validates before scoring (item 14).

@gemini-code-assist — both criticals **DECLINED**, with a measurement: `json.Unmarshal` of `{"fired":null,"matchCount":null}` into a struct with non-pointer `bool`/`int` returns `err=<nil>` and leaves the fields untouched — the Go spec makes a JSON null into a non-pointer scalar a no-op, not a failure. The probe read exactly that row (the shipped K5b error row) in every passing seed20 run before and after this fold; `shippedRow.Error *string` already carried the row into the error path. No change to `Fired`/`MatchCount` types; the `probe_test.go` dereference concern therefore does not arise.

Also folded (owner's rulings on two build-leg residues): the published `explanationClasses[MALFORMED-FACTS-CONTROL].fires` text in both `differential-report.json` and `wazero-pairs.json` now states the per-arm predicate (accuracy over the enumerated byte delta; listed in the PR body). Not in this round but worth naming: the one seam defect between the two halves (the probe refusing the shipped K5b error row as "carries no arms") was found by running the seam, not by review — the PR body records it.

Local lane: on the reviewed head the pre-fold round-0 line was `local-lane: 54577ee2 round=0 settled=false lanes=2/2` (its one CRITICAL refuted at source — the receiver it named is a `RecordRow`, which declares `seedEntry`). For the folded lineage `totem review --covariate` reports no verdict artifact recorded (the pre-push gate mints none), so there is no line to carry for `86a81bd9`; the scoped Opus falsification leg above is the review of record for the fold, and its five MATERIAL findings are folded in this commit.

Issue #2700, opened 2026-08-30
spec: `totem spec` refuses an unanchored topic + `--from <record>` binds the hand record; the strict pre-commit gate reads ANCHORED + non-empty-by-content — 4/4 slug-topic runs confabulated this week; 53/53 runs read `similarity-only:8` (the artifact records no anchor); an EMPTY draft satisfies the gate today (ruled with strategy 2026-08-30; queued behind the seed-20 run of record)


satur8d:
## Charter

**Ruled with strategy 2026-08-30 (strategy-claude `totem-strategy:.totem/orchestration/strategy-claude/outbox/2026-08-30T0059Z-totem-claude-re-totem-spec-confabulates-ruled-concur.md`, in reply to `totem:.totem/orchestration/totem-claude/outbox/2026-08-29T2248Z-strategy-claude-totem-spec-slug-confabulation.md`; operator's ask 2026-08-29 to raise it):** concur on items 1–2 below with two amendments; item 3 is doctrine and strategy authors it. **Queued behind the seed-20 run of record (mmnto-ai/totem-strategy#1154 § 6), with mmnto-ai/totem#2698. No build before the operator's word.** Slice owner: totem-claude.

Standing line this lands under (strategy): "`totem spec` writes a contract only when it is grounded; its output is one retrieval input, never the contract" and "spec-first is a ritual with a gate, not a text to obey".

## The class

`totem spec <slug>` on a free-text topic drafts a spec for the eight nearest code snippets, not for the topic — and the strict pre-commit gate (mmnto-ai/totem#2691) accepts the run artifact of that discarded draft as evidence that a spec preceded the commit. Measured 4/4 on slug topics this week; every hand-authored replacement carries a ritual note saying so.

### Mechanism, at `57f9af5a`

- **The resolver.** `packages/cli/src/commands/spec.ts:327-349`: an input that is a bare number, an issue URL, or `owner/repo#N` becomes `{issue, freeText: null}`; anything else becomes `{issue: null, freeText: input}` (`:344-347`). The prompt renders the first as `=== ISSUE #N ===` + `<issue_title>`/`<issue_body>` (`:123-131`) and the second as `=== TOPIC ===` + `<topic_text>` (`:133-135`). A slug is a topic; the model is asked for "a spec for `<slug>`" over eight similarity hits and specs the hits.
[...]

PR #2691, opened 2026-08-29
fix(mail,hooks): silent failure → loud — own-outbox undeliverable dispatch exits 4 (sender fault); strict pre-commit gates on the totem spec run artifact

satur8d:
Closes #2685

Closes #2690

Two silent-failure classes on the CLI's own coordination surfaces, one slice. Design record: `.totem/specs/silent-to-loud.md` (v3 — cohort design review by strategy-claude, CURES-REQUIRED → folded, one bounded delta round → concurred; every finding re-verified at source before folding).

## `totem mail` — sender faults (mmnto-ai/totem#2685)

A dispatch whose `to:` matches no roster agent is undeliverable to every seat-scoped poll. The roster sensor (mmnto-ai/totem#2344) existed but was warning-only and exit-neutral, so the sender's own poll exited 0 with a clean inbox verdict beside a dispatch that never left (the 2026-08-26 liquid-city blind round, `to: lc-agy, lc-codex, …`).

- **Own-hosted → sender fault.** When the dispatch sits in an outbox **this repo hosts for a seat in the poll's self-set** (path-keyed: the outbox resolves under `<repoRoot>/.totem/orchestration/`; a same-named residue dir in another workspace repo stays a foreign warning), the poll renders an `Error:` line, carries a structured `senderFaults[]` entry in `--json` (`{kind: 'undeliverable-to', repo, agent, file, to, hint}`), and **exits 4** — a new SENDER-FAULT arm: the verdict IS derived and still renders; `2` (NOT-DERIVED) wins when both hold. "YOUR outbox" is claimed only for a single explicit identity; gated/union polls say "an outbox this repo hosts".
- **Gate-wired, not by proxy.** `ecl-gc`'s A2.2 completeness gate and A2.4 verify take `senderFaults` as an explicit conjunct with a named `gateReasons` entry (today the janitor was red only by accident of the roster warning riding the gate-armed `warnings` channel).
- **Comma refused at the front door.** `AGENT_ID_UNSAFE_CHAR_PATTERN` (core) gains `,` — the whitespace-free list `--to "a,b"` previously passed the hard guard and was written.
- Help, wiki (`docs/wiki/cli-reference.md`), and the signon skill constant name the arm and the "propagate nothing from a fault line" rule (an `Error:` line names a dispatch basename, like a gated poll's warnings).

## Strict pre-commit gates on spec evidence (mmnto-ai/totem#2690)

[...]

PR #2701, opened 2026-08-30
fix(hooks): render the configured totemDir from ONE resolver into every managed git-hook writer — builders take a required options object; default render byte-identical to tools/*; the schema normalises + refuses unrenderable values; doctor senses a stale render (mmnto-ai/totem#2692, ruled (a) centralized; two falsification passes folded)

### LINKED ISSUE #2692: hooks: the managed git-hook family hardcodes .totem/ — a custom totemDir gets non-functioning pre-push gates and an always-blocking strict pre-commit (class; surfaced by Greptile on #2691's evidence read)

satur8d:
## Ruled: (a) centralized — operator, 2026-08-30

mmnto-ai/totem#2692 reserved the cure for a ruling between (a) render `totemDir` at install, (b) a `TOTEM_DIR` env the hooks honour, (c) document the bound + a doctor sensor. A read-only census (deposited on the issue) corrected two of the issue's premises — `spec-evidence-readers.test.ts` does not exist (the suites are `install-hooks.test.ts:1700-1912`) and `.gemini/hooks/BeforeTool.js` was deleted by the mmnto-ai/totem#2690 slice — and surfaced (d), a CLI-resolved reader for pre-commit. The operator ruled **(a), centralized**; (b)/(c)/(d) are declined with grounds in the record, `.totem/specs/2692.md` (§ 3).

## The class

`totem spec` writes its run artifact at `path.join(configRoot, config.totemDir)/artifacts/runs` (`utils.ts:1476`); the strict pre-commit reader was a hardcoded `const dir = ".totem/artifacts/runs"` inside a single-quoted `node -e '…'` (`install-hooks.ts:503` → `tools/pre-commit:28`). Under a custom `totemDir` the writer and the reader named different trees → `exit 2` → "Run `totem spec <issue>`" forever — the mmnto-ai/totem#2690 shape re-created. The rest of the family failed the other way, silently: pre-push's `[ -f ".totem/…" ]` guards and post-merge / post-checkout's `grep -q '\.totem/…'` filters never matched, so the manifest, lint, badge, lockfile and claim-discipline gates were SKIPPED. No `totemDir` resolver existed anywhere (~175 inline joins); three hook writers, ONE config read (`hooksCommand` read `hooks.tier`; `init` rendered from `--strict` only; the pre-push upgrade path read nothing and silently downgraded strict → standard).

## The cure (C1–C8 of the record)

- **C1 One resolver** — `resolveHookRenderOptions(cwd, flags?) → { tier, totemDir, fallbackCmd, configPath }`; EVERY writer of hook text calls it: `totem hook install`, `totem init` (after the config write), the non-interactive installer, `installPostMergeHook`, the hook-manager helper scripts, the silent pre-push upgrade, and `doctor --parity`'s regenerated canonical.
- **C2** The four builders take a **required** options object — no defaulted parameter, so the compiler catches a forgotten thread. 112 call sites migrated (14 src · 98 tests), grep-verified.
- **C3 Byte-identity at the default** — rendering at `{ standard, '.totem', 'pnpm dlx @mmnto/cli' }` reproduces `tools/{pre-commit,pre-push,post-merge}` byte-for-byte; `tools/` is untouched; no consumer's installed hooks drift on upgrade. Pinned by `tools-hook-parity.test.ts` and re-derived from `dist`.
- **C4 Per-site quoting** — a JSON literal inside the single-quoted reader; the plain double-quoted `[ -f ]` / `[ -d ]` words `tools/*` already ships; BRE-escaped `grep` patterns. The `@mmnto/totem` schema normalises the value (backslash → `/`, a leading `./`, trailing slashes) and refuses an empty one or one carrying a quote, `$`, a backtick, a non-ASCII character, a newline or a control character; the resolver refuses `.`, a `..` segment and a leading `-`; a config that resolves but will not load renders the defaults loudly.
- **C5** The TS-side siblings read the configured value: the `<totemDir>/hooks/` helper dir + manager guidance, `eject` (with a delete-target guard), `doctor`'s stale-rule walk, `link`.
- **C6** `doctor --parity` senses a stale render by construction; the always-on `Git Hooks` row gains ONE conditional compare when `totemDir` is non-default (block-scoped, ownership-aware remedy).
- **C7** Changeset `@mmnto/cli` minor · `@mmnto/totem` patch, stating the break and the behaviour changes. **C8** docs.

## Amendments after the build (§ 6 of the record — deposited by the build leg as deviations, ratified or folded by the owner; none applied silently)

- **A1** `totemDir` is read from the REPO-LOCAL config only; the global `~/.totem/` profile (which declares `totemDir: '.'` to describe itself) yields `.totem` — honouring it would re-render every config-less repo's hooks against the checkout root. `tier` keeps its any-config precedence; the asymmetry is deliberate.
- **A2** The refusal set gained `$` and a backtick so every `sh` guard keeps the one plain double-quoted form; the conditional single-quote regime and its two helpers were removed.
- **A3** `totem link` resolves the TARGET repo's own repo-local `totemDir` (else `.totem`), never this repo's setting or the global profile (`link.test.ts`).
- **A4** `checkGitHooks` compares the totem-owned BLOCK (start → end marker), not the whole file, and names the remedy by ownership — `--force` only for a totem-owned whole file; a user hook with an appended block gets the delete-and-re-append line, because `installGitHook --force` rewrites the whole file.
- **A5** `installPostMergeHook` added to the writer list. **A6** `eject` refuses to delete a `totemDir` normalising to `''`, `.` or containing `..` (the global profile's `.` would have aimed `rmSync` at the checkout).
- **A7** Path shapes: the schema normalises `totemDir` (backslashes, a leading `./`, trailing slashes) and refuses an empty value; the renderer refuses `.`, a `.`/empty segment, a `..` segment and a leading `-`, naming the reason; the resolver refuses on the configured value; `generateHookHelpers` refuses before its `mkdir`; `link` refuses a target `totemDir` that names the target or escapes it; the CLI and schema character predicates are pinned identical over every ASCII code point.
- **A8** A config that resolves but will not load renders the defaults LOUDLY (one line names the file and the failure; `configError` on the result) — before this a refined-out `totemDir` silently installed `.totem/` hooks.
- **A9** `eject` names `.totem/prepare.cjs` in its roster (init scaffolds it at `.totem/` regardless of `totemDir`); `resolveEjectTotemDir` gets its own test file.
- **A10** The `doctor` sensor compares on the `totemDir` axis only (canonical rendered at the tier the installed hook declares — a `--strict` hook is never reported as drift nor handed a downgrading `--force`) and classifies a legacy hook (no end marker → `--force` with a back-up-first note).

## Disclosed behaviour changes

- `totem init` and the pre-push upgrade path (`shieldCommand`) now honour `hooks.tier` from config.
- `totem hook install` resolves config at the git root, so a subdirectory install picks up the repo's config instead of the global profile.
- `doctor` warns (never gates) when the installed hooks were rendered for a different `totemDir`; after changing `totemDir`, run `totem hook install --force`.
- No migration shim, no env fallback — one truth + a loud line (the 2026-08-29 ruling on mmnto-ai/totem#2691).

## Gates (worktree, on the committed tree)

```
pnpm -r build                                          => 0
@mmnto/cli   vitest run   168 files · 4263 pass · 12 skip (pre-existing win32 skips) => 0
@mmnto/totem vitest run   148 files · 4194 pass · 5 skip  => 0
pnpm format:check                                      => 0
pnpm -r --filter @mmnto/cli --filter @mmnto/totem lint => 0 (pre-existing unused-directive warnings only)
node packages/cli/dist/index.js lint --branch          => 0  PASS, 0 errors (365 advisory warnings — the push-gate scope; the first push was refused on ONE Tenet-4 catch rule until the A8 catch carried its totem-context directive)
tools/{pre-commit,pre-push,post-merge} from dist at the default => BYTE-IDENTICAL (2876 / 5953 / 5571 B)
```

The `sh`-executed spec-evidence suites — the existing mmnto-ai/totem#2690 nine and the new custom-`totemDir` four — RAN on the Windows build machine (Git-Bash `sh`); the Linux arm runs them too.

## Legs

- Explore leg (read-only census, 68 reads) → the design record; every load-bearing claim re-verified by the owner at `57f9af5a`.
- Build leg (Opus, isolated worktree) → commits `35c97e36` · `da45d736` · `fcb6c4ef` · `410e4027`; 10 deviations deposited, dispositioned in § 6.
- Owner folds → `ed547202` (A1–A6), `d64af367` (A7–A10), `3395a6bb` (the pass-2 fold).
- Falsification leg 1 on the full diff → **0 blocks-merge · 5 fold-before-merge · 8 notes**: the path-shape class (`''` rendered an ABSOLUTE reader path; a trailing slash silently killed both sync greps; `.` / a leading `-`), the legacy-hook remedy naming the opposite cure, a tier-only mismatch reported as `totemDir` drift with a downgrading `--force`, the eject delete-guard untested, and the changeset's "refused loudly by the installer" false on the config path. All five folded as **A7–A10** (`d64af367`); the notes are recorded in § 6 with what was and was not folded.
- Falsification leg 2, re-armed on the fold diff (`ed547202..d64af367`) → **0 blocks-merge · 6 fold-before-merge · 11 notes**: three docs claimed the schema *refuses* a backslash it *normalises*; a non-ASCII `totemDir` passed both predicates while git C-quotes such bytes in `diff --name-only`, so the greps could never match (a live member of the class); the sensor read the tier and the end marker from the whole file rather than the totem block; the core-side schema change had no core-side test; one stale record note. All six folded, plus four of the notes, in `3395a6bb`; the rest recorded in § 6.
- **Convergence named:** two passes, 0 blocks-merge each; pass 2's folds were one-line scope corrections, one predicate clause and tests; a third pass would polish the same seam. Remaining notes (the loud line is swallowed on the pre-existing `shieldCommand` catch; an empty `.totem/` left after eject under a custom dir; `link`'s narrower refusal set; a lone space accepted) are recorded, not folded.

## Out of scope, named

The ~175 inline `totemDir` joins and their `cwd`-vs-`configRoot` split (mmnto-ai/totem#2510's class); the session-hook family's `.totem` literals (a separate slice, same resolver); mmnto-ai/totem#2700's gate upgrade (lands on the reader this PR renders, after the seed-20 run of record).

Closes #2692


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 disposition — `d84a7b7b` (9 fixed · 1 declined; each finding verified at source before its disposition)

@greptileai @coderabbitai @gemini-code-assist — thank you; the round was sharp. Every item below names the commit it landed in (`d84a7b7b`, the single review commit on the branch) or the ground for declining.

### Greptile (P1 ×2, 1 outside-diff; confidence 2/5)

1. **Unverified recursive deletion target** (`eject.ts`, outside diff) — **fixed.** `deleteArtifacts` now sweeps a CUSTOM `totemDir` only when `looksLikeTotemDir` holds (at least one of `TOTEM_DIR_ENTRIES` — `lessons/`, `cache/`, `compiled-rules.json`, `prepare.cjs`, …) and never `.git`; a `totemDir` naming `src`/`docs`/`.git` is a reported skip ("holds none of Totem's entries — remove it by hand if that is intended"). The default `.totem/` is Totem's by name and is swept as it always was (the pre-existing eject test sweeps an empty one). `looksLikeTotemDir` is unit-tested (`eject-totemdir.test.ts`).
2. **Config failures silently default** (`eject.ts:50`; and the doctor sensor's catch) — **fixed, both.** `resolveEjectTotemDir` now separates "no config anywhere" (silent default, supported) from "repo-local config that will not load" (a `log.warn` naming the file and the failure, then the default) — the same loud-default shape the resolver took for hooks. `hooksRenderedForWrongTotemDir` returns `{ failure }` instead of `[]`, and `checkGitHooks` renders it as a **warn** row ("could not be compared against the canonical … : <reason>") — never a pass on an unexamined tree. Tested (`eject-totemdir.test.ts` loud-default case).
3. **Directory names become globs** (`link.ts:86`) — **fixed.** A target `totemDir` carrying `*`, `?`, `[`, `]`, `{`, `}`, `(`, `)` or `!` is refused with the reason (it is written verbatim into an ingest glob) rather than escaped; the hooks accept those characters, a glob cannot carry them safely. Tested with `knowledge*`, `x[1]`, `a{b,c}`.

### CodeRabbit (1 major · 5 minor · 1 trivial)

4. **Thread `totemDir` into every other `totemDir`-aware doctor row** (`doctor.ts:2417`, major) — **fixed.** One resolved `totemDir` feeds `checkCompiledRules`, `checkSecretLeaks`, `checkSecretsFileTracked`, `checkUpgradeCandidates`, `checkStaleRules`, `checkGrandfatheredRules`, `checkFreezes` (each already took the parameter; each was called at the default). `checkIndex` stays on `lanceDir` as you noted. The two secret rows attesting a clean scan of a tree they never opened was the sharp end — agreed.
5. **Resolved-path containment for configured directories** (`eject.ts:47`, `link.ts:61-69`) — **fixed.** `path.resolve` + `path.relative` against the applicable root now backs the segment tests in both: eject falls back to the default on an empty, parent-escaping or absolute relative; link refuses.
6. **Isolate the pre-push upgrade test from the global profile** (`install-hooks.ts:341` / `install-hooks.test.ts:1390`) — **fixed.** The `upgradePrePushHookIfNeeded` suite writes a repo-local `totem.yaml` pinning `hooks.tier: standard`, so a developer's global `~/.totem/` profile can no longer flip the expected block.
7. **Quote the rendered `totemDir` in the hook-manager guidance** (`install-hooks.ts:591`) — **fixed.** Whitespace-bearing values are double-quoted in the husky/lefthook lines and quote-escaped in the simple-git-hooks JSON; the default guidance stays `sh .totem/hooks/…`. (`$` and a backtick are refused upstream, so double quotes are inert.)
8. **Preserve unlink after a target directory change** (`link.ts:86`) — **fixed.** `--unlink` is keyed by the target path: the generated globs are matched by shape (`<relative>/<any dir>/lessons…`) and by the stable `// Linked: <relative>` record, so a link made under `.totem` still unlinks after the target moved to `knowledge`. Tested.
9. **Drop the per-case `require('node:child_process')`** (`doctor.test.ts:198`, trivial) — **fixed** in the new cases (the top-level `execSync` import is used); the two pre-existing occurrences elsewhere in the file are untouched.
10. **Add the `totemDir` reinstall exception to `AI_PROMPT_BLOCK`** (`config-schema.ts:505`) — **declined.** The reflex block carries `REFLEX_VERSION` (a material text change bumps it and triggers a cohort-wide drift repair of every consumer's reflex file); the reinstall rule is already stated in the `totemDir` docstring, the `hooks.tier` docstring, `config-reference.md` and the changeset. A reflex-text change is its own slice; noted in the record as a follow-up.

### Gemini (high)

11. **Use `TotemError` for CLI-layer runtime validation; don't prefix `[Totem Error]`** (`install-hooks.ts:253`) — **fixed on the CLI entry, half-declined on the sync backstop, with the reason.** The resolver (`resolveHookRenderOptions`, what `totem hook install` / `totem init` actually hit) now throws `TotemError('CONFIG_INVALID', …, recoveryHint)`; the message no longer carries its own `[Totem]` prefix (`handleError` adds the tag — the double prefix was real). `assertRenderableTotemDir`, the backstop inside the four builders, stays a plain `Error`: the builders are synchronous and the commands layer lazy-loads `@mmnto/totem`, so `TotemError` cannot be imported there without a static import of core. Both forms now render the same unprefixed text.

### Gates on `d84a7b7b` (worktree, committed tree)

```
pnpm -r build                                          => 0
@mmnto/cli   vitest run   168 files · 4268 pass · 12 skip (pre-existing win32 skips) => 0
@mmnto/totem vitest run   148 files · 4194 pass · 5 skip  (core source unchanged this round) => 0
pnpm format:check                                      => 0
eslint (cli + core)                                    => 0
node packages/cli/dist/index.js lint --branch          => 0  PASS, 0 errors (the push-gate scope)
tools/{pre-commit,pre-push,post-merge} from dist       => BYTE-IDENTICAL at the default; every refusal fires, unprefixed
```

`totem review --covariate` (executed, zero-LLM): "Covariate: no verdict artifact recorded for the current lineage — run `totem review` with review.lanes configured to emit one (sensor; exit 0)." — so no `local-lane:` line is carried on this round.

PR #2705, opened 2026-08-30
spike(spine-adopt): the seed-20 run of record at 69b85544 — one unsplit seam, both artifact trees committed whole incl. the run's manifest.json; K1–K8 10 PASS · 0 FAIL · 0 NOT MEASURED (K7 21/21 vs Linux run 33298423197); scorer frozen 30e49ac2 @ strategy 0ed61838

satur8d:
## The seed-20 run of record — executed at `69b85544`, one unsplit seam, both artifact trees committed whole (incl. the run's `manifest.json`); K1–K8 10 PASS / 0 FAIL / 0 NOT MEASURED (K7 measured 21/21)

Operator's **GO** (direct, this session 2026-08-30 ~17:40Z; the strategy-session word relayed by strategy-claude at 17:21Z). Preconditions closed on both sides: charter v1.1; apparatus pin `57f9af5a` (an ancestor of `69b85544` with the spike source byte-identical outside the two artifact trees — `git diff --stat --no-renames 57f9af5a 69b85544 -- ':(top)spikes/spine-adopt' ':(top,exclude)spikes/spine-adopt/artifacts' ':(top,exclude)spikes/spine-adopt/wazero-probe/artifacts'` is empty); controls at the pin (mmnto-ai/totem#2702 → `159f4356`); the K7 seed20 Linux leg (mmnto-ai/totem#2703 → `69b85544`); `score.mjs` frozen at strategy `0ed61838`, sha256 `30e49ac295300a94745c2226a7dd6717cb10ffb68fa260bdab004287083f7ec2` (re-verified at source this session: strategy `origin/main` == `0ed61838`, `i/lf`, no commit touching it since).

### The seam, per-stage exit codes captured directly (never through a pipe)

| # | stage | env | rc | wall (UTC) |
|---|---|---|---|---|
| 1–4 | `manifest · lower · build-wasm · certify` — K3 arm B, the control-only build FIRST → `artifacts/k3-control/` | `SPIKE_CONTROL_RECORD=records/c-supp-astgrep-compound-failopen-catch.rule.yaml` (`SPIKE_RECORD_SET` unset) | 0 · 0 · 0 · 0 | 17:49:38 → 17:51:24 (incl. the cargo 1.96.1 build of `host/`) |
| 5 | `all` | `SPIKE_RECORD_SET=seed20` | 0 | 5 s |
| 6 | `differential` (lower → t5-sweep → build-wasm → certify → host:opa → host:regorus → compare) | `seed20` | 0 | 26 s |
[...]

Issue #2704, opened 2026-08-30
spike(spine-adopt): post-run apparatus slice — the 11 findings from the mmnto-ai/totem#2702 bot round (A1 lower.mts fileText type guard · A2 absolute path in records-verification.json · A3–A11 cosmetic/prose); NOT MEASURED gets a `measured: false` shape, never `ok: null` in a swept file — the scorer is fail-closed (ruled 2026-08-30; queued behind the run of record)


satur8d:
## Charter

**Ruled 2026-08-30** (strategy-claude, `totem-strategy:.totem/orchestration/strategy-claude/outbox/2026-08-30T0642Z-totem-claude-re-2702-bot-round-ruled-a1.md`, on my deposit `totem:.totem/orchestration/totem-claude/outbox/2026-08-30T0640Z-strategy-claude-controls-pr-round-deposit.md`; operator's word to file, direct): the eleven apparatus findings the mmnto-ai/totem#2702 bot round surfaced (Gemini 1, CodeRabbit 20, Greptile none — each verified against the committed artifacts and the seed before disposition) are **not load-bearing for the run of record and do not move the apparatus pin `57f9af5a`**. They are ONE post-run apparatus slice. **Queued behind the seed-20 run of record (mmnto-ai/totem-strategy#1154 § 6). No build before the operator's word.** Slice owner: totem-claude.

**Binding constraint from the scorer's side (carry into every fix here):** since mmnto-ai/totem-strategy#1160 (`score.mjs` at `3c2d1940`) every check row the scorer reads is **fail-closed** — a row passes only on boolean `ok: true`; anything else refuses, naming the row. So a NOT MEASURED row must never be published as `ok: null` in any artifact the scorer sweeps (`manifest-checks.json`, the deposit `checks[]`, `controls.json`'s refusing rows). Give NOT MEASURED its own shape — `measured: false` with `ok` absent under a distinct key — or keep such rows out of the swept artifacts. A4 below is the row that would have gone wrong.

## The eleven findings (all verified at `3a30dd2b` / run commit `638ca5f7`; none hand-edited)

[...]

PR #2710, opened 2026-08-30
spike(spine-adopt): the E24 apparatus slice — K3b pair fields · explicit error:null · T5 header at the scored 20 · records[].inlineFilePath required · manifest.scorerSha256 (43c8667c) · A12: INV 2 referent re-keyed + the specimens baseline re-homed (the next apparatus pin)

satur8d:
## The E24 apparatus slice — the first execution's four deposit-shape classes cured, `scorerSha256` bound, and the A12 HEAD-keying cured (specimens baseline re-homed) — the next apparatus pin

Operator-greenlit route (2026-08-30 ~20:28Z in the strategy session, relayed 20:30Z; taken directly in-session ~20:50Z), scope as restated once in the scorer's 20:40Z freeze mail. Charter of record: v1.3 § 7 E24 at mmnto-ai/totem-strategy main `a2cb78a`. Scorer of record: `score.mjs` frozen at **`43c8667c5a5ea4947b01b88b2269c816131717b54e768580361a843b664c96c9`** (re-derived at that commit before the slice was cut, and independently by the falsification leg).

### The six items (two commits: `31bab7f1` the slice, `4d03c5c1` the falsification-leg fold)

| # | item | cure | where |
|---|---|---|---|
| 1 | **K3b ×58** — E6 fixes the row at `{glob, ok, matchingProbe, nonMatchingProbe}` "on both sides" | the generated pair joined per glob from the ONE generator (`generatedSeedProbes`, also published as `manifest.probeGeneration.pairs[]`); counts ride as extra fields; `ok` now also requires the pair to EXIST (leg F4 — the sweep counts cannot catch a missing pair; a missing one is a BLIND row naming "NO GENERATED PAIR") | `controls.mts` |
| 2 | **K5 ×2** — § 5 K5's letter `error: null` | `VerdictRow.error: string \| null`, explicit on every row; the K5b malformed-facts control keeps its non-empty string; `isErrorRow` unchanged; the Go decoder already reads `*string` | `shipped-verdicts.mts` |
| 3 | **T5 header ×2** — E5's "the records that reach target lowering — 20 on this seed" | the K-control package is skipped BEFORE the header pushes: named in `controlPackagesSkipped` only, never double-listed; `packages`/`policies` 21 → 20 | `t5-sweep.mts` |
| 4 | **`records[].inlineFilePath`** — the (b) ruling: REQUIRED, non-empty string or explicit `null` | published on every scored row from the loaded row; the second half (non-empty) gated at the PUBLISHER by a new manifest FAIL check (leg F5); the E2 R1 inline category becomes evidence-derivable (the class-5 root cause) | `manifest.mts` |
| 5 | **`manifest.scorerSha256`** — § 5's binding fact, never realized before | the freeze-time sha verbatim; a re-freeze is a new apparatus pin by construction | `manifest.mts` |
| 6 | **A12 (mmnto-ai/totem#2704)** — two HEAD-keyed checks made a committed run of record un-runnable at its own pin | INV 2's committed referent keyed on what HEAD commits: a `control` run over a committed run of record reads `HEAD:…artifacts/k3-control/chains/`; the **specimens baseline is RE-HOMED** — the seven pre-slice chains committed byte-exactly (verified against `69b85544`) under `artifacts/specimens/chains/`, so the specimens leg keeps FULL committed-vs-published drift detection at every commit including run pins (A12 option (i), the charter's "a specimens baseline that survives a committed run of record" — leg F2); the specimens-baseline era behavior byte-for-byte otherwise; the arm-B binding (`manifest.mts:874`) needs no change — the seam's control-build-first order binds it at any HEAD once the control build can pass | `certify-verify.mts`, `artifacts/specimens/chains/*` (7), workflow comment |

No chain-emission byte changes. The spike is not a published package — no changeset. Workflow: comment only.

### Verification — the full seam at THIS branch's HEAD, a run-of-record-shaped commit (the A12 acceptance)

Per-stage rc captured directly, `4d03c5c1`, 21:16:44Z → 21:17:53Z: control-only build FIRST (manifest · lower · build-wasm · certify) → `SPIKE_RECORD_SET=seed20 all` → `differential` → Go arm → `controls` → **the slice's deposit-shape gate** (`slice-assert`: K3b fields present and equal to `probeGeneration.pairs[]` per glob; every verdict row has the `error` key, the two K5 rows `null`; header 20/20 with the control only in `controlPackagesSkipped`; `inlineFilePath` on all 22 rows, `.github/workflows/ci.yml` claimed by `be74c55c`, the E2 R1 union now covers `probePaths[]` from the manifest alone; `scorerSha256` == the freeze; `k3Capture.repinned: false` from arm B; controls **10 PASS · 0 FAIL · 0 NOT MEASURED**) → restore → the **specimens** leg (all · differential · Go · controls) → restore. **14/14 rc=0, 0 assertion failures, tree clean after both restores.** The specimens leg's INV 2 now reads `referent: artifacts/specimens/chains` and reports every chain "bytes identical" (drift-detection live again); the control build's INV 2 reads `referent: artifacts/k3-control/chains`. The same seam run at the pre-fold commit `fa1e34a2` was equally green. At `main` today (a committed run of record) both Linux legs are red; **this push should turn both green** — the first CI-green commit since the run pin.

### Falsification leg (dispatched on the slice diff; deposit folded in `4d03c5c1`)

- **Folded:** F1 prettier (the only red file in `prettier --check .`) · F2 the specimens baseline re-homed (the charter's letter was not met by a named skip) · F3 the seed20 skip row's self-contradicting text ("never has a committed referent" beside the referent it named) · F4 the false "ok would catch it" claim → pair presence asserted, premise stated · F5 the (b) ruling's non-empty half gated publisher-side; the `?? null` comment corrected · F6 a–g (stale § S5 paragraph; hardcoded `preSliceSource` provenance; per-chain `modeReason` blamed the record set; the "23rd rows[] entry" ordinal; K5 "fails toward flagging" → "refuses"; "byte-equal by construction" softened to the verified claim).
- **Deposited, not built (out of the greenlit scope):** F3's seed20 × committed-seed20 drift-detection extension, and the `fixtureFile` tolerate-absent twin of class 5 — **A13 on mmnto-ai/totem#2704**, proposed to the scorer with the run pin.
- **Verified clean by the leg, independently:** `scorerSha256` byte-exact vs `a2cb78a`; the referent state table (control × specimens-baseline = the old behavior byte-for-byte; absent/unreadable committed manifest fails loudly); every consumer of the changed shapes (`isErrorRow`, `compare.mts`, the Go `*string` decoder, the `runManifestSha256` preimage on both sides, the manifest-checks reader); the arm-B trace.

### Gates

Repo-wide `pnpm run format:check` clean · `pnpm -r build` green · full-branch `totem lint` PASS · both seams green (above). ESLint does not cover `spikes/**` (workspace = `packages/*`; the files carried pre-existing `any`s at `69b85544`) — unchanged by this slice.

### Then the loop (each step on the operator's word; the scorer's E24 sequence)

The merged commit is the **new apparatus pin** → mailed to strategy-claude → **controls at the pin** (one PR: `controls.json` · `manifest-checks.json` · `k3-control/` · `k4-swap/`, the seed run's own artifacts NOT committed) → **the run of record** (one unsplit seam from the pin, artifacts committed whole incl. `manifest.json`, `scorerSha256` binding `43c8667c…`) → the **new run pin** mailed with the K7 run id → the scorer's two clones (evidence at the run pin, replay at `manifest.runCommit` per E23, control-only build first) → `scoring.{json,md}` → codex blind replay → the verdict on mmnto-ai/totem-strategy#1138. **Sequencing:** nothing else lands on `main` between the run's `runCommit` and its run pin (B-1 is repo-wide); mmnto-ai/totem#2709 was merged first for exactly that reason and this branch is rebased on it.

Related: mmnto-ai/totem-strategy#1138 · mmnto-ai/totem-strategy#1162 · mmnto-ai/totem-strategy#1165 · mmnto-ai/totem#2704 · mmnto-ai/totem#2705 · mmnto-ai/totem#2709.

https://claude.ai/code/session_017BoXoT6Ecn1tpyVmk2nYRM

### DIFF
--- a/.github/workflows/spine-spike-linux.yml
+++ b/.github/workflows/spine-spike-linux.yml
@@ -12,9 +12,12 @@
 # too, so an edit to the workflow exercises itself.)
 #
 # TWO LEGS, one matrix (mmnto-ai/totem-strategy#1154 § 5 K7, ruled 2026-08-30):
-#   - `specimens` — the original arm, unchanged: shipped runtime re-derived on
-#     Linux, the OPA differential, the SMT floor + binding probe, the wazero probe,
-#     the K-controls; uploads `spine-spike-linux-artifacts`.
+#   - `specimens` — the original arm: shipped runtime re-derived on Linux, the
+#     OPA differential, the SMT floor + binding probe, the wazero probe, the
+#     K-controls; uploads `spine-spike-linux-artifacts`. Since the run of record
+#     landed at the top-level artifact tree, its INV 2 committed referent is the
+#     re-homed baseline `spikes/spine-adopt/artifacts/specimens/chains/` (A12,
+#     mmnto-ai/totem#2704 — the E24 slice).
 #   - `seed20` — the R14 seed-20 seam on Linux, in the apparatus's own order
 #     (control-only build FIRST so the seed manifest measures `repinned` from arm B,
 #     then `all` → `differential` → the Go arm → `controls`); uploads
--- a/spikes/spine-adopt/artifacts/specimens/chains/r0123456789abcdef_d_file.json
+++ b/spikes/spine-adopt/artifacts/specimens/chains/r0123456789abcdef_d_file.json
@@ -1,0 +1,120 @@
+{
+  "generatedBy": "spikes/spine-adopt/src/build-wasm.mts",
+  "contract": "rego/LOWERING.md § Lowering 1 — sha256(record yaml) → sha256(lowered .rego + fact schema) → sha256(bundle.wasm); Prop 310 Amendment R2 source→IR→artifact",
+  "specimen": "d-file",
+  "ruleId": "0123456789abcdef",
+  "package": "totem.spike.r0123456789abcdef_d_file",
+  "entrypoint": "totem/spike/r0123456789abcdef_d_file/result",
+  "recordSha256": "048f8da56316438929144ad60e5650ece885f44bf26172d1727a1f4ab14ed239",
+  "recordFile": "records/d-requires-file.rule.yaml",
+  "regoSha256": "3525bdd8ec94033d51117dad49e69dca531570c529e85d98cc75887b5623392a",
+  "regoComposition": {
+    "note": "sha256 over the three component digests, labelled and newline-joined in this exact order. Component digests are published so the composite is auditable without re-deriving it.",
+    "order": [
+      "policy.rego",
+      "globs.json",
+      "factSchemaLine"
+    ],
+    "components": {
+      "policy.rego": "dcbcd77a041737297322928b74b650911c6eda6b4552bec20c70df5188382085",
+      "globs.json": "e28b298ed0a8d93c6eb9c0e00c0af87cd39d070c642be98b7e0f6aa038fcf5c9",
+      "factSchemaLine": "3e9643cd65a2335b0b1de3c508e9cadf229ecdd3a29d7519ba33409ed1c3e58f"
+    },
+    "factSchemaLine": "input = { file, fileText: string|null, lines: [string], astMatches: [{lineNumber, lineText, startLineText, startPrecedingLineText}] }",
+    "preimage": "policy.rego:dcbcd77a041737297322928b74b650911c6eda6b4552bec20c70df5188382085\nglobs.json:e28b298ed0a8d93c6eb9c0e00c0af87cd39d070c642be98b7e0f6aa038fcf5c9\nfactSchemaLine:3e9643cd65a2335b0b1de3c508e9cadf229ecdd3a29d7519ba33409ed1c3e58f\n"
+  },
+  "wasmSha256": "07b771da11130e61082f28d41f209e7f44335db10c370b0ff4f33acdf2692aad",
+  "bundleTarballSha256": "f7537894fa9ecfd5a32ab9a7dcfac7047f87746499d5dc91f1f8d6f89851ec6b",
+  "manifest": {
+    "revision": "",
+    "roots": [
+      ""
+    ],
+    "wasm": [
+      {
+        "entrypoint": "totem/spike/r0123456789abcdef_d_file/result",
+        "module": "/policy.wasm"
+      }
+    ],
+    "rego_version": 1
+  },
+  "determinism": {
+    "buildCount": 2,
+    "wasmIdentical": true,
+    "tarballIdentical": true,
+    "wasmSha256": [
+      "07b771da11130e61082f28d41f209e7f44335db10c370b0ff4f33acdf2692aad",
+      "07b771da11130e61082f28d41f209e7f44335db10c370b0ff4f33acdf2692aad"
+    ],
+    "bundleTarballSha256": [
+      "f7537894fa9ecfd5a32ab9a7dcfac7047f87746499d5dc91f1f8d6f89851ec6b",
+      "f7537894fa9ecfd5a32ab9a7dcfac7047f87746499d5dc91f1f8d6f89851ec6b"
+    ]
+  },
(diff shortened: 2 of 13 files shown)


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
Round disposition (bot round invoked by the operator; one consolidated comment per bot-protocols § 8.1) — @gemini-code-assist @greptileai @coderabbitai

Pushed as `4d1183e0` (over `4d03c5c1`), verification seam re-run green at that commit (control-only build → seed20 all → differential → Go → controls → the slice's deposit-shape gate → specimens leg; every stage rc=0), repo-wide `prettier --check .` clean. Each finding was verified at source before its disposition.

1. **FIXED** — `certify-verify.mts`, the committed `artifacts/manifest.json` parse (GCA high · Greptile P2, the reason for its 4/5). Premise true: `catch { … = null }` collapsed a MALFORMED authoritative artifact into the absent/specimens-era fallback and selected a referent anyway — and with the re-homed specimens baseline now committed, a `specimens` run would have passed over it. Now throws `[Totem Error] … not valid JSON` with `cause`. Absence (pre-manifest history) remains the legitimate fallback, as the comment now states.
2. **FIXED** — `certify-verify.mts`, `git ls-tree` exit status (CodeRabbit minor). Premise true, and the same gap pre-existed on the `committedNames` listing. One `lsTreeAtHead` helper: a non-zero status throws (a no-match pathspec exits 0 with empty stdout — the honest "not committed" — so non-zero is a real git fault); `gitShow` already refused on status. A git fault can no longer become a named skip with `ok: true`.
3. **FIXED** — `manifest.mts`, the dead `?? null` on `inlineFilePath` (GCA high; the slice's own falsification leg had raised it as F5). `Specimen.inlineFilePath` is a required `string`; published verbatim now, the comment stating the future none-case as a type change to `string | null`.
4. **FIXED** — `manifest.mts`, the seed20 probe-coverage check (CodeRabbit major). Premise partly forward-looking: the check reads the loaded rows (string-typed today, so no `null` can enter now), but the (b) ruling made an explicit `null` a legal published value and the check would then fail manifest publication silently. Filtered to strings before the membership test; the `records[]` value is untouched.

Nothing else in the round: the 9 additional CodeRabbit comments are LGTMs (incl. "no control referent gating change is needed" — agreed), the Greptile summary and CodeRabbit walkthrough carry no further findings. CI at the previous head was CLEAN with both `Spine Spike (Linux)` legs green — the first green legs at a run-of-record-shaped commit (the A12 acceptance in CI); the re-run at this push is expected to match.

local-lane: (no line to carry — `totem review --covariate` at this lineage reports no verdict artifact and no admission record; the local review lane was not run on this branch)

satur8d:
Round addendum — the residual @greptileai raised after the disposition, verified and FIXED in `2cc4d75c` (verification seam green at that commit; repo-wide `prettier --check .` clean).

**Premise true:** `gitShow()` returns `null` for every non-zero `git show` status, so "path absent" and "repository/HEAD broken" were indistinguishable at the committed-manifest lookup, and the A12 logic read `null` as the legitimate pre-manifest absence — a git fault there could still select a referent or publish a named skip; the round's `lsTreeAtHead` fix did not cover this lookup, and the nearby comment overstated what `gitShow` "refuses". **Fix:** the committed manifest's presence is now established by `lsTreeAtHead(COMMITTED_MANIFEST)` (non-zero status throws; a no-match exits 0 and is the honest absence); `git show` runs only when present, and a `show` failure on a present path throws as a repository fault. The helper moved above its first use; the state comment names the three cases (absent · malformed · present) with how each is established. GCA and CodeRabbit had nothing further after the disposition; not re-tagged.

local-lane: (no line to carry — `totem review --covariate` at this lineage reports no verdict artifact and no admission record; the local review lane was not run on this branch)

PR #2760, opened 2026-09-04
fix(hooks,doctor): doctor compares every managed hook block to the canonical on every install; a bare install rewrites a stale block in place under an attested totem:fork extension and renders at the installed tier (mmnto-ai/totem#2753; the liquid-city 1.123.0 datum)

### LINKED ISSUE #2753: hooks/doctor: the managed hook rewrite refuses a hook with trailing attested blocks, so the 1.123.0 pre-push arm never reaches a consumer that extends its hooks through init; doctor reports hooks PRESENT, not CURRENT (LC datum, mmnto-ai/liquid-city#1174)
## Symptom (observed at liquid-city's 1.123.0 bump, mmnto-ai/liquid-city#1174; routed by strategy-claude 2026-09-03T21:51Z)

`totem init -y` on a consumer whose git hooks carry content AFTER the `# [totem] end` marker reported **"Upgraded reflexes to v13"** while `.git/hooks/pre-push` kept the 1.121.0 text (no legs arm) and `.git/hooks/pre-commit` kept the `.spec-completed` marker that 1.122.0 retired. `totem doctor` then printed **"All 4 hooks installed"** — presence, not currency. Caught by LC's Opus falsification leg as BLOCKING; anchored as LC lesson-b919c189. Cure per clone today: `totem hook install --force`, then the consumer's re-append.

## Mechanism

`totem hook install` without `--force` refuses to overwrite a hook that carries content past its end marker — the Proposal 289 / mmnto-ai/totem#2406 "no overwrite of user-owned or unbounded content" contract, working as written. LC appends four attested `totem:fork` extension blocks exactly there (`tools/git-hooks/install.cjs`). So for any consumer that extends its hooks past the marker, the channel the mmnto-ai/totem#2698 changeset names for the pre-push arm's arrival ("the existing channel — `totem init` / the managed hook refresh") is closed by the contract that protects the extension, and the reflex-block upgrade succeeds beside a hook that did not move. The arm's reach note ("the strict arm's live population is small") is smaller than stated by every such consumer.

The doctor half is the same class as reflex drift, which doctor already senses: the hook row answers "is a file present", never "is its text the CLI's current template".

## Cure shapes (two named by lc-claude, one by the sensor side; nothing ruled)

1. **Sensor first (every tier):** `totem doctor`'s hook row compares the installed hook's version stamp / managed-region text against the running CLI's template, so "installed" means current and a stale managed region is reported as drift with the one-command cure (`totem hook install --force`, then re-append) — presence-vs-currency, the reflex-drift shape.
2. **The managed rewrite preserves trailing attested blocks across an upgrade:** rewrite the managed region between the markers and carry the after-marker content through byte-for-byte (the mmnto-ai/totem#2138 drift-repair already distinguishes "after-end-marker user content" — extend it from refuse-to-overwrite to rewrite-and-preserve when the trailing content is attested).
3. **Consumer-side:** the consumer's installer spawns `--force` itself before re-appending (LC's local workaround; not a totem-lane change).

A test for (1): a hook file whose managed region is one template version behind, with attested trailing blocks, is reported as DRIFT by doctor and as "installed" by nothing. A test for (2): `totem init` on that file yields the current managed region AND the identical trailing blocks, and refuses (as today) when the trailing content is NOT attested.

## Not in scope

The legs gate itself (mmnto-ai/totem#2698) and the strict-arm reach ruling (mmnto-ai/totem#2706).

Related (none closed): mmnto-ai/liquid-city#1174 · mmnto-ai/totem#2406 · mmnto-ai/totem#2138 · mmnto-ai/totem#2698 · mmnto-ai/totem#2706 · mmnto-ai/totem#2649 (managed git-hook regen: skip-on-present reads stale as delivered).

satur8d:
Closes #2753 

**The datum.** A consumer that extends its managed hooks past the end marker (liquid-city appends five attested `totem:fork` blocks) never received a managed-hook upgrade through `totem init`, and `totem doctor` printed "All 4 hooks installed" over hooks frozen at an older template — its currency compare ran only when `totemDir` was non-default (mmnto-ai/liquid-city#1174).

**Three arms.** (1) `totem doctor` compares every managed block to the canonical the running CLI renders, on every install; a stale block is a `warn` with the one remedy that fits the file's shape. (2) A bare `totem hook install` / `totem init` rewrites a stale managed block in place when the trailer is an attested extension — a full `totem:fork` marker (reason, owner, attested) on a comment line within the trailer's leading comment run, the shape liquid-city actually produces — carrying the trailer through byte-for-byte; unattested trailers still decline; `--force` still rewrites the whole file. (3) A rewrite renders at the tier the installed hook declares unless a flag or `hooks.tier` pins one (flag > config > installed > standard), on `hook install`, `init` and the hook-manager helpers, so keeping a hook current never silently downgrades a `--strict` hook.

**Live proof.** The built predicates return true on both of liquid-city's real hooks (read-only). Fixtures are the verbatim LC blocks.

**Review of record.** Owner-verified at each commit. Slice leg: 1 BLOCKING (a barrel-import rule; resolved by a ticketed `totem-context` directive with its premise verified) → `74abbb3d`. Falsification leg: 1 BLOCKING — my first-line attestation ruling declined the founding datum — 3 MATERIAL · 7 MINOR → `2730adc7`. Re-arm: 1 BLOCKING — `tierPinned` dropped on two early returns, so `--strict` was ignored without a config, hidden by a machine-coupled test — 2 MATERIAL · 5 MINOR → `c793673f` (red-then-green proven; tests now isolated from the home config). Re-arm: 0 · 5 · 5, prose and assertion strength → `b1f389a5`. Changeset `@mmnto/cli` minor; MIGRATION none (a consumer-side `--force` workaround can be retired).

**The bot round and its ladder (2026-09-04).** Bots on `b1f389a5`: Greptile P1 + P2, Gemini 1 high, CodeRabbit 2 minor — all five accepted → `997485c3` (every hook write atomic through core's `writeFileAtomicSync`; attestation fields trimmed; the tier regex anchored). Re-arm 1: 0 · 3 · 4 → `d2b11792` (the atomic test could not distinguish the fix — now the inode; "every hook write" was an overclaim — now true for every git-hook writer, the append arms as whole-file replacements; doctor's unanchored twin of the tier regex — now the installer's `declaredHookTier`). Re-arm 2: **1 BLOCKING** · 2 · 2 → `8a35423a` (my fold-6 replacement re-encoded a user's non-UTF-8 hook as U+FFFD — the hook is now read raw and every arm that keeps it carries bytes; the rewrite arm's "byte-for-byte" made literal; the hard-link boundary named; `bigint` inodes). Re-arm 3: 0 · 1 · 2 → `c3a3ab1f` (a non-decoding managed region declined silently as `exists` while doctor prescribed the bare install — now a reported `skipped-non-utf8` action and a doctor `non-utf8` shape through one exported predicate, both pinned). Re-arm 4: 0 · 0 · 4 → `ec7938a9` (doctor senses the shape before the staleness compare, so a bad shebang byte warns; messages name the region; a vacuous regex, a dead branch, a stacked tail). Re-arm 5, bounded to fold 9: 0 · 0 · 1 → `a72f1112` (the warn headline names the encoding cause instead of calling a byte-identical block stale); the leg called the ladder terminated. Head `a72f1112`. Deposits for every head read sit in the legs store; the round comment carries each disposition.

**Gate.** build; prettier; `pnpm lint`; `totem lint` 0 errors; full `pnpm test` with both gitconfigs masked and a fresh `TEMP`: 12/12 green; the four hook/init test files give identical counts under a real and an empty home; `totem doctor` on this repo's shared hooks: pass. Re-run in full after every fold, last at `a72f1112` (fresh `TEMP`): 12/12, core 4290, cli 4578, mcp 230, packs 102 / 58 / 19; prettier, ESLint and `totem lint --base main` clean each time.

Bots: one round, operator-invoked 2026-09-04T07:03Z on `b1f389a5` (CodeRabbit · Gemini · Greptile) — Greptile P1 + P2, Gemini 1 high, CodeRabbit 2 minor, all five accepted and folded in `997485c3`; every disposition in the round comment below. Greptile's confidence line said not to merge until the extension rewrite was atomic; it now is, through core's `writeFileAtomicSync`.

https://claude.ai/code/session_015M6AwdHorDacnmh4MQQgn5


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Bot round disposition — 2026-09-04 (operator-invoked on `b1f389a5`)

Every premise verified at source before the fold; all five findings accepted and folded in `997485c3`.

**Greptile — P1, hook rewrite is non-atomic (`install-hooks.ts:1547`): ACCEPTED.** True as stated: `writeExecutableHook` was `fs.writeFileSync` + chmod, a truncate-and-write, and the block-rewritten arm hands it a file whose trailer is the consumer's own lines — the one part no template can regenerate. `writeExecutableHook` now routes through core's `writeFileAtomicSync` (`packages/core/src/fs-atomic.ts`, the Tenet 4 user-file mutation helper from mmnto-ai/totem#2620) with the executable mode: same-directory temp, fsync, mode on the temp, rename last; symlink identity preserved by the helper. That covers all four write sites, not only the new arm — the whole-file overwrite had the same window. The pre-existing append arm (`fs.appendFileSync` onto a user hook) is unchanged and disclosed. The helper's two declared boundaries now reach hooks, and the changeset names them: the hooks directory itself must be writable, and a cross-uid hook the process cannot chown fails loud rather than being overwritten in place. Test: after a block rewrite the hooks directory holds the hook alone and the mode is 0o755 on POSIX.

**Greptile — P2, blank attestations accepted (`install-hooks.ts:1405`): ACCEPTED.** Verified: core's `parseForkMarker` captures the quoted value raw (`"([^"]*)"`), so `reason=" "` passed a length check. All three fields are trimmed before the non-empty test; a whitespace-only value in any position is pinned false, and the changeset's attestation clause says so.

**Gemini — high, tier regex unanchored (`install-hooks.ts:418`): ACCEPTED.** The templates emit `TOTEM_HOOK_TIER="…"` unindented, so the read is now anchored to a line start with the multiline flag; a comment inside the block that quotes the assignment can no longer steer the render. Tests: the template assignment reads; a quoting comment above a `standard` assignment reads `standard`; a comment alone reads `undefined`; a CRLF block still reads.

**CodeRabbit — minor, `cli-reference.md:34-35` re-parented the flag list: ACCEPTED.** `--strict` / `--standard` are back under the flags bullet, the two new bullets follow, and the tier bullet now carries the atomic-write sentence.

**CodeRabbit — minor, `doctor.test.ts:223` `require` in an ESM test: ACCEPTED.** The three `const { execSync } = require(...)` lines shadowed the file's own line-1 import; removed. The tests had been passing under vitest's module wrapper, so the finding's crash premise was not reproducing, but the shadowing was real and the import is the right binding.

**Re-armed falsification leg over `997485c3`** (0 BLOCKING · 3 MATERIAL · 4 MINOR — all seven folded in `d2b11792`). The three MATERIAL ones sharpened this round's own fixes: (1) my atomic-write test could not distinguish the fix from the old code — cleanup and mode were satisfied by `fs.writeFileSync` + chmod too — so both the rewrite test and the append test now capture the hook's inode before the call and assert it changed (rename-over replaces the directory entry; an in-place write keeps it; proved on Windows as well as POSIX); (2) "every hook write is atomic" was an overclaim — the helper scripts, the post-merge install, the append arm and the pre-push upgrader still wrote in place — so rather than carve the claim down, every git-hook writer now routes through the atomic helper, the two append arms as whole-file replacements (an interrupted append leaves a hook truncated mid-block, which git still runs), with session hooks named as the out-of-scope writer; (3) `doctor.ts` kept an unanchored twin of the tier regex Gemini's finding fixed in the installer, so the two arms read the fold's own fixture at different tiers — doctor now calls the installer's exported `declaredHookTier`, one predicate for both. The four MINOR: the pre-push upgrader's splice was the P1 shape too (closed the same way); the trim rule reached the docblock and both consumer docs; the changeset names all four of the helper's declared boundaries plus the one undeclared delta (a dangling symlinked hook now fails loud instead of silently creating its target); a commit had cited the parent's line numbers. The leg also checked CodeRabbit's `ReferenceError` premise: false under vitest 4, which injects `require`; the shadowing was real, and the fix stands on that.

**Re-armed once more over `d2b11792`** (1 BLOCKING · 2 MATERIAL · 2 MINOR — all five folded in `8a35423a`). The BLOCKING one was a regression of my own fold: building the whole-file replacement from the UTF-8 decode re-encoded a user's non-round-tripping bytes as U+FFFD — the leg executed it through the built dist (a cp1252 `0xE9` in a comment came back as `EF BF BD`) where the pre-round append never touched existing bytes, and it is exactly the silent-corruption class the mmnto-ai/totem#2620 eject ruling skips-with-reason on `.git/hooks/*`. Cure: the hook is read raw once, the decoded text serves only the probes, the append arms concatenate bytes, the block-rewrite arm slices the trailer from the raw file after proving the totem-owned region round-trips (declining otherwise), and the legacy pre-push upgrader declines a non-round-tripping hook with its ruled silent `false`. Byte-level tests pin both arms, including idempotence on the same bytes. The MATERIAL two: the rewrite arm's "byte-for-byte" had the same decode under it (now literal), and the changeset had omitted the helper's one silent boundary — hard-link identity is not preserved by rename-over — which the branch made newly reachable. MINOR: a wording contradiction in the boundaries bullet, and the inode reads now use `bigint` (the plain `ino` is lossy above 2^53 on Windows; measured, no collision, but retired).

**Re-armed a third time over `8a35423a`** (0 BLOCKING · 1 MATERIAL · 2 MINOR — all folded in `c3a3ab1f`). The leg verified the byte fix end to end through the built dist, established that the block-rewrite arm had been lossy at every commit before the fix (the pre-round string path rewrote a stale block with U+FFFD in the trailer whenever it fired), and found the one thing my fold-7 decline got wrong: a managed region that does not decode returned `exists` — "already installed" — while doctor kept prescribing the bare install it would never perform. Three silent runs against a permanently inert instruction, reachable by one ANSI-editor save that turns the template's em dash into a single byte. Cure, the eject precedent: the installer returns a new `skipped-non-utf8` action with its own line on `hook install`, `init` and the enforcement installer; doctor reads the raw bytes, runs the same exported predicate (`ownedTrailerByteStart`) on the attested shape, and classifies it `non-utf8` with a re-save-as-UTF-8-or-`--force` clause instead of the bare install. Both decline branches and the doctor row are pinned; the changeset and both wiki pages name the action and the shape.

**Re-armed a fourth time over `c3a3ab1f`** (0 BLOCKING · 0 MATERIAL · 4 MINOR — all folded in `ec7938a9`). The predicate proves the region above the extension, shebang line included, while doctor's staleness compare covers the block alone, so a bad byte on the shebang line alone read `pass` in doctor and `skipped` on install; doctor now runs the predicate before the compare and warns, and every message names the region. Also: a negative regex in the doctor test that could not match any composer output (replaced with one that fails on both regressions it guards), a dead re-scan branch under a stale comment (removed), and two stacked `--force` tails in a mixed remedy (the re-save clause now precedes the `--force` clause with its tail inside its parenthetical).

**Re-armed a fifth and final time over `ec7938a9`** (0 BLOCKING · 0 MATERIAL · 1 MINOR — folded in `a72f1112`, prose and one assertion, no re-arm). The warn headline had said "the managed block is stale" for the shape where the block is byte-identical and the cause is encoding; it now names the encoding cause when every flagged hook is that shape, and appends it as a clause in a mixed set. The leg confirmed the doctor reorder introduces no false warn on any shape and called the ladder terminated: five re-arms measured BLOCKING → BLOCKING → MATERIAL → MINOR → MINOR, each a strictly weaker class. Head: `a72f1112`.

**Gate (full workspace, fresh `TEMP`):** build 6/6; `pnpm test` 12/12 tasks — core 4290, cli 4572, mcp 230, packs 102 / 58 / 19; prettier clean; ESLint clean; `totem lint --base main` PASS, 0 errors.

satur8d:
@coderabbitai
@greptileai
@gemini-code-assist

## Bot round disposition — 2026-09-04 (operator-invoked on `b1f389a5`)

Every premise verified at source before the fold; all five findings accepted and folded in `997485c3`.

**Greptile — P1, hook rewrite is non-atomic (`install-hooks.ts:1547`): ACCEPTED.** True as stated: `writeExecutableHook` was `fs.writeFileSync` + chmod, a truncate-and-write, and the block-rewritten arm hands it a file whose trailer is the consumer's own lines — the one part no template can regenerate. `writeExecutableHook` now routes through core's `writeFileAtomicSync` (`packages/core/src/fs-atomic.ts`, the Tenet 4 user-file mutation helper from mmnto-ai/totem#2620) with the executable mode: same-directory temp, fsync, mode on the temp, rename last; symlink identity preserved by the helper. That covers all four write sites, not only the new arm — the whole-file overwrite had the same window. The pre-existing append arm (`fs.appendFileSync` onto a user hook) is unchanged and disclosed. The helper's two declared boundaries now reach hooks, and the changeset names them: the hooks directory itself must be writable, and a cross-uid hook the process cannot chown fails loud rather than being overwritten in place. Test: after a block rewrite the hooks directory holds the hook alone and the mode is 0o755 on POSIX.

**Greptile — P2, blank attestations accepted (`install-hooks.ts:1405`): ACCEPTED.** Verified: core's `parseForkMarker` captures the quoted value raw (`"([^"]*)"`), so `reason=" "` passed a length check. All three fields are trimmed before the non-empty test; a whitespace-only value in any position is pinned false, and the changeset's attestation clause says so.

**Gemini — high, tier regex unanchored (`install-hooks.ts:418`): ACCEPTED.** The templates emit `TOTEM_HOOK_TIER="…"` unindented, so the read is now anchored to a line start with the multiline flag; a comment inside the block that quotes the assignment can no longer steer the render. Tests: the template assignment reads; a quoting comment above a `standard` assignment reads `standard`; a comment alone reads `undefined`; a CRLF block still reads.

**CodeRabbit — minor, `cli-reference.md:34-35` re-parented the flag list: ACCEPTED.** `--strict` / `--standard` are back under the flags bullet, the two new bullets follow, and the tier bullet now carries the atomic-write sentence.

**CodeRabbit — minor, `doctor.test.ts:223` `require` in an ESM test: ACCEPTED.** The three `const { execSync } = require(...)` lines shadowed the file's own line-1 import; removed. The tests had been passing under vitest's module wrapper, so the finding's crash premise was not reproducing, but the shadowing was real and the import is the right binding.

**Re-armed falsification leg over `997485c3`** (0 BLOCKING · 3 MATERIAL · 4 MINOR — all seven folded in `d2b11792`). The three MATERIAL ones sharpened this round's own fixes: (1) my atomic-write test could not distinguish the fix from the old code — cleanup and mode were satisfied by `fs.writeFileSync` + chmod too — so both the rewrite test and the append test now capture the hook's inode before the call and assert it changed (rename-over replaces the directory entry; an in-place write keeps it; proved on Windows as well as POSIX); (2) "every hook write is atomic" was an overclaim — the helper scripts, the post-merge install, the append arm and the pre-push upgrader still wrote in place — so rather than carve the claim down, every git-hook writer now routes through the atomic helper, the two append arms as whole-file replacements (an interrupted append leaves a hook truncated mid-block, which git still runs), with session hooks named as the out-of-scope writer; (3) `doctor.ts` kept an unanchored twin of the tier regex Gemini's finding fixed in the installer, so the two arms read the fold's own fixture at different tiers — doctor now calls the installer's exported `declaredHookTier`, one predicate for both. The four MINOR: the pre-push upgrader's splice was the P1 shape too (closed the same way); the trim rule reached the docblock and both consumer docs; the changeset names all four of the helper's declared boundaries plus the one undeclared delta (a dangling symlinked hook now fails loud instead of silently creating its target); a commit had cited the parent's line numbers. The leg also checked CodeRabbit's `ReferenceError` premise: false under vitest 4, which injects `require`; the shadowing was real, and the fix stands on that.

**Re-armed once more over `d2b11792`** (1 BLOCKING · 2 MATERIAL · 2 MINOR — all five folded in `8a35423a`). The BLOCKING one was a regression of my own fold: building the whole-file replacement from the UTF-8 decode re-encoded a user's non-round-tripping bytes as U+FFFD — the leg executed it through the built dist (a cp1252 `0xE9` in a comment came back as `EF BF BD`) where the pre-round append never touched existing bytes, and it is exactly the silent-corruption class the mmnto-ai/totem#2620 eject ruling skips-with-reason on `.git/hooks/*`. Cure: the hook is read raw once, the decoded text serves only the probes, the append arms concatenate bytes, the block-rewrite arm slices the trailer from the raw file after proving the totem-owned region round-trips (declining otherwise), and the legacy pre-push upgrader declines a non-round-tripping hook with its ruled silent `false`. Byte-level tests pin both arms, including idempotence on the same bytes. The MATERIAL two: the rewrite arm's "byte-for-byte" had the same decode under it (now literal), and the changeset had omitted the helper's one silent boundary — hard-link identity is not preserved by rename-over — which the branch made newly reachable. MINOR: a wording contradiction in the boundaries bullet, and the inode reads now use `bigint` (the plain `ino` is lossy above 2^53 on Windows; measured, no collision, but retired).

**Re-armed a third time over `8a35423a`** (0 BLOCKING · 1 MATERIAL · 2 MINOR — all folded in `c3a3ab1f`). The leg verified the byte fix end to end through the built dist, established that the block-rewrite arm had been lossy at every commit before the fix (the pre-round string path rewrote a stale block with U+FFFD in the trailer whenever it fired), and found the one thing my fold-7 decline got wrong: a managed region that does not decode returned `exists` — "already installed" — while doctor kept prescribing the bare install it would never perform. Three silent runs against a permanently inert instruction, reachable by one ANSI-editor save that turns the template's em dash into a single byte. Cure, the eject precedent: the installer returns a new `skipped-non-utf8` action with its own line on `hook install`, `init` and the enforcement installer; doctor reads the raw bytes, runs the same exported predicate (`ownedTrailerByteStart`) on the attested shape, and classifies it `non-utf8` with a re-save-as-UTF-8-or-`--force` clause instead of the bare install. Both decline branches and the doctor row are pinned; the changeset and both wiki pages name the action and the shape.

**Re-armed a fourth time over `c3a3ab1f`** (0 BLOCKING · 0 MATERIAL · 4 MINOR — all folded in `ec7938a9`). The predicate proves the region above the extension, shebang line included, while doctor's staleness compare covers the block alone, so a bad byte on the shebang line alone read `pass` in doctor and `skipped` on install; doctor now runs the predicate before the compare and warns, and every message names the region. Also: a negative regex in the doctor test that could not match any composer output (replaced with one that fails on both regressions it guards), a dead re-scan branch under a stale comment (removed), and two stacked `--force` tails in a mixed remedy (the re-save clause now precedes the `--force` clause with its tail inside its parenthetical).

**Re-armed a fifth and final time over `ec7938a9`** (0 BLOCKING · 0 MATERIAL · 1 MINOR — folded in `a72f1112`, prose and one assertion, no re-arm). The warn headline had said "the managed block is stale" for the shape where the block is byte-identical and the cause is encoding; it now names the encoding cause when every flagged hook is that shape, and appends it as a clause in a mixed set. The leg confirmed the doctor reorder introduces no false warn on any shape and called the ladder terminated: five re-arms measured BLOCKING → BLOCKING → MATERIAL → MINOR → MINOR, each a strictly weaker class. Head: `a72f1112`.

**Gate (full workspace, fresh `TEMP`):** build 6/6; `pnpm test` 12/12 tasks — core 4290, cli 4572, mcp 230, packs 102 / 58 / 19; prettier clean; ESLint clean; `totem lint --base main` PASS, 0 errors.

PR #2758, opened 2026-09-04
feat(core,cli,mcp): searchRelevanceFloor has no default — unset is no floor; the shipped gate compares a retrieval's best relevance and the former default fired on none of the 55; DEFAULT_SEARCH_RELEVANCE_FLOOR removed (mmnto-ai/totem#2727; after #2757)

satur8d:
Closes #2727 

**Merge order:** after #2757 (the R4 experiment record) — this PR cites `.totem/fixtures/floor-arm-2026-09-03/` by path.

**The ruling this codes.** The R4 record (#2757) established that both shipped floors are whole-run gates on a retrieval's BEST vector-leg relevance — never per-item — and that on the measured profile (gemini-embedding-2-preview, 768-d, squared L2 on unit vectors) the lowest best-relevance over 55 recorded `totem spec` queries is 0.559 (0.5687 among refusal-eligible runs), so the shipped default of 0.25 fired on none of them and any value below a repo's own measured floor is inert. A calibrated value is repo-local; the synthesis's second branch (§ 8.2.3, § 9.8) applies to the DEFAULT.

**What changed.** `searchRelevanceFloor` is optional with no default; unset means no floor: `search_knowledge` never answers `no_useful_hits` by floor (envelope prints `floor="none"`, selection manifests record `null`), `totem spec` refuses only on the zero-hit arm (its floor line says `floor none — …`), the run artifact records `grounding.floor` only when one applied; a configured value or a per-call `min_relevance` (including an explicit `0`) restores today's gating byte for byte. `DEFAULT_SEARCH_RELEVANCE_FLOOR` is REMOVED from `@mmnto/totem`, no shim (mmnto-ai/totem#2691), and the bump is **MAJOR** — ruled at the bot round 2026-09-04: the root barrel's own header promises zero removals until a major (`packages/core/src/index.ts`, mmnto-ai/totem#2336), and a deleted public export is breaking by the semver definition; no tracked file in the five cohort repos imports it, which is why the build had first called it minor. `.changeset/config.json` pins the six `@mmnto/*` packages as one `fixed` group, so `@mmnto/cli`, `@mmnto/mcp` and the three packs take the same 2.0.0 with it. No lesson-pool gate (lesson relevances ≈ 0.34 on this index would be zeroed by any reachable value); lessons never ground a run — ruled final. This repo sets no value.

[...]

Issue #2727, opened 2026-09-02
config: the default searchRelevanceFloor (0.25) is unreachable against 1/(1+distance) on unit-normalized embeddings — measured nonsense 0.548 / relevant 0.706; 107 MCP selections, zero exclusions — calibrate the DEFAULT (mmnto-ai/totem#2700 fold, leg M3)


satur8d:
## The datapoint

`searchRelevanceFloor` (default `0.25`, `packages/core/src/config-schema.ts:443,570`, mmnto-ai/totem#2463) is compared against the per-hit vector-leg relevance `1/(1+_distance)` (`packages/core/src/store/lance-search.ts:199-202`). For L2 distance over unit-normalized vectors the distance is bounded by 2, so relevance is bounded below by `0.333` — the default floor can never fire on such an index. Measured 2026-09-02 on the `mmnto-ai/totem` checkout's index (gemini-embedding, 768-d, hybrid):

| query | bestRelevance | per-hit spread |
|---|---|---|
| a real query (the mmnto-ai/totem#2700 preflight) | 0.706 | — |
| nonsense: `purple trombone lasagna quarterly elephant zeppelin marmalade` | 0.548 | 0.541–0.548 |

Corroboration from the strategy seat (2026-09-02T01:45Z dispatch): 107 `search_knowledge` selections in its `selection-manifests.ndjson`, all at floor `0.25`, all `status: ok`, **zero** below-floor exclusions.

So the `no_useful_hits` envelope (`packages/mcp/src/tools/search-knowledge.ts:855-905`) and the new `totem spec` refusal arm (mmnto-ai/totem#2700, item 1) both name a floor that, at the default, is a mechanism claim without a reachable mechanism (Tenet 19). Users can configure a reachable value today (`0.6` would refuse the measured nonsense case on this embedder); the DEFAULT is uncalibrated.

## What is asked (its own ruling — this issue exists so mmnto-ai/totem#2700 does not fold a config-wide change)

1. Calibrate `DEFAULT_SEARCH_RELEVANCE_FLOOR` from a measured spread (relevant vs nonsense queries across the cohort's embedders — gemini, openai, ollama), or make the relevance formula metric-aware (the leg's open question: are `gemini-embedding` vectors unit-normalized through `table.vectorSearch`? no normalization step exists in `packages/core/src/embedders/`).
2. Say in `docs/wiki/config-reference.md` what the floor is measured against and what values are reachable.
[...]

PR #2761, opened 2026-09-04
feat(core): metric-bound relevance — a named-metric map pinned by SDK measurement replaces the hard-coded 1/(1+d); both vector queries set distanceType('l2'); the manifest records the metric; one warning per query when a relevance leaves [0, 1] (mmnto-ai/totem#2738; after #2758)

### LINKED ISSUE #2738: store: metric-bound relevance — a named-metric normalization map + invariant tests in lance-search.ts, warn outside [0, 1] (round mmnto-ai/totem-strategy#1193 § 2.1 C1)
## Charter

§ 8.2 item 4, verbatim:

> **Metric-bound relevance** (C1 LEARN): a named-metric normalization map + invariant tests in `lance-search.ts`; warn outside [0, 1].

The candidate it borrows from is LangChain's score normalization (C1). § 2.1 states what is borrowed and what is not, verbatim:

> **LEARN C1** LangChain score normalization | the doctrine, not the constants: the map depends on metric + norm; warn when scores leave [0, 1]; report "no relevant docs" explicitly; its euclidean constant presumes a non-squared distance and would be WRONG on LanceDB's squared `l2` | slice: a metric/normalization-bound adapter test in `lance-search.ts` (replace the hard-coded `1/(1 + _distance)` with a named-metric map + invariant tests)

Read the emphasis: **the doctrine, not the constants.** The borrow is the *shape* — a map keyed by metric, with an invariant on the output range — not LangChain's euclidean constant, which the synthesis records as wrong for the metric Totem actually runs on. Three seats reached the row: "totem-codex · harness (both dated 2026-09-02; `langchain-core` 1.6.1) · strategy-codex (LEARN, blind)".

This slice is **not** the floor. The floor is § 8.2 item 3 (mmnto-ai/totem#2727) and carries its own open decision. What this slice changes is how a raw distance becomes a `relevance` number and what happens when that number leaves its metric's range.

## Ground

**The current formula site, on totem `main` at `7e247984`** — `packages/core/src/store/lance-search.ts:193-203`, inside `rowToSearchResult`:

```ts
  // Vector search returns _distance (lower = better); FTS returns _score (higher = better).
  // The vector-derived `1/(1+_distance)` similarity is ALSO the true relevance
  // signal (mmnto-ai/totem#2463): captured into `relevance` here so it survives
  // the RRF fusion that later overwrites `score` with a rank artifact. FTS rows
  // (only `_score`, an incomparable BM25 magnitude) carry no `relevance`.
  let score = 0;
  let relevance: number | undefined;
  if (row['_distance'] != null) {
    relevance = 1 / (1 + (row['_distance'] as number));
    score = relevance;
  } else if (row['_score'] != null) {
    score = row['_score'] as number;
  }
```

One hard-coded transform, no metric in scope, no range assertion. `relevance` is attached at `lance-search.ts:221-223` only when a `_distance` was present; FTS-only rows carry none.

**The metric is never chosen and never recorded.** Both vector call sites request a search with no distance type — `packages/core/src/store/lance-search.ts:37` (`table.vectorSearch(queryVector!).limit(maxResults)`) and `:85` (`table.vectorSearch(queryVector).limit(limit).withRowId()`) — and no vector index is created at all: the only `createIndex` in the store is the FTS index on `content` (`packages/core/src/store/lance-store.ts:276`). So the metric in force is whatever LanceDB defaults to, which R1 determined empirically rather than read from configuration. **This means "read the metric from the store" is not possible today** — the slice has to introduce the metric as a recorded fact before a map can be keyed on it. That is part of the work, not a precondition for it.

**The R1 measurement (`.totem/fixtures/spec-runs-2026-09-02/r1-summary.md` at `14daff4d`, index `totem_chunks`, 5,903 rows, pin `8d5e2691`, embedder `gemini-embedding-2-preview` 768-d).** Each figure with its denominator:

| fact | figure | denominator / source |
|---|---|---|
| the metric LanceDB actually returns | `l2sq` — **squared** L2 | determined 3/3 trials, `|diff| < 1e-5`; `r1-summary.md` § Distance metric, with the three self-hit traces printed |
| stored vector norms | min 0.999999 · median 1.000000 · max 1.000001 — unit-norm | n = **2,000 of 5,903** rows; `r1-summary.md` § Stored vector norms |
| query vector norms | 1.000000, dims 768 | 5 sampled queries, all unit-norm; `r1-summary.md` § Query vector norms |
| the resulting range | `_distance ∈ [0, 4] ⇒ 1/(1+_distance) ∈ [1/5, 1] = [0.2, 1]` | `r1-summary.md` § Distance metric, verbatim |
| minimum relevance observed | **0.5022** (the `spec` leg's worst) | over **3,795** raw vector hits across 55 re-run queries; `r1-summary.md` § Vector-leg relevance |
| hits below 0.333 | **0** | of 3,795 |
| hits below 0.25 | **0** | of 3,795 |
| delivered-hit relevance | min 0.5100 · median 0.6189 · max 0.7541 | 440 delivered hits, of which **410 carry `relevance`** (vector leg present) and **30 are FTS-only** (no `relevance`); `r1-summary.md` |
| SDK and its documented default | 0.26.2; `query.d.ts VectorQuery.distanceType`: "Set the distance metric to use ... By default `l2` is used." | `r1-summary.md` § Distance metric |

The synthesis's § 1 contact table records this claim at two independent contacts on different instruments — "harness: `lance-linalg` source, `diff * diff` with no sqrt (2026-09-02)" and "totem-claude: 3/3 measured trials; min hit 0.5022 over 3,795" — with totem-codex concurring at weight one. It is one of the rows the round's ≥ 2-contact floor is **met** on (§ 1).

The consequence the synthesis draws (§ 0.1), verbatim: "On unit vectors `relevance = 1/(1 + d) = 1/(3 − 2·cos)` ranges [0.2, 1]; the 0.25 default corresponds to cos −0.5." And beside it, that the same signal still orders: "The same relevance orders relevant above not-relevant at AUC 0.846 — better than the cross-encoder the PRIOR proposed (0.629)." Relevance is a **ranking signal** here, not a gate input — which is why this slice is the normalization map and the range warning, and not a threshold change.

## Falsifier

**The synthesis registers no falsifier specific to this slice.** § 2.1's C1 row carries a `trigger / seam` (the slice text quoted under Charter) and no pre-registered refutation condition; § 8.2 item 4 adds none. Ledger rule 2 puts the falsifier on the building seat — "the falsifier is the deposit" (§ 4) — so it is owed and not written here.

What the synthesis *does* bind, verbatim, and what any implementation must therefore reproduce (§ 1, contact table, row 1):

> `_distance` is squared L2; relevance ∈ [0.2, 1]; the 0.25 default cannot fire

and (§ 2.1, C1 row):

> its euclidean constant presumes a non-squared distance and would be WRONG on LanceDB's squared `l2`

An implementation whose map, run on this index at this pin, produces a relevance outside [0.2, 1] for a unit-norm `l2` row, or which silently reproduces LangChain's euclidean constant for the `l2` case, contradicts the measurement above.

## Done when

- The hard-coded `1 / (1 + _distance)` is replaced by a **named-metric map** — the metric is an explicit input, not an assumption, and each supported metric has its own entry.
- The map **warns when a computed score leaves [0, 1]** (§ 8.2 item 4's "warn outside [0, 1]"). A warning, not a throw and not a filter — nothing here actuates.
- **Invariant tests** in `lance-search.ts`'s test surface assert the per-metric range, including the squared-`l2` case with the measured bound: unit-norm rows produce relevance in [0.2, 1].
- The metric becomes a **recorded fact** rather than an empirical discovery: it is set explicitly at the query sites (`lance-search.ts:37`, `:85`) and/or recorded with the index, then read by the map. A metric with no entry in the map is a loud case, not a silent fallthrough.
- FTS-only rows continue to carry no `relevance` (the existing contract at `lance-search.ts:221-223`), and RRF fusion continues not to overwrite it — the mmnto-ai/totem#2463 property the current comment states.
- A changeset for the releasable slice.

## Sequencing

- **§ 8.2 item 4** — after item 1 (mmnto-ai/totem#2735, filed), item 2 (the reader defects) and item 3 (mmnto-ai/totem#2727, the floor).
- **Its relation to the floor is the open question, and the synthesis leaves it open.** § 8.2 item 3 records that mmnto-ai/totem#2727's answer is "either the metric-aware per-profile floor on `distanceRange` with a calibration procedure (a repo-local borderline set) and a profile record, or remove the default floor and keep `status="no_useful_hits"` for the explicit `min_relevance` override only" — an unresolved either/or, with § 9 item 8 recording the calibrated-floor arm as still owed. This slice can land on either branch, because it changes normalization and range reporting rather than any threshold; **whether it should land before or after that decision is the operator's to rule, and is not chosen here.**
- Every slice is a legs-owed diff under the mmnto-ai/totem#2698 gate: one falsification leg before it is presented.
- Per agent-decision-points § 2c: replace, one truth, a loud line — no shim, no compat reader. Concretely: one relevance transform in the codebase, the old constant deleted rather than kept behind a flag, and the metric-unsupported case loud.

## Refs

- Synthesis: `mmnto-ai/totem-strategy:audits/internal/2026-09-03-spec-prior-art-round-synthesis.md`, sha256 `5408f3670e7989605540f366ea31d852d04d041521a6d25ba97e3d28c256e383`, at strategy commit `66737d61` (§ 2.1 LEARN C1; § 0.1; § 1 contact table; § 8.2 item 4).
- Round: mmnto-ai/totem-strategy#1193.
- Fixture: mmnto-ai/totem#2734 — `.totem/fixtures/spec-runs-2026-09-02/` on totem `main` at `14daff4d`; `SHA256SUMS` digest `49c87ae59ce644396bf21ae46b0c1afef802cce756bde72da0bfd90a93d3f29c`. Record read here: `r1-summary.md`.
- Formula site: `packages/core/src/store/lance-search.ts` `rowToSearchResult`, read on totem `main` at `7e247984`.
- mmnto-ai/totem#2727 — the floor (§ 8.2 item 3); it names this same formula site and asks for a metric-aware relevance in its item 1.
- mmnto-ai/totem#2463 — the origin of the `relevance` field the current comment cites.
- The gate this slice's diff is measured under: mmnto-ai/totem#2698.
- Sibling slices: mmnto-ai/totem#2735 (§ 8.2 item 1).

satur8d:
Closes #2738 

**Merge order:** after #2758 (both edit `spec.ts`; this branch orders the floor evaluator's parameters so the rebase is trivial).

**The borrow (C1: the doctrine, not the constants).** The hard-coded `1/(1 + _distance)` becomes a named-metric map keyed by the SDK's three spellings — `l2` (squared L2; on unit vectors relevance ∈ [0.2, 1]), `cosine` and `dot` (`1 − d/2`) — pinned by an SDK-measured test that builds a real LanceDB table and asserts the `_distance` each metric returns (dot distance is `1 − a·b`, confirmed). Both vector queries set `distanceType('l2')` explicitly; the index manifest records the metric; a once-per-query warning names the metric-specific cause when a relevance leaves [0, 1] (under `l2` only a negative `_distance`, an SDK or data fault). The `l2` numbers are byte-identical to before.

**Disclosed behavior changes.** A `_distance` that is NaN, ±Infinity or non-numeric now yields no relevance (before: NaN or 0). An out-of-range relevance is omitted from the run artifact instead of throwing at the schema, and `totem spec`'s floor treats it as FAULTED — neither signal nor exemption: it cannot raise the best relevance, cannot be withheld as a measurement, and cannot save a run; a run whose every hit is faulted refuses, and a refusal beside a fault names the count. That is TIGHTER on the fault path than before this PR. Correction of record, twice over: the fold-2 commit body said the exemption "restored the refusal" (it cannot — an exemption only loosens), fold 3 then disclosed the loosening honestly, and the bot round (Greptile P1) refuted the loosening itself — an SDK fault is not evidence the way a keyword match is — so the fold on `66f26246` replaced the exemption with the fault class. The MCP reader still filters relevance on `typeof` only — a follow-up.

**Review of record.** Owner-verified. Slice leg 0 · 2 · 5 (the map confirmed by measurement) → `b52812ca`; re-arm 0 · 3 · 4 → `f16dc8f1`; re-arm 1 BLOCKING (my false cure claim) · 4 · 3 → `9127adbc`. **Bot round (2026-09-04, on `9127adbc`):** Greptile P1 + P2, CodeRabbit 2 minor + an out-of-scope check, Gemini none — P1 accepted and it refuted the slice's exemption design (a fault could defeat a refusal); P2 declined with basis; the two minors accepted; the scope check answered on the issue. Folded on `66f26246`; leg over that fold 1 BLOCKING · 4 · 3 (a stale migration line, a "mirrors MCP" docblock, l2-specific refusal causes, a missing lessons clause, a misgrounded decline basis; the freeze unpinned; a two-arm doc) → `9f0e2136`, on a rebase onto main `b17e348b` with mmnto-ai/totem#2758 in (the evaluator is `(context, isInRange, floor?)`; that PR's eight tests take the predicate). Final leg over `9f0e2136`: 0 · 4 · 3, prose and one test (the refusal says "tallied", not "warned" — `totem spec` wires no warning sink, a CLI-wide follow-up; both floor-free arms named; the no-floor all-faulted case pinned) → `d63f8310`; ladder terminates. Head `d63f8310`. Deposits for every head a leg read sit in the legs store; each disposition is in the round comment. Changeset `@mmnto/totem` minor, `@mmnto/cli` patch.

**Gate.** build; prettier; `pnpm lint`; `totem lint` 0 errors; full `pnpm test` with both gitconfigs masked and a fresh `TEMP`: 12/12 green. Re-run in full on the rebased head `66f26246` (fresh `TEMP`): 12/12, core 4341, cli 4600, mcp 230, packs 102 / 58 / 19; prettier, ESLint and `totem lint --base main` clean. Bots: one round, operator-invoked 2026-09-04T16:41Z on `9127adbc` (CodeRabbit · Gemini · Greptile); every disposition in the round comment.

https://claude.ai/code/session_015M6AwdHorDacnmh4MQQgn5


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@greptileai @coderabbitai @gemini-code-assist — disposition of the 2026-09-04 round (invoked on `9127adbc`). Every premise verified at source before the fold; three findings accepted, one declined with basis, the scope check answered on the issue. Folded in `66f26246`, then a falsification leg over that fold (1 BLOCKING · 4 MATERIAL · 3 MINOR — six folded, one re-grounded, one declined, in `9f0e2136`) and a rebase onto main `b17e348b` with mmnto-ai/totem#2758 in: `evaluateGroundingFloor` is `(context, isInRange, floor?)`, the predicate preceding the now-optional floor as this branch's parameter order anticipated, and `refuse = nothingUsable || (floor !== undefined && …)`; the two doc hunks keep main's no-default text plus this branch's metric-bound sentences; mmnto-ai/totem#2758's eight two-argument test calls take the predicate.

**Greptile — P1, invalid relevance bypasses refusal (`spec.ts:737`): ACCEPTED, and it refutes this slice's own fold-3 design.** A relevance failing `isRelevanceInRange` was counted floor-EXEMPT — the same class as a keyword-only hit — so a negative relevance beside a below-floor sibling made `floorExempt` nonzero and the run proceeded into synthesis. The code, the changeset and the spec had disclosed this as "a loosening on the fault path"; the disclosure was honest and the design was wrong — an SDK fault is not evidence the way a keyword match is. Now three classes: no vector leg → EXEMPT (one saves the run, as ruled); a relevance in range → SIGNAL; a relevance failing the predicate → FAULTED, which cannot raise `bestRelevance`, cannot be withheld as a measurement, and cannot save the run. A run whose every hit is faulted refuses ("nothing usable grounds this run"); a refusal beside a fault names the faulted count and its cause. `GroundingFloorVerdict` gains `faulted`. The record and the judgment still agree (both refuse to treat the value as a measurement); the judgment keeps the distinction the artifact's absence loses. Tests: each invalid value alone refuses as faulted; your exact case (negative beside weak) refuses and withholds the weak sibling; NaN beside weak refuses; a fault beside an at-or-above-floor hit proceeds on the real signal; a fault beside a keyword-only hit proceeds on the exemption; the refusal text is pinned for both shapes.

**Greptile — P2, warn contract in a pre-acceptance artifact (`.totem/specs/2738.md:53`): DECLINED, with basis.** The rule cited ("pre-acceptance artifact discipline") has its referent stated in the accepted proposal that settled it, mmnto-ai/totem-strategy `proposals/accepted/301-authoring-discipline-scaffold.md` § 77: "A proposal names the gate; the ADR writes the contract … advisory/warn-only acceptance predicates must not be frozen into a pre-acceptance artifact" — the artifact class is the PROPOSAL, and the ADR is the post-acceptance home. No file under `doctrine/**` or `design-tenets.md` requires a warn-only runtime contract to live in an ADR (swept by the falsification leg on this fold). Files under `.totem/specs/` in this repo are point-in-time build sheets, not living documents — the convention the operator recorded as a CodeRabbit learning on mmnto-ai/totem#1553 (quoted in CodeRabbit's own review on this PR), and the binding runtime contract for the range warning lives where this repo's post-acceptance contracts live: the `relevance.ts` docblocks and the changeset. The paragraph the finding anchored on is rewritten regardless, because the P1 fold changed what it described. If the rule is meant to reach build sheets too, that is a rule-scope change to raise with its author, not a move for this PR.

**CodeRabbit — minor, document invalid-relevance omission (`schema.ts:219`): ACCEPTED.** Both the schema field doc and `buildGroundingBundle`'s docblock now state the two omission causes (no vector leg; non-finite or out-of-range) and what the floor makes of each — exempt for the first, faulted for the second.

**CodeRabbit — minor, freeze `DISTANCE_METRICS` (`relevance.ts:19`): ACCEPTED.** `Object.freeze` at runtime, the `readonly` type kept; `assertDistanceMetric`'s `includes`/`map` are unchanged on a frozen array.

**CodeRabbit — pre-merge "Out of Scope Changes" (the floor arm belongs to mmnto-ai/totem#2727 per the issue text): answered on the issue.** The floor arm is a consequence of the range predicate this slice introduces — the record and the judgment must share one predicate, and the judgment must decide what a failing value means — so it stays here and is recorded on mmnto-ai/totem#2738 as a scope extension.

**Gemini — no findings.** Nothing to disposition.

**Falsification legs on the fold.** Over `66f26246` (pre-rebase): 1 BLOCKING · 4 MATERIAL · 3 MINOR — the changeset's migration line still said "can newly PROCEED" (fixed: REFUSE); the evaluator docblock claimed to mirror the MCP predicate (it does not: the MCP reader's bare `typeof` is the disclosed follow-up); the refusal strings hard-coded an `l2` cause (now the fault class only); the all-faulted arm lacked the delivered-lessons clause (added); the P2 decline's basis was misgrounded (re-grounded above); the freeze was unpinned (pinned); the CLI reference listed two arms (three now); one MINOR declined — a fault is not a measurement, so it is never listed as one. Over `9f0e2136` (post-rebase): 0 BLOCKING · 4 MATERIAL · 3 MINOR, folded in the final prose commit — the refusal now says "tallied out of range by the search layer", not "warned", because `totem spec` constructs its stores with no warning sink (the default is a no-op), which the changeset names as a CLI-wide follow-up; the two "zero-hit arm" docblocks and the CLI reference name both floor-free arms; a NaN is "not a finite number in [0, 1]"; the all-faulted arm with no floor is pinned byte-exactly; the config sample's 0.2 bound is scoped to unit-norm vectors; one latent MINOR (the arm calls an out-of-range value a fault, which `relevance.ts` scopes to `l2`; the metric is a module constant) declined with that disclosure. The leg called the ladder terminated. Head: `d63f8310` (force-pushed over `9127adbc`, since the branch was rebased).

**Gate (full workspace on the final head, fresh `TEMP`):** build 6/6; `pnpm test` 12/12 tasks — core 4345, cli 4611, mcp 234, packs 103 / 59 / 20; prettier clean; ESLint clean; `totem lint --base main` 0 errors. Rebased onto main at `45d7dff1` (mmnto-ai/totem#2759 and mmnto-ai/totem#2760 in); a second rebase follows mmnto-ai/totem#2758, which this PR merges after — `evaluateGroundingFloor`'s parameter order was chosen for that landing.

Issue #2748, opened 2026-09-03
release: the OIDC-provenance verify step's effective retry budget is ~25 s against a ~2.5 min npm propagation (1.123.0 cut RED after a complete publish), and the v0 tag repoint rides that verdict


satur8d:
## Symptom (measured 2026-09-03, the 1.123.0 cut)

The Release run for `19e08f56` (mmnto-ai/totem#2746 merged 19:46:31Z) went **red at `Verify OIDC provenance on published packages`** while the publish had already **succeeded for all four packages**. The verify step polled `@mmnto/totem@1.123.0` five times, 5 s apart, and gave up at 19:48:24Z with `E404 No match found for version 1.123.0`; the registry's own `time` field puts that version's visibility at **19:50:44Z** — 2 min 20 s after the step exhausted its budget, 2 min 47 s after the GitHub release for the same package was created (19:47:57Z). The other three packages (`@mmnto/cli`, `@mmnto/mcp`, `@mmnto/pack-rust-architecture`) were visible within the window; the script never reached them because it fails on the first miss.

Run: https://github.com/mmnto-ai/totem/actions/runs/33798402670. Verified afterwards from the seat: `npm view` 4/4 at 1.123.0; `PUBLISHED_PACKAGES=… node tools/verify-oidc-provenance.mjs` run locally **PASS 4/4** (attestations + signatures); per-package tags and GitHub releases 4/4.

## Mechanism (at source, `19e08f56`)

`tools/verify-oidc-provenance.mjs`: `MAX_ATTEMPTS = 5`, `RETRY_DELAY_MS = 5000`, `SPAWN_TIMEOUT_MS = 20_000`. A fast 404 returns in well under a second, so the effective wait for a not-yet-visible version is **~25 s**. The workflow comment above the step (`.github/workflows/release.yml`) says the arithmetic "bounds it to ~2 min/package (8 min worst case across 4 packages)" and pins `timeout-minutes: 4` — that bound is only reached when every attempt hits the 20 s spawn timeout; for the actual failure shape (a prompt 404 while the CDN propagates) the budget is 25 s. The comment describes a ceiling, not the floor the failure shape sees.

## Second coupling — the tag step rides the verify step

[...]

Issue #2737, opened 2026-09-03
spec gate: the strict reader's two defects — a level-aware body predicate and heading-text tolerance, with a frozen mutation suite (round mmnto-ai/totem-strategy#1193 § 2.5)


satur8d:
## Charter

The synthesis's § 2.5 disposition row, in its own words:

> **Two reader defects, a slice in totem's lane:** the shipped gate is weaker than the promise (7/7 pass while 3/7 miss or empty promised sections) AND stricter-than-correct if naively extended to 9 headings (the body-terminates-at-any-heading predicate would falsely block the longest section)

and its seam, verbatim:

> a level-aware body predicate and heading-text tolerance (parentheticals) — with a frozen mutation suite

Two defects, one slice, in this order:

1. **A level-aware body predicate.** A `####` inside a section is not that section's end. Today the body scan terminates at a heading of *any* level, so a section that opens with a sub-heading reads as empty.
2. **Heading-text tolerance for parentheticals.** An exact-match reader over a promised heading string blocks a draft that carries the section under a heading with the parenthetical dropped.

**The all-9-headings extension comes only after both defects are fixed.** The synthesis is explicit that a naive extension to 9 is *stricter-than-correct*: extending first would block correct drafts. § 8.2 item 2 states the order — "level-aware body predicate + heading-text tolerance, with a frozen positive / negative mutation suite; the all-9 extension only after it."

The reader itself is a KEEP, not a candidate for replacement. § 2.5's first row: it "demotes to (a) the fallback for every arm not rendered from a schema (override prompts, records, non-Gemini backends) and (b) the integrity sensor over a hand-editable artifact (`install-hooks.ts:790`) — it does not retire".

## Ground

**The reader, as it stands on totem `main` at `7e247984`.** The predicate is generated into the pre-commit hook by `buildPreCommitHook` in `packages/cli/src/commands/install-hooks.ts`. The lines below are the TypeScript source of the generated hook (a template literal — the escapes are doubled in source and emit as single backslashes in the hook body):

`packages/cli/src/commands/install-hooks.ts:865-871` — `isHeading`, level-blind by design:

```js
function isHeading(line) {
  let n = 0;
  while (n < line.length && ["#"].indexOf(line.charAt(n)) > -1) n = n + 1;
  if (n < 1 || n > 6) return false;
  if ([" ", "\\t"].indexOf(line.charAt(n)) < 0) return false;
  return line.slice(n + 1).trim().length > 0;
}
```

[...]

Issue #2325, opened 2026-07-08
feat(release): VP-PR cut sensor — Consumer-impact body-tag scan + path-heuristic classifier + single choreography comment (strategy#839 couple)


satur8d:
## Couples to

[mmnto-ai/totem-strategy#839](https://github.com/mmnto-ai/totem-strategy/pull/839) — `doctrine/packaging-conventions.md` § "Release planning — the meaningful-cut discipline" (merged `3c9b906`, 2026-07-08). The single-writer split ruled there: doctrine text = strategy lane; the cut-side sensor/comment tooling = totem lane, this issue (filed on the merge ping, ECL `2026-07-08T1945Z`). Doctrine amendments route through strategy-claude — this issue owns only the mechanism.

## What (doctrine § Mechanism, items 1–3)

A **VP-PR-only** job (PRs from `changeset-release/main`) that classifies the pending cut and posts/updates **one** sticky comment in front of the operator's cut word.

### 1. `Consumer-impact:` body-tag scan

Collect `Consumer-impact: <class>` lines declared in changeset prose bodies. At VP-PR time the changesets are already consumed, so the scan reads the VP-PR diff — the tag line appears both in the deleted `.changeset/*.md` hunks and the added `CHANGELOG.md` hunks; read the **added-CHANGELOG side**, so the sensor sees exactly what ships to consumers. Doctrine fixes the tag as prose-body-only (the changesets parser reads frontmatter as a package→bump map) — a frontmatter-placed tag is itself a flag-worthy authoring miss the comment should call out.

### 2. Path-heuristic classifier for untagged entries

For each entry folded into the VP-PR changelog (each source PR), fetch its changed paths and flag **untagged contract-shaped** entries. Seed the seam list from the doctrine's class catalog:

| Contract-bearing class | Path seam (seed heuristics) |
|---|---|
| Consumer config schema | `packages/core/src/config-schema.ts` |
| ECL / file / wire formats | mail/ECL modules (dispatch filename + frontmatter + processed-mark semantics), `compiled-rules.json` schema, index formats |
| Init-distributed content | `packages/cli/src/commands/init-templates.ts`, `SKILL_CONTENT` skills, hooks, managed-block reconcile targets |
| CLI surface / exit-code contracts | command definitions, `--json` payload shapes, `process.exit` seams in command modules |
| Machine surfaces | `action.yml` (the floating `@v0` tag is a consumer surface), MCP tool schemas (`packages/mcp/**`) |
[...]

PR #2762, opened 2026-09-04
Version Packages

satur8d:

# Releases

> The changelog information of each package has been omitted from this message, as the content exceeds the size limit.


satur8d:
Release-cut sensor fill (mmnto-ai/totem#2325; `packaging-conventions.md` § "Release planning — the meaningful-cut discipline"), read on the refreshed head `6c216b18` after mmnto-ai/totem#2761 merged at `ba22ea1f`. Each flagged entry verified against its commit's hunk on that file; the choreography follows. Advisory only — the cut word stays the operator's, and it is sequenced after the `strategy-doctrine-v0.1.42` tag.

**`b17e348` · consumer config schema · `packages/core/src/config-schema.ts` — CARRIED BY THE MAJOR, not dismissed.** The hunk removes `DEFAULT_SEARCH_RELEVANCE_FLOOR` and drops the `.default()` from `searchRelevanceFloor` (bounds unchanged, field optional). Two consumer effects, both in the Major Changes entry this PR carries: a `totem.config.ts` that never set the field now runs unfloored (on the measured profile the removed default fired on none of 55 recorded queries, so delivered results do not change); a consumer that imported the constant from `@mmnto/totem` fails at type-check and replaces it with an explicit repo-local floor or nothing. That second effect is the reason this cut is 2.0.0; the migration paragraph in the changelog is the tag.

**`b17e348` · machine surfaces · `packages/mcp/src/tools/search-knowledge.ts` — additive, carried by the same entry.** The retrieval envelope's `floor` attribute gains one closed token, `floor="none"`, when no floor applies; the attribute shape is unchanged, so a wrapper agent's single-regex read still holds. The selection manifest records `floor: null` in that case instead of inventing a number. `status="no_useful_hits"` is unreachable with no floor configured and no per-call `min_relevance`. No consumer edits a file for this; a wrapper that enumerates the token set adds one value.

**`45d7dff` · init-distributed content · `packages/cli/src/commands/init.ts` — DISMISSED as a path hit; the behavior is the point of the release for one consumer.** The hunk adds six summary lines to `totem init`'s report (per hook: drift-repaired / block-rewritten with the attested extension carried through / skipped-non-utf8 with the remedy). Nothing new is distributed into a consumer's tree. What a consumer sees is the installer it runs: a stale managed hook block is now regenerated in place on a bare install with the extension after the end marker carried through byte-for-byte, and a hook whose managed region does not decode as UTF-8 is reported and left byte-identical. That is the liquid-city 1.123.0 datum (mmnto-ai/totem#2753) getting its repair on its next `totem init` or `totem hook install` after the bump.

**`9c7f9cb` · consumer config schema · `packages/core/src/config-schema.ts` — DISMISSED as a path hit.** The hunk adds one exported predicate, `hasUnrenderableHeadingChar`, and moves no schema field. Its only caller is the unit-tested invariant over the built-in `SPEC_REQUIRED_SECTIONS` constant (`spec-templates.ts`), which holds the CLI's own promised headings renderable into the pre-commit reader; a consumer's config is not judged by it. Consumers gain the fix itself: the strict spec-gate reader requires all nine promised headings and is level-aware (mmnto-ai/totem#2737).

### Choreography

- **Which consumers move:** the three cohort consumers of the fixed `@mmnto/*` group — mmnto-ai/totem-strategy, mmnto-ai/totem-status, mmnto-ai/liquid-city — bump 1.124.0 → 2.0.0 (cli, totem, mcp, and the three packs where installed). Each runs `totem sync` after the bump: the boot cross-check reads a pack's engine range from `.totem/installed-packs.json` as written at the last sync, and the packs' `engines['@mmnto/totem']` widened to `>=1.26.0 <3.0.0` in this batch (ADR-097 § 11 keeps that field out of the version bump). A consumer importing `DEFAULT_SEARCH_RELEVANCE_FLOOR` follows the migration paragraph. Wrapper agents reading the retrieval envelope accept `floor="none"`.
- **In what order:** (1) `strategy-doctrine-v0.1.42` tag + publish, strategy-claude's lane, in flight on mmnto-ai/totem-strategy#1221; (2) the cut word here → publish 2.0.0, with the OIDC verify polling every package to its visibility deadline (mmnto-ai/totem#2748); (3) this machine's global `npm i -g @mmnto/cli@2.0.0`; (4) the consumer bumps, strategy first, then status and liquid-city, each followed by its `totem sync`; nothing in 2.0.0 must carry anything for 0.1.42 — the only coupling is the pin (strategy-claude's 17:56Z reply).
- **Who owns which half:** totem-claude owns the cut verify, the publish verify and this repo's global update; strategy-claude owns the cohort consumer bumps and carries the sync-after-bump note into liquid-city's bump (their word, 17:56Z); the operator's merge word on this PR is the gate.

Issue #2763, opened 2026-09-04
feat(cli,mcp,core): a shared LanceStore warning sink — every command, the ingest pipeline and the MCP server construct their stores on the no-op default, so the store's own warnings (the out-of-range relevance tally, a failed FTS leg, a failed FTS index build) reach no one (follow-up disclosed in mmnto-ai/totem#2761)


satur8d:
## What is wrong

`LanceStore`'s constructor takes an optional warning sink and defaults it to a no-op (`packages/core/src/store/lance-store.ts:117-126` on main at `ba22ea1f`):

```ts
  constructor(
    dbPath: string,
    embedder: Embedder,
    sourceContext: SourceContext,
    onWarn?: (msg: string) => void,
  ) {
    …
    this.onWarn = onWarn ?? (() => {});
  }
```

Every store the CLI commands, the ingest pipeline and the MCP server construct takes that default. Enumerated on `ba22ea1f` (`grep -rn "new LanceStore" packages/cli/src packages/mcp/src packages/core/src/ingest`), the constructions that pass NO sink:

| Caller | Site |
|---|---|
| `totem spec` (primary + linked) | `packages/cli/src/commands/spec.ts:991`, `:1011` |
| `totem search` (primary + linked) | `packages/cli/src/commands/search.ts:24`, `:53` |
| `totem stats` | `packages/cli/src/commands/stats.ts:31` |
| `totem triage` | `packages/cli/src/commands/triage.ts:182` |
| `totem review-learn` | `packages/cli/src/commands/review-learn.ts:188` |
| `totem lesson extract` family | `extract-local.ts:166`, `extract-pr.ts:174`, `extract-scan.ts:78`, `extract-shared.ts:322` |
| `totem sync` (core ingest pipeline) | `packages/core/src/ingest/pipeline.ts:428` |
| MCP server (primary + linked) | `packages/mcp/src/context.ts:223`, `:427` |

The one caller that wires a sink is the session hook: `packages/cli/src/hooks/auto-context.ts:192` and `:214` pass a fourth positional argument that writes to stderr under an `[auto-context]` prefix.

## What is lost on those paths

Everything the store emits through `this.onWarn`, read from `lance-store.ts` and `lance-search.ts` on `ba22ea1f`:

- the out-of-range relevance tally, one warning per query (`lance-search.ts:71-86`, mmnto-ai/totem#2738) — the reason this issue exists;
[...]

PR #2773, opened 2026-09-04
fix(mcp,core): search_knowledge classifies a faulted relevance as FAULTED on core's isRelevanceInRange — never signal, never a withheld measurement, never an exemption; an all-faulted retrieval answers no_useful_hits; the envelope and the selection manifest carry the count (mmnto-ai/totem#2770)

### LINKED ISSUE #2770: fix(mcp): search_knowledge selects relevances with a bare typeof === 'number' check, so NaN or an out-of-range value reads as signal — bestRelevance, the floor verdict and the disclosures inherit it; the CLI classifies the same value FAULTED since #2761
## What is wrong

`packages/mcp/src/tools/search-knowledge.ts` on main `895379a4` selects relevances with a bare type check:

- `:730-733` — `relevances = results.map(r => r.relevance).filter(x => typeof x === 'number')`; `hasRelevanceSignal = relevances.length > 0`; `bestRelevance = Math.max(...relevances)`.
- `:877-878` — under a configured floor, `floorExempt = typeof r.relevance !== 'number'` and `withheld = typeof r.relevance === 'number'`.
- `:883`, `:920`, `:952` — the disclosure lines format any number as a relevance.

So a `NaN` or an out-of-range value from the store is **signal** here. Since mmnto-ai/totem#2761 the CLI evaluator classifies the same value **FAULTED**: it cannot raise `bestRelevance`, cannot be withheld as a measurement, cannot save a run, and an all-faulted retrieval refuses. The changeset disclosed this reader as the surface not touched: "it would still read a NaN or an out-of-range value as signal".

## Consequences

- With a floor configured (or a per-call `min_relevance`): one `NaN` among the hits makes `Math.max` return `NaN`, `NaN < floor` is false, the floor never fires, and the tool answers `status="ok"` with the noise it should have withheld — a fault defeats the refusal, the same class Greptile's P1 refuted in the CLI on mmnto-ai/totem#2761.
- With no floor (the default since mmnto-ai/totem#2758): a negative or above-1 relevance is rendered as a measurement in the envelope and the selection manifest, and counted in `bestRelevance`.
- The retrieval envelope and manifest have no vocabulary for a faulted hit, so the record and the judgment cannot agree the way the CLI's now do.

## Fix shape

- Use core's `isRelevanceInRange` (root barrel, `packages/core/src/index.ts:210`) as the single predicate: in range → signal; a number failing it → faulted; no number → exempt (keyword-only).
- `bestRelevance` over in-range values only; `floorExempt` excludes faulted; faulted hits are never withheld as measurements and never satisfy the floor; an all-faulted retrieval answers `no_useful_hits` with the faulted count and cause disclosed, mirroring the `totem spec` wording.
- The envelope and the selection manifest carry a `faulted` count; the disclosure line prints `relevance faulted` rather than a number.
- Tests: each invalid value alone; a negative beside a weak sibling under a floor (must withhold, not proceed); `NaN` beside weak; a fault beside an at-or-above-floor hit proceeds on the real signal; a fault beside a keyword-only hit proceeds on the exemption — the CLI's five shapes, on the MCP surface, each failing on today's code.

Scope note: the floor semantics do not move (whole-run gate on the best in-range relevance, exempt hits unchanged, `min_relevance` overrides both ways).

satur8d:
## `search_knowledge` classifies a faulted relevance as FAULTED — neither signal nor exemption — on the one predicate the CLI gate and the grounding-bundle builder already use; an all-faulted retrieval answers `no_useful_hits`; the envelope and the manifest carry the count

Closes #2770

## Mechanical Root Cause

`packages/mcp/src/tools/search-knowledge.ts` selected relevances with a bare `typeof === 'number'` at three sites (the `bestRelevance` fold, the floor arm's exempt/withheld split, and the disclosure lines). A `NaN` from the store therefore poisoned `Math.max` — `NaN < floor` is false, the floor never fired, and the tool answered `status="ok"` with the noise it should have withheld: a fault defeating the refusal, the class the mmnto-ai/totem#2761 bot round refuted in the CLI. With no floor (the default since mmnto-ai/totem#2758), a negative or above-1 value rendered as a measurement in the per-hit field and the envelope (and, under a floor, in the selection manifest's exclusion reason — the leg's F4 corrected the issue's wider claim). The mmnto-ai/totem#2738 changeset disclosed this reader as the surface not touched.

## Fix Applied

**Three classes on one predicate.** Core's `isRelevanceInRange` (finite, in [0, 1]) — the same predicate `evaluateGroundingFloor` classifies with and the grounding-bundle builder omits on — now decides in the MCP reader: in range → SIGNAL; no relevance (keyword-only) → EXEMPT; a number failing it → FAULTED. `bestRelevance` is the max over signal only; the below-floor arm judges signal only; exempt hits carry a batch exactly as before.

| input (floor 0.4)                       | before                          | after                                                        |
| --------------------------------------- | ------------------------------- | ------------------------------------------------------------ |
| one hit, relevance NaN / −0.1 / 1.5 / ∞ | `ok`, rendered as a number      | `no_useful_hits`, `bestRelevance="n/a"`, `faulted="1"`, path disclosed as `relevance faulted` |
| NaN beside a weak 0.2 hit               | `ok` (NaN poisoned `Math.max`)  | `no_useful_hits`, `bestRelevance="0.200"`, both withheld     |
| −0.1 beside a weak 0.2 hit              | `ok`                            | `no_useful_hits`                                             |
| NaN beside a good 0.83 hit              | `ok`                            | `ok`, the NaN hit returned with `**Relevance:** faulted (…)` |
| −2 beside a keyword-only hit            | `ok`                            | `ok` on the exemption, `bestRelevance="n/a"`                 |
| no floor, every hit faulted             | `ok`                            | `no_useful_hits` — the one way that status arises with `floor="none"` |
| any fault-free batch                    | —                               | byte-identical apart from the trailing `faulted="0"`         |

**A new arm.** Every hit faulted → `no_useful_hits` with or without a floor, content withheld, candidates disclosed by path, with the CLI refusal's sentence: `Retrieval returned N hits, but every one carried a relevance that is not a finite number in [0, 1] (tallied out of range by the search layer) — nothing usable to return.` Under a floor, a fault beside real hits adds the CLI's other sentence (`N hit(s) carried … and did not count as signal or as exemption.`) and a `Faulted candidates` list.

**Records.** The `<retrieval-envelope>` gains a trailing `faulted="N"`, always present so the line keeps one closed shape. **Bump level is your ruling (the leg's F2):** the mesh page promises the envelope "parses with a single regular expression", and a regex anchored on the documented terminal form `hits="N" />` stops matching — the suite's own regex did, and mmnto-ai/totem#2727 last touched this line while deliberately keeping its shape. No cohort code parses it (searched all four repos), and `@mmnto/mcp` is public on npm. I shipped it as **minor** on the reading that the contract is closed attribute values a regex can rely on, not a frozen attribute set, and the page now says a wrapper must match by attribute name and admit appended attributes; if you read the documented line as the contract, it is a **major** and the changeset is a one-word edit. The selection manifest records a faulted hit as `excluded` (`relevance-faulted (not a finite number in [0, 1])`) on the withheld arms and as `selected` with `returned rank=N (relevance faulted)` on the ok arm; its context carries `faulted`, which is why `@mmnto/totem` moves: `faulted` joins the CLOSED `SELECTION_CONTEXT_KEYS` set — a deliberate schema change by that set's own rule. The `min_relevance` description, `config-reference.md` and `cross-repo-mesh.md` say so; a wrapper regex anchored on `hits="N" />` must admit the new attribute (no cohort code parses the envelope — searched totem, totem-status, liquid-city, totem-strategy; only prose and journals mention it).

**Unchanged:** the floor semantics (whole-run gate on the best in-range relevance, exempt hits, `min_relevance` overriding both ways). Out of scope, untouched: the store's warning sink (mmnto-ai/totem#2763).

## Verification

In the per-agent worktree on `895379a4` + this branch:

- `pnpm build` 6/6; `search-knowledge.test.ts` 75/75 (the CLI's five shapes on the MCP surface, plus the no-floor and no-fault pins and two manifest cases); core `selection-manifest` + `relevance` suites 58/58. Full workspace `pnpm test`: 12/12 tasks (mcp 246 passed / 1 skipped, core 4345, cli 4611 — the last two unchanged from main, as this branch touches no cli code and one core constant).
- **Discrimination:** with the fix swapped out for origin/main's reader and the new tests kept, 24 of 75 fail — the two manifest cases fail on semantics alone (the old reader's reasons read `relevance=NaN` / `returned rank=2`), and every envelope-parsing case fails at least on the envelope's new attribute; the faulted-class cases additionally assert the status and `bestRelevance` the old reader got wrong (`ok` and `NaN`).
- `totem lint --branch` PASS (0 errors); `prettier --check .` clean; ESLint 3/3; `totem doctor --strict` exit 0.
- **Falsification leg (Opus, one pass over `origin/main...104ddad2`):** no BLOCKING. Executed: the classification held on every arm against constructed inputs; fault-free batches diffed byte-identical apart from `faulted="0"`; the all-faulted gate cannot be reached by the federation-failure or fts-fallback paths; 9/9 mutants killed; of the 24 old-reader failures, 11 are semantic (the issue's own reasons) and 13 attribute-only. F1 MATERIAL — the shipped `searchRelevanceFloor` JSDoc still said `no_useful_hits` can never fire unset (folded, two sentences). F2 MATERIAL — the bump-level ruling above (yours). F3 MINOR — the all-faulted body stated the cause twice where the CLI's arm states it once (folded). F4 MINOR — the changeset's no-floor sentence over-claimed the manifest (folded). F5 MINOR — config-reference said a faulted hit is always disclosed by path; on a returned batch it is returned with its field reading `faulted` (folded). F6/F7 MINOR — assertion hygiene and coverage (folded: both renderings covered, −∞ added, the floor is now a table axis, the `min_relevance` route pinned). One open question the leg raised for BOTH readers, not this diff: the inherited sentence "tallied out of range by the search layer" may not hold for the NaN class, which the store's tally skips — left as is, disclosed. Not re-armed: the fold changes disclosure text, docs and tests, not the classification; the suite is green on the fold. Deposit at `104ddad2`, the head the leg read; the fold's `config-schema.ts` edit is the leg's own F1 remedy, so the gate will read `covers 3/4 owed paths`.

https://claude.ai/code/session_01LM7M3MuvwsqbdEMauLq891


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Bump level — ruled (operator, 2026-09-04): minor.** The `<retrieval-envelope>` contract is its closed attribute values a regex can rely on, not a frozen attribute set; `faulted="N"` is appended, always present, and the mesh page now says a wrapper must match by attribute name and admit appended attributes. No cohort code parses the line (searched totem, totem-status, liquid-city, totem-strategy). The changeset stays `@mmnto/mcp: minor`.

satur8d:
**Disposition — Gemini round.** @gemini-code-assist one finding, declined with the operator's ruling above: the bump stays `minor`. The envelope's contract is closed attribute values a regex can rely on, not a frozen attribute set; `faulted="N"` is appended, always present, and `docs/wiki/cross-repo-mesh.md` now says a wrapper matches by attribute name and admits appended attributes. Verified before declining: no code in the four cohort repositories parses the line (only prose and journals mention it), and this package's own log line matches `bestRelevance` by name. Ruled minor by the operator on 2026-09-04.

satur8d:
**Disposition — CodeRabbit + Greptile rounds** (four findings: one fixed, three declined with the verification). Fix push: `e973c5c8`.

@greptile-apps **Overflow faults undercounted — fixed by scoping and labelling.** A faulted hit cut by the federation final limit now carries `federation-final-limit overflow (relevance faulted)` as its exclusion reason, and the `faulted` context key is documented as the RETURNED pool's count — the number the envelope reports — so the row's two fault surfaces say where each fault went instead of one number claiming both. Pinned: a NaN hit ranked second in a linked store falls past the final limit with the labelled reason while the context reads `faulted: 0`. (Counting overflow faults into the same key would have made the manifest's `faulted` disagree with the envelope's for one response, which is the exact one-name-two-numbers shape the key exists to avoid.)

**Minor release breaks wrappers — declined, ruled.** The operator ruled this minor on 2026-09-04 (the comment above): the envelope's contract is its closed attribute values a regex can rely on, not a frozen attribute set; `faulted="N"` is appended and always present; the mesh page says a wrapper matches by attribute name and admits appended attributes. Verified before declining: no code in totem, totem-status, liquid-city or totem-strategy parses the line; the suite's own full-shape regex was updated as part of the change, which is what a full-shape consumer would do.

**Old readers drop new rows — declined, with the deployment shape as the evidence.** The hazard needs a manifest WRITER newer than its READER inside one repository. Both are the same `@mmnto/totem`: the MCP server runs from the workspace build (`.mcp.json` → `./packages/mcp/dist/index.js`) and the hooks read through the repo's `node_modules/@mmnto/totem`; consumer repos pin `@mmnto/mcp` and `@mmnto/totem` to one fixed-group version, so the two cannot drift apart without a broken install. And the closed key set is the point: `SELECTION_CONTEXT_KEYS`'s own docblock says adding a key is a deliberate schema change, made so a shared relevance score cannot be smuggled into the row — loosening the reader would give that guard back. A row an older reader cannot parse is not dropped silently either: `readSelectionManifests` reports the aggregate `skipped N malformed line(s)` on its accounting channel, and reject-counting consumers read the file raw by contract. A `schemaVersion` bump would not help the case you name — a v1 reader rejects a v2 row the same way.

@coderabbitai **Treat the envelope shape change as breaking — declined, the same ruling** as above; the changeset stays `minor`.

Issue #1884, opened 2026-05-11
init: GEMINI_SESSION_START template should call `totem describe`, not `totem status`


satur8d:
## Problem

`packages/cli/src/commands/init-templates.ts:GEMINI_SESSION_START` currently calls `totem status` at session start:

```js
execSync('totem status', { timeout: 30000, stdio: ['ignore', 'inherit', 'inherit'] });
```

But the family-canonical SessionStart hook calls `totem describe`:

- `mmnto-ai/totem-strategy:.gemini/hooks/SessionStart.js` → `totem describe`
- `mmnto-ai/totem-substrate:.gemini/hooks/SessionStart.js` → `totem describe`
- `mmnto-ai/arhgap11:.gemini/hooks/SessionStart.js` → `totem describe`
- `mmnto-ai/totem-status:.gemini/hooks/SessionStart.js` → `totem describe`

The two commands produce different output. `totem describe` emits the `[Describe] Project: ... Lessons: N Targets: N Hooks: ...` orientation banner that consumers have integrated against at session start. `totem status` emits "current project health" (manifest freshness, shield staleness) — useful for ad-hoc checks, not orientation.

The Claude-side `CLAUDE_SESSION_START` hook (added in mmnto-ai/totem#1845 slice 1, shipped via `mmnto-ai/totem#1862`) already calls `describe`. The Gemini side is the asymmetric residue — it drifted to `status` at some point and the family-canonical pattern moved on.

## Empirical evidence

Per strategy-Claude's parity-matrix run (handoff `2026-05-09T0620Z`):
- 4 of 4 family repos use `describe` in their actual SessionStart hooks
- Only the init template emits `status`
- This means every fresh `totem init` produces a SessionStart hook that diverges from the install pattern surveyed in the codebase

## Fix shape

1. `init-templates.ts:GEMINI_SESSION_START` — change `execSync('totem status', ...)` → `execSync('totem describe', ...)`. Update the leading comment from "Runs \`totem status\`" to "Runs \`totem describe\`" so the generated hook is self-documenting.

[...]

PR #2774, opened 2026-09-04
fix(core,cli,mcp): describe counts the ACTIVE compiled-rule set through the one predicate lint loads with, and reports the inert split — the banner read 485 compiled against a lint that enforces 385 (mmnto-ai/totem#2765)

### LINKED ISSUE #2765: fix(core,cli,mcp): describe reports the RAW compiled-rule total (485) while status and lint count the ACTIVE set (385) — the session banner and describe_project quote a number lint never enforces (the describe half of the mmnto-ai/totem#2388 parity)
## Observed

On main `895379a4`, `@mmnto/cli` 2.0.0 (the workspace build and the global install agree byte-for-byte on this):

```
$ totem status
[Status] Rules: 385 compiled
$ totem describe
[Describe] Rules: 485 compiled
```

The SessionStart hook banner is the describe number, so every session on this repo opens on "485 compiled", and the MCP `describe_project` tool's `legacy` block comes from the same core function, so MCP consumers read 485 as well. `totem lint` loads 385.

The split on this repo's `.totem/compiled-rules.json`, read directly:

| status field | count |
|---|---|
| none (active) | 385 |
| `archived` | 93 |
| `untested-against-codebase` | 7 |
| **raw total** | **485** |

## Source

- `packages/cli/src/commands/status.ts:34-45` counts through `loadCompiledRules`, which applies the mmnto-ai/totem#1345 status filter. Its comment (from the mmnto-ai/totem#2388 parity fix) says the count goes "through the SAME loader lint/describe use … The raw file total would be a third, different number (485 raw vs 387 active at the time of this fix)".
- `packages/core/src/describe.ts:47-53` does not use that loader. It parses the file and reports `parsed.rules.length` — the raw total the status comment names as the number to avoid.

So the mmnto-ai/totem#2388 parity holds between status and lint, and describe is the half it did not reach (or regressed): the orientation surface reports the raw total, the enforcement surfaces count the active set.

## Why it matters

Describe is the orientation banner. An operator who reads "485 compiled" at session start and a lint run that enforces 385 has no surface that reconciles the two, and the 100-rule gap is exactly the archived-plus-untested set that lint deliberately does not run. A sensor whose number a reader cannot trace to the enforced set is the sensor-trust class (Tenet 4 display honesty), the same class mmnto-ai/totem#2388 fixed for status.

## Proposed

1. `describeProject` counts the active set through `loadCompiledRules` — the same loader status, lint and the MCP reader use — so the three surfaces agree by construction.
2. Keep the raw total visible but labeled: render the split when it is non-trivial (`Rules: 385 active (93 archived, 7 untested)`), so the archived mass does not vanish from the banner, it just stops posing as enforced.
3. A test that fails on today's code: a fixture `compiled-rules.json` with archived and untested entries, asserting `describeProject(...).rules` equals `loadCompiledRules(...).length` and equals what status reports. MCP `describe_project` inherits through the shared function; its snapshot test moves with it.

**Scope note.** Nothing here changes what "active" means (mmnto-ai/totem#1345) or status's manifest `rule_count` fallback for an absent compiled-rules file; it only makes describe count what the others count.

satur8d:
## `describe` counts the ACTIVE compiled-rule set through the one predicate lint loads with, and reports the inert split — the banner read 485 compiled against a lint that enforces 385

Closes #2765

## Mechanical Root Cause

`packages/core/src/describe.ts` counted `parsed.rules.length` — every entry in `compiled-rules.json` — while `totem status` and `totem lint` count through `loadCompiledRules`, which applies the mmnto-ai/totem#1345 status filter (archived / untested-against-codebase / pending-verification excluded). The SessionStart banner runs `totem describe`, and the MCP `describe_project` tool's `legacy` block comes from the same core function, so every session on this repo opened on `Rules: 485 compiled` and every MCP consumer read 485, against a lint that enforces 385. The 100-rule gap is exactly the archived (93) + untested (7) set lint deliberately does not run, and no surface reconciled the two. mmnto-ai/totem#2388 fixed this for status; describe was the half it did not reach.

## Fix Applied

- **One predicate.** The status filter becomes a named function, `isActiveCompiledRule` (`packages/core/src/compiler.ts`, exported from `@mmnto/totem`), and `loadCompiledRules` filters through it — no second spelling of the status list anywhere.
- **describe counts through it.** `describeProject` reads the file through the same schema-validating loader lint and status use (`loadCompiledRulesFile`) and counts through the predicate. `ProjectDescription.rules` is now the active count; five new required members carry the rest: `rulesCompiled` (the raw total), `rulesArchived`, `rulesUntested`, `rulesPendingVerification` (with `rules + the three inert counts === rulesCompiled`), and `rulesSource` (`compiled-rules` | `absent` | `unreadable`) so a zero is never ambiguous. Readers gain them for free; a TypeScript consumer that constructs a `ProjectDescription` must supply them (the only in-repo double is an untyped `vi.mock` literal).
- **The banner line.** `formatRulesLine` renders `Rules: 385 active of 485 compiled (93 archived, 7 untested-against-codebase)` where the split is non-trivial, `Rules: N active` where nothing is inert, `Rules: 0 active (no compiled-rules.json)` for a missing file, and `Rules: 0 active — compiled-rules.json unreadable (…)` for one the loader refused. Describe never prints a bare "N compiled" again (status keeps its own wording, out of scope).
- **One consequence, stated (and corrected by the leg).** A file that is valid JSON but fails the compiled-rules schema now reports zeros from describe, labelled `unreadable` (it previously counted any `rules` array); lint refuses it outright. `totem status` handles that file differently — it falls back to the compile manifest's `rule_count`, a raw total (this repo's manifest carries 485) — and that pre-existing fallback is untouched. My first draft claimed status reports zeros there; the leg executed 0 vs 42 and the claim is gone from every surface.
- **The MCP surface moves with it.** `describe_project`'s output schema (`LegacyProjectDescriptionSchema`, a declared mirror of `ProjectDescription`) gained the five members; and its rich state's `ruleCounts.active` now counts through the same predicate — before this it counted `status !== 'archived'`, so one call answered `legacy.rules` 385 beside `ruleCounts.active` 392 on this repo (the leg's F5). `ruleCounts` gains `untested` and `pendingVerification` so its parts still sum to the file total.
- **Unchanged.** What "active" means (mmnto-ai/totem#1345), status's manifest fallback and wording, and the shape of everything else `describe_project` returns.

Changeset **minor** for `@mmnto/totem` (the predicate export, five new members, the active count), `@mmnto/cli` (the banner line) and `@mmnto/mcp` (the schema mirror and `ruleCounts`).

## Verification

In the per-agent worktree on `895379a4` + this branch:

- **Measured on this repo's own `compiled-rules.json` after the change:**

  ```
  [Describe] Rules: 385 active of 485 compiled (93 archived, 7 untested-against-codebase)
  [Status]   Rules: 385 compiled
  ```

  and `totem describe --json` → `rules 385, rulesCompiled 485, rulesArchived 93, rulesUntested 7, rulesPendingVerification 0` — the issue's table, reproduced by the sensor itself. **The four rules-file shapes, built CLI after the second fold:**

  ```
  no file:         Rules: 0 active (no compiled-rules.json)
  schema-invalid:  Rules: 0 active — compiled-rules.json unreadable (run 'totem lint' for the parse error)
  not JSON:        Rules: 0 active — compiled-rules.json unreadable (run 'totem lint' for the parse error)
  1 + 1 + 1 inert: Rules: 1 active of 3 compiled (1 archived, 1 untested-against-codebase)
  ```

- `pnpm build` 6/6; core `describe` + `compiler` suites green; cli `describe` 18/18 (+ `status`); mcp `schemas` + `state-extractors` + `describe-project` green. Full workspace `pnpm test` after the fold: 12/12 tasks (core 4348 passed, cli 4614, mcp 235 — +3, +3 and +1 over main are this PR's cases).
- **Discrimination:** the core mixed-status case asserts `rules` 5 of 10 with the split (3 / 1 / 1) and `rules === loadCompiledRules(...).length` — the pre-fix code returns 10; the schema-invalid case asserts zeros labelled `unreadable` where the pre-fix code returned 1; the cli parity case runs `statusCommand` against a scaffold with 4 of 10 archived and asserts describe's `rules` equals the 6 status prints — the issue's own second conjunct, which the first draft had not asserted (the leg's F2); the mcp extractor case (2 active, 1 archived, 1 untested, 1 pending) returned 4 active on the pre-fold extractor; the cli scaffold's rule entries are schema-valid now (the old `{ id }` stubs never were).
- `totem lint --branch` PASS (0 errors); `prettier --check .` clean; ESLint 3/3; `totem doctor --strict` exit 0.
- **Falsification leg (Opus, two passes).** Pass 1 over `origin/main...b0fe73fc`: the arithmetic held across 13 constructed fixtures and 6 cohort manifests (no input where describe's `rules` differs from lint's loader; the split invariant held; the pre-fix tests fail for the issue's reasons), and nine findings landed on claims and mirrors — F1 MATERIAL the status-fallback claim (executed 0 vs 42), F2 MATERIAL the untested second conjunct, F3 MATERIAL the MCP schema mirror, F4 MATERIAL the changeset's missing package, F5 MATERIAL the MCP extractor's second spelling (392 vs 385, executed), F6–F9 MINOR (wording reach, a pin, "additive", the ambiguous zero). All nine folded in `d1bebafe`. Pass 2, re-armed on the fold: F1–F8 verified complete by execution (385/385/385 parity, the old extractor's 392 reconstructed, no stripped schema members, status prints `Rules: 6 compiled` on the parity fixture); F10 MATERIAL — F9's remedy covered only the schema-invalid path: a file that is not JSON made the loader WARN and return an empty envelope instead of throwing, so it reported zeros labelled as a real file (folded: the label is read off the loader's `onWarn` seam; the core malformed-JSON test now asserts `unreadable`); F11 MINOR — two `RuleCountsSchema` guard tests had gone vacuous on the missing members (folded); F12 MINOR — the `unreadable` line's advice clause was unpinned (folded, exact); F13 MINOR — a code comment over-generalized status's fallback (folded: schema-invalid falls back, non-JSON prints 0); F14 MINOR — the `minor` bump for five required members on a public interface is a disclosure-vs-semver judgment (kept minor; the changeset now carries the migration line: add the five fields or build through `describeProject`; no downstream constructor exists in the cohort; no shim, per the no-legacy ruling); F15 MINOR — the parity fixture now carries an untested entry as the issue's clause names (folded); F16 MINOR — the zero branches now render the description they are given (folded). Not re-armed a third time: the second fold is the leg's own three-line remedy plus tests and comments; the four rules-file shapes were re-executed with the built CLI after it (below). Deposit at the head the re-armed leg read.

https://claude.ai/code/session_01LM7M3MuvwsqbdEMauLq891


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
**Disposition — Greptile round** (one P2, declined with the verification).

@greptile-apps **Validation errors become zeroes — declined.** The premise is accurate about what the code does; the conclusion misreads which surface owns the throw. `describeProject` is the read-only orientation sensor behind the SessionStart banner and the MCP `describe_project` tool, and its not-throwing contract predates this PR (mmnto-ai/totem#1884's tests: "reports 0 when malformed (does not throw)"); it is unchanged here. Tenet 4's loud half is what this PR adds: the zeros are now LABELLED — `rulesSource: 'unreadable'`, rendered as `Rules: 0 active — compiled-rules.json unreadable (run 'totem lint' for the parse error)` — so the state can no longer pose as an honestly empty rule set, and the reader is told which surface throws. That surface is `totem lint`, whose loader raises `TotemParseError` on the same file and always has; a schema-invalid rules file fails loud there, at the enforcement boundary, which is where Tenet 13 (sensors, not actuators) puts it. Throwing from the banner would instead break every session start on a machine whose rules file is malformed, with no gate crossed and no cure nearer. Verified: `totem status` degrades on the same file (to the manifest's raw count), `totem lint` throws, and all six cohort manifests parse today, so the labelled-zero path is reachable only at the next schema version bump.

satur8d:
**Disposition — CodeRabbit round** (five findings: four fixed, one put to the operator). Fix push: `6e44b12a`.

@coderabbitai **Inaccessible rules file labelled absent — fixed:** `describeProject` now stats the file and maps only `ENOENT` to `absent`; any other failure reaches the catch and is `unreadable`. **Absent-state assertions — fixed:** a case pins `rulesSource: 'absent'`, `rulesCompiled: 0`, `rules: 0` for a missing file. **Exact `formatRulesLine` assertion — fixed:** `Rules: 12 active of 20 compiled (8 archived)` is asserted whole. **Tautological range checks — fixed:** removed; `counts.active > 0` and the loader-parity assertion stay.

**Major release level for the five required members — put to the operator, not decided here.** The premise is exact: `ProjectDescription` is publicly exported and a TypeScript consumer that CONSTRUCTS it must now supply five members. What the bump should be turns on which contract the type is: the return shape of `describeProject` (readers gain the members for free; no cohort code constructs the type — the only in-repo double is an untyped mock) or an implementer contract (then a required addition is breaking). The operator ruled the analogous question on mmnto-ai/totem#2773 as minor on the first reading, and this PR's changeset carries the migration line either way; the ruling for this PR is theirs and the changeset is a one-word edit if it goes the other way.

PR #2778, opened 2026-09-05
Version Packages

satur8d:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

- b8fe7ed: `describeProject` counts the ACTIVE compiled-rule set — the set `totem lint` enforces and `totem status` reports — instead of the raw file total, and reports the inert split beside it (mmnto-ai/totem#2765; the describe half of the mmnto-ai/totem#2388 parity).

  **What was wrong.** `totem describe` (and so the SessionStart banner every session opens on) and the MCP `describe_project` tool's `legacy` block printed `parsed.rules.length` — every entry in `compiled-rules.json`, archived and untested ones included — while `totem status` and `totem lint` count through the loader that applies the mmnto-ai/totem#1345 status filter. On this repository that read `Rules: 485 compiled` at session start against a lint that enforces 385: the 100-rule gap is exactly the archived-plus-untested set lint deliberately does not run, and no surface reconciled the two.

[...]

PR #2772, opened 2026-09-04
feat(hooks,ci): adopt the legs-gate parity row — the CI arm (a hard `totem legs gate` step in the lint workflow over tracked deposits) plus the per-arm `hooks.legsOwed.enforce` knob (mmnto-ai/totem#2771)

satur8d:
## Adopt the `legs-gate` parity row here — the CI arm (a hard `totem legs gate` step in the lint workflow over tracked deposits) plus the per-arm `hooks.legsOwed.enforce` knob

Closes #2771

## Mechanical Root Cause

`@mmnto/strategy-doctrine` 0.1.42 (pinned since mmnto-ai/totem#2764) carries the `legs-gate` parity row. Its census reads **totem = neither arm** (`tools/pre-push` at tier standard, advisory) — adoption owed. The row names two enforcement arms over the same evidence, a repo adopts one: (a) a `totem legs gate` step in the lint workflow AND `.totem/artifacts/legs/` re-included from the artifacts ignore AND `hooks.legsOwed.globs` declared, or (b) hooks tier strict. Strategy's broadcast of 2026-09-04T0005Z also asked for a per-arm enforcement knob so the legs arm can block locally while spec-evidence and shield stay at the repo's tier; their 17:56Z reply holds the 0.1.43 codification until the knob ships.

Before this PR: fourteen globs declared (`totem.config.ts`), `.gitignore` ignored `.totem/artifacts/` wholesale (deposits per-checkout, invisible to CI), no workflow step, and the managed hook's advisory arm ran `legs gate --advisory` and **ignored its exit status entirely**.

## Fix Applied

**1. The knob — `hooks.legsOwed.enforce: 'block' | 'advisory'`** (`packages/core/src/config-schema.ts`; read by `buildLegsGateDeps` and applied in `runLegsGate` at the one point the tier was already applied).

| knob    | `legs gate` (bare) | `legs gate --advisory` | stdout                                     |
| ------- | ------------------ | ---------------------- | ------------------------------------------ |
| unset   | derived (0/2/3)    | 0                      | unchanged from main                        |
| `block` | derived            | **derived**            | + `[Totem] legs: hooks.legsOwed.enforce = block` (appended, either flag) |
| `advisory` | **0**           | 0                      | + `… = advisory`                           |

One formatting path: the lines are composed once; the knob, like the tier, changes only the exit code. The knob line is appended last so every verdict line keeps its index, and printed under either flag so the two tiers' stdout stays identical for one config.

[...]

PR #2746, opened 2026-09-03
Version Packages

satur8d:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

- 01a9c74: The falsification-leg floor gets a reader: `totem legs deposit` records what a leg read, `totem legs gate` answers whether THIS head was read, and the strict pre-push hook blocks a legs-owed push carrying no fresh deposit (mmnto-ai/totem#2698).

  **The break it closes.** The review-leg floor was a duty a seat remembered. Nothing read it, so nothing caught it when it was skipped: mmnto-ai/totem#2694's 74-file diff was presented for merge without the leg it owed, and the round in mmnto-ai/totem-strategy#1193 needed six legs over one synthesis — legs 2, 3 and 5 each caught an error the PREVIOUS fold had introduced. A floor that holds only when someone remembers it is not a floor.

  **The deposit artifact (`@mmnto/totem`).** `<totemDir>/artifacts/legs/<diffSha>.json`, `schemaVersion` `1.0.0`: `diffSha` (a 40-hex commit, and the file's own name), `readAt` (ISO-8601, the leg's own instant), `findings[]` (`id`, `severity` — `BLOCKING` / `MATERIAL` / `MINOR` — `file`, `line`, `claim`, `counterexample`), `folded[]` (finding ids, a subset of `findings[].id`), and a one-line `verdict`. It is keyed by the READ sha rather than content-addressed, deliberately: the question is _was this head read_, so two legs at one head are the same answer and a rerun supersedes by intent — the writer is create-exclusive and `--replace` DISCLOSES the `readAt` it overwrote instead of silently deduping. Validation runs on the way out, so a writer bug never poisons the store. The loader never throws: a deposit is a hand-editable file a pre-push hook reads on every push, so an unreadable, non-JSON, schema-invalid, newer-major or misnamed file becomes a named one-line `corrupt` row the caller prints, and never masks a valid sibling. Every string that reaches a hook's stdout refuses C0/C1 bytes at the schema boundary, at write and at read alike — a newline in a `verdict` would forge a second `[Totem]` line.

[...]

PR #2780, opened 2026-09-05
fix(ci): the legs-gate CI arm skips the release train — changesets/action's Version Packages PR deletes the consumed .changeset/** files and authors nothing, so every VP-PR was red at the hard step and the 2.1.0 cut could not execute (mmnto-ai/totem#2779)

### LINKED ISSUE #2779: ci: the legs-gate CI arm blocks the bot-authored Version Packages PR — `.changeset/**` is owed and the release train only deletes the consumed changesets; exempt `changeset-release/*` heads from the hard step and disclose the arm's reach (mmnto-ai/totem#2778 red at 2.1.0)
## Observed

- VP-PR mmnto-ai/totem#2778 (`changeset-release/main`, head `fee03d17`, generated by `changesets/action` on the push of `b8fe7edd`): Lint run 33934362266 fails at the step `totem legs gate (HARD — the review-leg floor)` with exit 3; `mergeStateStatus` is BLOCKED; `Totem Lint` is a required check of ruleset `main-required-checks`. The prior VP head `c44a5851` (run 33934284260) failed the same way. Every Version Packages PR from mmnto-ai/totem#2772 on is red at the hard arm, so the 2.1.0 cut cannot execute.
- The verdict line, verbatim:

```
[Totem] BLOCKED: this push is legs-owed (.changeset/** → .changeset/describe-active-rules-2765.md, .changeset/** → .changeset/legs-gate-adoption-2771.md, .changeset/** → .changeset/mcp-faulted-relevance-2770.md, .changeset/** → .changeset/spec-gate-fences-2769.md) and carries no fresh falsification-leg deposit for head d69b7f1e
```

- The owed match is `.changeset/**` against the four DELETED changeset files. The other twelve changed paths (six `CHANGELOG.md`, six `package.json`) match no owed glob. The `totem lint` step that follows PASSes on the same diff.
- The head the gate names, `d69b7f1e`, is the synthetic merge ref of the `pull_request` event, not the branch tip (status-claude's datum 2 from mmnto-ai/totem-status#236, observed here too). A deposit at the branch tip could never rank `exact` under the CI arm, and no deposit can be placed on this branch at all: the action force-pushes it on every push to main.
- The four `stale deposit … unknown to this repo` lines are the mmnto-ai/totem#2776 class (yesterday's squash-merged deposits), disclosed, not the cause.

## Why the arm should not fire here

`.changeset/**` is owed because a changeset carries an authored bump-level judgment (the minor-vs-major ruling on mmnto-ai/totem#2773 is the exhibit). The release train's PR authors nothing: it deletes the consumed changesets and renders their already-merged, already-legged text into `CHANGELOG.md` and `package.json`. There is no judgment for a leg to falsify and no seat is the diff's author. The arm's reach as written (`doctrine/model-tiering.md` § The gate; the `legs-gate` parity row: the CI arm runs where the repo's workflow triggers) does not yet name the release train.

## Fix

1. **Workflow (this repo):** the hard step carries `if: ${{ !startsWith(github.head_ref, 'changeset-release/') }}` and a ticket-ref comment naming this issue. `Totem Lint` triggers on `pull_request` only, so `github.head_ref` is always the PR's head branch and the guard is complete for this workflow; the branch prefix is `changesets/action`'s default (`changeset-release/<base>`), which `release.yml` does not override. The `totem lint` step is untouched.
2. **Doctrine (strategy's lane, dispatched with the 2.1.0 cut notice):** one clause on the arm's reach: a bot-authored release-train PR is outside the arm, and a `push` trigger is not a firing arm for this verb (status-claude's datum 4). 0.1.43 can codify both with the knob clause.

## Considered and not taken

- Dropping deleted paths from the owed scope (`--diff-filter`): a deleted doctrine file, ADR or skill is a judgment-dense change and stays owed.
- A deposit on the VP-PR: the branch is rebuilt by the action on every push to main, and the CI head is the merge ref.
- `hooks.legsOwed.enforce: 'advisory'` on main: defeats the arm on every PR to unblock one.
- Removing `.changeset/**` from the owed globs: takes the bump-level judgment out of the floor.

## Done when

The regenerated VP-PR's `Totem Lint` is green with the legs-gate step skipped and the `totem lint` step run; a feature PR that touches `.changeset/**` without a deposit still fails the step; the reach clause is on strategy's 0.1.43 list.

satur8d:
Closes #2779

## What

The hard `totem legs gate` step in `Totem Lint` (adopted in mmnto-ai/totem#2772, the legs-gate parity row's CI arm) is skipped for the release train and nothing else. A job-level predicate, defined once,

```yaml
env:
  LEGS_GATE_RELEASE_TRAIN: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.head_ref == 'changeset-release/main' }}
```

SELECTS the branch (exact name AND same repository, both load-bearing: a prefix test is widenable by any author, and a fork's `head_ref` is the bare branch name; GitHub's `==` and `startsWith` are both case-insensitive, so case is not the distinction). It does not exempt it. A `release-train check` step (`id: release_train`), gated on that predicate, runs `tools/release-train-shape.sh`, which writes `exempt=true` only when HEAD's diff against the base is exactly the release train's work: every changed path is a deleted `.changeset/*.md`, an added or modified `packages/<pkg>/CHANGELOG.md` (a new package's first release ADDS one: mmnto-ai/totem#2514) or a modified `packages/<pkg>/package.json` (`packages/*` is the workspace glob); the deleted changesets are exactly the pending set at the merge base, never a subset and never the README; a CHANGELOG was rendered whenever changesets were consumed; and with nothing pending only an empty diff is exempt. Rename detection is off so a deleted changeset never pairs with an added CHANGELOG. What is proven is that the OWED changes (the changeset deletions, `.changeset/**` being the owed glob) are precisely the train's consumption; the content of the rendered CHANGELOG and package.json is not an owed path, is not checked, and stays the cut sensor's remit at the cut word. That is the shape of every Version Packages PR to date (259 by head ref, 2,646 file entries, zero out of shape, the leg's census). The gate's `if:` is `${{ steps.release_train.outputs.exempt != 'true' }}`, so on any other branch the check is skipped, no output exists, and the gate runs. `shell: bash` with `set -euo pipefail` in the step and the script: a fetch or diff that fails aborts the step and the job rather than reading as an empty, in-shape list. Either outcome is announced with a static annotation and the reasons in the log; no derived text rides an annotation. The release PR is PAT-actored by design (`release.yml`), so no actor test could authenticate it; the diff can. The trust boundary is the arm's own: the deposit a feature PR commits is author-written too, so the floor was never a defence against a writer forging evidence.

Executed, not asserted, in `packages/cli/src/commands/release-train-shape.test.ts` (a throwaway git repository, thirteen cases plus `bash -n`, each not-exempt case pinning its reason line): the train is exempt; a new package's added CHANGELOG is exempt, even one whose text is the deleted changeset's verbatim; deleting a SUBSET of the pending changesets is not (the bypass Greptile named); deletions with nothing rendered are not; a deleted `.changeset/README.md` is not; the train plus an added doctrine file, plus the root `package.json`, or plus a whitespace-carrying path is not; an empty diff with changesets pending is not; rewritten CHANGELOG and package.json with nothing pending is not; an empty diff with nothing pending is exempt; an unresolvable base fails the derivation and writes nothing. The script was also run against the real release branch (`fee03d17` vs `origin/main`, merge base `b8fe7edd`): `exempt=true`.

The parity test in `packages/cli/src/commands/tools-hook-parity.test.ts` pins the shape on each step's own comment-filtered slice: the env predicate on the job header; the check step's id, its single `if:`, the diff command, both awk clauses, both output writes and both annotations; the gate step's single `if:`, no `startsWith(`, and its run block being exactly the base fetch plus the bare gate line (a `|| true` would otherwise pass unseen); the check preceding the gate. The test fails on every prior shape of the workflow. One sentence in `docs/wiki/enforcement-model.md` names the one diff shape outside the arm.

## Why

- mmnto-ai/totem#2778 (the 2.1.0 cut) is red at the step: Lint run 33934362266 exits 3 with `BLOCKED: this push is legs-owed (.changeset/** → …)` for head `d69b7f1e`. `Totem Lint` is a required check of ruleset `main-required-checks`, so the cut cannot execute. Every VP-PR since mmnto-ai/totem#2772 fails the same way.
- The VP-PR only DELETES the four consumed changesets and renders their already-merged, already-legged text into six `CHANGELOG.md` and six `package.json` files. `.changeset/**` is owed because an authored changeset carries a bump-level judgment; the release train authors nothing, so there is no judgment for a leg to falsify and no seat is the diff's author.
- No deposit can ride the branch: the action force-pushes it on every push to main, and the CI head is the synthetic merge ref, so `exact` is unreachable there anyway.

## What does NOT change

- A feature PR touching `.changeset/**` (or any owed glob) without a deposit still fails the step.
- The `totem lint` step still runs on the release train (it PASSes on the VP diff today, 16 warnings, 0 errors).
- `github.head_ref` is always set: this workflow triggers on `pull_request` only, so the guard is complete for it. The skipped step's `git fetch origin "$BASE_REF"` is not load-bearing for the later lint step, which does its own `--depth=1` fetch.
- The branch prefix is `changesets/action`'s default (`changeset-release/<base>`); `release.yml` does not override it.

## Considered and not taken

Dropping deleted paths from the owed scope (a deleted doctrine file or ADR stays owed); a deposit on the VP-PR (the branch is rebuilt each run); `hooks.legsOwed.enforce: 'advisory'` on main (defeats the arm on every PR); removing `.changeset/**` from the globs (takes the bump-level judgment out of the floor).

## After merge

The release workflow regenerates `changeset-release/main` on the push of this merge; the regenerated VP-PR's checks run under the amended workflow. The arm's reach clause (a bot-authored release-train PR is outside the arm; a `push` trigger is not a firing arm) goes to strategy with the 2.1.0 cut notice so 0.1.43 codifies it.

## Gate

Branch `totem lint` PASS (385 rules, 0 errors; the warnings are the frozen-lesson advisory class on the test file); `prettier --check` clean on all three files; the workflow parses with the notice and gate steps in order and the predicate on the `lint` job; ESLint clean; `tools-hook-parity.test.ts` 8/8; workspace build exit 0. No changeset: no package changes.

**Review leg (doctrine/model-tiering § Review legs):** four Opus passes, re-armed on every fold that rewired semantics. Pass 1 on `bcfc26b5`: no BLOCKING, two MATERIAL (the prefix guard was author-controlled and fork-reachable; nothing pinned the `if:`), two MINOR, all folded in `f8a1a109`. Pass 2 on that fold: no BLOCKING, seven text and test-scoping items, six folded in `c90f630f`, one disclosed as the residual that Greptile's P1 then named as reachable. Pass 3 on the shape-check fold `5b271cb5`: no BLOCKING, three MATERIAL holes in the shape check (an ADDED CHANGELOG in history; fail-open without pipefail; the test did not pin which branch writes `exempt=true`) and five MINOR, all folded in `963e7239`, convergence not named. Pass 4 on that fold: no BLOCKING, convergence named; one latent MATERIAL (`grep -q` under pipefail on a very large diff) and three accuracy items folded with executed probes. Greptile's follow-up then showed shape is not provenance (a subset deletion of pending changesets was exempt), answered by the mechanism in `d97023b7`, so the leg re-armed. Pass 5 on that fold: no BLOCKING, **convergence named**, Greptile's follow-up closed on both points; two MATERIAL (a narration overclaim on the nothing-pending class, no floor lost; assertion strength) and four MINOR folded with executed cases, no further re-arm. A bot round on `4f197868` (Gemini, Greptile, CodeRabbit) then yielded four folds with no change to the predicate: derivations fail loud in their own stages with `|| true` only on single grep stages, `printf` over `echo`, the step lookup scoped to the `lint` job with an exact-count match, and one bookkeeping fix to the deposits (a disclosed finding removed from the folded lists, rewritten in place with their original read instants). Deposits ride the branch under `.totem/artifacts/legs/`: `f8a1a109….json` (passes 1 and 2), `963e7239….json` (passes 3 and 4) and `d97023b7….json` (pass 5, keyed to the head it read); the pre-push gate credits the nearest ancestor and discloses the commits since.

https://claude.ai/code/session_018wCSG9uto8ffJVbGfZumRk


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@coderabbitai review
@greptileai review

satur8d:
@greptileai — disposition of your round on `7962f155`, addressed in `5b271cb5`. Every finding verified against the current code before acting.

**P1 · Branch Name Bypasses Gate — FOLDED, as a mechanism.** The premise held: a writer with push access could put legs-owed changes on the existing release branch and the branch-name predicate would skip the hard step. Actor authentication is not available by design: `release.yml` PAT-actors the Version Packages PR (`RELEASE_TOKEN`) so its pushes trigger workflows, which makes the PR's actor the PAT owner, the same identity as a human seat. So the fix validates the exempt diff's provenance by its SHAPE. The job-level `LEGS_GATE_RELEASE_TRAIN` predicate now only SELECTS the branch; a new `release-train check` step (`id: release_train`) derives `git diff --name-status "origin/$BASE_REF...HEAD"` and exempts the gate only when every changed path is a deleted `.changeset/*.md`, a modified `CHANGELOG.md` or a modified `package.json`, which is the shape of every Version Packages PR to date (mmnto-ai/totem#2778, mmnto-ai/totem#2762, mmnto-ai/totem#2752, mmnto-ai/totem#2746, read from their file lists). It writes `exempt` to `$GITHUB_OUTPUT`; the gate's `if:` is `steps.release_train.outputs.exempt != 'true'`, so a skipped check (any other branch) writes no output and the gate runs, and a diff that cannot be derived fails the step. Either outcome is announced (`::notice` on exemption, `::warning` naming the out-of-shape paths when the gate runs). Executed, not asserted: the same awk over the real VP diff (`origin/main...origin/changeset-release/main`, 16 paths) keeps nothing; over this PR's own diff it keeps 4 paths; over the VP diff plus a simulated `A doctrine/new-rule.md` it keeps that line. The parity test pins the step's id, its single `if:`, the diff command, both awk clauses, both output writes and both annotations. The residual that remains is the one the shape defines: a writer who pushes ONLY consumed-changeset deletions and CHANGELOG/package.json edits onto the release branch is, by construction, performing the release train's own work.

**P2 · Notice Assertions Lack Scope — FOLDED.** `stepNamed(workflow, namePrefix)` slices one step from its `- name:` to the next step's, comment lines dropped; `ifLinesOf()` returns a step's trimmed `if:` lines. The check step's assertions (id, exactly one `if:` equal to `env.LEGS_GATE_RELEASE_TRAIN == 'true'`, the run's diff/awk/output/annotation lines, no `continue-on-error`) and the gate step's (exactly one `if:` equal to the output test, no `startsWith(`, the run block being exactly the fetch plus the bare gate line) all read their own step's slice, never the whole workflow.

**Leg:** this fold rewires semantics, so the falsification leg is re-armed on `7962f155..5b271cb5`; its deposit follows as its own commit. Until then the credited evidence is the ancestor deposit at `f8a1a109`, which the pre-push gate disclosed as `+3 commits since the leg read`.

satur8d:
@coderabbitai — disposition of your round on `7962f155` (2 actionable), addressed in `5b271cb5`. Each finding verified against the current code before acting.

**Minor · Assert the predicate on the notice step — FOLDED.** The notice step became the `release-train check` step (see below), and every assertion now reads one step's own slice: `stepNamed(workflow, namePrefix)` cuts from the step's `- name:` to the next step's with comment lines dropped, and `ifLinesOf(step)` returns its trimmed `if:` lines. The check step is asserted to carry `id: release_train`, exactly one `if:` equal to `${{ env.LEGS_GATE_RELEASE_TRAIN == 'true' }}`, the `git diff --name-status` command, both awk clauses, both `$GITHUB_OUTPUT` writes and both annotations; the gate step exactly one `if:` equal to `${{ steps.release_train.outputs.exempt != 'true' }}`. Nothing is searched in the full workflow any more except the job-header env line, which reads the comment-filtered job header.

**Trivial · Replace repeated magic values with named constants — FOLDED.** Module-level `LF` (built with `String.fromCharCode`, the banked decode-trap idiom, with the reason in the comment), `NOT_FOUND`, `GATE_STEP_NAME` and `RELEASE_TRAIN_STEP_NAME`; every raw `-1` and `10` in the helpers and assertions replaced.

**Also in the same commit (Greptile's P1, for your next incremental pass):** the release train is now identified by its diff shape, not its branch name. The check step derives `git diff --name-status "origin/$BASE_REF...HEAD"` and exempts the gate only when every path is a deleted `.changeset/*.md`, a modified `CHANGELOG.md` or a modified `package.json`; anything else pushed onto the release branch meets the gate, a skipped check writes no output so the gate runs, and a failed diff fails the step. The awk was executed over the real Version Packages diff (keeps nothing), this PR's diff (keeps 4 paths) and a simulated hijack (keeps the added doctrine file).

**Leg:** the fold rewires semantics, so the falsification leg is re-armed on `7962f155..5b271cb5`; its deposit follows as its own commit.

satur8d:
@greptileai — disposition of your follow-up (comment 5550374978) on `5b271cb5`, addressed in `d97023b7`. Both points verified against the code before acting.

**P1 remains open (shape ≠ provenance; deleting a pending changeset) — FOLDED, as a mechanism, with your regression case.** The premise held: a diff of only deleted `.changeset/*.md` satisfied the shape, so a same-repository push to the release branch could drop one pending changeset and be exempted. The check now proves the diff is EXACTLY the release train's work, in `tools/release-train-shape.sh`, and writes `exempt=true` only when all three hold: every changed path is what the action writes (a deleted `.changeset/*.md`, an added or modified `packages/<pkg>/CHANGELOG.md`, a modified `packages/<pkg>/package.json`); the deleted changesets are exactly the pending set on the base ref, read from `git ls-tree`, never a subset and never the README; and a CHANGELOG was rendered whenever changesets were consumed. Eleven executed cases in `packages/cli/src/commands/release-train-shape.test.ts` build a throwaway git repository and read what the script writes to `$GITHUB_OUTPUT`: your subset-deletion case is not exempt (`not exactly the pending set`), deletions with nothing rendered are not, a deleted README is not, the train plus an owed doctrine file / the root `package.json` / a whitespace-carrying path is not, an empty diff with changesets pending is not; the train and a new package's first (ADDED) CHANGELOG are exempt; an unresolvable base fails the derivation and writes nothing. The script run against the real release branch (`fee03d17` vs `origin/main`) writes `exempt=true`.

On the attestation ask, declined with the boundary stated in the workflow comment: the release PR is PAT-actored by design (`release.yml`), so no actor or run identity distinguishes it from a human seat, and any attestation a workflow could write (a status, a manifest) is writable by the same repository writer. The floor's own evidence, the leg deposit a feature PR commits, is author-written too; the CI arm was never a defence against a writer forging evidence, it is the sensor that a leg read the head. Within that boundary, what the exemption now proves is that the owed changes in the diff (the changeset deletions, `.changeset/**` being the owed glob) are precisely the train's consumption of main's pending set. CHANGELOG and package.json content are not owed paths; their correctness is the cut sensor's remit at the cut word, not the legs floor's.

**`$other` in the `::warning` line — FOLDED.** No derived text rides an annotation any more: both annotations are static and the reasons print to the step log as plain lines. For the record, the runner identifies a workflow command by the line it reads from stdout and only then decodes `%0A` inside the message, so a `%0A` in a path could not have started a second command; the surface is removed regardless.

**Leg:** this fold rewires semantics, so the falsification leg is re-armed on `a71e65a6..d97023b7`; its deposit follows as its own commit.

satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@gemini-code-assist — disposition of your round on `4f197868`, addressed in `982ff4b0`. Verified against the code before acting.

**High · `printf '%s\n'` over `echo "$diff"` in `tools/release-train-shape.sh` — FOLDED, with a correction to the premise.** The script is bash (`#!/usr/bin/env bash`, invoked as `bash tools/…`), and bash's builtin `echo` does not interpret backslash sequences unless `xpg_echo` is on; measured on the seat's bash (`shopt xpg_echo` → off; `echo "a\tb"` prints the four characters). The leg's pass 4 had verified the same. So no filename could have been altered on the path the script runs on. The fold is taken anyway because it removes the dependency on that shell option: every stage that carries git output to a filter now uses `printf '%s\n'`, and the script's header says why.

satur8d:
@greptileai — disposition of your round on `4f197868`, addressed in `982ff4b0`. Verified against the code before acting.

**P2 · Derivation failures are suppressed — FOLDED.** The premise held: `pending`, `deleted` and `other` were each one pipeline ending in `|| true`, so a failing `git ls-tree`, `sed` or `sort` would have read as an empty set, and with an empty diff the empty pending set is the exempt path. Derivations and filters are now separate stages: every git call and every `sed` / `sort` stage runs without a `|| true` and aborts the step under `set -euo pipefail`, and a `|| true` sits only on a single `grep` stage, where exit 1 is the expected "no line matched" state and nothing else can fail. The 13 executed cases still pass, the bad-ref case still fails the derivation and writes nothing, and the script re-run against the real release branch writes `exempt=true`.

Your issue comment acknowledging the P1 closure is read and noted with thanks; the re-armed leg's pass 5 also closed it on both points and named convergence.

satur8d:
@coderabbitai — disposition of your round on `4f197868` (2 actionable), addressed in `982ff4b0`. Verified against the code before acting.

**Minor · Remove `p3-f8` from `folded` in both deposits — FOLDED.** Correct: that finding's counterexample reads "DISCLOSED, no change", so it is not addressed and did not belong in `folded`. Both deposits are rewritten in place through `totem legs deposit --replace` with their original read instants (`963e7239…`: folded 22 → 21; `d97023b7…`: 29 → 28), and the verdict text records the bookkeeping.

**Minor · Reject duplicate workflow-step matches — FOLDED.** `stepNamed` now slices the `lint` job first (`lintJobOf`: from the job key to the next top-level job key or end of file, comments dropped) and returns empty unless exactly one step in that job carries the name, so a decoy step in another job, or a second step with the same name, fails the `not.toBe('')` assertion instead of satisfying the pins. `lintJobHeaderOf` reads the same slice.

PR #2850, opened 2026-09-08
Version Packages

satur8d:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

- 7716929: A third gate in the gate engine: `merge-ready` (mmnto-ai/totem#2800) — a PreToolUse gate on the **Bash and PowerShell** tools that reads a five-point deterministic pre-merge floor off GitHub before `gh pr merge` runs, with zero LLM calls and zero writes. `totem gate install merge-ready` (or `--all`, or `totem init --gates=merge-ready`) installs it under a `Bash|PowerShell` matcher beside `transport-shield`.

[...]

PR #2844, opened 2026-09-08
feat(gates): merge-ready — a deterministic five-predicate pre-merge floor as a PreToolUse gate on gh pr merge and a gate check verb, evaluated from GitHub with zero LLM calls; exact-by-marker severity with its ADR-109 budget corpus; one bot-identity definition; tier split in the engine (mmnto-ai/totem#2800)

satur8d:
## What

Promotes the user-level `merge-gate.cjs` prototype into the product as `merge-ready`, a third deterministic gate in the core registry: a PreToolUse hook on `gh pr merge` (Bash and PowerShell) and the verb `totem gate check --event merge-ready` that evaluate the charter's five-predicate pre-merge floor from GitHub with zero LLM calls — green checks, no unresolved bot review thread, no standing changes-requested, no bot HIGH inline still applying to the head commit, a mergeable state. Serves mmnto-ai/totem#2800; the two preflight specs (`.totem/specs/2800.md`, `.totem/specs/2801.md`) ride in as the first four commits.

## How

- **Evaluator in core** (`packages/core/src/merge-ready.ts`), a pure function of (payload, gh responses, env, tier) behind an injected `GhRunner`; one GraphQL document with explicit cursors on reviews, threads and the checks rollup, a second document only to resolve a branch to its PR number; the head sha is read once and every later page is checked against it; a failed or incomplete read is unevaluable, never clean (R2). Zero checks passes as a fact with the count in provenance and one stderr line (R5). Predicate 4 compares the commit a bot HIGH inline currently applies to (`commit.oid`) against the head, resolution ignored — the prototype's semantics, ruled on the first leg's finding; witness capture on mmnto-ai/totem#2827.
- **Tier split in the engine, per gate** (R1): the verb gains `--tier strict|pilot` (default strict); only merge-ready returns `warn` under pilot for the unevaluable class; freeze-check and transport-shield fail closed at every tier. The wrapper forwards `--tier` only when its baked tier is not strict, so a strict install stays runnable on a 2.2.x CLI; a pilot install discloses its 2.3.0 floor at install time.
[...]

PR #2843, opened 2026-09-08
feat(signon,mail): signon derives the seat FIRST — totem mail --derive-seat corroborates the inherited identity, the cohort map is keyed on the git origin so worktrees resolve, assignment mail precedes orient, a BLIND-round kit ends signon (mmnto-ai/totem#2801)

satur8d:
## What

Signon derives the seat FIRST. The managed signon skill opens with a step that reads the inherited `TOTEM_SELF_AGENT`, uses it and never writes a different one, stops and asks the operator when it is empty, and surfaces a refusal when the declaration and the repo disagree; the seat-anchored mail poll moves ahead of orient, journal and every knowledge search on hook-less seats; a BLIND-round kit (the subject prefix strategy-claude ruled) ends signon at the kit. The probe behind step 0 is a new `totem mail --derive-seat`, and the cohort seat map is now keyed on the git origin's repository so a per-agent worktree resolves the same seats as the resident checkout. Serves mmnto-ai/totem#2801; the two preflight specs ride in as the first four commits (shared with the mmnto-ai/totem#2800 branch, identical content).

## How

- **`totem mail --derive-seat`** (`packages/cli/src/commands/mail.ts`, wired in `index.ts`): one stdout line `seat=<id> source=env` when the inherited env names exactly one seat this repo hosts; otherwise a stderr refusal naming the supplied value (or "unset") and every hosted seat, exit 2; "this repo hosts no seat" when the structural union is empty; contradictory with `--as` and `--all-seats` (exit 2); never reads an outbox or a processed dir (a filesystem spy pins it on both branches). Hosted-ness is the structural union only — config `host_agents`, else the seat dirs, else the cohort map — the same derivation `--as` uses; the env can never vouch for itself. Duplicate env entries are normalised as the poll normalises them; the config-omits-a-present-dir case names its cause (config replaces the dir set, mmnto-ai/totem#2141) instead of contradicting itself.
- **Cohort map keyed on the origin repository** (`packages/core/src/orchestration-resolver.ts`, `sys/git.ts`): layer 3 of `resolveSelfAgents` reads `git config --local --get remote.origin.url` through core's exec helper and parses the repository name (https, ssh, with or without `.git`), falling back to the directory basename when there is no origin. Layers 1–2 (env, config) return before the spawn. This is what makes a worktree named `totem-totem-claude-…` resolve `totem`'s seats; before it, the structural union there was empty and the mmnto-ai/totem#2204 gate fired for the wrong reason.
[...]

Issue #2800, opened 2026-09-06
gate(merge-ready): a PreToolUse gate on gh pr merge + a verb evaluating the deterministic pre-merge floor (CI green at the current head sha · no unresolved bot threads / CHANGES_REQUESTED / HIGH inline · branch current) — folds the user-level merge-gate.cjs prototype (2026-05-19) into the product


satur8d:
## Charter (strategy-claude, 2026-09-06; operator: "Greenlight your recs")

A gate event `merge-ready` in the gate engine: a PreToolUse hook on `gh pr merge` (Bash and PowerShell, command-position match) that evaluates the **deterministic pre-merge floor** for the named PR and denies the merge while any predicate fails, plus the same predicate as a verb — `totem gate check --event merge-ready --payload '{"repo":"owner/name","pr":N}'` — emitting the JSON `GateVerdict` for a human, a CI step, or another tool. Zero LLM. The fold source is the user-level prototype `~/.claude/hooks/merge-gate.cjs` on the crown seat's machine (219 lines, wired 2026-05-19 as a Tier-1 PreToolUse hook on Bash), whose own header says: "Promote to cohort-portable (.claude/hooks/ in repos) if validated across N>=5 PRs without false-positives." It has run on every strategy merge since; this issue is that promotion, into the product instead of a hand-copied script.

### Why now, and why the product

- The failure class the prototype was written for: a merge judged on the status-check rollup alone misses un-actioned inline bot findings, because `gh pr view --json statusCheckRollup` and `gh api .../pulls/N/comments` are different API surfaces. Empirical anchors in the prototype's header: mmnto-ai/totem#366 (R-1) and mmnto-ai/liquid-city#363 (R0), both missed a GCA HIGH inline despite green checks. Twice since, a merge moved past fresh CodeRabbit comments, once because of a subshell defect in a guard script (the `;`-separated chain class, `lesson-cea217b9`).
- Third-party evidence: Anthropic's Claude Code Insights report on this operator (2026-09-05) ranks a **self-verifying merge gate** as its first horizon item and describes it as an agent; it is a deterministic predicate, which is totem's thesis (rules you can enforce, state you can derive). The report is a datum that the market wants this verb.
[...]

PR #2852, opened 2026-09-08
feat(cli): totem resolve-threads <pr> — resolve only the bot review threads a round disposition evidenced, dry-run by default and --apply on the operator's go; the review-reply skill's done ends with that operator-gated step (mmnto-ai/totem#2841)

### LINKED ISSUE #2841: review-reply(cli): a deterministic verb resolves bot review threads as the disposition's last step — never a thread without a human reply, dry-run by default — so the merge-ready gate's predicate 2 meets the practiced routine (calibration: 5/5 recent PRs deny under strict)
## Why

The merge-ready gate (mmnto-ai/totem#2800) denies on any unresolved, non-outdated bot review thread — the charter's predicate 2, no severity floor. A calibration run of the built evaluator under strict against the five most recently merged totem PRs denied all five: two as unevaluable (a merged PR reports its merge state as unknown, an artifact), three on unresolved bot threads, including minor CodeRabbit findings that a disposition comment had declined with a reason but nobody had resolved in the thread. The gate is working as chartered; the practice is one step short. Ruled 2026-09-08 (operator, on totem-claude's recommendation): resolve the threads rather than weaken the gate, and mechanise the step so it is one command inside the disposition routine.

## What

A deterministic, zero-LLM verb that resolves bot review threads on a PR as the last step of a round disposition, plus the review-reply skill step that calls it.

- Reads the PR's review threads through GraphQL (`reviewThreads`, with `isResolved`, `isOutdated`, the root comment's author and the thread's comments) — the same read shape `spine-review-thread-source.ts` already wraps behind the `ReviewThreadSource` port, and the same bot-identity source the merge-ready gate single-sources in core (`BOT_REVIEWER_IDENTITIES`).
- Selects threads rooted by a bot identity that are unresolved and not outdated.
- **Invariant: never resolves a thread that carries no human reply.** A folded finding's thread must carry the fix-push acknowledgement or a reply; a declined finding's thread must carry its reason. A silent resolve would hide a finding from the next reader, which is the failure the gate exists to catch. Threads with no human reply are listed as "needs a reply first", never resolved.
- Resolves the selected threads through the `resolveReviewThread` mutation, one call per thread, and prints one line per thread (resolved / skipped: no human reply / skipped: outdated / skipped: already resolved).
- Dry-run by default; `--apply` mutates. The review-reply skill's `done` step assembles the list, shows it, and applies only on the operator's explicit go, like every other mutation in that skill.
- Optional selection by thread id or by finding number from the `totem triage-pr` inbox, so a partial round can resolve only what it addressed.

## Not in scope

- Any change to the merge-ready gate's predicates (predicate 2 stays: no severity floor).
- Auto-resolving on a fix push without a disposition.
- Bots' own resolution behaviour (CodeRabbit marks addressed findings in its summary but leaves the thread open; that is why this verb exists).

## Sequence

This lands before merge-ready is installed anywhere. The install order ruled the same day: fixture record on mmnto-ai/totem#2800 → merge → publish → `totem gate install merge-ready --pilot` on totem for two or three PRs under the new routine → strict once a pilot run shows zero warnings on a PR that went through the routine.

## Related

mmnto-ai/totem#2800 (the gate and its calibration measurement), the review-reply skill (`.claude/skills/review-reply/SKILL.md`, managed constant in `packages/cli/src/commands/init-templates.ts`), `packages/cli/src/commands/triage-pr.ts` (the inbox), `packages/cli/src/commands/spine-review-thread-source.ts` (the existing GraphQL thread reader), mmnto-ai/totem#2626 (ghcq has no listener; its threads are audit-trail only).

satur8d:
## What

A deterministic, zero-LLM verb, `totem resolve-threads <pr>`, resolves the bot review threads a round disposition has answered, as the last operator-gated step of the review-reply skill's `done` flow. It exists because the merge-ready gate's predicate 2 (mmnto-ai/totem#2800) denies on any unresolved, non-outdated bot thread, and a strict run over the five most recently merged totem PRs denied all five, three of them on minor findings a disposition had declined in prose but nobody had resolved in the thread. Ruled on 2026-09-08: resolve the threads, do not weaken the gate, mechanise the step. Serves mmnto-ai/totem#2841; the preflight spec and its R1–R4 rulings ride in as the two spec commits.

## How

- **Reads** the PR's review threads through one paginated GraphQL document (thread id, resolution, outdated flag, path, each comment's database id, author and instant; a second document by node id for a thread's comments past its first page) and the PR-level comments through the REST issue-comments endpoint, both behind one injected exec seam so the suite runs offline. An unfollowable page or a blown page budget is a hard error with no partial plan — the same no-silent-shrink rule the spine's thread reader carries.
- **Selects** threads rooted by a review-bot identity (`BOT_REVIEWER_IDENTITIES` from core, the exact GraphQL login test) that are unresolved and not outdated, and derives each one's evidence: any non-bot comment after the root in the thread, OR any non-bot PR-level comment created later than the root (R2 — the only lawful answer to a Gemini thread is at the PR level, and the round disposition is that comment). "Non-bot" on the REST side excludes GitHub Apps and `[bot]` logins as well as the four review bots, so the release cut sensor's comments never count as a human disposition.
- **Plans** one row per bot-rooted thread — `resolve`, `skip:already-resolved`, `skip:outdated`, `skip:not-selected`, `skip:no-evidence` — printed with the REST root id beside each; `--json` carries the same rows. A thread with no evidence is listed as needing a reply first and is never resolved under any flag.
- **Mutates** only under `--apply`: `resolveReviewThread` once per `resolve` row, continuing on a per-thread failure and exiting 2 at the end naming the count; a selected no-evidence row makes the exit 2 after the other rows were applied. Dry-run is the default and performs zero mutations. The verb never posts a comment, reply or review (an exec spy asserts every argv the verb can emit).
- **Selection** (R4): every evidenced bot thread by default; `--ids <REST root ids>` narrows; an unmatched or malformed id aborts with exit 2 before any mutation.
- **Skill step** (R3): the review-reply `done` flow, after the disposition comment posts, runs the verb dry, shows the plan, and applies only on the operator's explicit go; the three distributed surfaces stay byte-locked to the constant with string locks on the new step.

## Verification

Built stacked on the mmnto-ai/totem#2800 branch, then brought forward onto main at `6b984c33` (after mmnto-ai/totem#2843, mmnto-ai/totem#2844 and mmnto-ai/totem#2848 merged) with `git rebase --onto`: fifteen commits replayed, no conflicts, `git range-diff` reports every one `=` (patch-identical). On the rebased tree: full workspace build + test 12/12 tasks green (`pnpm build` then `pnpm test`), `pnpm format:check` clean, ESLint 0 errors, `totem lint` PASS. The verb's own suite: `resolve-threads.test.ts` (the fixture capture of a real round plus the synthetic arms — dry-run performs zero mutations, `--apply` resolves only evidenced threads, the allowlist exec spy names every argv the verb can emit, the failure and success `--json` documents locked), `resolve-threads-cli-wiring.test.ts` (the mmnto-ai/totem#2097 seam), and `init.test.ts` (the three review-reply surfaces byte-locked to the constant).

Three legs, every deposit committed under `.totem/artifacts/legs/` (read 56fcab96, b8c65ec1 and — after the rebase onto main — 70351e0e, which the gate reads as the nearest ancestor of HEAD). The third, scoped to the rebase's interactions, named termination: the registry the verb consumes (`bot-identity.ts`) is blob-identical at the stack base, the head and main, so the verb was written against the final module; the three review-reply surfaces are byte-locked on this branch; the CLI seam holds; the wiki entry is placed and accurate; this `@mmnto/cli` minor folds into the 2.3.0 cut if merged before mmnto-ai/totem#2850, else rides 2.4.0. Two minor follow-ups it named are deferred to the pilot-install PR that re-renders this skill: the bot-identity parity sensor scans two consumers by name and this verb is a third (its `[bot]` suffix rule would need to move to core to join the scan), and the skill's step 4 should say in one clause that a clean resolve run is not an allow verdict — the gate's other predicates, a HIGH inline on the head commit or an unreadable read, still decide. Leg 1 (eleven findings, none blocking, two material): the never-writes spy was a denylist a `gh pr review` mutant escaped, rebuilt as an allowlist over the verb's five exact exec shapes and run in the dry-run test too; an App's in-thread reply counted as human evidence because the read never fetched the author's typename, so both documents now select it and a Bot typename, a `[bot]` suffix or core's exact list all mean bot. Nine minors folded: the mutation response is parsed and `isResolved` required, the narrative states R2's rule rather than "the disposition comment", a disclosed limit on the loose CodeRabbit pattern, the capture carries all seven PR-level comments, not-found fixtures shaped like the observed exit-1 body, the failure JSON document locked, the ordering locks assert presence first, the deleted-account rule recorded as a judgement on the design. Leg 2 (four minors, termination named): the row set is the four review-bot roots while the human-reply test is the three-arm one, stated in the wiki and the changeset; the fixture bot list derives from core's identities; the success JSON document's key set and every summary field asserted; one narration corrected in a test comment. Declined: none.

## Not in this PR

- Any change to merge-ready's predicates (predicate 2 keeps no severity floor).
- Auto-resolving on a fix push without a disposition.
- Selection by inbox number (R4: not stable across re-runs).
- The pilot install of merge-ready on totem, which follows this PR and the 2.3.0 cut.

## Related

Closes mmnto-ai/totem#2841

See mmnto-ai/totem#2800 (the gate and the calibration measurement), mmnto-ai/totem#2626 (ghcq has no listener), mmnto-ai/totem#2097 (the global `--json` collision the wiring test now locks against).

https://claude.ai/code/session_01MKX3XdibxYeiV3AwETsykH


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round 1 disposition — 4 findings (greptile P1 + P2, CodeRabbit major + trivial; gemini none), each verified against source; 2 folded, 2 declined with the measurement or the source named; one scoped falsification leg on the fold, termination named for the round

@greptileai @coderabbitai

Commits on this branch since the round: `80ef9cae` (the fold), `06410a2b` (round 2 — the leg's four folds: a structural test on the registration source so the action cannot regain the precondition, two probe-shaped mutant rows, the probe's 60 s timeout and the three old stderr lines named in the changeset), `441907f7` (the leg's deposit: eight findings, one material, three held under measurement). Head is `441907f7`; full workspace build + test 12/12 on the folded tree, the cli suite 4879 green after round 2, prettier and ESLint clean, `totem lint` PASS, the legs gate reading the deposit as the nearest ancestor covering 5/5 owed paths.

### Folded

| # | Finding | Fold |
|---|---|---|
| 1 | greptile P1 `index.ts:704` — the JSON failure contract bypassed when gh is missing | Real against the changeset's own promise ("`{error, rows, exitCode}` on a failure"): the action's `requireGhCli()` exited before the command ran. The precondition now lives inside the verb — `gh --version` through its own seam, the sixth and first shape in the exec allowlist — and fails in contract in both modes with no further gh call; the action drops its precondition and says why. A test drives gh-absent in text mode and under `--json` and pins `calls === [['--version']]`. |
| 2 | CodeRabbit trivial `resolve-threads.test.ts:143` — the `MAX_PAGES` exhaustion branches uncovered | Folded as proposed: `threadsAlwaysHasNext` / `commentsAlwaysHasNext` on the harness; two tests drive both readers past the cap and pin exit 1, no rows, no mutation, exactly 20 pages read, the detail naming the cap. |

### Declined, with the measurement or the source

| # | Finding | Why |
|---|---|---|
| 3 | CodeRabbit major `resolve-threads.ts:781` — `gh api --paginate` emits one JSON array per page, so the read fails on a second page | Measured, not read: on this gh (`gh version 2.99.0 (2026-09-01)`) `gh api repos/mmnto-ai/totem/issues/2844/comments?per_page=2 --paginate` over a nine-comment PR — five pages — answered ONE JSON array of nine, with zero `][` boundaries; `JSON.parse` of the whole stdout succeeds. `--slurp` is what produces the array-of-page-arrays shape the finding describes (measured: outer 5, inner 2,2,2,2,1), so adding it and flattening would be the change that broke the read. The REST arm of `--paginate` merges array pages; the per-page emission is the GraphQL arm, which this verb never uses (it pages GraphQL with its own cursors). The read parses correctly as shipped. |
| 4 | greptile P2 `resolve-threads-cli-wiring.test.ts:16` — the wiring test mirrors the registration instead of exercising it | The repo's standing shape for the mmnto-ai/totem#2097 seam: `packages/cli/src/index.ts` registers every command at import with side effects and is not importable in a test, so each seam is sensed by a mirror that "must carry it to sense it" — `mail-cli-wiring.test.ts` is the same shape and was strengthened this week on mmnto-ai/totem#2843 for exactly the drift this finding names. The mirror here carries `optsWithGlobals` and both flag orders; a divergence in `index.ts` is caught by the CLI's own end-to-end runs and by review, as for every other verb. |

The leg also recorded, as a fact and not a fold: `totem orient` is the only sibling action carrying both `--json` and the action-level precondition, and its wiki entry promises no failure document, so this fold's rationale does not indict it.

`local-lane: none leg: 80ef9cae blocking=0 material=1 folded=4`

Issue #2845, opened 2026-09-08
legs(deposit + gate): record the leg's runtime-attested model in the deposit and mark the floor unmet without it — the declared model: is a request (Claude Code 2.1.257 SUBAGENT_MODEL_FORCE); the gate proves coverage, not the Opus claim (CodeRabbit on mmnto-ai/totem-strategy#1291)


satur8d:
## What

`totem legs deposit` records the runtime-attested model of the leg it deposits, read from the harness record (the subagent transcript's `message.model` series — the same surface strategy's `tokens:` line derives `legs=` from), and `totem legs gate` marks the floor UNMET when a deposit covering an owed path carries no model attestation, or one below the floor the doctrine names.

## Why

doctrine/model-tiering.md § Review legs (strategy's 0.1.47 batch, mmnto-ai/totem-strategy#1291) now states the rule the gate cannot enforce: a declared `model:` is a request — a harness can force another model over per-spawn and definition picks from outside the definition (Claude Code 2.1.257's `CLAUDE_CODE_SUBAGENT_MODEL_FORCE`) — so where a runtime record exists, the model of record is that record's attestation, never the declaration. `LegDepositSchema` today is `{ schemaVersion, diffSha, readAt, findings, folded, verdict }` (packages/core/src/artifacts/legs.ts): the gate proves coverage of the head and the owed paths and nothing about the model, so a deposit satisfies the floor without proving an Opus leg. CodeRabbit raised exactly this on the batch (the gate's model claim is unattested); the doctrine line discloses it and points here.

## Shape

- Deposit: an OPTIONAL `model` field — `{ id: '<vendor model id>', basis: 'attested' | 'declared' }` — written by `legs deposit` when it can read the leg's transcript (Claude Code: `<session>/subagents/agent-*.jsonl`, `message.model` per request; one model per deposit, the mode of the series, a mixed series disclosed), else `basis: 'declared'` from `--model <id>` when the seat passes it, else absent.
- Gate: a new arm beside coverage — `legs gate --model-floor <class>` (default from config, e.g. `opus`) marks the floor UNMET when a covering deposit's `model` is absent or `declared`, or attested below the floor; exit 3 owed-and-unanswered as today. Cross-seat / free-lane legs (another meter) carry `basis: 'declared'` by construction and are admitted only when the config names them as an accepted class — the doctrine's "stages a cross-seat or free-lane leg instead".
[...]

Issue #2833, opened 2026-09-07
tests(gates): nothing in CI pins the committed .claude/hooks/gate-wrapper.cjs against CLAUDE_GATE_WRAPPER — the install test pins a scaffolded temp copy and doctor --parity covers only SessionStart.cjs; add the byte lock beside the SessionStart one (and, optionally, a parity row)


satur8d:
## Observed

Nothing in CI compares the repo's committed `.claude/hooks/gate-wrapper.cjs` to the `CLAUDE_GATE_WRAPPER` template it is scaffolded from. `packages/cli/src/commands/gate-install.test.ts` pins a scaffolded TEMP copy against the template; `doctor --parity` covers only `SessionStart.cjs` (`packages/cli/src/commands/doctor-parity.ts`, the `SessionStart` comparator). The tracked `SessionStart.cjs` has a byte lock in `init.test.ts` (`tracked === CLAUDE_SESSION_START`); the tracked gate wrapper has none.

So a template change that forgets to regenerate the committed wrapper, or a hand edit to the committed wrapper, passes CI. The mmnto-ai/totem#2822 round verified identity by hand (imported the built template, `cmp` + `sha256`, 14106 bytes, match) — a check a test should own.

## Ask

One lock beside the existing `SessionStart.cjs` lock in `init.test.ts` (or in `gate-install.test.ts`): read the repo's `.claude/hooks/gate-wrapper.cjs` and assert it equals `CLAUDE_GATE_WRAPPER` byte-for-byte, with a message naming the regeneration path (`totem gate install <event>` drift-repairs it). Optional second half: a `doctor --parity` row for the gate wrapper across cohort repos, the same shape as the `SessionStart` comparator, so a cohort repo's stale wrapper reads as drift rather than silence.

## Done when

A deliberate one-byte edit to the committed wrapper fails the test; regenerating it passes. The parity half, if built, reports a stale wrapper on a fixture repo.

## Prior art

mmnto-ai/totem#2799 (the wrapper), mmnto-ai/totem#2804 (the build), mmnto-ai/totem#2821 (the install on totem), mmnto-ai/totem#2822 (the round that verified identity by hand).

Issue #2822, opened 2026-09-07
gates: a Bash|PowerShell-matched gate fails closed on a missing repo-local CLI and so blocks the bootstrap commands that would cure it (pnpm install / pnpm build / totem eject are all Bash) — a fresh clone cannot bootstrap through Claude Code; add a PATH resolution arm, name the real exits, disclose at install


satur8d:
## Observed (a reviewer's derivation, confirmed at source)

A `Bash|PowerShell`-matched gate installed in `.claude/settings.json` fails closed when the repo-local CLI is absent. On a fresh clone opened in Claude Code, the commands that would make the CLI present — `pnpm install`, then `pnpm build` in this monorepo — are Bash commands, so the gate applies to them and blocks them. The cure the wrapper prints, "Reinstall totem or run `totem eject`", is also a Bash command the same gate blocks. The only exits are a terminal outside Claude Code or a hand edit of `settings.json` through the Write tool. A session started in a fresh worktree is plausibly the same case (the hook resolves against its own cwd); not verified here.

The Write|Edit-matched `freeze-check` never had this property: a checkout needs no Write to bootstrap. The Bash matcher makes the fail-closed arm self-blocking.

## Mechanism, at source (4cf8a2b9; the generated wrapper is byte-identical)

- `.claude/hooks/gate-wrapper.cjs:146-154` (from `packages/cli/src/commands/init-templates.ts`, the gate wrapper template): `cliPath = join(process.cwd(), 'node_modules', '@mmnto', 'cli', 'dist', 'index.js')`; absent ⇒ stderr `<event> applies but the totem CLI is not resolvable; failing closed. Reinstall totem or run totem eject to remove the gate.` ⇒ `process.exit(2)`. The comment above it records the ruling: an APPLICABLE gate that cannot be evaluated for any reason fails closed (mmnto-ai/totem#2799; Tenet 4).
- In this monorepo `node_modules/@mmnto/cli` is the workspace link to `packages/cli`, so `dist/index.js` exists only after a build — two blocked commands, not one.
- No PATH arm: the wrapper never tries a `totem` binary on PATH, though `@mmnto/cli` is installed globally on every cohort machine and carries `gate check --payload -` from 2.2.0.

## Ask

[...]

PR #2855, opened 2026-09-09
feat(gates): merge-ready installed at the PILOT tier on this repository — the PreToolUse entry on gh pr merge at 2.3.0, the committed wrapper refreshed to the template it had drifted from; the two minors deferred from the resolve-threads round folded, and the parity sensor gains the escaped-suffix arm and the fourth consumer its leg found missing (mmnto-ai/totem#2841)

satur8d:
## What

The merge-ready gate (mmnto-ai/totem#2800) installed at the PILOT tier on this repository, as ruled after the 2.2.1 calibration read (strict denied five of five recently merged PRs on unresolved bot threads): `totem gate install merge-ready --pilot` at 2.3.0 adds one PreToolUse entry on `Bash|PowerShell` behind the transport shield — the wrapper evaluates `gh pr merge` commands only, forwards `--tier pilot` to `gate check`, and maps a deny to exit 0 with the verdict on stderr (advisory), a strict install being the follow-on after two or three clean pilot PRs. The install also re-scaffolded the committed `.claude/hooks/gate-wrapper.cjs`, which had drifted from the 2.3.0 template — the gap mmnto-ai/totem#2833 names (nothing in CI pins the committed copy).

## How

**The two minors deferred from the resolve-threads round** (mmnto-ai/totem#2841; the leg's F3 and F5, named in mmnto-ai/totem#2852's body), folded here as that body promised:

- **F3 — the parity sensor's missing consumers and its missing arm.** `bot-identity-parity.test.ts` scanned a hardcoded two-entry list while `resolve-threads.ts` spelled the App suffix `/\[bot\]$/i` in code of its own. The deferred finding said that line could not join the scan as-is; this PR's own leg refuted that premise — the sensor's literal scan looks for `[bot]` and a regex source spells it `\[bot\]`, so the line had always passed, unscanned in substance. The cure is therefore two-fold: the rule now lives in `@mmnto/totem`'s `bot-identity.ts` as `hasBotAppLoginSuffix(login)` (a new export, with its own tests in core), the verb injects it as a third predicate (`hasAppSuffix`), and the sensor gains an escaped-suffix arm with two mutant rows AND two more consumers — `resolve-threads.ts` and `merge-ready.ts`, the gate engine, which the leg found importing the module unscanned. Four consumers, four scanned; the "both consumers" prose in the module, the barrel and the sensor now counts four.
- **F5 — the skill's step-4 clause.** A clean `resolve-threads --apply` run is not an allow verdict, and it clears the unresolved-bot-threads predicate only when no unresolved, non-outdated thread rooted by a known review bot remains (a `skip:no-evidence` row stays unresolved and, under `--apply`, makes the run exit 2; a run narrowed with `--ids` leaves its unnamed rows as `skip:not-selected` and exits 0): the gate re-reads the PR on `gh pr merge`, and a bot HIGH inline whose commit cannot be read makes the evaluation UNEVALUABLE once every earlier predicate passes (pilot warns, strict denies — verified by the leg against the engine's call site). The review-reply skill's step 4 now says so and ends by reading the floor itself (`totem gate check --event merge-ready …`) and reporting that verdict beside the resolve rows before the merge word is asked for. Re-rendered through the CLI's own marker-based scaffolder (the leg reproduced the render byte for byte); the `.agents` twin byte-copied (mmnto-ai/totem#2788 is the durable fix).

**The wrapper refresh, characterized.** The committed wrapper was last written by mmnto-ai/totem#2822 (337 lines); mmnto-ai/totem#2844 changed the template without refreshing the committed copy. The leg ran the ORIGINAL wrapper on a merge-ready event: `no payload projection for event "merge-ready"; failing closed`, exit 2 — so the settings entry without the refresh would have blocked every Bash call, which is why the two ride one commit. The same file serves the live strict transport-shield entry; old and new measured identical on its allow and deny paths (0 bytes stderr, exit 0 on allow; the strict deny arm unchanged; strict forwards no `--tier`, so a 2.2.x CLI still answers it). Cost of the new entry on a non-merge Bash call: node startup plus the argv walk, about 60–80 ms, exiting before any CLI is resolved — the installer's fresh-clone fail-closed disclosure therefore applies only to an actual `gh pr merge`. Open question the leg could not ground in this repo: whether the host runs two same-matcher PreToolUse entries sequentially; strategy has carried the same shape since mmnto-ai/totem-strategy#1302.

**The bot round's fold (round 1).** CodeRabbit's Major held: the template's heredoc scanner lacked the preceding-character guard core's `transport-shield.ts` carries, so the second `<` of a here-string (`<<<`) opened a heredoc whose body swallowed every later line — a merge on the next line went unjudged, a fail-open path. Folded in the template, the committed wrapper re-scaffolded, and a regression row added to `gate-install.test.ts` that fails on the unfixed template ("the wrapper never spawned the CLI") and passes with the guard, plus the complement (a here-string as the merge's own operand still fires). CodeRabbit's Minor held: the step-4 clause now says the predicate clears only when no unresolved, non-outdated bot-rooted thread remains (a `skip:no-evidence` row stays and makes an `--apply` run exit 2; `--ids` leaves `skip:not-selected` rows and exits 0), re-rendered on all three copies and in the changeset. Greptile's P1 and P2 re-name two limits of the shipped projection (a wrapper program's operand, `command -p`, `env …`, and the executable spelled with an extension or a path — `gh.exe`, which was disclosed nowhere until this round and is now a disclosed-miss row and a template comment; and the projection's git reads running before the 30 s deadline, disclosed in the wrapper's own comment): the safe direction under the pilot, a bypass under strict — filed as mmnto-ai/totem#2856, the strict install's precondition, rather than folded into the install PR. **The scoped leg over that fold** (six findings) caught my own clause claiming both skip shapes exit 0 — the no-evidence arm exits 2, pinned by the verb's tests — and a second scanner divergence from core of the same fail-open class: a `#` glued to `$(` was not read as a comment, so a `<<word` inside it opened a heredoc that swallowed the merge line; the template gains core's substitution arm and a regression row that fails without it. The changeset and this body were carrying the pre-fold clause; both now carry the shipped one. **The leg over that second fold** confirmed every cure (the clause true on both arms of the verb, both scanner rows genuine falsifiers, no regression on any enumerated shape) and found two MORE divergences from core's scanner of the same class, both pre-existing in the 2.3.0 template: no paren-boundary arms (a `#` after `(` or an operator `)` is text, so `(true)#<<note` opens a heredoc) and a bare heredoc-delimiter class narrower than core's (`<<E:F` parsed as the prefix `E`, so the terminator never matches). Porting arm by arm is the gradient that produced three rounds; the durable cure is one shared scanner. Both are disclosed in the template's comment, locked as miss rows in the suite (a fix flips them to spawned rows), and filed as mmnto-ai/totem#2857 — the seat names convergence for this PR there: it ships the 2.3.0 template plus the two cures the round found, and the scanner's remaining parity is the gate's lane.

## Verification

- `pnpm install --frozen-lockfile`, `pnpm build` (6/6), the full `pnpm test` (12/12 tasks), `pnpm lint` (ESLint 0 errors), `format:check` clean, `totem lint` PASS (the leg counted nine advisory warnings on the first commit, all false positives — the anchored-regex lesson on the new `.test(login)` line, boolean `||` arms read as numeric defaults, test-file noise; after the fold and prettier it reports one, on the parity test's own source-scan line).
- The installed wrapper, smoke-tested end to end against the open mmnto-ai/totem#2854 with a synthetic `gh pr merge` envelope: a real GitHub read; at the seat's run two checks were still running, so `deny`, exit 0 under the pilot tier (advisory); by the leg's re-read the PR was green and the verdict `allow` — the deny-to-exit-0 mapping is pinned by the leg on an UNEVALUABLE case instead (pilot warn → exit 0, strict deny → exit 2). A non-merge command passes silently.
- A changeset (`@mmnto/totem` and `@mmnto/cli` minor; the fixed group takes every `@mmnto/*` package to the same minor).

One falsification leg over the diff before this PR opened (`init-templates.ts`, `.claude/skills/**`, `.agents/skills/**` and `.changeset/**` are legs-owed): eleven findings — two material on the sensor's own claims, folded above; the rest minor or notes, folded or disclosed here. Its deposit rides the branch at `49dbe6c6`, and the legs gate reads it as the nearest ancestor covering every owed path. The agent definition declares Opus; the runtime model is unattested in the deposit (mmnto-ai/totem#2845).

## Not in this PR

- No strict install and no `--gates` change to `totem init`'s defaults.
- No CI pin for the committed wrapper (mmnto-ai/totem#2833 stands).
- No `.agents/skills` write in `totem init` (mmnto-ai/totem#2788 stands).

## Related

mmnto-ai/totem#2800 (the gate), mmnto-ai/totem#2841 / mmnto-ai/totem#2852 (the resolve-threads round and its deferred minors), mmnto-ai/totem#2856 (the projection limits greptile re-named; the strict install's precondition), mmnto-ai/totem#2857 (the scanner's remaining parity with core; one shared scanner), mmnto-ai/totem#2833, mmnto-ai/totem#2788, mmnto-ai/totem-strategy#1302 (strategy's pilot install, the cohort's first).

https://claude.ai/code/session_01UEFs7GcBowjFEjbEKSGC8r


satur8d:
/gemini review
@greptileai review
@coderabbitai review

satur8d:
## Round 1 disposition — 4 distinct findings (CodeRabbit major + minor, greptile P1 + P2; gemini none), each verified against source; 2 folded at `a499c97c` and re-folded on two scoped legs' findings at `37f26786` and `e0b4e05a`, 2 declined as disclosed limits and filed

@coderabbitai — both findings held at source and are folded (`a499c97c`, then `37f26786` and `e0b4e05a` after the scoped legs over the folds):

1. **Major, `gate-wrapper.cjs` heredoc scanner (fixed).** Confirmed: the wrapper template (`init-templates.ts`) opened a heredoc at the SECOND `<` of a here-string, while core's `transport-shield.ts:401` carries the `command[i - 1] !== '<'` guard — `grep x <<< bar` followed by a merge line blanked the merge, a fail-open path. The guard is in the template, the committed wrapper is re-scaffolded from it by `totem gate install merge-ready --pilot`, and `gate-install.test.ts` gains a regression row that fails on the unfixed template ("the wrapper never spawned the CLI") and passes with the guard, plus the complement (a here-string as the merge's own operand still fires).
2. **Minor, review-reply step 4 (fixed).** The clause now reads: a clean `--apply` run is not an allow verdict, and it clears the unresolved-bot-threads predicate only when no unresolved, non-outdated thread rooted by a known review bot remains — a `skip:no-evidence` row stays unresolved and makes the run exit 2 (the verb's own tests pin that), a run narrowed with `--ids` leaves its unnamed rows as `skip:not-selected` and exits 0. Re-rendered from the template on all three copies through the CLI's own scaffolder (the `.agents` twin byte-identical) and carried into the changeset. My first fold of this line said both shapes exit 0; the scoped leg over the fold refuted it against the verb's exit-code arm before it posted.
5. **Beyond the round, from the legs over the folds:** a second divergence from core's scanner of the same fail-open class as the Major — a `#` glued to `$(` was not read as a comment, so a `<<word` inside it opened a heredoc that swallowed the merge line — is cured (core's substitution arm; a regression row that fails without it). The leg over that cure then found two more, both pre-existing in the 2.3.0 template: no paren-boundary arms (`(true)#<<note` then a merge line) and a bare heredoc-delimiter class narrower than core's (`<<E:F` read as the prefix `E`). Porting arm by arm is the gradient that produced three rounds; the cure is one shared scanner. Both are disclosed in the template's comment, locked as miss rows in the suite, and filed as mmnto-ai/totem#2857 — convergence for this PR: it ships the 2.3.0 template plus the two cures this round found; the scanner's remaining parity is the gate's lane.
3. **Description check (fixed).** The body now carries the template's headings (`## How`, `## Verification`, `## Review leg`).
4. **Static notes, not actionable:** SkillSpector's "hidden instructions" on both skill copies is the managed-block marker comment every distributed skill carries; the ast-grep non-literal-path warnings are the parity test's own source reads.

@greptileai — both findings re-name limits the shipped template discloses; declined for this PR and filed as mmnto-ai/totem#2856, the strict install's precondition:

1. **P1, exact `gh` token.** The projection's disclosed-miss rows in `gate-install.test.ts` already lock `sudo`, `timeout 30`, `env GH_TOKEN=x`, `command -p`, `exec -a x`, `eval "…"` and `time -p` as misses that never spawn — the safe direction under the pilot (no false deny; one lost advisory read), a bypass only under a strict install. `gh.exe` was disclosed nowhere until you named it: it and `./gh` are now disclosed-miss rows and a line in the template's comment, and the issue names the class. Widening the executable token is a change to the gate's projection (mmnto-ai/totem#2844's lane), not to its install.
2. **P2, projection before the deadline.** The wrapper's own comment above the deadline discloses it: up to three `gitRead`s per merge, each on a 10 s timeout, run before the 30 s aggregate deadline exists. Bounding them is the same follow-up.

Not addressed: gemini posted no finding.

local-lane: none leg: 49dbe6c6 blocking=0 material=2 folded=7

satur8d:
## Calibration replay line — mmnto-ai/totem#2861, Option B (operator-ruled 2026-09-16)

Under the ruling, the merge-ready gate's predicate 4 reads a resolved HIGH as dispositioned only when a non-bot PR-level comment after its root NAMES it on a per-thread line. The round disposition on this PR (2026-09-09T20:36:15Z) declined the Greptile P1 as a disclosed limit and filed it as mmnto-ai/totem#2856; this comment carries the machine line the review-reply template now emits for it, so the calibration replay can read this row:

disposition: 3972344503 declined

Issue #2875, opened 2026-09-16
orchestration-resolver: COHORT_AGENT_MAP lacks the four Kimi seats the roster of record seats — a to: lc-kimi dispatch is invisible to a fresh-clone poll while the signoff table names the seat (mmnto-ai/totem#2865 leg f1; the map's own docstring says it is kept in sync with that table)


satur8d:
## The drift

`COHORT_AGENT_MAP` in `packages/core/src/orchestration-resolver.ts` (the bootstrap fallback `resolveSelfAgents` layer 3 unions with the seat dirs, and `knownCohortAgents()` flattens for `mail send`'s recipient check) carries six keys — the four cohort repositories with their Claude and Gemini pair, `arhgap11` with its pair, and `totem-playground` empty — and no Kimi seat anywhere:

```
totem:            totem-claude, totem-gemini
totem-strategy:   strategy-claude, strategy-gemini
liquid-city:      lc-claude, lc-gemini
totem-status:     status-claude, status-gemini
arhgap11:         arhgap11-claude, arhgap11-gemini
totem-playground: (none)
```

The roster of record (`doctrine/cohort-roles.md` § 1.1 in mmnto-ai/totem-strategy) seats a Kimi lane in all four cohort repositories: `strategy-kimi`, `totem-kimi`, `lc-kimi` (2026-07-18, § 1.4) and `status-kimi` (2026-07-29, § 1.4). The map's own docstring says it is kept in sync with the `signoff` skill's step-2a table; that table now names the four Kimi seats (mmnto-ai/totem#2865), so the two copies disagree in the direction the map's job cannot absorb.

## Why it matters

The map exists for fresh-clone visibility: the orchestration tree is gitignored, so on a checkout with no seat dirs and **no `TOTEM_SELF_AGENT` declared** (the env layer resolves first and is exempt from the identity gate — a per-shell seat binding makes a Kimi poll work today), a dispatch addressed `to: lc-kimi` is invisible to `totem mail` until that seat has written once — the failure the map's `status-claude` comment records for the seat it DID add. Surfaced by the skill-text falsification leg on the mmnto-ai/totem#2865 fix (its f1), with that counterexample; the env qualifier is its re-armed leg's r2-f1.

## Why it is not in the mmnto-ai/totem#2865 fix

[...]

Issue #2865, opened 2026-09-13
skills(signoff): the step-2a cohort table lists liquid-city's and totem-strategy's Kimi seats as not seated while lc-kimi and strategy-kimi are seated (template drift at init-templates.ts:2734-2735; credit lc-kimi)


satur8d:
## The drift (credit: lc-kimi, signoff observation 2026-09-13T03:40Z, routed to the LC bus for the upstream touch)

The `signoff` skill template's step-2a cohort table — the hardcoded basename→agent-id map that says "keep in sync with the ADR-106 cohort list" — still lists the `liquid-city` row's Kimi column as `_(not seated)_`:

- template source: `packages/cli/src/commands/init-templates.ts:2735` at `1620d1c8` (`| \`liquid-city\` | \`lc-claude\` | \`lc-gemini\` | _(not seated)_ |`); the same row at `:2734` (`totem-strategy`) and `:2737` (`totem-status`) carries the same word.
- stamped copies: `mmnto-ai/totem:.claude/skills/signoff/SKILL.md:24` and every consumer's `.claude/skills/signoff/SKILL.md` (LC's at `:24`).

`lc-kimi` IS seated: `.totem/orchestration/lc-kimi/` exists in mmnto-ai/liquid-city, `totem mail --derive-seat` corroborates it from that seat, and the cohort roster of record lists it — `mmnto-ai/totem-strategy:doctrine/cohort-roles.md:50` (`| mmnto-ai/liquid-city | lc-claude | lc-gemini | lc-codex | lc-agy | lc-kimi |`) with the 2026-07-18 seating entry at `:85`. `strategy-kimi` is likewise seated per the same file (`:89`, "`strategy-kimi` dirs + `host_agents` entry live 2026-07-18"), so the `totem-strategy` row's `_(not seated)_` is the same drift; `totem-status` has no Kimi seat ("no status seat", `:85`) and its cell is correct.

The cohort-roles ledger already names this table as an owed follow-on: `cohort-roles.md:90` — "Remaining: registry propagation — … `mmnto-ai/totem:.claude/skills/signoff/SKILL.md` agent table (totem-claude lane)". This issue is that item, made a tracked artifact.

## Why it matters

[...]

PR #2871, opened 2026-09-13
feat(init): the public AGENTS.md floor — totem init scaffolds a sterile, marker-bounded floor and refreshes exactly its span; this repository's AGENTS.md rewritten to the T0 floor per the mmnto-ai/totem-strategy#619 inventory ruling, the sterility test locking the committed span to the template, the unauthenticated resolvability script, eject's scrub, and the sweep of the redirects, the Junie twin, the .claude/docs pages and the styleguide's deployment and operator lines (five falsification legs folded, convergence named)

satur8d:
The T0 public floor of `AGENTS.md`, per the inventory ruling on mmnto-ai/totem-strategy#619 (strategy-claude, 2026-09-12; rows 1–24 plus the class-5 acceptance test), applied after the overlay cut reached this repository's pin (`@mmnto/strategy-doctrine` 0.1.48, mmnto-ai/totem#2863): overlay-first, strip-second, as design v1 § 4 sequences it.

## What changed

**The floor is now the init scaffold.** `totem init` scaffolds an `AGENTS.md` when a repository has none: a title carrying the project name, a managed span between `<!-- totem:agents-floor:start -->` and `<!-- totem:agents-floor:end -->`, and a stub for the repository's own conventions (`AGENTS_FLOOR_BLOCK`, `renderAgentsFloorScaffold` in `init-templates.ts`; `scaffoldAgentsFloor` and `deriveProjectName` in `init.ts`). The span is the product-valid instruction set any consumer gets — session start, the working rules, review-bot conduct, the installed skills, the overlays line — and names no vendor, no private repository, no deployment, no doctrine tag and no agent identifier. Every later init refreshes exactly the bytes of the span and touches nothing else: the span is the first end marker paired END-anchored with the last start marker before it (the reflex scrub's pairing), so an orphan marker above the span never widens the refresh; the span is rendered in the file's own line terminator, so a CRLF checkout converges in one write with no bare LF; a second complete span is left alone and named; a second pass is a byte no-op; the template is prettier-stable (proved below). A repository's own `AGENTS.md` without the markers is never rewritten and is told once, dimly, how to adopt the span; an unpaired marker is named and left alone. It runs in every init mode, bare included, because agent instructions are not developer tooling. `totem eject` removes every complete span with the reflex scrub's seam contract (one blank line where the span sat, in the file's own line terminator; the file itself when the span was all it held; marker residue named, never attributed).

**This repository's `AGENTS.md` is the first render**, rewritten line by line to the ruling:

[...]

PR #2876, opened 2026-09-16
chore(skills): the review-reply own-line tagging sentence in the practice form, the signoff table's three Kimi cells and a truthful bootstrap-fallback sentence, a table-to-map sync test (mmnto-ai/totem#2865); the map's Kimi propagation filed as mmnto-ai/totem#2875

satur8d:
## Two distributed skill-text fixes, re-stamped

Per strategy-claude's dispatches of 2026-09-15 (item 2 of 19:48Z, corrected at 20:04Z, final form at 23:16Z after mmnto-ai/totem-strategy#1337 merged) and lc-claude's mmnto-ai/totem#2865. Both live in `packages/cli/src/commands/init-templates.ts`; the stamped copies in `.claude/skills/` were refreshed by `totem init` and the `.agents/skills/` copies mirrored byte-for-byte.

**`review-reply` step 2 — each tag on its OWN line, in the blessed short form.** "each tag on its OWN line (operator-ruled 2026-09-15; the line shape is practice, not a GCA guarantee, and position within the comment is unruled)". It binds the practice `doctrine/bot-protocols.md` § Tagging binds and asserts no mechanism; the observed datum stays in the doctrine, so the template adds no private reference a consumer cannot resolve (the first cut had added four liquid-city issue refs to a file that had none — the leg's f8 — and quoted a "two days earlier" interval the doctrine deliberately pins to dates — f4).

**`signoff` step 2a — the three Kimi cells (mmnto-ai/totem#2865).** `strategy-kimi` and `lc-kimi` (seated 2026-07-18) and `status-kimi` (2026-07-29; the issue's premise that totem-status has no Kimi seat was the 07-18 entry, superseded eleven days later — the leg's f2) now fill the cells that read "not seated". The table is named for what it is: the bootstrap fallback — three of the roster's five vendor columns plus two archived repositories ADR-106 still names — that rots when a seat is added; `totem mail --derive-seat` is the derivation and no cell overrides it.

**The map half, measured and routed (mmnto-ai/totem#2875).** Core's `COHORT_AGENT_MAP` says it is kept in sync with this table and carries the Claude and Gemini pair only, so on a fresh clone a dispatch addressed to a Kimi seat is invisible until that seat has written once (the leg's f1). Adding the four seats was built and measured here first: it moves 21 assertions in the resolver suite and 6 mail-accounting tests (broadcast denominators, the suspended-marker exclusion, the identity gate's per-seat floor) — a runtime change in `@mmnto/totem` with its own review surface, so it is its own issue rather than a rider on a skill-text patch. The table's sentence says the Kimi column leads the map until it lands, and a new test (`packages/cli/src/commands/signoff-table-sync.test.ts`) holds the two copies to each other: every seat the map carries is in the table, and the table leads by exactly the four Kimi seats — tightening to equality when the map gains them.

**Not changed, noted:** the adjacent clause "tags must be present when the comment is POSTED, never edited in (GCA's listener fires on comment-created only)" is carried verbatim by the doctrine's § Tagging, so it stays; the issue's second ask (replace the table with a derive-by-reference line) is a larger rewrite of step 2a and is left with mmnto-ai/totem#2875's roster-complete question.

**Gate:** workspace build, ESLint, prettier, the full workspace test run and `totem lint` on the branch; two falsification legs — the first on the cut (five material, four minor: folded or routed as above) and a re-armed one on the fold (two material, seven minor: the fresh-clone sentence's missing env qualifier and a private doctrine pointer the fold had added, both folded; the leg named convergence, mechanically closed) — deposited under `.totem/artifacts/legs/`; a patch changeset for `@mmnto/cli` (the fixed group takes every `@mmnto/*` package).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
@greptileai
@coderabbitai

## Round 1 disposition — mmnto-ai/totem#2876 (bots invoked by the operator at 02:10Z; three inline findings, one review with none)

| # | finding | disposition |
| --- | --- | --- |
| 1 | Greptile P2 — `signoff-table-sync.test.ts:83`: the parity check flattens every cell and the map into global sets, so a seat filed under the wrong repository row or vendor column keeps both sets unchanged and passes | **fixed** in `16172600` — the test groups the map's union back by the roster's `<shorthand>-<vendor>` naming and checks each row's cells by shorthand and by column, then holds per repository that the map's seats ⊆ the row and that the table leads the map by exactly its Kimi seat; a mis-filed seat now fails by name |
| 2 | Greptile P2 — `init-templates.ts:2740`: the distributed sentence said a Kimi seat resolves only where its seat dir exists without `TOTEM_SELF_AGENT`, omitting the `config.json` `host_agents` layer the resolver reads before the dirs | **fixed** in `16172600` — the sentence now reads "with no `TOTEM_SELF_AGENT` declared and no `config.json` `host_agents` naming it"; re-stamped into `.claude/skills/signoff/SKILL.md` and mirrored |
| 3 | CodeRabbit Major — `signoff-table-sync.test.ts:19-23`: replace the static imports with dynamic imports per the `packages/cli/**` lazy-loading instruction | **declined** — that instruction is the CLI command-file cold-start discipline: command modules must not load core eagerly because `index.ts` registers every command at start. A vitest file is never on that path; the sibling `config-drift.test.ts` and `init.test.ts` import the same modules statically and are the precedent. The mmnto-ai/totem#2871 round declined the same finding for the same reason and CodeRabbit recorded it as a learning then; nothing to fix |

Gemini Code Assist reviewed and reported no findings; no tag.

disposition: 4021893903 fixed
disposition: 4021893908 fixed
disposition: 4021900801 declined

local-lane: none leg: 2ce0b83f blocking=0 material=2 folded=8

Issue #2893, opened 2026-09-19
gates(merge-ready): the wrapper's tokenizer treats a double-quoted string as one word, so a $( … ) or backtick substitution inside double quotes runs unjudged — a strict-tier bypass; port core's tokenizeShell with a token-parity lock (the 2856 leg's F4)


satur8d:
### Symptom

`ghPrMergeArgvs` (the projection in `CLAUDE_GATE_WRAPPER`, `packages/cli/src/commands/init-templates.ts`) consumes a double-quoted string as a single token and never looks inside it for `$( … )` or a backtick pair. bash executes both inside double quotes:

```
bash -c 'echo "$(echo RAN_DOLLAR)"'   → RAN_DOLLAR
bash -c 'echo "`echo RAN_BACKTICK`"'  → RAN_BACKTICK
```

so `echo "$(gh pr merge 5)"` and ``echo "`gh pr merge 5`"`` merge the PR, while the wrapper projects nothing for either (measured through the export seam on the mmnto-ai/totem#2856 branch and on the pre-2856 hook alike). The unquoted forms `echo $(gh pr merge 5)` and ``echo `gh pr merge 5` `` project since 2856. Under PILOT this is one lost advisory read; under STRICT it is a bypass. A single-quoted string is genuinely data and stays a false-deny control.

### Where it was mis-recorded

The 2856 branch's first cut locked ``echo "`gh pr merge 5`"`` as a never-spawns row with a comment calling it "data, not a segment of its own", and the rendered hook said "a backtick inside quotes never reaches here" — a fail-open labelled as the safe direction. The falsification leg on that branch caught it (F4); the fold corrects the disclosure and moves the row to the disclosed-miss list pointing here. The `$( … )`-in-double-quotes twin was disclosed nowhere before this issue.

### Cure

[...]

Issue #2851, opened 2026-09-08
core(exports): @mmnto/totem's exports map carries import conditions only — a CommonJS require() is refused with ERR_PACKAGE_PATH_NOT_EXPORTED, which keeps strategy's gh-parity-twins.cjs a hand copy of parity-label-canon.ts and the prepare wrapper on a manual node_modules walk


satur8d:
## Datum

From strategy-claude's 2026-09-08T22:04Z dispatch, out of mmnto-ai/totem-strategy#1275's cure (the two label-canon readers' parity test): `@mmnto/totem` 2.2.1's `exports` map carries `import` conditions only, so a CommonJS `require('@mmnto/totem')` is refused with `ERR_PACKAGE_PATH_NOT_EXPORTED`, while a dynamic `import()` resolves `parseLabelCanon` / `namespaceTokensOf` fine.

## What it costs

That is what keeps `tools/gh-parity-twins.cjs` in mmnto-ai/totem-strategy a hand copy of `packages/core/src/parity-label-canon.ts` instead of a consumer of it — the retire-the-copy cure Tenet 21 prefers. The strategy parity test is the interim coupling and names this in its header. The same shape already bit this repo's own distributed CommonJS wrappers: `.totem/prepare.cjs` documents that `require.resolve('@mmnto/cli/package.json')` and `require.resolve('@mmnto/cli')` both throw `ERR_PACKAGE_PATH_NOT_EXPORTED` and walks `node_modules` by hand instead (init-templates.ts, the prepare-wrapper comment, mmnto-ai/totem#2410 PR-B).

## Cure shape

A `require` condition on `@mmnto/totem`'s exports (a CJS build beside the ESM one, or a small CJS entry for the pure readers), or a documented `./package.json` and `./readers` subpath — whichever the build can carry without splitting the barrel. Then the strategy twin consumes totem's reader and deletes its copy, and the prepare wrapper's manual walk can go too.

## Not decided here

Whether the whole package gets a dual build or only the dependency-free readers; the cost is a bundling step and a second entry to keep in parity. Filed as the datum; the design is a preflight.

Filed by totem-claude on strategy-claude's datum (the mmnto-ai/totem-strategy#1275 lane), 2026-09-08.

satur8d:
## Datum: the conditions objects disagree on order across the three published maps (the 2918 leg's F7)

Read from source at b3373d9a (2.9.2):

| package | `.` entry, conditions in declaration order |
|---|---|
[...]

PR #2894, opened 2026-09-19
fix(gates): the merge-ready wrapper judges the projection shapes that bypassed it (wrapper programs and their grammars, gh.exe and paths, eval, env -S, backtick substitution, redirections anywhere by the shell's operator-quoting rule), bounds its whole run by one budget with a lower-only --budget-ms and two named fail-closed arms, and carries core's heredoc scanner verbatim with a span-parity lock (mmnto-ai/totem#2856, mmnto-ai/totem#2857)

### LINKED ISSUE #2856: gates(merge-ready): the wrapper's projection misses executable forms of gh (a wrapper program's operand, gh.exe, command -p) and runs its git reads before the 30 s deadline — both disclosed at the pilot; the strict install's precondition (greptile P1 and P2 on mmnto-ai/totem#2855)
## Two disclosed limits of the merge-ready wrapper's projection, named by greptile on the pilot install (mmnto-ai/totem#2855) — filed as the strict tier's precondition

The committed `.claude/hooks/gate-wrapper.cjs` (the `CLAUDE_GATE_WRAPPER` template in `packages/cli/src/commands/init-templates.ts`, shipped at 2.3.0 by mmnto-ai/totem#2844) projects a `gh pr merge` from the Bash/PowerShell envelope and evaluates it through `totem gate check --event merge-ready`. Two limits of that projection are disclosed in the template's own comments and locked as "disclosed miss" test rows in `gate-install.test.ts`; greptile re-named them on the pilot-install round (P1 and P2, both on the refreshed wrapper). Under the PILOT tier every miss is the safe direction (no false deny, one lost advisory read); under a STRICT install the same miss is a bypass. So they are a precondition for the strict install that follows two or three clean pilot PRs, not a defect of the pilot.

### 1. The detector recognizes only a bare `gh` at command position

`ghPrMergeArgvs` (the wrapper's projection; find it by name — the fold rounds move line numbers) strips the command-position words (`exec`, `command`, `eval`, `time`, `if`, …) and assignment prefixes (`GH_TOKEN=x`), then requires `tokens[0] === 'gh'`. Forms that therefore run unjudged: a wrapper PROGRAM whose operand is the merge (`sudo gh pr merge 5`, `timeout 30 gh pr merge 5`, `env GH_TOKEN=x gh pr merge 5`), a builtin carrying a flag before `gh` (`command -p gh pr merge 5`, `exec -a x gh pr merge 5`), a merge handed over as one quoted word (`eval "gh pr merge 5"`), a backtick substitution — each a disclosed-miss row in the suite since the mmnto-ai/totem#2844 rounds — and greptile's new example on mmnto-ai/totem#2855, the executable spelled with an extension or a path (`gh.exe pr merge 5`, `./gh pr merge 5`), which was disclosed nowhere until that round and is now a disclosed-miss row and a line in the template's comment (`gh` resolves through PATH on this cohort's shells, so the spelling has not been observed in a hook envelope). Hardening candidates: treat a closed list of wrapper programs (`sudo`, `env`, `timeout`, `nice`, `command -p`, `exec -a NAME`) as transparent and read their operand; accept `gh.exe` / a path ending in `/gh` as the executable token. Any widening must keep the false-deny direction closed (the heredoc-body and quoted-string rows) and add a mutant row per form.

### 2. Projection runs before the 30 s aggregate deadline starts

The wrapper's evaluation loop (`const deadline = Date.now() + 30000`, find it by that line) opens a shared 30 s deadline across every projected payload, but the projection above it runs up to three `gitRead`s per merge, each on its own 10 s timeout, BEFORE that deadline exists — disclosed in the template's comment (round 3, F3 of the mmnto-ai/totem#2844 round). A multi-merge envelope with a hung git can therefore reach the host's hook timeout through the reads; if the host kills the hook there, the fail-closed exit never applies. Candidate cure: start the deadline before projection and pass the remaining budget into each `gitRead`, or bound the reads' total.

### Verification for either fix

`packages/cli/src/commands/gate-install.test.ts` drives the built template through `runWrapper` with a stub CLI: a fix row must show the projected payload for each newly recognized form and must keep the disclosed-miss rows that remain misses explicit; the deadline change needs a row with a stubbed slow `git` proving the fail-closed arm is reached inside the budget. The committed wrapper is re-scaffolded by `totem gate install merge-ready --pilot` (mmnto-ai/totem#2833 still leaves it unpinned in CI).

### Related

mmnto-ai/totem#2800 (the gate), mmnto-ai/totem#2844 (the build and its rounds), mmnto-ai/totem#2855 (the pilot install; greptile P1 and P2 there), mmnto-ai/totem#2833 (the wrapper's CI pin).

### LINKED ISSUE #2857: gates(merge-ready): the wrapper template's shell scanner is a hand copy of core's transport-shield scanner and has diverged in the fail-open direction (no paren-boundary arms; a narrower bare heredoc-delimiter class) — port the remaining arms or share one module; two divergences cured and two disclosed on mmnto-ai/totem#2855
## The merge-ready wrapper template's shell scanner is a hand copy of core's `transport-shield` scanner and has diverged from it in the fail-open direction — port the remaining arms, or share one module

`CLAUDE_GATE_WRAPPER` (`packages/cli/src/commands/init-templates.ts`) carries its own heredoc blanker for the merge-ready projection, described in its comment as "the shape core's `transport-shield` scanner uses" (`packages/core/src/transport-shield.ts`). It is a second copy of one grammar, and three rounds of falsification legs on the pilot-install PR (mmnto-ai/totem#2855) found four divergences, every one in the SAME direction: the template opens a heredoc core does not, the unterminated body blanks every later line, and a `gh pr merge` on one of those lines runs without a gate read. Under the PILOT tier that is one lost advisory read; under STRICT it is a bypass. Two were cured on that PR (the here-string guard `command[i - 1] !== '<'`; the command-substitution boundary arm) with regression rows that fail on the unfixed template. Two are disclosed there as misses and locked as rows, and are this issue:

1. **No paren-boundary arms.** Core (`transport-shield.ts:213-219`, implemented at `:367-377`) tracks which `(` each `)` closes: an opening `(` and an operator `)` begin a word, the `)` that closes a `$( … )` continues one. The template lets `(` and `)` fall to the default word-boundary rule, so a `#` after `(` or after an operator `)` is text, and a `<<word` in that text opens a heredoc. Reproduced by the leg with bash 5.3.15 executing the merge line: `(true)#<<note` ⏎ `gh pr merge 5` · `(#<<note` ⏎ `gh pr merge 5` ⏎ `)` · `(true; echo a)#<<note` ⏎ `gh pr merge 5` — the wrapper never spawns; core's `findHeredocs` returns no heredoc for any of them.
2. **The bare heredoc-delimiter class is narrower than core's.** Core's `HEREDOC_AT` reads a bare delimiter as `[^\s'"\\<>()|&;]+`; the template's bare-word loop accepts `[A-Za-z0-9_.\-\/]` only, so a delimiter carrying any other character (`<<E:F`, `E*F`, `E+F`, `E=F`, `E,F`, `E@F`, `E!F`, `EOF~`) is parsed as a prefix, the real terminator line never matches, and the body blanks to end of input. Reproduced: `cat <<E:F` ⏎ `body` ⏎ `E:F` ⏎ `gh pr merge 5` — bash terminates at `E:F` and runs the merge; the wrapper never spawns; core parses `delimiter="E:F", unterminated=false`.

### The cure

Arm-by-arm porting is the gradient that produced four rounds; the durable fix is ONE scanner. Candidates, in order: (a) have the template `require` core's scanner at run time — the wrapper already resolves the repo-local CLI, and `@mmnto/totem`'s `dist/transport-shield.js` exports `findHeredocs`; the wrapper's fallback when no CLI resolves is already fail-closed, so an absent module can take the same arm; (b) generate the template's scanner FROM core's source at build time so the two are byte-locked by a parity test (the `bot-identity-parity.test.ts` shape); (c) if the copy must stay, port core's paren arms and `HEREDOC_AT` verbatim and add a parity test that runs both scanners over the same corpus (every row in `gate-install.test.ts`'s heredoc and disclosed-miss lists plus the eight delimiters above) and asserts equal blanking.

### Verification

`gate-install.test.ts` drives the built template through `runWrapper` with a stub CLI: the two disclosed-miss rows added on mmnto-ai/totem#2855 (`(true)#<<note` ⏎ merge; `cat <<E:F` … ⏎ merge) currently assert NO spawn — a fix flips them to `spawnedPayload()` rows; a scanner-parity test over the shared corpus is the lock that ends the class. The committed wrapper is re-scaffolded by `totem gate install merge-ready --pilot` (mmnto-ai/totem#2833 leaves it unpinned in CI).

### Related

mmnto-ai/totem#2800 (the gate), mmnto-ai/totem#2844 (the build), mmnto-ai/totem#2855 (the pilot install; the two cures and the two disclosures), mmnto-ai/totem#2856 (the projection's own limits: the executable token and the deadline), mmnto-ai/totem#2833 (the wrapper's CI pin).

satur8d:
## The merge-ready wrapper judges the shapes that bypassed it, bounds its whole run by one budget, and carries core's heredoc scanner verbatim (mmnto-ai/totem#2856, mmnto-ai/totem#2857)

Both issues were the strict tier's precondition on every consumer (the status seat named them as such on mmnto-ai/totem-status#260): under PILOT each miss was one lost advisory read, under STRICT each was a bypass. The ruled design is `.totem/specs/2856.md` (sections A–G, the Implementation Design — scope, data model, lifecycle, the failure table, the invariants, three open questions ruled — and the four fold contracts the review round produced) and `.totem/specs/2857.md` (the scanner contract), both committed with the diff.

**Two design calls for the operator's review, ruled under the standing greenlight of 2026-09-19 and revisable here.** (1) The scanner cure is option (c) of mmnto-ai/totem#2857 — a verbatim port with an executable parity lock — not the issue's first-ranked run-time require. Measured on this machine by the census leg: `@mmnto/totem`'s exports map carries `import` conditions only and no scanner subpath (mmnto-ai/totem#2851), so `require('@mmnto/totem/...')` fails from anywhere; the package is linked only under `packages/cli/node_modules/` in this monorepo; a run-time require would need install-layout probing and a new fail-closed class on every layout it does not know. Re-openable when 2851 lands a `require` condition. (2) The budget is lowered — never raised — by a new hook argument `--budget-ms <n>` (either spelling; last-wins on a repeat; the installed settings.json line never passes it; the suite does), rather than an environment variable, so no ambient process can widen the window.

### What is judged now (2856 item 1)

- **The executable** may be `gh`, `gh.exe`, or either behind a path (the basename after the last `/` or `\`, `.exe` case-insensitive). A variable executable (`$GH`, `${GH}`) and an unquoted win32 path stay locked misses.
- **A closed table of transparent wrapper programs** is stripped with its option grammar before the anchor is read — `sudo`, `env`, `timeout`, `nice`, `nohup`, `command`, `exec`, `time` — which options take a separate operand (short and long forms), how many positionals the word eats, whether `--` ends its options, which options DESCRIBE rather than execute (`command -v`/`-V`; `sudo -l`/`--list`/`-v`/`--validate`/`-V`/`-K`) and therefore end the segment with no projection; the assignment strip re-runs after each; `time` is bash's reserved word (`time [-p] [--] pipeline`), so any other flag ends the segment; `eval` re-tokenizes its operand once (depth 1); `env -S` / `--split-string` SPLITS its operand on whitespace (env's own rule — a `;` inside it is not a separator to env) and splices the words in place of the option, in every spelling including the attached short form (`env -S'gh pr merge 5'`, `env -Sgh pr merge 5`). Locked misses, disclosed: `npx`, `xargs`, `bash -c` (not on the table); a table word spelled by path (`/usr/bin/time -f x …`); `timeout` with no duration; `eval` nested twice; env's own escapes and nested quotes inside an `-S` string (`env -S 'gh\_pr\_merge\_5'`, `env -S 'env -S "gh pr merge 5"'` both run the merge and read as fewer words here).
- **Redirections anywhere in the segment** are dropped with their file token before the wrapper strip and the anchor, by the shell's own rule: what makes a redirection real is the quoting of the OPERATOR, never of the word it sits in — `>"out.txt" gh pr merge 5` is a real redirection and projects `pr: 5`; `gh pr merge --squash ">"out.txt` hands gh the argument `>out.txt` and projects the current branch with that argument; `gh pr merge -b "<br>" 5` keeps its body and projects `pr: 5`; `gh pr merge --squash >"merge.log"` projects the current branch, never a fabricated one. The tokenizer records, per token, the index of its first character that came from quotes or an escape, and the strip drops a token only when the matched operator prefix lies entirely before it. Residue, disclosed and locked: the operators carrying `|` or `&` (`>|`, `2>&1`, `>& file`, `<& 3`, `exec 3>&1 …`) — the tokenizer's own separators end the token first, and ALL FIVE are merges the shell runs (a fail-open, not text the shell ignores); `&> file …` splits the same way but leaves a readable `>` and projects.
- **A backtick** becomes a segment separator in bash mode, so the operand of a backtick substitution is a segment of its own, and the backtick rides on as a token so a merge whose TARGET is a substitution reports `unresolvedTarget` rather than the current branch. In PowerShell mode a backtick-newline at a word boundary is that shell's line continuation (the next line continues the command); inside a word it embeds an escaped newline in the argument, which PowerShell also does, so `gh pr merg`+backtick+newline+`e 5` projects nothing.
- **A substitution INSIDE double quotes is a disclosed miss, not a control:** bash executes ``echo "`gh pr merge 5`"`` and `echo "$(gh pr merge 5)"`, and the tokenizer's quote arm swallows the string whole. Filed as mmnto-ai/totem#2893 (the cure is the tokenizer's own port of core's grammar, the twin of what 2857 did for the scanner).
- **Disclosed FALSE FIRES (the deny direction, each asserted by a row):** an array assignment, a `case` pattern and a function body (pre-existing); PowerShell's backtick escape inside a double-quoted string (this walk reads POSIX quoting for both tools); an INVALID OPTION to a table word (`command -x`, `timeout -Z 30`, … — the program runs nothing, the wrapper judges the merge behind it; ruled disclose-not-cure, because a closed flag list per program turns every real flag it omits into a MISS, the wrong direction); a trailing backtick at end of input in PowerShell mode (a parse error there); `$VAR` inside an `env -S` string (env refuses the whole command). `TOTEM_MERGE_GATE_OVERRIDE=1` is the audited way past any of them.
- **Rows:** the twelve pilot-round misses flipped to projecting; every wrapper form has a mutant row; every positive loop asserts exactly `{ repo, pr: 5, headSha }`; the false-deny controls (single-quoted strings, heredoc bodies, comments carrying a separator or a backtick) still never spawn; the strip-table cells, the executable table, the clamp's edges, the redirection strip and the ps continuation are exercised through the export seam with exact argv.

### The budget (2856 item 2)

The deadline is set at the entry — right after the baked-arg parse, before stdin is read and before any projection — so the projection's git reads are bounded by what is left of it (10 s ceiling, 250 ms floor, no spawn once it is spent) instead of running up to three 10-second reads ahead of a deadline that did not yet exist, where a hung git reached the HOST's hook timeout and a killed hook's exit code never applies. A spent budget exits 2 at every tier with a named line and no CLI spawn; the stdin read is inside the same budget (a timer, cleared when the envelope arrives, fails closed with its own line if it does not); the hook's exit-code header names both arms. Rows: a copy of the node binary named `git` on a temp PATH with a `NODE_OPTIONS` preload that sleeps only when its own argv0 is `git` — with `--budget-ms 1500` the wrapper exits 2 in about 1.5 s (on the pre-budget template the same row ran 16.6 s, spawned the CLI and exited 0 — re-run by the seat in a detached worktree at `7841a154`); an envelope written without closing stdin exits 2 in about 1 s; `--budget-ms=1500` behaves as the separate-token form.

### The scanner (2857)

The template's blanker was a hand copy of core's `findHeredocs` that had diverged in the fail-open direction: no paren-boundary arms, a narrower bare-delimiter class, and a double-quote backslash that escaped any next character. It is now a verbatim port of core's walk and its three tables (`findHeredocSpans`, returning core's spans minus the unused `body`), the blanker replaces a body with spaces so every offset is preserved, both sites carry a `sync-anchor` comment, and a parity test holds the two span-for-span equal over the suite's whole command corpus and a seeded 3 000-string fuzz corpus in both bash and PowerShell modes (the first falsification leg added 40 072 comparisons of its own over a richer alphabet and a second seed: 0 divergences). The four paren and delimiter rows now project; on the pre-port template all four projected nothing. The rendered `HEREDOC_AT` in the committed hook is byte-identical to core's. Core is unchanged apart from the one anchor comment. One deliberate difference from core, recorded in the spec: the wrapper's blanker also blanks comment regions because the wrapper's tokenizer, unlike core's, does not read comments itself; two rows discriminate that arm.

### Also in the diff

An export seam: the hook runs its entry only as the main module and otherwise exports `blankHeredocBodies`, `findHeredocSpans`, `ghPrMergeArgvs`, `isGhExecutable`, `projectMergeReady` and `clampBudgetMs`. The template's disclosed-miss comment and disclosed-false-fires paragraph are read from the suite (every entry has a row). The committed `.claude/hooks/gate-wrapper.cjs` is re-scaffolded and equals the rendered constant — verified by script at the head, since no test reads the committed hook (that gap is mmnto-ai/totem#2833). A `minor` changeset for `@mmnto/cli` (the new hook argument); the fixed group takes every `@mmnto/*` package with it. No docs change: `docs/wiki/cli-reference.md` documents `gate install`'s matchers only, not the wrapper's arguments.

### Verified

Build leg (Opus, against the two specs, tests first): 9 commits, each row RED before its implementation; six folds (7 + 5 + 2 + 6 + 1 + 1 commits), each new row RED by a pre-fold copy of the rendered hook or a scratch mutant, named per row in the legs' reports. The seat's own run of the real gate at the head `06299840`: `pnpm build` and `pnpm test` exit 0 (cli 182 files / 5022 tests, core 159 / 4630, mcp 12, packs 5+5+2), `format:check` clean, ESLint 0 errors (8 pre-existing warnings in untouched files), `totem lint --branch` PASS (0 errors; the advisory warnings are prose matches in the spec and the suite), the hook-equals-constant and `HEREDOC_AT`-identity checks by script.

### Review of record: four falsification legs, thirty-two findings, every one folded

Each leg executed its claims against bash 5.3, GNU coreutils 8.32, pwsh 7, the engine, or a scratch mutant of the rendered hook; the deposits ride the branch under `.totem/artifacts/legs/`.

- **Leg 1 at `e152e37b`** (0 BLOCKING, 5 MATERIAL, 8 MINOR): two new false denies from the wrapper strip (sudo's describe-only flags; bash's `time` reserved word modelled as GNU time), an undisclosed miss family inside the closed table (long options with separate operands), a locked row whose comment inverted what bash does with substitutions inside double quotes, a trailing redirection riding into argv as a fabricated branch, the stdin read outside the budget, the `=` spelling ignored, three spec citations of a test that does not exist, and four test-strength gaps. Fold 1.
- **Leg 2 at `c52ae094`** (0 BLOCKING, 3 MATERIAL, 4 MINOR): fold 1's whole-segment redirection strip was quote-blind (a quoted `-b "<br>"` was dropped and the flag swallowed the PR — a regression against main), a residue sentence understated the `&`/`|` operators, the PowerShell line-continuation bullet was in the wrong list and unlocked, an invalid option to any table word still projected, the exit-code header omitted the budget arms. Fold 2 — and one of its choices ruled back into fold 3 (`env -S` re-entered rather than consumed).
- **Leg 3 at `47fab6ce`** (1 BLOCKING, 1 MATERIAL, 5 MINOR): fold 2's literal rule as the seat had ruled it ("any part of the token came from quotes") was not the shell's, so `>"out.txt" gh pr merge 5` became an unjudged merge under strict and a trailing quoted file fabricated a branch again; PowerShell's backtick-newline is a join only at a word boundary; env residue clauses without rows; the changeset's five-operator clause; spec markers. Fold 3 itself sound on every probe. Fold 4.
- **Leg 4 at `9b42de08`** (0 BLOCKING, 1 MATERIAL, 4 MINOR): every fold-4 arm true by execution and RED-derivable by mutant; a 10 000-input fuzz over the projection with no throw (worst 0.301 ms); an 8 000-input scanner parity with no divergence; a 3 768-case quoting sweep against real bash isolating exactly one fail-open family left on the branch's redirection strip — a fused redirection whose quoted filename carries whitespace, which the filename class `[^\s]+` could not match — plus the `-S=` route, a PowerShell backtick-whitespace-newline false fire, clustered short options, and the contrived quoting shapes. Fold 5: one regex character, one `=` guard, the disclosures with rows; verified by re-executing the leg's own corpus — and the build leg, partitioning the corpus by provenance (regenerated from its generator, byte-identical), found the fourth leg's classifier had labelled every miss "whitespace" by construction: 54 were (54 → 0 after fold 5, no over-strip), and 106 were a SECOND fail-open family the leg had not named — a quote or escape inside the operator prefix (`>">"out.txt gh pr merge 5`: bash's operator is the bare `>`, the wrapper's greedy `>>` reached past the first literal character and the token survived the strip). Fold 6: the operator prefix is the longest one that is itself an operator and ends at or before the first literal index (`>>` → `>`, `<<<` → `<<` → `<`, `2<>` → `2<` — bash extends an operator token over unquoted characters only; the seat's contract had carved `<<<` and `2<>` out as all-or-nothing, and the corpus showed that carve-out would have left 72 bypasses, so the bold rule shipped and the carve-out is withdrawn in the spec). After fold 6 the corpus reads: both fail-open families 0 (106 → 0, 54 → 0), OVERSTRIP 0; 405 of 3 768 cases still diverge and every one is a text divergence or a bogus deny — 146 false fires where bash runs nothing (an empty or missing redirection file), 217 extra-text divergences on a merge bash did run, 42 where the wrapper names a target the engine denies — none a fail-open, each kind named in the hook's residue sentence and locked as rows.

**Convergence, named by the seat on fold 6:** the sixth fold is one rule the shell corpus measures end to end, so no fifth leg was armed; the corpus (3 768 quoting positions against real bash) and the scanner parity fuzz are the two mechanical verifiers, and every remaining bucket in the corpus is a text divergence in the deny direction, disclosed and locked. Residue left disclosed and locked, all of it in the safe direction or a bogus deny, never a silent wrong-PR judgment: the `|`/`&`-carrying redirection operators; substitutions inside double quotes (mmnto-ai/totem#2893); env's own escapes, nested quotes and clustered short options inside an `-S` string; a variable executable; an unquoted win32 path; a wrapper program not on the table or spelled by path; the false fires listed above. The strict tier's remaining precondition on this repository is mmnto-ai/totem#2833 (nothing in CI pins the committed hook against the template; verified by script here).

Closes mmnto-ai/totem#2856

Closes mmnto-ai/totem#2857


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round disposition — the 2026-09-19 bot pass at bcdb766a, folded at b3268336

@coderabbitai
@greptileai

Head confirmation first, by the four reads the merged review-reply section prescribes: the head read bcdb766a, all three review objects carried that sha, and both summary comments named it (Greptile's `Last reviewed commit`, CodeRabbit's `coveredCommitId`). Two findings, both real, both fixed in b3268336; GCA reviewed and reported nothing, so it is not addressed here.

**1. CodeRabbit minor, `packages/cli/src/commands/init-templates.ts` (thread 4055075417) — fixed.** Correct as stated: `isGhExecutable` lower-cased only the suffix, so `GH.EXE pr merge 5` and `Gh.exe pr merge 5` — the same file on win32, which resolves names without case — ran unjudged, a bypass of the very class the § A widening claimed to close. The whole `gh.exe` basename now compares without case (`base.length === 6 && base.toLowerCase() === 'gh.exe'`), the bare `gh` stays exact, and the committed `.claude/hooks/gate-wrapper.cjs` is re-scaffolded from the rebuilt constant (byte-equal, 85 184 bytes). Rows: `GH.EXE` and `Gh.exe` in the § A spelling loop and in the basename unit rows (`C:\tools\GH.EXE` too); the bare `GH` remains locked as a non-match, a disclosed miss on win32 recorded in the spec's § A sentence.

**2. Greptile P1, `packages/cli/src/commands/init-templates.ts` (thread 4055076150) — fixed.** Correct as stated: the sudo grammar carried `--chroot` but not `-R`, so the generic path dropped `-R` alone and left `/chroot` standing at command position, where the anchor never saw `gh`. Read against sudo(8): `-R directory, --chroot=directory` is real, and two more separate-operand short options were missing by the same route, `-a type, --auth-type=type` and `-c class, --login-class=class` (BSD-only, but the table's rule is the program's manual). All three join the operand list with their long spellings. Rows: `sudo -R /chroot`, `sudo --chroot /chroot`, `sudo -a bsdauth`, `sudo -c staff` in the § B strip loop, and `sudo -R /chroot` plus `sudo -a bsdauth -c staff` as strip-table cells; the spec's § B row and the changeset carry the same words.

RED on the pre-fold constant and hook (stashed back to bcdb766a, the extended suite run against them): four failures — the § A spelling loop, the § B strip loop, the `sudo -R /chroot` cell (`[]` where `[['5']]` is owed), and the `GH.EXE` unit row. Gate at b3268336: workspace build and test exit 0 (cli 182 files, 5022 passed, 20 skipped; core 159 files, 4630 passed), `format:check` clean, ESLint 0 errors, `totem lint --branch` PASS with 0 errors, hook byte-equal to the rendered constant. Not re-legged: a case-insensitive compare of the same rule and three entries added to a closed table by the program's manual change no clause of the ruled design, so convergence stays as named on fold 6.

disposition: 4055075417 fixed
disposition: 4055076150 fixed

local-lane: none leg: 9b42de08 blocking=0 material=1 folded=5

Issue #2890, opened 2026-09-19
review-reply skill constants: carry the post-trigger handling doctrine now states — confirm by review object OR the bot's summary comment for that sha; a chat reply or silence is not a pass (standalone re-trigger on the word is); wait the ~12 min acknowledgement window or merge and record the late ack on the thread


satur8d:
## The gap

The distributed `review-reply` skill (rendered from CLI constants into `.claude/skills/review-reply/SKILL.md` and its `.agents` twin on every cohort repo) states the trigger rule at its line ~75: triggers are standalone comments, one trigger and no prose. It carries nothing about what a seat does AFTER a trigger. strategy's doctrine now does, at `doctrine/bot-protocols.md` § Invocation (merged 2026-09-19 as PR mmnto-ai/totem-strategy#1355, four falsification legs), and the two surfaces should agree:

- confirm a review against the review object for THAT sha, or the bot's summary comment for it (Greptile's verdict arrived as a summary comment only on mmnto-ai/totem-strategy#1180 and mmnto-ai/totem-strategy#1182), never the head's green;
- a chat reply or silence with no review to confirm is not a pass under the cadence (one external pass per chosen bot), so a standalone re-trigger on the operator's word is the pass, not a re-invoke;
- before merging on the other reviewers, either wait one acknowledgement window (~12 min, the mmnto-ai/totem-strategy#1182 bound) or merge and record the late acknowledgement as a line on the PR thread naming the bot and the time.

## Why now

A co-batched multi-trigger comment on mmnto-ai/totem-status#260 (2026-09-19) drew reviews from GCA and Greptile and a chat reply from CodeRabbit, no review; the seat's disposition read the green CodeRabbit status as a review before catching it. The doctrine line is the corrected rule; the skill is the surface a seat actually runs a round from, and it is rendered from constants this repo never hand-edits (the 2026-09-19 narrow leg on 1355 flagged it as a propagation pointer, not an edit).

## Ask

Fold the three clauses above into the review-reply constants at the next cut, so the rendered SKILL.md on every cohort repo carries them. Doctrine of record: mmnto-ai/totem-strategy `doctrine/bot-protocols.md` § Invocation and § Working the findings at `16c5694`.

satur8d:
## Closed: shipped in mmnto-ai/totem#2892, cut in 2.8.0

[...]

PR #2892, opened 2026-09-19
fix(skills): the distributed review-reply text carries the post-trigger handling — confirm each invoked bot's review against the review object or the summary comment's named sha, never a green status; silence is not a pass, the pass is a standalone re-trigger; wait one acknowledgement window or record the late acknowledgement (mmnto-ai/totem#2890)

satur8d:
## What

The rendered `review-reply` skill (from `REVIEW_REPLY_SKILL_CONTENT`, distributed by `totem init` to every consumer as `.claude/skills/review-reply/SKILL.md` and its `.agents` twin) stated the trigger rule (standalone comments, one trigger, no prose) and nothing about what follows a trigger. The doctrine of record now does (bot-protocols § Invocation and § Working the findings, merged 2026-09-19), and the skill is the surface a seat actually runs a round from, so the two must agree. One new section, "Before Phase 1: confirm each invoked bot's review is on the head", carries the three clauses (part of mmnto-ai/totem#2890):

1. Confirm every invoked bot's review against the review object for THIS head sha, or the bot's summary comment matched to the head by the sha its own body names (Greptile's `Last reviewed commit` link; CodeRabbit's `final_review_risk_coverage` marker), never by the comment's timestamp (an in-place re-review does not advance it) and never a green commit status on the head: CodeRabbit's status settles green on every push head whether or not it reviewed that head (`Review completed` on the sha it reviewed, `Review skipped` on every push head it did not), a PENDING status means still reviewing; Greptile's green check run does mark the sha it reviewed; GCA posted neither a status nor a check run on any GCA-reviewed sha measured so far. One fenced block prints both halves of every comparison: the head sha, every review with the sha it was submitted against, and the sha each summary comment names.
2. A chat reply or silence with no review to confirm is not a pass under the one-external-pass-per-bot cadence; the pass is a standalone re-trigger posted by the operator — that bot's first pass, not a re-invoke, which the cadence reserves for risky rework with the reason recorded in the round comment.
3. Before merging on the other reviewers, wait one acknowledgement window (about 12 minutes from the trigger) or merge and record the late acknowledgement as one line on the PR thread naming the bot and the time.

## How

[...]

PR #2895, opened 2026-09-19
Version Packages

satur8d:

# Releases
### Minor Changes

- 61f009c: fix(gates): the merge-ready gate wrapper now judges the projection shapes that bypassed it, bounds its whole run by one budget, and carries core's heredoc scanner verbatim (mmnto-ai/totem#2856, mmnto-ai/totem#2857). Under the pilot tier each of these was one lost advisory read; under the strict tier each was a bypass, so both issues were the strict install's precondition. WHAT IS JUDGED NOW: the executable may be spelled `gh`, `gh.exe`, or either behind a path (the basename after the last `/` or `\`; the whole `gh.exe` basename compares without case, since win32 resolves `GH.EXE` and `Gh.exe` to the same file, while the bare `gh` stays exact); a CLOSED table of transparent wrapper programs is stripped with its option grammar before the anchor is read (`sudo`, `env`, `timeout`, `nice`, `nohup`, `command`, `exec`, `time` — which options take a separate operand, how many positionals the word itself eats, whether `--` ends its options), with the assignment strip re-running after each so `env -u X A=1 gh pr merge 5` and `sudo -u root timeout 30 gh pr merge 5` both resolve; `eval` re-tokenizes its operand once (depth 1); a leading redirection is skipped with its file token; and a backtick becomes a segment separator, so the operand of a backtick substitution is a segment of its own (the backtick rides on as a token, so a merge whose TARGET is a substitution still reports `unresolvedTarget` rather than falling back to the current branch). In POWERSHELL mode a trailing backtick is instead that shell's LINE CONTINUATION — at a WORD BOUNDARY the backtick and the newline after it are consumed and the next line continues the command, the twin of bash's trailing backslash there — where before it separated in both modes, which made the backtick itself the merge's target (`unresolvedTarget`: strict denied a target nobody wrote, pilot warned, and the real PR merged unjudged in the next segment). INSIDE a word the two shells part company, so the arm stops at the boundary: bash's backslash-newline really joins its halves, while PowerShell's backtick is that shell's ESCAPE character and `gh pr merg` + backtick + newline + `e 5` passes the single argument `merg<LF>e`, which is not `merge` and which pwsh does not run. The tokenizer embeds that newline in the token and projects nothing, where joining it projected a merge the shell never performs. THE STRIP TABLE, CORRECTED BY THE FALSIFICATION ROUND: `sudo`'s describe-only options (`-l`/`--list`, `-v`/`--validate`, `-V`/`--version`, `-K`/`--remove-timestamp`) run nothing, so a merge behind one is no longer projected — that was a FALSE DENY; every long option that takes a SEPARATE operand joins its word's operand list (`sudo --user|--group|--prompt|--chdir|--chroot|--host|--role|--type|--other-user|--command-timeout|--auth-type|--login-class`, `env --unset|--chdir`, `timeout --kill-after|--signal`, `nice --adjustment`), without which the operand itself read as the program and the merge behind it went unjudged — and so do sudo's three short spellings the table had omitted, `-R directory`, `-a type` and `-c class`, where `sudo -R /chroot gh pr merge 5` dropped the `-R` alone and left `/chroot` standing at command position; and `time` is read as bash's RESERVED WORD (`time [-p] [--] pipeline`), where any `-` token that is not `-p` means the shell runs nothing — GNU `/usr/bin/time`'s `-o`/`-f` grammar belonged to a different program. And `env -S` / `--split-string` no longer CONSUMES its operand: env splits that string into words and PREPENDS them to the arguments that follow, so the words take the option's place and the strip reads on from them — `env -S 'gh pr merge 5'`, `env -S "gh pr merge" 5`, `env -S gh pr merge 5`, `env --split-string gh pr merge 5`, `env --split-string='gh pr merge 5'`, `env -u X -S 'gh pr merge 5'` and `env -S 'A=1 gh pr merge 5'` each RUN `gh pr merge 5` (measured, coreutils 8.32), and each projected NOTHING while the operand rode off with the option, which is a bypass under the strict tier. The ATTACHED SHORT spellings are read the same way: a short option carries its operand with no separator, and the quote arm joins `-S'…'` into that same token, so `env -Sgh pr merge 5` and `env -S'gh pr merge 5'` — both of which coreutils runs — are judged now instead of dropped as a flag of env. An `=` is NOT a separator for a short option, so the `=` split belongs to LONG options only — and where an `=` follows a short option's letter the whole token is dropped as a flag of that word, which is the route that reads both spellings right. `env -S=X gh pr merge 5` is env splitting the operand `=X` into one word, an assignment with an EMPTY NAME, so the command is `gh pr merge 5` and the merge RUNS (measured, coreutils 8.32): reading `=X` as an attached operand had left it standing at the front of the strip, where it is neither an assignment this walk accepts nor a command, and the merge behind it went unjudged — a bypass under the strict tier. `env -S='gh pr merge 5'` is the same rule the other way: env's words are `=gh`, `pr`, `merge`, `5`, the command is coreutils `pr` (`pr: merge: No such file or directory`), no merge runs and none is projected. THE REDIRECTIONS: one strip over the WHOLE segment now runs ahead of the wrapper strip and the anchor test, so a trailing redirection never rides into argv (`gh pr merge --squash > merge.log` reached the engine as a pull request on branch `>`, a deny with a reason no one wrote) and a redirection between the executable and its verb no longer breaks the anchor; an operator alone takes the file token with it, a fused one (`>out.txt`, `2>/dev/null`, `<<<bar`, `2<>file`) goes alone, and a `<<EOF` head drops harmlessly because the scanner blanked its body first. That strip reads a token's TEXT, so the tokenizer now records, per token, the INDEX of the first character that came from inside quotes or from a backslash escape (`-1` when the whole word is bare), and the strip drops a token only when an operator prefix lies ENTIRELY BEFORE that index — the LONGEST prefix that does, which is not always the longest the two patterns match, because bash extends an operator token over UNQUOTED characters only: `>">"out.txt gh pr merge 5` is the operator `>` with the filename `>out.txt`, a real redirection with a real merge behind it, where the greedy `>>` ended past the index, the word was kept, stood in front of `gh` and broke the anchor — a bypass under the strict tier, and the same shape in `<<"<"bar` and `2<">"out.txt`, all three cured by taking the bounded prefix (`>>` → `>`, `<<<` → `<<` → `<`, `2<>` → `2<`). That is the shell's own rule: what makes a redirection real is the quoting of the OPERATOR, never that of the word it sits in. `>"out.txt" gh pr merge 5` truncates the file and merges PR 5, while `gh pr merge --squash ">"out.txt` hands gh the ARGUMENT `>out.txt` and redirects nothing. The FILENAME may hold anything, whitespace included: a quoted name with a space in it is ONE token, and while the fused pattern's filename class excluded whitespace that token matched neither pattern — so `>"out file.txt" gh pr merge 5` stood in front of `gh`, broke the anchor and went unjudged though bash truncates the file and merges PR 5, and trailing a merge the same word rode into argv, where `gh pr merge --squash >"merge log.txt"` reached the engine as a pull request on a branch named `>merge log.txt`. The class is `[\s\S]+` now; what keeps a QUOTED operator out is the first-literal index, not the filename's shape. Locked beside it as residue, re-measured over a 3 768-case quoting corpus run against real bash, where both fail-open families read 0 and every divergence left is one of three kinds: a quote pair that opens at INDEX 0 of the word (`"">out.txt`, `">">out.txt`, `"2">file`) leaves no operator prefix before it, so the word is data here while bash passes the empty string, `>` or `2` as an ARGUMENT and applies the redirection that follows it; a `$VAR` in a kept filename is not expanded, so this walk names `>$FILE` where bash wrote `>varfile.txt`; and where the quote sits INSIDE the operator prefix the word strips although bash sometimes runs nothing behind it — an empty filename (`>"">out.txt`) or a file that does not exist (`<"<"<out.txt`) fails the redirection, so the walk judges a merge the shell never ran. The first two are confined to the argv's TEXT and every word this walk keeps is a target the engine denies; the third costs one bogus deny. All three are the deny direction, no bypass. Without any such rule, `gh pr merge -b "<br>" 5` had been losing its body to the strip and then PR 5 to the `-b` that swallowed the next token, and `gh pr merge -t ">>" --repo owner/name 5` had been losing its repository, because a `>>` ALONE takes the token after it and that token was `--repo`; with the coarser "any part of the token came from quotes" spelling of it, the quoted-filename forms rode into argv instead and `gh pr merge --squash >"merge.log"` reached the engine as a pull request on a branch named `>merge.log`. Unquoted spellings keep stripping. THE BUDGET: the wrapper's deadline is now set at its entry, before stdin is read and before any projection, so the projection's git reads are bounded by what is left of it (10 s ceiling, 250 ms floor, no spawn at all once it is spent) instead of running up to three 10-second reads ahead of a deadline that did not yet exist — where a hung git reached the HOST's hook timeout and a killed hook's exit code is never applied, a fail-open on a fail-closed gate. A spent budget writes `[totem gate-wrapper] the <n> ms budget was spent before gate "<event>" could be evaluated …` and exits 2 at every tier, without spawning the CLI. The READ of the envelope is inside the budget too: a timer armed at the entry writes `[totem gate-wrapper] the <n> ms budget was spent before the envelope arrived on stdin — blocking (fail-closed).` and exits 2 if stdin never closes, where before the hook waited unbounded until the host killed it — a killed hook's exit code is never applied. MINOR: a new hook argument `--budget-ms <n>`, in either spelling (`--budget-ms 1500` or `--budget-ms=1500`, last-wins on a repeat), which the installed settings.json line never passes and which can only LOWER the budget (a value below 1000 reads as 1000; above 30000 or unparseable reads as the 30000 default). THE SCANNER: the wrapper's `blankHeredocBodies` was a hand copy of core's `findHeredocs` that had diverged in the fail-open direction — no paren-boundary arms (so a `#` after `(` or after an operator `)` was text and a `<<word` inside it opened a body), a narrower bare-delimiter class (so `<<E:F` read as the prefix `E` and the terminator line never matched), and a double-quote backslash that escaped any next character where core escapes only `$`, backtick, `"`, `\` and newline. Each of those blanked a `gh pr merge` on a later line. It is now a verbatim port of core's walk and its three tables, returning core's spans minus the unused `body`, with the blanker replacing a body with spaces so every offset is preserved; both sites carry a `sync-anchor` comment and an executable parity test holds the two span-for-span equal over the suite's whole command corpus and a seeded 3 000-string fuzz corpus in both bash and PowerShell modes. Core is unchanged apart from that one comment line. The committed `.claude/hooks/gate-wrapper.cjs` is re-scaffolded from the new template. Disclosed and locked as rows, unchanged in the safe direction: a variable executable (`$GH`), an unquoted win32 path (this walk reads POSIX quoting for both tools), `timeout` with no duration, a program not on the closed table (`npx`, `xargs`, `bash -c`), a table word spelled by PATH (`/usr/bin/time -f x gh pr merge 5` — a program with its own option grammar, not the reserved word), env's OWN splitting rules inside a `-S` / `--split-string` operand (the operand is read now, but split on WHITESPACE and nothing else: env's escapes — `\_` is a SPACE — its `$VAR` expansion inside the string and its `#` comment are not modelled, and quotes inside the string are not stripped, so `env -S 'gh\_pr\_merge\_5'`, which really runs `gh pr merge 5`, reads as one word here, and `env -S 'env -S "gh pr merge 5"'` runs the merge too because env strips the quotes inside its own operand while this walk keeps them and reads `"gh` as the executable), CLUSTERED short options on a table word (`env -vu X gh pr merge 5`, `env -iS '<cmd>'`: the option test reads the WHOLE `-` token, so a cluster matches no entry of that word's operand list, is dropped as one flag, and the operand belonging to the cluster's last letter is left standing at the front of the strip, where it blocks the anchor — while coreutils runs the merge in both, measured on 8.32; ATTACHMENT is not the gap, since `env -uX`, `nice -n10`, `timeout -k5 30` and `timeout -sTERM 30` all project and all run), `eval` nested deeper than one level, a SUBSTITUTION INSIDE DOUBLE QUOTES (``echo "`gh pr merge 5`"``, `echo "$(gh pr merge 5)"` — bash executes both, the tokenizer's quote arm swallows the string whole, so the merge runs unjudged: a fail-open filed as mmnto-ai/totem#2893), the redirection operators whose own separator ends the segment first (`>|`, `2>&1`, `>& file`, `<& 3`, `exec 3>&1 …` — `|` and `&` are the tokenizer's own segment separators and end the token before the operator is whole; ALL FIVE of those are merges the SHELL RUNS, measured with a stub on bash 5.3, each applying its redirection and then running `gh pr merge 5`, the `<& 3` form once that descriptor is open, and the segment each leaves starts at the FILE — `out.txt`, `1`, `file`, `3` — so nothing of the merge is read: a fail-open miss, not text the shell ignores; `&>` splits the same way but leaves a readable `>` at the front of the next segment, so that one IS read), and PowerShell's OWN QUOTING — its backtick escapes outside double quotes, its here-strings — which is not modelled at all, because the walk reads POSIX quoting for both tools (PowerShell's call operator is not among these: `& gh pr merge 5` projects, since `&` is one of the separators and the segment after it starts at `gh`). Disclosed FALSE FIRE, the deny direction on contrived text: in PowerShell mode a double-quoted string whose backtick escapes a quote (``Write-Output "a `"; gh pr merge 5`"b"``) is one string to PowerShell and merges nothing, but this walk reads POSIX quoting for both tools and judges the merge inside it; and a backtick that is the LAST CHARACTER of the input (`gh pr merge 5` + backtick, with nothing after it, not even a newline) is a continuation with nothing to continue, so pwsh answers with a parse error and runs nothing while the backtick — not being followed by a newline — falls through to the segment-separator arm and the merge in front of it is judged (give that same input a trailing newline and pwsh runs the merge, which the continuation arm projects: only the last-character spelling diverges). A backtick followed by WHITESPACE and then a newline is the third of PowerShell's own: the backtick escapes that space, pwsh runs `gh pr merge` with NO target — the current branch's PR merges — and reads the next line as its own statement, while the arm here wants the newline immediately, so the separator arm takes the backtick and the walk names it as an `unresolvedTarget` the strict tier denies. A third, env's: a `$VAR` inside a `-S` operand (`env -S 'gh pr merge $PR'`) makes env refuse the whole command (`only ${VARNAME} expansion is supported`) and run nothing, while the split reaches the anchor and `$PR` rides as an `unresolvedTarget` the strict tier denies. And one that is not contrived at all: an INVALID OPTION to a word on the strip table (`command -x`, `exec -x`, `timeout -Z 30`, `nice -Z`, `env -Z`, `nohup -x`, `sudo -Z`) makes the program answer "invalid option" and run NOTHING, while the strip reads the unknown `-` token as one of that word's own options and judges the merge behind it. Disclosed rather than cured: the cure is a closed flag list per program — the shape `time` carries, whose reserved-word grammar really is two flags — and measured on a mutant with that list everywhere it turns every real flag the list omits (`sudo -n`, `sudo -E`, `timeout --foreground`) into a MISS. A miss is a bypass under the strict tier; a false fire on a command that runs nothing costs one bogus deny, which `TOTEM_MERGE_GATE_OVERRIDE=1` clears.

### Patch Changes

- fa21640: review-reply skill: the distributed text now says what a seat does AFTER a review trigger, in a section before Phase 1 — confirm each invoked bot's review against the review object for the head sha or the bot's summary comment matched by the sha its body names (one fenced block of paginated `gh` reads prints the head sha, every review with the sha it was submitted against, the sha each Greptile summary names as its `Last reviewed commit` and the sha each CodeRabbit summary names as covered), never a green commit status on the head (CodeRabbit's status settles green either way; Greptile's check run does mark its sha; GCA has posted neither on any GCA-reviewed sha measured so far); a chat reply or silence with no review to confirm is not a pass, so the pass is a standalone re-trigger posted by the operator, that bot's first pass rather than a re-invoke; before merging on the other reviewers, wait one acknowledgement window (about 12 minutes from the trigger) or merge and record the late acknowledgement as a line on the PR thread (mmnto-ai/totem#2890).
  - @mmnto/totem@2.8.0
### Patch Changes

- @mmnto/totem@2.8.0
## @mmnto/totem@2.8.0

## @mmnto/pack-rust-architecture@2.8.0

## @mmnto/pack-agent-security@2.8.0


satur8d:
## Cut-sensor fill for the 2.8.0 cut (mmnto-ai/totem#2325) — both entries verified, contract-bearing, the minor is the right size

**Entry `61f009c` (init-distributed content, `packages/cli/src/commands/init-templates.ts`) — verified.** The merge-ready gate wrapper template `CLAUDE_GATE_WRAPPER`, which `totem gate install merge-ready` scaffolds into a consumer's `.claude/hooks/gate-wrapper.cjs` (mmnto-ai/totem#2894, closing mmnto-ai/totem#2856 and mmnto-ai/totem#2857). Minor is right: a new hook argument `--budget-ms` (lower-only, never passed by the installed settings line), and the projection widens in the deny direction only (wrapper programs, executable spellings, `eval`, `env -S`, redirections, backtick substitution, the scanner port); the exit contract and the settings line are unchanged. A consumer on 2.7.x keeps working; its committed hook stays on the old projection until it re-runs `totem gate install`, which is the step that makes the strict tier safe there — mmnto-ai/totem-status#260 named both issues as strict preconditions.

**Entry `fa21640` (init-distributed content, same file) — verified.** The review-reply skill's "Before Phase 1" section in `REVIEW_REPLY_SKILL_CONTENT`, which `totem init` renders into a consumer's `.claude/skills/review-reply/SKILL.md` and its `.agents` twin (mmnto-ai/totem#2892). One added section, no removed export, no changed verb or flag; a consumer gains it on its next re-init, and its drift parity row names the stale copy until then.

### Choreography

- **Which consumers move:** every repo that renders the skill and installs the gate — mmnto-ai/totem-strategy (its strict flip mmnto-ai/totem-strategy#1358 is open; the re-scaffold from 2.8.0 belongs before or with it, since the widened projection is what closes the bypass classes the strict tier would otherwise inherit), mmnto-ai/totem-status, mmnto-ai/liquid-city (still on ^2.6.0; it picks both up with its pending pin bump). Each moves by bumping its `@mmnto/cli` pin, re-running `totem init` (re-stamps both skill copies) and `totem gate install merge-ready` at its current tier (re-scaffolds the hook).
- **In what order:** this cut first; then each consumer on its own seat's cadence, with no ordering between consumers. On strategy, the hook re-scaffold precedes or accompanies the strict flip.
- **Who owns which half:** totem-claude cuts (this PR, on the operator's word) and mails the cut to the three consumer seats; strategy-claude, status-claude and lc-claude each own their pin bump, re-init, hook re-scaffold and receipt.

Issue #2899, opened 2026-09-20
init: totem init re-stamps the .claude/skills copies on a consumer but leaves their .agents/skills twins untouched — two liquid-city syncs hand-copied the twins forward; rule whether init carries both (recommended) or the twin is consumer-owned


satur8d:
### Symptom

On mmnto-ai/liquid-city, `totem init` at @mmnto/cli 2.8.0 (the ^2.8.0 pin sync, PR mmnto-ai/liquid-city#1294) re-stamped the two `.claude/skills/` copies (review-reply's "Before Phase 1" section, the signoff Visiting-case sentence) and reported only those two merges in its summary; the `.agents/skills/` twins were left on the previous text. The consumer's build leg caught the drift by `cmp` and byte-copied the twins forward by hand. The same hand step was needed at that repository's ^2.6.0 sync (PR mmnto-ai/liquid-city#1286). Reported by lc-claude in its 2.8.0 receipt of 2026-09-20T01:40Z; the datum is routed here as this repository's row, per that dispatch.

### What the source of truth says today

In this repository the `.agents` twin is a byte-copy of the `.claude` copy, and `packages/cli/src/commands/init.test.ts` locks that equality (the constant-equals-file test and the `.agents` byte-copy test). Consumers carry no such lock: liquid-city's `tools/agents-contract.test.mjs` does not check twin equality, so nothing on the consumer side names the drift when init leaves one copy behind.

### The question to rule

Either init is meant to carry both copies, and leaving the `.agents` twin un-re-stamped is a defect of the scaffolder (a file that only repeats another file and is maintained by hand will drift, which this repository's own doctrine calls a bug), or the twin is consumer-owned by design, in which case the consumer-side equality guard is the consumer's to add and the 2.7.0 line "the .agents twin stays a byte-copy" is the answer. The recommendation from this seat is the first: init already owns the `.claude` copy, so it should write the twin in the same pass and report both merges in its summary, and a consumer-side drift row (doctor --parity) should name a twin that differs from its `.claude` sibling.

### Reproduce

On a consumer checkout with both `.claude/skills/review-reply/SKILL.md` and `.agents/skills/review-reply/SKILL.md` present at an older text, bump the pin and run `totem init`; compare the two files afterwards with `cmp`.

### Verification, when fixed

[...]

Issue #2788, opened 2026-09-05
init: write the four distributed skills to `.agents/skills/` as well as `.claude/skills/` from the same constants and markers — every cohort `.agents` copy is a hand byte-copy that drifts at each cli bump (mmnto-ai/totem#2532 slice 2; measured on the 09-03 bump)


satur8d:
## Observed (strategy-claude, cohort census 2026-09-05)

`totem init` writes the four distributed skills to `.claude/skills/` only (`packages/cli/src/commands/init.ts:575-577`: the loop over `DISTRIBUTED_CLAUDE_SKILLS` renders `.claude/skills/<name>/SKILL.md`). Every `.agents/skills/` copy in the cohort is a hand byte-copy: totem (mmnto-ai/totem#2559, 2026-08-03), totem-strategy (mmnto-ai/totem-strategy#1054, 2026-08-12), totem-status (mmnto-ai/totem-status#200, 2026-08-21); liquid-city has none (`.agents/` gitignored whole, mmnto-ai/liquid-city#846 open since 2026-07-21).

Consequence measured on 2026-09-03: the cli bump re-rendered strategy's `.claude` twins to covariate format v1.2 and left the `.agents` byte-copies at v1.1; `doctor --parity` reported WARN drift on 2 of 4 (re-synced by hand in mmnto-ai/totem-strategy#1238). Every bump will do this on every byte-copied repo until init writes both surfaces.

## Ask (slice 2 of mmnto-ai/totem#2532)

- `totem init` writes the same four skills to `.agents/skills/<name>/SKILL.md` from the same constants, with the same managed-block markers, and preserves each surface's own extension tail outside the markers (the after-marker seam already exists for `.claude`).
- The doctor's `agents-skills` detector already reads the surface; no detector work, only the writer.
- Sized as the write-both step of the existing skill distribution: the constants, markers and seam exist; the change is the second target path plus its idempotent re-render and tests (fresh init, re-init over a byte-copy, re-init over a surface carrying an extension tail).

## Done when

A `totem init --yes` on a cohort checkout leaves `.claude/skills/*` and `.agents/skills/*` byte-equal inside the markers and their tails untouched; `doctor --parity` reads both surfaces conforming after a bump with no hand re-sync; a test per shape.

satur8d:
## Closed as delivered by mmnto-ai/totem#2902 (in @mmnto/cli 2.9.0), on the operator's ruling of 2026-09-21

[...]

Issue #2532, opened 2026-07-31
charter: baseline-skill parity across all five vehicles — signon/signoff/review-reply/review-loop (consolidates the 07-23 ALL-FOUR ruling; absorbs the mmnto-ai/totem#1891 gemini slice)


satur8d:
## The need (operator-directed 2026-07-30: "that has to change")

The four baseline session skills — `signon` · `signoff` · `review-reply` · `review-loop` — are fully distributed to **Claude vehicles only** (`totem init` marker-block distribution, source-locked by the mmnto-ai/totem#1890 invariants). The ALL-FOUR cross-vendor parity ruling (2026-07-23) never became build charters: the gemini half (mmnto-ai/totem#1891) has been dormant since 2026-05-11, and the codex / agy / kimi halves have no tracker at all. This issue is that consolidated charter.

**The live cost, two exhibits from this week:** lc-codex executed lc-claude's signoff duties (GH project update + derived-surface refresh) **by imitation** — right behavior, no contract, pure divergence risk (operator-observed 2026-07-30; feeds the mmnto-ai/totem-strategy#992 census). And six operator-run vehicle seats hold zero processed marks across all live broadcasts partly because their session bring-up/close rituals aren't skill-encoded (mmnto-ai/totem-status#135's per-edge lag measurement).

**New dependency raising the priority:** the primary-seat ("crown") formalization being chartered strategy-side names a **codex/Sol crown trial** gated precisely on the codex vehicle carrying these skills. Skill parity is now the unblock for that trial, which argues for queueing this ahead of mmnto-ai/totem#2525 / mmnto-ai/totem#2106 — operator word on final order.

## Scope

1. **Per-vehicle adapters under the ruled base-plus-extensions contract** (seat steps outside markers; adapter extracts after ≥2 instances): codex, gemini (absorbs mmnto-ai/totem#1891's slice-4 scope), agy, kimi. Known vehicle facts to design against: kimi does not scan `.claude/skills` and requires `TOTEM_SELF_AGENT`; gemini's skill surface was the gate-5 blocker; agy mechanics were census-stage as of 07-23 — complete the census as step zero rather than guessing.
2. **The four baseline skills only** — no per-repo extensions in scope (those stay seat-owned via the after-end-marker seam).
3. **Parity sensing:** a `doctor --parity` row class per vehicle so honest-absent is visible, mirroring the existing Claude-skill rows.

## Not in scope

[...]

PR #2299, opened 2026-07-04
feat(cli): promote /signon to a distributed Claude Code skill (strategy#536)

satur8d:
## What

Promotes the session-start **`/signon`** skill from the strategy-local pilot to a distributed Claude Code skill, per the operator ruling recorded on [mmnto-ai/totem-strategy#536](https://github.com/mmnto-ai/totem-strategy/issues/536) (Proposal 295 d2; post-R2 gate fired, pilot n=2 trip-free). `totem init` now installs `.claude/skills/signon/SKILL.md` — the read-twin of `/signoff` — through the existing mmnto-ai/totem#1890 marker-based distribution mechanism.

## Changes

- **`SIGNON_SKILL_CONTENT`** added beside `SIGNOFF_SKILL_CONTENT` in `init-templates.ts` and registered in `DISTRIBUTED_CLAUDE_SKILLS`. New source-of-truth file `.claude/skills/signon/SKILL.md`, locked to the constant by the #1890 match-source invariant (`SIGNON_SKILL_CONTENT matches .claude/skills/signon/SKILL.md`).
- **Both recorded pilot amendments folded** into the canonical content before freeze:
  - **Roster (amendment 1):** the frozen cohort-shorthand → repo-slug map (`totem` / `strategy` / `status` / `lc` → `mmnto-ai/{totem, totem-strategy, totem-status, liquid-city}`) carried in step 3 for cross-repo gate re-derivation — viable while the roster is frozen (mmnto-ai/totem-strategy#611 gates changes).
  - **Content-date `--since` (amendment 2):** step 2's mail-poll cutoff derives from the newest journal's **content date** (filename stamp / frontmatter), **never file mtime** — mtime lies on fresh clones/worktrees and silently reports "inbox clean" over waiting mail (mmnto-ai/totem-strategy#813; settles that ticket's canonical half).
  - The strategy-local pilot preamble is stripped; the canonical steps move inside the managed marker block.
- **`eject`** now derives `DISTRIBUTED_CLAUDE_SKILL_NAMES` from the canonical `DISTRIBUTED_CLAUDE_SKILLS` array instead of a hand-maintained literal, so a newly distributed skill can never orphan on `totem eject` because the two lists drifted.

## Release

Changeset: `@mmnto/cli` **minor**. Rides the open 1.89.0 train (VP mmnto-ai/totem#2286) so one release carries the R-arc slices + the lc Tier-1 intake gate + cohort `/signon` parity.

[...]

PR #2902, opened 2026-09-20
fix(init,eject): totem init writes each distributed skill to its .agents/skills twin from an always-run step independent of tool detection wherever the init cwd carries an .agents/ directory, reporting each twin; eject removes the twins it wrote; managed-block equality with per-surface tails (mmnto-ai/totem#2899; the mmnto-ai/totem#2788 charter)

satur8d:
## What

`totem init` now writes each distributed skill to its `.agents/skills/<name>/SKILL.md` twin beside `.claude/skills/<name>/SKILL.md`, wherever the repository carries an `.agents/` directory, and the install summary lists both rows. Before this, init refreshed only the `.claude` copy and left a consumer's twin on the previous text: two liquid-city syncs found the drift by `cmp` and hand-copied the twin forward, and the datum reached this repository in lc-claude's 2.8.0 receipt of 2026-09-20T01:40Z (mmnto-ai/totem#2899, filed and ruled on the operator's word).

## How

The skill loop moves out of `installClaudeHooks` into the exported `distributeClaudeSkills(cwd, roots, opts)` with `SKILL_TWIN_ROOTS = ['.claude', '.agents']`, so the behaviour is testable against a temp directory. The Claude installer calls it for `.claude` only; `initCommand` runs its own always-run step for `.agents` after the tool installers, independent of which tools were detected, because that surface is the vendor-neutral one Gemini CLI, Antigravity, Kimi and Codex read, so a repository with no Claude surface and an `.agents/` directory gets its twins too. The twin is written only where the init cwd carries an `.agents/` directory (init writes every surface at its cwd, so it is run from the repository root): a repository without it gets one disclosed skipped row instead of silence, and an `.agents` that is not a directory gets one disclosed row instead of four write errors. Each written twin gets its own row with its own label, "Scaffolded" or "Refreshed vendor-neutral skill twin (.agents/skills)", through the renderer's disclosed-override path. The two copies are equal inside the markers and each keeps its own extension tail below the end marker, which is the contract the parity manifest states; byte equality is the zero-tail case. `totem eject` now walks both roots with the same marker test, prunes each skills root, and carries both paths in its dirty-tree roster, so init no longer writes a surface eject cannot see. The force-refresh warning and the action mapping are unchanged and apply to both roots. The docs page, `AGENTS.md`, the parity hint and the changeset say the same, and reference the charter for this change, mmnto-ai/totem#2788 (mmnto-ai/totem#2532 slice 2).

## Verification

Six tests in `init.test.ts` (the charter's three shapes and three more) and one in `eject.test.ts`: a stale marker-bearing twin (the liquid-city shape) is refreshed to the same managed block as the `.claude` copy and both rows are reported, the twin as merged with its own label and the fresh `.claude` copy as created; an extension tail on each surface survives a re-run while the managed blocks stay equal and the files are not byte-equal; a marker-less twin is preserved with an err row unless the refresh is forced, which overwrites it with the same warning label as the `.claude` copy; a cwd without `.agents/` gets every `.claude` copy, no twin, and exactly one skipped row; an `.agents` that is a file is one disclosed skipped row; a second run reports every twin as exists and the `.claude`-only root never emits an `.agents` row; and eject removes both managed copies while a marker-less twin stays. Each init test calls a function the previous module did not export, so each fails on the pre-fix module, and the leg re-implemented the old loop to show the first fails for the behavioural reason. At the fold head: `pnpm build` exit 0; the init, eject and doctor-parity test files green (the parity hint's three locked assertions kept their asserted prefix); the full workspace `pnpm test` exit 0 (cli 182 files, 5030 passed with the seven new; core 159 files, 4630 passed); `format:check` clean; ESLint 0 errors; `totem lint --branch` PASS with 0 errors. One order-dependent floor-scanner failure seen once at `206e38a4` passed alone on this tree and on main's and the file passed twice in full after. A minor changeset for `@mmnto/cli`.

Owed: `docs/wiki/**` is a legs-owed path and the placement of the writer is a judgment. One falsification leg over `206e38a4` (deposit `.totem/artifacts/legs/206e38a4…json`, read at 2026-09-20T02:40:01Z, 62 tool uses): 1 BLOCKING, 5 MATERIAL, 4 MINOR, every one folded in fold 1. The blocking one: eject scrubbed `.claude/skills/` only and left the four twins init now writes unnamed in its lists, the orphan class the derived scrub list exists to prevent, reproduced by an init-then-eject run. The material ones: the writer lived inside the Claude installer, so a `.gemini`-only repository carrying `.agents/` received no twins, reproduced on three headless init runs; "byte-equal twin" overclaimed against the managed-block-plus-tail contract the parity manifest states, reproduced with tails on each surface; the charter mmnto-ai/totem#2788 was unreferenced and three prose sites it falsifies stood; the charter's tail shape and the force path were untested; patch understated a change that adds two exports and a new file surface. The minor ones: the twin rows rendered as "Scaffolded Claude Code hook"; the presence test admitted an `.agents` file and four write errors behind "Init complete"; the skipped row asserted a repository-level fact from a cwd probe; an unrelated eslint directive had been dropped by an autofix. Folds: the writer moved to an always-run step in `initCommand` for `.agents`, the Claude installer writes `.claude` only; eject walks both roots with a new test; managed-block wording everywhere; the charter referenced and the three prose sites corrected with the parity hint's asserted prefix kept; tail, force and `.claude`-only-root tests; minor bump; twin labels; a directory test with one disclosed row; the row says "at the init cwd" rather than adding a git-root anchor the rest of init does not have; the directive restored. **The fold rewires the writer's placement, so a re-armed leg scoped to the fold is owed before this PR is presented for merge; it has not run yet.**

## Not in this PR

A git-root anchor for the twin probe (init writes every surface at its cwd; a twin anchored elsewhere than its `.claude` sibling would be a new inconsistency, so init is run from the repository root). The pre-existing "Scaffolded Claude Code hook" label on the `.claude` skill rows, which are skills, not hooks. The consumer-side parity row for a stale twin already exists (`agents-skills` in doctor-parity, shipped by mmnto-ai/totem#2532 slice 1 and wired on liquid-city), so the second half of the issue's recommendation needed no change.

## Related

Part of mmnto-ai/totem#2899; the consumer exhibits are mmnto-ai/liquid-city#1286 and mmnto-ai/liquid-city#1294; the datum arrived with the 2.8.0 cut receipts (mmnto-ai/totem#2895).


satur8d:
/gemini review
@coderabbitai review
@greptileai review

satur8d:
## Round disposition (totem-claude, 2026-09-20)

@gemini-code-assist
@coderabbitai
@greptileai

The round's three reviews sit on `3c6cdad1`; the fold that answers them is `9d75d3ef` (fold 3), on top of the re-armed leg's fold `004a5252` (fold 2), which had landed after the trigger and already answered one finding. Every premise was checked against source before the verdict.

1. **Greptile P2, twin writes are non-atomic (`init.ts`): FIXED in `9d75d3ef`.** Premise confirmed: `scaffoldClaudeSkill` wrote in place with `fs.writeFileSync` at its create, force and refresh sites, the same path the `.claude` copy has always taken, and this PR reused it for the twin. All three sites now go through `writeFileAtomicSync`, the helper the Tenet 4 user-file contract names and this file already used for its other writes. Test: with the atomic seam armed to throw, a stale marker-bearing twin keeps its bytes, no temp file sits beside it, the three absent twins are not created, and every row reports the error; with the seam disarmed the same run refreshes one and creates three. Red with the in-place writes (the armed seam had nothing to intercept), green with the helper.

2. **Greptile P2, skill roots are mirrored (`eject.ts`): FIXED in `004a5252`.** The re-armed falsification leg found the same thing (its F3) before the trigger landed: `SKILL_TWIN_ROOTS` now lives in `init-templates.ts` and `totem eject` reads it for the scrub loop, the prune loop and the dirty-tree roster; the roster test derives its expectation from the same constant.

3. **CodeRabbit Major, outside the diff, marker order (`eject.ts`): FIXED in `9d75d3ef`.** Premise confirmed at `init.ts` (`existingStart > existingEnd` is the no-marker guard on the write side) against a presence-only test on the scrub side, pre-existing for `.claude` and extended to `.agents` by this PR. The scrub now applies the same three-way test, so a file whose end marker precedes its start marker is preserved and named under Skipped on either root. Test on both roots, red before the fix (eject deleted both files), green after.

4. **CodeRabbit Minor, prune failures unreported (`eject.ts`): FIXED in `9d75d3ef`.** The catch around the empty-skills-root prune now appends `<root>/skills/ (could not prune the empty directory: <message>)` to the skipped list; best-effort, accounted. No test forces an `rmdir` failure; disclosed.

5. **GCA HIGH, filter `.agents` out of eject's roots when the directory is absent (`eject.ts`): DECLINED, measured.** The four `(not found)` rows are the shape the scrub has always printed for `.claude`: on `main` the same loop pushes `.claude/skills/<name>/SKILL.md (not found)` for every distributed skill that is absent, whatever `.claude/` holds. The twin rows are that same accounting on the second root, one row per surface item, and the re-armed leg raised the noise question (its F11) and it was ruled that the rows stay symmetric rather than one root being filtered and the other not. The prune loop the suggestion would shorten is two `existsSync` probes.

6. **CodeRabbit Trivial, lazy-load the test bindings (`init.test.ts`): DECLINED.** The lazy-load instruction is the command modules' startup discipline (the comment at `eject.ts` cites mmnto-ai/totem#2299); this test file, like every test file in the package, imports its subjects statically, and the block in question already held `initCommand` and the other exports the same way. Fold 2 also moved `SKILL_TWIN_ROOTS` to `init-templates.ts`, so the thread's line is outdated.

Verification at `9d75d3ef`: workspace build exit 0; full `pnpm test` exit 0 (cli 182 files, 5033 passed, 20 skipped; core 159 files, 4630 passed); `format:check` clean; ESLint 0 errors; `totem lint --branch` PASS, 0 errors; the changeset carries both safety folds.

disposition: 4058269575 fixed
disposition: 4058269578 fixed
disposition: 4058280020 fixed
disposition: 4058267743 declined
disposition: 4058280025 declined

local-lane: none leg: 3c6cdad1 blocking=0 material=3 folded=10

